PR #11, opened 2018-12-20
Add unassessed activities list for the evaluation hub

CarolBusman:
This is the base of the activities list. Not yet implemented, but planned:
- filtering (waiting on design)
- sorting (waiting on api)
- paging (waiting on api)
- `d2l-list` styling (instead of using `d2l-table`)
- translations


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -0,0 +1,100 @@
...
+			function loadAndWait(id, done) {
+				list = fixture(id);
+				function waitForLoad() {
+					if (list._loading) {
+						setTimeout(waitForLoad, 250);
+					} else {
+						done();
+					}
+				}
+				waitForLoad();
```
CarolBusman: Unsure if this is the best way to ensure that the data has been loaded into the list. When using `flush()` the test were still failing sometimes, not having the data loaded before the assert.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,286 @@
...
+
+	getHref(entity, rel) {
+		if (entity.hasLinkByRel(rel)) {
+			return entity.getLinkByRel(rel).href;
+		}
+		return '';
+	}
+
+	getActivityLink(/*entity*/) {
+		// TODO: figure out how to construct this link!!!
```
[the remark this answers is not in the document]
CarolBusman: Yup! The api just isn't returning the link yet, so this is a placeholder for now (forgot to add that to the summary when I made the PR).

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,286 @@
...
+										<d2l-td>[[getProperty(s, h.key)]]</d2l-td>
+									</template>
+								</dom-repeat>
+							</d2l-tr>
+						</template>
+					</dom-repeat>
+				</d2l-tbody>
+			</d2l-table>
+			<d2l-offscreen>
+				<button onclick="[[loadMore]]">Load more</button>
```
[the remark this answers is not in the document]
m6jones: This repo hasn't been set up with serge yet. But I am planning on doing that today.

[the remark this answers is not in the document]
m6jones: Serge support pr https://github.com/BrightspaceHypermediaComponents/activities/pull/12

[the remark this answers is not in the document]
CarolBusman: This button is temporary for now, as a placeholder for when the api implements paging.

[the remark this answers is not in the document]
CarolBusman: Translations/Serge will be addressed in a future PR.

CarolBusman:
This PR is ready for re-review; fixes for current comments have been checked in.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+		// window.location.href = e.model.s.activityLink;
+		// console.log(e.model.s.activityLink);
+	}
+
+	loadData(entity) {
+		if (!entity) {
+			return Promise.resolve();
+		}
+
+		this.set('_loading', true);
```
CodeBaboon: `.set` is not necessary for manipulating the value of simple types. I believe it is only necessary when setting sub-properties on objects: https://polymer-library.polymer-project.org/2.0/docs/devguide/model-data#set-path

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+			.catch(function() {
+
+			}.bind(this))
+			.then(function() {
+				this.set('_loading', false);
+			}.bind(this));
+	}
+
+	followLink(data, rel) {
+		var entity = data.entity || data;
```
CodeBaboon: This feels a bit weird to me, and is also technically open to a null reference error if `data` is `null`. Consider maybe passing in the correct value from the caller so that this is not necessary (either `data.entity` or `me.entity`, I'm unsure which).

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+		if (entity.hasLinkByRel(rel)) {
+			return this.fetch(entity.getLinkByRel(rel).href);
+		}
+		return Promise.resolve(data);
+	}
+
+	async parseActivities(act) {
+		var self = this;
+		var promises = [];
+		var entity = act.entity || act;
```
CodeBaboon: Similar to my concern over the implementation of `followLink`, this is open to a null reference error. Also, with only one caller that I can see, I'm not sure when/why the `act` would be different between calls such that the check is necessary?

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+				});
+			}));
+		});
+
+		self._filterHref = self.getHref(entity, Rels.filters);
+		//self._sortHref = self.getHref(entity, Rels.sort);
+		//self._pageNextHref = self.getHref(entity, Rels.pageNext);
+
+		const result = await Promise.all(promises);
+		self.set('_data', self._data.concat(result));
```
CodeBaboon: This is another case where I don't believe `.set` is necessary, and can just be `self._data = self._data.concat(result);`.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+
+	getActivityLink(/*entity*/) {
+		// TODO: get this link from api once enabled!!!
+		return 'link to activity';
+	}
+
+	getActivityPromise(entity, item) {
+		if (entity.hasLinkByRel(Rels.assignment)) {
+			var assignment = entity.getLinkByRel(Rels.assignment);
+			return this.fetch(assignment.href)
```
CodeBaboon: You could probably make use of the `getHref` function from a few lines up, so something like

```
var href = this.getHref(Rels.assignment);
if (href) {
    return this.fetch(href)
                     .then(function(a) {
                         item.activityName = a.entity.properties.name;
                    });
}
return Promise.resolve();
```

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+					item.activityName = a.entity.properties.name;
+				});
+		}
+		return Promise.resolve();
+	}
+
+	getCoursePromise(entity, item) {
+		if (entity.hasLinkByRel(Rels.organization)) {
+			var org = entity.getLinkByRel(Rels.organization);
+			return this.fetch(org.href)
```
CodeBaboon: Another possible re-use of `getHref`.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+					item.courseName = o.entity.properties.name;
+				});
+		}
+		return Promise.resolve();
+	}
+
+	getUserPromise(entity, item) {
+		if (entity.hasLinkByRel(Rels.user)) {
+			var user = entity.getLinkByRel(Rels.user);
+			return this.fetch(user.href)
```
CodeBaboon: Another possible re-use of getHref.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,260 @@
...
+										<d2l-td>[[getProperty(s, h.key)]]</d2l-td>
+									</template>
+								</dom-repeat>
+							</d2l-tr>
+						</template>
+					</dom-repeat>
+				</d2l-tbody>
+			</d2l-table>
+			<d2l-offscreen>
+				<button onclick="[[loadMore]]">Load more</button>
```
dbatiste: Offscreen focusable elements are a little odd because for sighted keyboard users, the user "loses" the focus when focus goes offscreen.  Addressing likely requires some design.

dbatiste: Also, VoiceOver on iOS _might_ force a scroll to that element, which could really confuse people.  Should be checked.

svanherk: Yeah this will be totally different, our plan is to create an "infinite-loader" shared component/behaviour that wraps a table or list element.  We don't really have any designs for this yet, I _assume_ we'll want to implement something similar to what MVC grids uses right now?

CarolBusman: This is just a temporary placeholder until the paging is implemented. It is definitely not in its final form yet.

dbatiste: Yeah, a few of us have talked about this a bit.  There are a few components that all need to play nice together, and definitely we don't want various apps repeating the code or reinventing.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
CarolBusman: On first load of the component, we get data from the root of the api, but after coming back from the filter/sort/paging, we'll have a list of unassessed activities. I believe the problem was when writing the test for these two scenarios, I wasn't accurately representing what state the entities would be in when being passed into `loadData()`. I've updated the tests to better reflect how the data comes in in the two scenarios.

CarolBusman:
I believe all comments have been addressed, ready for re-review.

components/d2l-evaluation-hub-activities-list/d2l-evaluation-hub-activities-list.js
```diff
@@ -0,0 +1,248 @@
...
+			}.bind(this))
+			.catch(function() {
+			}.bind(this))
+			.then(function() {
+				this._loading = false;
+			}.bind(this));
+	}
+
+	followLink(entity, rel) {
+		if (entity && entity.hasLinkByRel && entity.hasLinkByRel(rel)) {
```
CodeBaboon: Not super important, but this seems like another potential place for calling out to `getHref` to avoid re-writing logic (especially if you add the `entity` check condition into that method implementation). So then you'd have something like:

```
followLink(entity, rel) {
    var href = getHref(entity, rel);
    if (href) {
        return this.fetch(href);
    }
}
```

CarolBusman: I'll add this in the next PR, if there are no further requested changes in this one.

PR #12, opened 2018-12-21
Added Serge support and added the enroll term.

m6jones:
This adds serge support to this repo.
https://docs.dev.d2l/index.php/Serge-Localize
Important files here are: 
- package.json
- templates/lang-mixin.ejs
- components/d2l-activity-list-item/ActivityListItemLocalize.js
- components/d2l-activity-list-item/build/lang/en.js
- gulpfile.js
- components/d2l-activity-list-item/d2l-activity-list-item.js
- activities.serge.json
- components/d2l-activity-list-item/d2l-activity-list-item-enroll.js

### DIFF
--- a/components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
+++ b/components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
@@ -1,5 +1,5 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
-import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
+import {ActivityListItemLocalize} from './ActivityListItemLocalize.js';
 import 'd2l-typography/d2l-typography.js';
 import 'd2l-button/d2l-button.js';
 import SirenParse from 'siren-parser';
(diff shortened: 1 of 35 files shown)


components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
```diff
@@ -9,15 +9,17 @@ import 'd2l-fetch/d2l-fetch.js';
  * @customElement
  * @polymer
  */
-class D2lActivityListItemEnroll extends mixinBehaviors([D2L.PolymerBehaviors.FetchSirenEntityBehavior], PolymerElement) {
+class D2lActivityListItemEnroll extends ActivityListItemLocalize(PolymerElement) {
```
AKobets: Could you please explain removing `D2L.PolymerBehaviors.FetchSirenEntityBehavior` from here ?

m6jones: Oh, I forgot about that. It was cleanup that I missed. The behaviour was no longer used as I went to a more compatible lower level way of handling it.
It would handle these two methods https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-list-item/d2l-activity-list-item-enroll.js#L77
However, the process that they used doesn't work in IE 11.

PR #16, opened 2019-01-09
Hover and Focus for list item.

m6jones:
Update the behaviour of the activity list item for hover and focus.

A/C:
- A item is able to have focus and hover in AND/OR combinations
A user can have focused on a search result item with the keyboard and hover with the mouse on another (AKA don't move focus due to the user's hovering)
- When normal, the top item doesn't have a top border. When focused/hovered, it does
   Ensure that the layout doesn't change due to hover/focus
- When the cursor is in the "bounds" of the search result, the hover effect is activated
NOT just on the link
The background of the list item changes to #F9FAFB on hover
The link appears underlined when hovered

image

### DIFF
--- a/components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
+++ b/components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
@@ -16,6 +16,9 @@ class D2lActivityListItemEnroll extends ActivityListItemLocalize(PolymerElement)
 				:host {
 					display: block;
 				}
+				:host([hidden]) {
+					display: none;
+				}
 			</style>
 			<d2l-button id="d2l-activity-list-item-enroll" primary$=[[_hasActionBoolean(actionEnroll)]] disabled$=[[!_hasActionBoolean(actionEnroll)]]>
 				[[localize('enroll')]]
--- a/components/d2l-activity-list-item/d2l-activity-list-item.js
+++ b/components/d2l-activity-list-item/d2l-activity-list-item.js
@@ -35,9 +35,27 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 					border-color: rgba(0, 111, 191, 0.4);
 					box-shadow: 0 0 0 4px rgba(0, 111, 191, 0.3);
 				}
-				.d2l-activity-list-item-container:hover,
-				.d2l-activity-list-item-container:focus {
-					background-color: #FAFBFC; /* Pending Colin */
+				.d2l-activity-list-item-top-line,
+				.d2l-activity-list-item-bottom-line {
+					display: none;
+					border-top: 1px solid var(--d2l-color-mica);
+					margin: 0;
+				}
+				.d2l-activity-list-item-top-line {
+					margin-top: -2px;
+				}
+				.d2l-activity-list-item-bottom-line {
+					margin-bottom: -2px;
+				}
+				.d2l-activity-list-item-container:hover .d2l-activity-list-item-top-line,
+				.d2l-activity-list-item-container:hover .d2l-activity-list-item-bottom-line {
+					display: block;
+				}
+				.d2l-activity-list-item-container:hover {
+					background-color: var(--d2l-color-regolith);
+				}
+				.d2l-activity-list-item-container a.d2l-focusable:focus {
+
 				}
 				.d2l-activity-list-item-container {
 					position: relative;
@@ -45,7 +63,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 				}
 				.d2l-activity-list-item-link-container {
 					overflow: hidden;
-					display: flex;
+					display: inline-flex;
 					flex-direction: row;
 					flex-grow: 1;
 					flex-shrink: 1;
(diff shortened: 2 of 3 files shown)


components/d2l-activity-list-item/d2l-activity-list-item-enroll.js
```diff
@@ -16,6 +16,9 @@ class D2lActivityListItemEnroll extends ActivityListItemLocalize(PolymerElement)
 				:host {
 					display: block;
 				}
+				:host([hidden]) {
```
m6jones: This is clean up to allow the enrollment button to be hidden.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -31,21 +31,38 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 					display: block;
 					max-width: 842px;
 				}
-				:host([active]) {
+				:host([active]) a.d2l-focusable {
```
m6jones: This puts the focus border on top of everything else.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -146,6 +163,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 				}
 			</style>
 			<div class="d2l-activity-list-item-container">
+				<hr class="d2l-activity-list-item-top-line" />
```
m6jones: I would have prefer to add these lines by using a shadow or a border. However, these lines should be added outside of the element and this was the best solution I came up with.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -173,8 +191,9 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 						</div>
 					</div>
 				</div>
+				<hr class="d2l-activity-list-item-bottom-line" />
 			</div>
-			<d2l-activity-list-item-enroll hidden$="[[!_actionEnroll]]" action-enroll="[[_actionEnroll]]"></d2l-activity-list-item-enroll>
+			<d2l-activity-list-item-enroll hidden$="[[actionEnrollHide]]" action-enroll="[[_actionEnroll]]"></d2l-activity-list-item-enroll>
```
m6jones: This is to hide the enrollment button and is more of clean up.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -232,6 +251,10 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 				type: String,
 				value: ''
 			},
+			actionEnrollHide: {
```
m6jones: cleanup, default is to hide the enrollment button.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -317,8 +340,8 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 
 		this._currentResponsiveConfig = currentConfig;
 		window.fastdom.mutate(() => {
-			const container = this.shadowRoot.querySelector('.d2l-activity-list-item-container');
-			container.style.padding = currentConfig.padding;
+			const container = this.shadowRoot.querySelector('.d2l-activity-list-item-link-container');
+			container.style.margin = currentConfig.padding;
```
m6jones: When this was padding it would push the link down as well. So I went down to a child and made it margin.

demo/d2l-activity-list-item/index.html
```diff
@@ -32,6 +32,9 @@
 				h3 {
 					text-align: center;
 				}
+				d2l-activity-list-item {
+					border-bottom: 1px solid var(--d2l-color-mica);
```
m6jones: Adding this mimics how the list will look.

AdmiralCoco: (Question) Do you know how the colors get resolved given that it isn't imported anywhere? 
My assumption is that it's included by some other package that depends on it?
 
(Just wondering as I came across this when I used colors in discovery-fra and it wasn't installed)

components/d2l-activity-list-item/d2l-activity-list-item.js
AdmiralCoco: That seems reasonable 👍 

So adding `border-top/bottom` of `d2l-activity-list-item-container` wouldn't work as it's within the element? 
(Sorry not 100% sure which lines are referenced here)

demo/d2l-activity-list-item/index.html
m6jones: It comes from `"d2l-colors": "BrightspaceUI/colors#semver:^4",`

AdmiralCoco: Right, but we don't need to `import` it in this file? Or is it just available to use automatically?

components/d2l-activity-list-item/d2l-activity-list-item.js
m6jones: Exactly.

demo/d2l-activity-list-item/index.html
dbatiste: It should be imported in this file.

PR #18, opened 2019-01-10
hypermedia calls can start directly from https://activities.api.prodev.d2l/my-unassessed-activities, don't have to go through root.

CarolBusman:
Also simplified some Promises to async/await for readability, added some null checks, rewrote `followLink()` to use `getHref()`.

### DIFF
--- a/demo/d2l-evaluation-hub/index.html
+++ b/demo/d2l-evaluation-hub/index.html
@@ -35,7 +35,6 @@
 			<demo-snippet>
 				<template strip-whitespace>
 					<d2l-evaluation-hub-activities-list href="data/activities/activitiesList/unassessedActivities.json" token="token"></d2l-evaluation-hub-activities-list>
-					<!-- <d2l-evaluation-hub-activities-list href="https://activities.api.proddev.d2l/my-unassessed-activities" token="token"></d2l-evaluation-hub-activities-list> -->
 				</template>
 			</demo-snippet>
 		</div>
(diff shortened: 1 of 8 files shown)


demo/d2l-evaluation-hub/index.html
```diff
@@ -34,7 +34,8 @@ <h3>Basic d2l-evaluation-hub demo</h3>
 			<h3>d2l-evaluation-hub-activities-list demo</h3>
 			<demo-snippet>
 				<template strip-whitespace>
-					<d2l-evaluation-hub-activities-list href="data/activities/activitiesList/root.json" token="whatever"></d2l-evaluation-hub-activities-list>
+					<d2l-evaluation-hub-activities-list href="data/activities/activitiesList/unassessedActivities.json" token="token"></d2l-evaluation-hub-activities-list>
+					<!-- <d2l-evaluation-hub-activities-list href="https://activities.api.proddev.d2l/my-unassessed-activities" token="token"></d2l-evaluation-hub-activities-list> -->
```
svanherk: Do we need this comment here? Could we instead add a comment about this to the README, and use `<d2l-evaluation-hub>` instead?

test/d2l-evaluation-hub/data/root.json
```diff
@@ -1,8 +0,0 @@
-{
```
svanherk: Is the test data the exact same as the demo data?  Should we instead combine them into the same place, or do we imagine them being different someday?

CarolBusman: The intent is to have the demo have more data to give a good example of what the table is going to look like, whereas the tests would have just enough data to perform the tests.

svanherk:
FYI Jamie - For monolith work

svanherk:
Corresponding LE PR: https://git.dev.d2l/projects/CORE/repos/le/pull-requests/13627/overview

CarolBusman:
@CodeBaboon @pbrown-d2l Are there any concerns with this PR, or is it be approved?

PR #21, opened 2019-01-14
Rearrange Repo

svanherk:
- Moving the `d2l-evaluation-hub-activities-list` inside the `d2l-evaluation-hub` folders
- Other small cleanup, should be no functional changes


index.html
```diff
@@ -2,7 +2,7 @@
 <html>
 	<head>
 		<meta charset="utf-8">
-		<meta http-equiv="refresh" content="0;url=demo/" />
+		<meta http-equiv="refresh" content="0;url=/components/d2l-activities/demo/" />
```
svanherk: This should help with the problem Phil was having with the demo, we'll now bump people to the correct demo level if they go `localhost:<port>`.

PR #23, opened 2019-01-14
Adding activity link to table.

CarolBusman:


package.json
```diff
@@ -50,7 +50,7 @@
 		"d2l-colors": "BrightspaceUI/colors#semver:^4",
 		"d2l-course-image": "Brightspace/course-image#semver:^3",
 		"d2l-fetch": "Brightspace/d2l-fetch.git#semver:^2",
-		"d2l-hypermedia-constants": "^6.1.0",
```
svanherk: Does pointing to `^6.5.0` not work?  I saw it being picked up properly in the latest BSI release

CarolBusman: I was getting an error that the version didn't exist when I was trying to use just the version and going with this ended up getting the correct version. However I believe I may have made some changes afterward that may have made using `^6.5.0` work, but I didn't go back and change it to check.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -34,10 +35,17 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				<d2l-tbody>
 					<dom-repeat items="[[_data]]" as="s">
 						<template>
-							<d2l-tr on-click="selectActivity">
+							<d2l-tr>
 								<dom-repeat items="[[_headers]]" as="h">
 									<template>
-										<d2l-td>[[getProperty(s, h.key)]]</d2l-td>
+										<d2l-td>
+											<template is="dom-if" if="[[h.canLink]]">
```
CodeBaboon: This might be old knowledge (Polymer 1), but back then the finding was that `dom-if` was slower than just using `hidden` classes on elements when the internals in question were simple types (like not other components that would need to be loaded and go fetch data and whatnot). So in this scenario it might be easier to just apply CSS to show/hide the anchor/span.

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -46,20 +46,34 @@
...
 				expected.push(item.submissionDate);
+				expectedLink.push(item.activityLink);
 			});
 
 			await loadPromise('data/unassessedActivities.json');
 
 			var data = list.shadowRoot.querySelectorAll('d2l-td');
 			for (var i = 0; i < expected.length; i++) {
-				assert.equal(expected[i], data[i].innerHTML);
+				if (i % 4 === 0) {
```
CodeBaboon: If another column gets added, or the link moves, this logic breaks, right? Maybe consider an alternate structure that could allow for each item to be tested/compared without it's relative placement being important (like `expected.push({ type: 'link', thing: item.displayname});`, etc).

package.json
svanherk: We've seen issues with npm grabbing different versions of repos and causing clashign errors in the LMS (which is why we had to change to `semver:^2` instead of `semver:^2.0.0` for all our unregistered stuff).  This might confuse npm if someone else is using `^6.5.0`, I think we should try that again and see if that works

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -34,10 +38,13 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				<d2l-tbody>
 					<dom-repeat items="[[_data]]" as="s">
 						<template>
-							<d2l-tr on-click="selectActivity">
+							<d2l-tr>
 								<dom-repeat items="[[_headers]]" as="h">
 									<template>
-										<d2l-td>[[getProperty(s, h.key)]]</d2l-td>
+										<d2l-td>
+											<a href="[[s.activityLink]]" hidden$="[[!h.canLink]]">[[getProperty(s, h.key)]]</a>
```
svanherk: We should use `d2l-link` here

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -48,18 +48,30 @@
...
+				expected.push({ text: item.courseName, link: '' });
+				expected.push({ text: item.submissionDate, link: '' });
 			});
 
 			await loadPromise('data/unassessedActivities.json');
 
 			var data = list.shadowRoot.querySelectorAll('d2l-td');
 			for (var i = 0; i < expected.length; i++) {
-				assert.equal(expected[i], data[i].innerHTML);
+				var link = data[i].querySelector('a');
```
svanherk: I think you need to update this to `d2l-link` as well

CarolBusman:
@CodeBaboon Are there any further concerns, or can this PR be merged?

PR #25, opened 2019-01-15
Add lang terms for Evalualtion Hub via serge.io

jeroach:
Fixed :)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -46,7 +47,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				</d2l-tbody>
 			</d2l-table>
 			<d2l-offscreen>
-				<button onclick="[[loadMore]]">Load more</button>
+				<button onclick="[[loadMore]]">[[localize('loadMore')]]</button>
```
jeroach: This works (I've tried changing the lang term).

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -56,10 +57,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			_headers: {
 				type: String,
 				value: [
-					{ key: [ 'displayName' ], sortKey: 'displayName', displayName: 'Submitter'},
-					{ key: [ 'activityName' ], sortKey: 'activityName', displayName: 'Activity Name'},
-					{ key: [ 'courseName' ], sortKey: 'courseName', displayName: 'Course Name'},
-					{ key: [ 'submissionDate' ], sortKey: 'submissionDate', displayName: 'Submission Date'}
+					{ key: [ 'displayName' ], sortKey: 'displayName', displayName: localize('displayName') },
```
jeroach: These do not work currently. The `localize` method cannot be found.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -12,7 +13,7 @@ import {Rels} from 'd2l-hypermedia-constants';
  * @polymer
  */
 
-class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior], PolymerElement) {
+class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior], EvaluationHubLocalize(PolymerElement)) {
```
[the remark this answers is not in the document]
m6jones: It is proper. You can only extend one class so to extend multiple you chain it.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
m6jones: This will need to be `this.localize` It might have to be wrapped in a method. I've never seen a `string` default value be an array of objects.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -26,7 +27,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					<d2l-tr>
 						<dom-repeat items="[[_headers]]">
 							<template>
-								<d2l-th on-click=sort>{{item.displayName}}</d2l-th>
+								<d2l-th on-click=sort>[[localize(item.sortKey)]]</d2l-th>
```
jeroach: We can just pass the `sortKey` in the HTML block here, which lets us remove the `displayName` property

m6jones: I guess one thing to note is that `item.displayName` is a two way binding. We want to get rid of that type of behaviour so removing the two way bind is good. However, we might want to double check it doesn't change anything.

CarolBusman: The only thing about using the `sortKey` to grab the localization is that they aren't finalized yet. The intention is to use them with the API, so if we need a different key for that and have to change it, all the localization would need to be updated as well. I'm wondering if we should have a `sortKey` and a `localizationKey`?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
CarolBusman: Yes, this should be `type: Array`. This was missed in previous PRs, since it does still work.

jeroach: Maybe I'm not understanding - is there action required here on this PR?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
jeroach: Hmmm, yeah if they're subject to change then we should have separate keys.

I'll add a `localizationKey` for now. If we find in the future that it's the same as the `sortKey` then we can just change the code, without needing to update the lang terms.

jeroach:
@m6jones @CarolBusman Any last concerns, or good to merge this? Thanks!

components/d2l-evaluation-hub/build/lang/zh.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+import {dedupingMixin} from '@polymer/polymer/lib/utils/mixin.js';
+
+/* @polymerMixin */
+const LangZhImpl = (superClass) => class extends superClass {
```
svanherk: Does this work properly in IE11?

jeroach: I'll check

jeroach: Works for `en` - is there a way to force another language?

Also, the EH demo is broken in IE11. No records load in the table.

The lang terms only affect the headers.

ie11_en

svanherk: You can force other languages by changing the `html` tag to use a different `lang`, like <html lang="fr">`.

jeroach: Translations work in IE11 for other languages - example below:

ie11_zh

Demo still broken in IE11

svanherk: Do you still have your local instance setup?  Is our demo broken in IE11, or is evaluation hub broken in IE11?

jeroach: Let me check

svanherk: Ok, not seeing any obvious fixes.  Tried including a number of different polyfills but that didn't help

m6jones: I've always found demos broken in IE11. I believe it is the `demo-snippet` isn't supported?

svanherk: IE11 works with other Polymer 3 repos, but I haven't been able to figure out what part of these demos it hates yet.  I can keep investigating and apply a fix to all three, doesn't block this PR.  Just wanted to make sure the arrow functions were being polyfilled properly, IE11 hates those.

jeroach: Sounds good - merging this PR then :)

PR #26, opened 2019-01-16
Fix demos

svanherk:
Our demos weren't actually applying the typography styles correctly.  Now they do!

### DIFF
--- a/demo/d2l-activity-list-item/index.html
+++ b/demo/d2l-activity-list-item/index.html
@@ -13,29 +13,37 @@
 			import '@polymer/iron-demo-helpers/demo-snippet';
 			import 'd2l-typography/d2l-typography.js';
 			import '../../components/d2l-activity-list-item/d2l-activity-list-item.js';
-		</script>
 
-		<style is="custom-style" include="demo-pages-shared-styles"></style>
-		<style is="custom-style" include="d2l-typography">
-				html {
-					font-size: 20px;
-				}
-				.fixedSize {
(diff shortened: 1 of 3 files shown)


demo/d2l-activity-list-item/index.html
```diff
@@ -13,29 +13,37 @@
 			import '@polymer/iron-demo-helpers/demo-snippet';
 			import 'd2l-typography/d2l-typography.js';
 			import '../../components/d2l-activity-list-item/d2l-activity-list-item.js';
-		</script>
 
-		<style is="custom-style" include="demo-pages-shared-styles"></style>
```
svanherk: @maya I fixed your demo to use the typography font as well, is that ok?  If not I can just change ours.

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -32,8 +41,4 @@ <h3>Basic d2l-evaluation-hub-activities-list demo</h3>
 			</demo-snippet>
 		</div>
 	</body>
-	<!-- The class="d2l-typography" on the body doesn't always work -->
```
svanherk: This is not a problem in Polymer 3, from what I've seen

demo/d2l-activity-list-item/index.html
m6jones: Thank you! I thought mine was working eek!

PR #27, opened 2019-01-16
Fix for quizzes.

CarolBusman:
Adding in logic to pull quiz name for quizzes. Also changing `_headers` to an `Array` to better reflect the actual data.

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -6,7 +6,7 @@ import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
 import 'd2l-polymer-siren-behaviors/store/siren-action-behavior.js';
 import 'd2l-link/d2l-link.js';
 import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
-import {Rels} from 'd2l-hypermedia-constants';
+import {Rels, Classes} from 'd2l-hypermedia-constants';
 
 /**
  * @customElement
@@ -62,7 +62,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 	static get properties() {
 		return {
 			_headers: {
-				type: String,
+				type: Array,
 				value: [
 					{ key: [ 'displayName' ], sortKey: 'displayName', displayName: 'Submitter', canLink: true},
 					{ key: [ 'activityName' ], sortKey: 'activityName', displayName: 'Activity Name', canLink: false},
@@ -140,6 +140,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		if (href) {
 			return this.fetch(href);
 		}
+		return Promise.resolve();
 	}
 
 	async parseActivities(entity) {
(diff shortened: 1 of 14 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -140,6 +140,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		if (href) {
 			return this.fetch(href);
 		}
+		return Promise.resolve();
```
CarolBusman: Currently, if there is no href, this returns undefined, which is causing the entire table not to load if there's an issue with one record; returning a resolved promise so that the table continues to load.

AlexBedley:
Note that Amaury just merged [the PR to add discussions](https://git.dev.d2l/projects/CORE/repos/le/pull-requests/13622/overview), so the UI shouldn't choke on those either.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
jmurray87: We just recently added discussions as well, which also has a different activity entity :(

svanherk: Was this the issue Josh was seeing on quad this morning?

jeroach: @svanherk Yep

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -181,12 +182,22 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
-			});
+		if (entity.hasClass(Classes.activities.userQuizActivity)) {
+			return this.followLink(entity, Rels.quiz)
+				.then(function(a) {
+					if (a && a.entity && a.entity.properties) {
+						item.activityName = a.entity.properties.name;
+					}
+				});
+		} else if (entity.hasClass(Classes.activities.userAssignmentActivity)) {
+			return this.followLink(entity, Rels.assignment)
```
jmurray87: For discussion we'll be looking for this rel https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/index.js#L87 - let me get back to you on what the class will be on the entity

svanherk: From what I understand though, this can go in as is.  The discussion rows just wouldn't be present, but the assignments and quizzes will appear

jmurray87: Yep, doesn't need to block this PR - just FYI for now

jmurray87: This will be the class! https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/index.js#L168 should have known, but had to double check lol

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -181,12 +182,22 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
 	getActivityPromise(entity, item) {
-		return this.followLink(entity, Rels.assignment)
-			.then(function(a) {
-				if (a && a.entity && a.entity.properties) {
-					item.activityName = a.entity.properties.name;
-				}
-			});
+		if (entity.hasClass(Classes.activities.userQuizActivity)) {
+			return this.followLink(entity, Rels.quiz)
+				.then(function(a) {
```
svanherk: Since the `then` code will be the same for all three, we could do something like:
```
var rel;
if (entity.hasClass(Classes.activities.userQuizActivity)) {
  rel = Rels.quiz;
} else if (entity.hasClass(Classes.activities.userAssignmentActivity)) {
  rel = Rels.assignment;
} else {
  return Promise.resolve();
}
return this.followLink(entity, rel)
...
```

svanherk:
I'm no longer seeing Activity Name being shown in the demo, and we should also add some data for quizzes and discussions as well

svanherk:
Things still look good 👍

PR #30, opened 2019-01-30
Fixed issue where contents of list item would overflow

m6jones:
Fixed issue where image on the list item wasn't hidden to screen reader.

Image is decorative.
> Decorative images don’t add information to the content of a page. For example, the information provided by the image might already be given using adjacent text, or the image might be included to make the website more visually attractive.

NAMELY:

> the information provided by the image might already be given using adjacent text,

https://www.w3.org/WAI/tutorials/images/decorative/

### DIFF
--- a/components/d2l-activity-list-item/d2l-activity-list-item.js
+++ b/components/d2l-activity-list-item/d2l-activity-list-item.js
@@ -174,8 +174,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 						<d2l-course-image
 							image="[[_image]]"
 							sizes="[[_tileSizes]]"
-							type="narrow"
-							aria-hidden>
+							type="narrow">
 						</d2l-course-image>
 					</div>
 					<div class="d2l-activity-list-item-content">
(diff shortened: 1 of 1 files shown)


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -173,7 +174,8 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 						<d2l-course-image
 							image="[[_image]]"
 							sizes="[[_tileSizes]]"
-							type="narrow">
+							type="narrow"
+							aria-hidden>
```
dbatiste: I notice that [d2l-course-image](https://github.com/Brightspace/course-image/blob/master/d2l-course-image.js#L39) does not provide an `alt` attribute on the `img`.  With the proposed change, we're delegating the responsibility to the consumer, yet there appears to be no way to provide alternate text, lead one to believe that `aria-hidden` should always be specified.  If this is the case, then I may suggest moving this into `d2l-course-image` itself because this attribute will easily be omitted.  If this is not the case, then I suggest `d2l-course-image` be updated to support specifying accessible description.

m6jones: That is a good idea! I'll go do that.

m6jones:
Fixed issue in d2l-course-image.
The other issue with overflow did not act the same today. So more investigation required.

PR #34, opened 2019-02-08
First Version of Infinite Loading - Just a button

svanherk:
There's a ton to consider with loading and paging, so for now I was going to leave the logic in here and then pull it out if we get to infinite-loading, or once the full table/list behaviours library is tackled.


components/d2l-evaluation-hub/build/lang/en.js
```diff
@@ -10,6 +10,7 @@ const LangEnImpl = (superClass) => class extends superClass {
 			'activityName': 'Activity Name',
 			'courseName': 'Course',
 			'displayName': 'First Name, Last Name',
+			'loading': 'Loading data',
```
svanherk: You guys are more familiar with Serge than I am - do I just ad it to `en` and a PR for the rest will be opened?  Or do I add it to all, and just set it to blank/the English term?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -23,11 +26,25 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					height:100%;
 					width:100%;
 				}
+				d2l-th {
```
svanherk: The `d2l-th` and `d2l-td` styles are temporary, until the list is created

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -23,11 +26,25 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+					margin-top: 1rem;
+					text-align: right;
+					width: 100%;
+				}
 				[hidden] {
 					display: none;
 				}
 			</style>
-			<d2l-table>
+			<d2l-table hidden$="[[_initialLoading]]" aria-colcount$="[[_headers.length]]" aria-rowcount$="[[_data.length]]">
```
svanherk: The `aria-colcount` and `aria-rowcount` are to solve a screen reader issue, but the screen reader is ignoring them.  I'm going to create a story to go back over the table accessibility (we should also have a summary).

svanherk:
I fixed some mixed line endings, so probably best to look at https://github.com/BrightspaceHypermediaComponents/activities/pull/34/files?w=1

components/d2l-evaluation-hub/build/lang/en.js
m6jones: Add it to all and set it to the english term. At least that is what I've been doing. :D

dbatiste: Nit: are we certain we want to use the "data" verbiage?  It's kinda of technical.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -54,9 +71,13 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					</dom-repeat>
 				</d2l-tbody>
 			</d2l-table>
-			<d2l-offscreen>
-				<button onclick="[[_loadMore]]">[[localize('loadMore')]]</button>
-			</d2l-offscreen>
+			<template is="dom-if" if="[[_pageNextHref]]">
+				<d2l-button class="d2l-evaluation-hub-activities-list-load-more" secondary onclick="[[_loadMore]]" hidden$="[[_loading]]">[[localize('loadMore')]]</d2l-button>
```
dbatiste: Note: `secondary` is not a property of `d2l-button`.  All `d2l-button`s are secondary unless they have the `primary` attribute or property set.

svanherk: Ah good to know, thanks!  I swear I saw this elsewhere... 🤔

components/d2l-evaluation-hub/build/lang/en.js
svanherk: Good point, I'll get an official term from the designer.  Was thinking this as a placeholder, but it's easier to just get it right the first time

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Here lol: https://github.com/Brightspace/d2l-class-stream-ui/blob/d00e9d3dfe2d358a3e1682e9105da83cca3e068e/components/d2l-cs-activities.html#L154

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -132,12 +155,44 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		} catch (e) {
 			// Unable to load activities from entity.
 		} finally {
+			this._initialLoading = false;
 			this._loading = false;
 		}
 	}
 
+	async _loadMore() {
```
dbatiste: What happens if `_loadMore` is invoked a second or third time before the first load-more is completed?  ex.  if the user clicks the load-more button more than once?

dbatiste: ... it looks like the load-more button will be hidden after the response is returned in `_loadData`, so the user could click more than once before that happens.  If the focus is on the load-more button and we hide it, focus will be moved to the body I think, and that might get announced to screen reader.  Might need a little testing of the focus interaction here.  Also, it looks like the load-more button may disappear, and then reappear - but perhaps that would happen outside the view of the user since it would reappear below the fold?  I think I may be tempted to only hide the button when we know there are no more "pages" of items, and while loading we disable the button.  Not sure.  I think I would want this behavior to be the same anywhere we use load-more.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -132,12 +155,44 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+		if (this._pageNextHref) {
+			this._followHref(this._pageNextHref).then(async function(u) {
+				if (u && u.entity) {
+					var tbody = this.shadowRoot.querySelector('d2l-tbody');
+					var lastFocusableTableElement = D2L.Dom.Focus.getLastFocusableDescendant(tbody, false);
+
+					try {
+						await this._loadData(u.entity);
```
dbatiste: I think focus will be moved to the body in `_loadData` as a result of hiding the button.  If there is an error, we'll need to decide where to put focus.

components/d2l-evaluation-hub/build/lang/en.js
dbatiste: Cool.  They might want to make a substitution.  The simplest might be `Loading`, `Loading more`, or `Loading more items`.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
dbatiste: Haha wow.  Yeah, it's not causing any harm, but it's not doing anything either.  ;)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Hiding the load-more button is the first thing that `_loadData` does: http://localhost:8081/components/d2l-activities/test/.  I would assume doing that immediately triggers the hiding and doesn't wait for the response, am I misunderstanding that?

I do really like the idea of disabling the button instead of hiding it, and only removing if there is nothing left to load, to handle some of the focus stuff.  I'll see if that's what we do elsewhere.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Good point - I don't know if design has finalized what we want to do on error state, but I'll highlight that to them and see if they have a preference.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
dbatiste: I don't think `_loadData` will be called until `this._followHref` (which calls fetch) is complete, but maybe I am missing something?

dbatiste: If disabling is the way we want to go that's cool.  I think we want to be sure there isn't any weird announcements, focus loss, and duplicate loading happening.

dbatiste: Oh, the other small thing I was going to mention - `_loadMore` is marked as `async`, but a promise is not being returned.  The caller `d2l-button`'s on-click doesn't care so it's not an issue, but if a test or any other code called this directly it should be able to expect a promise.

components/d2l-evaluation-hub/build/lang/en.js
svanherk: We're going with `Loading` for now, until Design has a chance to go through it

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: We've made a story about the error states and I've added it the questions for design.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: We both looked into this offline.  This function doesn't currently need to be `async` from what I can tell, and making it so doesn't help with the test issue.  Added a retry mechanism for that instead.  And then I think we're good for the other async functions in this file.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -169,19 +222,12 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 
 		self._filterHref = self._getHref(entity, Rels.filters);
 		//self._sortHref = self._getHref(entity, Rels.sort);
-		//self._pageNextHref = self._getHref(entity, Rels.pageNext);
+		self._pageNextHref = self._getHref(entity, 'next');
```
CarolBusman: Do we want to add this to to `d2l-hypermedia-constants`?

svanherk: I wasn't sure, `self` also wasn't there

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Attribute removed

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: I added a check to not run if `_loading` is already true, and I set `_loading` to true early.  Also using the `disabled` button now, but I am hearing a weird announcement of "unavailable" sometimes, if the "alert Loading" doesn't kick in soon enough (or at all).

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -102,19 +122,17 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		super.ready();
 		var self = this;
```
dbatiste: The `self` assignment could be avoided here by using `bind`.
```
this.addEventListener('d2l-siren-entity-error', function() {
	this._initialLoading = false;
	this._loading = false;
}.bind(this));
```

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
dbatiste: Similar to previous comment, the `self` assignment could be avoided above by using `bind` for the anonymous functions.  Aside, these statements (lines 221-223) could have just referenced `this`, but maybe `self` was used for clarity.  I'd avoid the `self` assignment altogether though.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -2,9 +2,12 @@ import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {EvaluationHubLocalize} from './EvaluationHubLocalize.js';
 import 'd2l-table/d2l-table.js';
 import 'd2l-colors/d2l-colors.js';
```
dbatiste: Is `d2l-colors` needed in this file?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -18,16 +21,25 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
 				}
 				[hidden] {
 					display: none;
 				}
 			</style>
-			<d2l-table>
+			<d2l-table hidden$="[[_initialLoading]]" aria-colcount$="[[_headers.length]]" aria-rowcount$="[[_data.length]]">
 				<d2l-thead>
 					<d2l-tr>
 						<dom-repeat items="[[_headers]]">
```
dbatiste: Looks like sorting is supported just yet, but it looks like the interaction will rely on a click of the `d2l-th`.  Are those keyboard accessible?

svanherk: Not yet, but I can add in the column sort buttons now so they're ready to go when we hook that up.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Nice catch, not anymore

svanherk:
Ok, I think I've addressed all comments or made/updated follow-up stories.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Spoke with Dave about this yesterday, we decided not to disable the button to avoid the screenreader from doing this.  The button is still not active (before we start loading again, we wait until we've finished loading).

svanherk:
@dbatiste @CarolBusman Any further comments?

PR #38, opened 2019-02-13
- use quiz attempt activity

jmurray87:


CarolBusman:
You'll also need to update the class in the demo and test data in `unassessedActivities.json`.

PR #43, opened 2019-02-21
Jroach/us103571

jeroach:
~~WIP after our very productive pairing session :)~~

More or less ready for review. I'm having some issues with the unit tests that I would appreciate some input/thoughts on. Here's what's going on:
- tests pass locally via `npm test` on command line
- tests pass in web browser via `http://127.0.0.1:8081/components/d2l-activities/test/` however, `AssertionError` messages and promise failures show up in the console (see attached image).
- tests pass on Travis CI - Branch
- tests **fail** on Travis CI - PR

I think it's a timing thing, but I'm running out of ideas. I've tried:
- increasing the delay on `window.setTimeout`
- wrapping calls in a `flush()` with a `done();` callback
- entirely re-writing the tests in the `.js` file format here (with an `async function loadPromise(url)` function): https://github.com/BrightspaceHypermediaComponents/activities/blob/master/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js

Please let me know if you have any suggestions! Thanks! @mtjandra @svanherk @CarolBusman 

uncaught_assertion_error_with_tests_in_browser


components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,118 @@
...
+		return [
+			'_loadData(entity, href)'
+		];
+	}
+	ready() {
+		super.ready();
+	}
+	constructor() { super(); }
+
+	_fetch(url) {
```
svanherk: @mtjandra ran into an issue where we're this is overriding the `d2l-polymer-siren-behaviors`'s `_fetch` method (https://github.com/Brightspace/polymer-siren-behaviors/blob/bed996403a40dbb92675224df4fa2dc25e58ba5f/store/siren-action-behavior.js#L84).  We don't use actions in this file currently but to be safe I think we should change the name of this function to something else (but not `_fetchEntity` lol: https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L62).  I assume Miles is updating the main file as part of his PR

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,118 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import 'd2l-table/d2l-table.js';
```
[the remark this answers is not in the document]
svanherk: What you actually want, instead of `table`, is to also import `d2l-icons/d2l-icon.js`

components/d2l-activity-name/d2l-activity-name.js
mtjandra: Yeah it was conflicting with https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js#L84. Yeah I am planning to rename ours to `__fetch`

jeroach: Sounds good - I'll opt for `__fetch` here as well (double underscore)

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,113 @@
...
+		return [
+			'_loadData(entity, href)'
+		];
+	}
+	ready() {
+		super.ready();
+	}
+	constructor() { super(); }
+
+	__fetch(url) {
```
svanherk: Based on a convo with Batiste (https://github.com/BrightspaceHypermediaComponents/activities/pull/44#discussion_r259904925), Miles went with `_myEntityStoreFetch`

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,113 @@
...
+				value: ''
+			}
+		};
+	}
+	static get observers() {
+		return [
+			'_loadData(entity, href)'
+		];
+	}
+	ready() {
```
svanherk: I don't think we need `ready` or `constructor`  here since we don't add any functionality to them.

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,113 @@
...
+	__fetch(url) {
+		return window.D2L.Siren.EntityStore.fetch(url, this.token);
+	}
+
+	async _loadData(entity) {
+		if (!entity) {
+			return Promise.resolve();
+		}
+
+		this._loading = true;
```
svanherk: Doesn't look like we use the `loading` property for anything, I think we can just remove it (in the main list, it's used to block "reloading" the table if the load is still in progress, but there's no reloading here).

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,113 @@
...
+		try {
+			return this._getActivityPromise(entity);
+		} catch (e) {
+			// Unable to load activities from entity.
+		} finally {
+			this._loading = false;
+		}
+	}
+
+	_followLink(entity, rel) {
```
svanherk: Minor: I broke this function out in the main list because we needed to sometimes use just `_getHref` and `_followHref` separately.  Doesn't look like that's the case here so they could just be combined back into `_followLink`, but up to you.

test/d2l-activity-name/d2l-activity-name.html
```diff
@@ -0,0 +1,75 @@
...
+		<script type="module">
+			let activityName;
+			suite('d2l-activity-name base', function() {
+				setup(function() {
+					activityName = fixture('assignment');
+				});
+				test('instantiating the element works', function() {
+					assert.equal(activityName.tagName.toLowerCase(), 'd2l-activity-name');
+				});
+				test('activityName is set correctly', function() {
```
svanherk: So anytime you ever use a `setTimeout`, `flush`, `requestAnimationFrame`, etc in a test, you need to use the `done()` callback.  That's why you're seeing the tests pass, but then the assertions fail later.  The test is completing and not waiting for the `setTimeout` to finish to see if it actually passed.

jeroach: oh cool - good to know :) will update

components/d2l-activity-name/d2l-activity-name.js
jeroach: We debated this when we were implementing it in the pairing session.

I'm okay to remove, but I want to confirm my understanding first. We left it in so that it was available for the parent component to have access to. If that's not needed functionality, then we can take it out. @CarolBusman Am I remembering our discussion correctly?

components/d2l-activity-name/d2l-activity-name.js
jeroach: Sounds good, will update accordingly

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,113 @@
...
+
+	async _loadData(entity) {
+		if (!entity) {
+			return Promise.resolve();
+		}
+
+		this._loading = true;
+
+		try {
+			return this._getActivityPromise(entity);
```
mtjandra: If promise fails shouldn't it throw a `reject`, then handled by chaining a `catch`? Unsure if try/catch works, but don't think I've tried it before

components/d2l-activity-name/d2l-activity-name.js
svanherk: The parent shouldn't be accessing private properties anyways (properties that start with `_`).  The only use case I can think of would be that the parent wants to wait to show the table until all of these activity names are fetched and displayed.  I expect that'd actually be a perf hit, and the better way forward would be to show the table and then have some kind of "loading skeleton" for this component.  But I see that being a later design decision, when we're free of the ~14~ 13 day rush

test/d2l-activity-name/d2l-activity-name.html
svanherk: Also, instead of the `setTimeout`s, I'd probably copy and use the function Carol created for the list tests: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js#L6.  Instead of setting an href (https://github.com/BrightspaceHypermediaComponents/activities/pull/43/files#diff-b58077cbff970f3af3e93542b7e9034eR17), you set it to nothing and then you call the `_loadData` function and wait on it, then execute the tests: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js#L178

That way you don't have to guess at a time, and watch it fail on slow IE11

components/d2l-activity-name/d2l-activity-name.js
jeroach: Cool. Will take it out for simplicity - we can always add it back later if we decide to use it in the parent.

test/d2l-activity-name/d2l-activity-name.html
jeroach: Haha cool, will do - oh IE11 - will you just do everyone a favour and die 😛

components/d2l-activity-name/d2l-activity-name.js
jeroach: Hmmm, I'm honestly not sure. I replicated the behavior from here: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js#L170

If it's a problem here, then it's likely an issue there as well.

Or we're all good :)

jeroach:
Addressed comments (except Miles') in main component, test refactoring is up next

components/d2l-activity-name/d2l-activity-name.js
mtjandra: Ahh, because there it is using try/catch around `await`, this one is a bit different I think. But would have to double check.

dbatiste: This is a great article on async/await: https://javascript.info/async-await

jeroach:
Tests still broken but curious if anything looks glaringly wrong.

components/d2l-activity-name/d2l-activity-name.js
jeroach: Updated the method to be `async` and the usage to use `await` - this is consistent with `d2l-evaluation-hub-activities-list.js`

Good catch! I missed that originally.

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,100 @@
...
+			},
+			_activityIcon: {
+				type: String,
+				value: ''
+			}
+		};
+	}
+	static get observers() {
+		return [
+			'_loadData(entity, href)'
```
mtjandra: `href` is being passed in, but I think it will be ignored since the function definition is `_loadData(entity)`. I'm not certain but don't you only need to observe for `entity`. Or is it some weird web-component thing where this component wouldn't notice entity changing, so you need to observe for href

jeroach:
😀

components/d2l-activity-name/d2l-activity-name.js
jeroach: I think there is some weird magic going on given:
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js#L126

and
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js#L163

Perhaps @CarolBusman or @svanherk could enlighten us :)

svanherk: Looking at how the store works, it's watching for changes to the `href` and `token` for us (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L48), so we probably don't need to watch for `href`.  If it did change, the call might actually run twice, so best to remove it here and in the list.

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -0,0 +1,100 @@
...
+class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior], PolymerElement) {
+	static get template() {
+		return html`
+			<style>
+				:host {
+					display: block;
+					font-weight: normal;
+				}
+			</style>
+			<d2l-icon icon="[[_activityIcon]]"></d2l-icon>
```
svanherk: We may want to talk to design about adding some padding between the icon and the text

jeroach: Sounds good - I'll reach out to Erin, but not going to hold up this PR on waiting to hear back.

PR #44, opened 2019-02-22
* US103318 -  Supply and fetch master teacher data

mtjandra:
image

* The new column shows up if the correct flags are on, I testing in LMS (see image above)

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -136,7 +136,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 	}
 	constructor() { super(); }
 
-	__fetch(url) {
+	_myEntityStoreFetch(url) {
 		return window.D2L.Siren.EntityStore.fetch(url, this.token);
 	}
 
@@ -214,7 +214,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 
 	_followHref(href) {
 		if (href) {
-			return this.__fetch(href);
+			return this._myEntityStoreFetch(href);
 		}
 		return Promise.resolve();
 	}
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -235,7 +235,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				var getUserName = this._getUserPromise(activity, item);
 				var getCourseName = this._getCoursePromise(activity, item);
 				var getActivityName = this._getActivityPromise(activity, item);
-				var getMasterTeacherName = this._getMasterTeacherPromise(activity, item);
+				var getMasterTeacherName =
+					this._shouldDisplayColumn('masterTeacher')
+					? this._getMasterTeacherPromise(activity, item)
+					: Promise.resolve();
 
 				Promise.all([getUserName, getCourseName, getActivityName, getMasterTeacherName]).then(function() {
 					resolve(item);
@@ -395,7 +398,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 
 	_shouldDisplayColumn(columnKey) {
 		if (columnKey.includes('masterTeacher')) {
-			return this['masterTeacher'];
+			return this.masterTeacher;
 		}
 		return true;
 	}
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -15,7 +15,7 @@ import {Rels, Classes} from 'd2l-hypermedia-constants';
  * @polymer
  */
 
-class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior, D2L.PolymerBehaviors.Siren.SirenActionBehaviorImpl], EvaluationHubLocalize(PolymerElement)) {
+class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior, D2L.PolymerBehaviors.Siren.SirenActionBehavior ], EvaluationHubLocalize(PolymerElement)) {
 	static get template() {
 		return html`
 			<style include="d2l-table-style">
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -10,13 +10,12 @@ import 'd2l-polymer-behaviors/d2l-dom-focus.js';
 import 'd2l-link/d2l-link.js';
 import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
 import {Rels, Classes} from 'd2l-hypermedia-constants';
-
 /**
  * @customElement
  * @polymer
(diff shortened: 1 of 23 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -137,7 +136,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 	}
 	constructor() { super(); }
 
-	_fetch(url) {
+	__fetch(url) {
```
mtjandra: Updating method name to prevent conflict, see https://github.com/BrightspaceHypermediaComponents/activities/pull/43#discussion_r259395050

demo/d2l-evaluation-hub/data/activities/activitiesList/activities/enrollments/filters/filters.json
```diff
@@ -0,0 +1,37 @@
...
+        "item",
+        "https://api.brightspace.com/rels/filters/collection"
+      ],
+      "title": "Role Markers",
+      "href": "data/activities/activitiesList/activities/enrollments/filters/roleMarkerFilter.json"
+    }
+  ],
+  "actions": [
+    {
+      "href": "components/d2l-activities/demo/d2l-evaluation-hub/data/activities/activitiesList/activities/enrollments/course.json",
```
mtjandra: Notice these are specific for the `d2l-evaluation-hub` demo only, don't think there is anything I can do about it. Rubrics do not mock out this functionality either. Reason we can't do anything about this is because in https://github.com/Brightspace/polymer-siren-behaviors/blob/bed996403a40dbb92675224df4fa2dc25e58ba5f/store/siren-action-behavior.js#L45 it does 
```js
var url = new URL(action.href, window.location.origin);
``` 
and as far as I can tell I cannot mock out the window object.

demo/d2l-evaluation-hub/d2l-evaluation-hub.html
```diff
@@ -39,7 +39,7 @@
 			<h3>Basic d2l-evaluation-hub demo</h3>
 			<demo-snippet>
 				<template strip-whitespace>
-					<d2l-evaluation-hub href="data/activities/activitiesList/unassessedActivities.json" token="whatever"></d2l-evaluation-hub>
+					<d2l-evaluation-hub master-teacher href="data/activities/activitiesList/unassessedActivities.json" token="whatever"></d2l-evaluation-hub>
```
mtjandra: Decided to have the `master-teacher` column show up in datahub demo by default so it isn't something neglected. If needed to view without this column you can use the `d2l-evaluation-hub-activities-list` demo or just remove during testing

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -10,13 +10,12 @@ import 'd2l-polymer-behaviors/d2l-dom-focus.js';
...
 import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
 import {Rels, Classes} from 'd2l-hypermedia-constants';
-
 /**
  * @customElement
  * @polymer
  */
 
-class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior], EvaluationHubLocalize(PolymerElement)) {
+class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior, D2L.PolymerBehaviors.Siren.SirenActionBehaviorImpl], EvaluationHubLocalize(PolymerElement)) {
```
svanherk: This should just be `D2L.PolymerBehaviors.Siren.SirenActionBehavior` as convention, or in case we combine other behaviors later (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js#L191)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -250,6 +254,80 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+							}
+						];
+
+						return this.performSirenAction(action, fields);
+					}
+				}
+			}.bind(this))
+			.then(function(filterOptions) {
+				if (filterOptions) {
+					var masterTeacherOption = filterOptions.getSubEntityByRel('https://api.brightspace.com/rels/filter');
```
svanherk: Should we add this to the constants file?  What about `role-markers` and `search`?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -250,6 +254,80 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+			}.bind(this))
+			.then(function(enrollment) {
+				if (enrollment && enrollment.hasSubEntityByRel(Rels.userEnrollment)) {
+					var userEnrollment = enrollment.getSubEntityByRel(Rels.userEnrollment);
+					if (userEnrollment.href) {
+						return this._followHref(userEnrollment.href);
+					}
+				}
+			}.bind(this))
+			.then(function(userEnrollment) {
```
svanherk: Could we use the `_getUserPromise` function at the end here?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -250,6 +254,80 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		this._data = this._data.concat(result);
 	}
 
+	_getMasterTeacherPromise(entity, item) {
```
svanherk: As discussed offline, I feel like the amount of logic here makes this better off as its own component.  But I could plan to do this as part of the user component work as well

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
dbatiste: I think in the past I've used `__` (double) in the behaviors since components would typically use `_` (single).  That seems to be a bigger change though.  Alternatively there's the option of choosing a different name.  I'm not 100% sure what the best practice is but avoiding single underscore in behaviors seems to make sense.  Of course, there's always the risk that two behavior could collide.  All the more reason to avoid behaviors when they aren't needed.

m6jones: If a behaviour or a mixin has a private method use the `__` (double)
protected methods in a behaviour/mixins should use `_` (single)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Yeah I will need to add it in at some point, I'll put you on the PR when I update

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Since I don't have much context with the `SirenActionBehavior` I'm not sure if it should be considered private or protected. As a middle ground I will rename this method to `_myEntityStoreFetch`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Since the methods are doing mutation (I wish it actually gave a value instead) it doesn't work that great. Because it will assign the `item.displayName` not `item.masterTeacher`

svanherk: Ah, I see what you mean.  This could also be solved going the component route (since it would just assign its own internal `text` value), so I'm fine with leaving for now and fixing as part of that move later.

PR #45, opened 2019-02-26
Refactor demo data to be generated by JS to allow easier data creation for sorting/filtering

AlexBedley:
**Problem:** ~evaluation hub~ quick eval has many possible states due to sorting, filtering, paging, etc.
**Problem:** This means for demoing/testing we would need to create static JSON files for each permutation to be served by `polymer serve`
**Problem:** This large amount of files will quickly become unmaintainable

**Solution:** Instead let's intercept requests by overriding `window.d2lfetch.fetch` and dynamically create and return demo data. This will make it substantially easier to add sorting/filtering demos, and hopefully just make it easier to create demo data in general for quick eval.


demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -15,7 +16,19 @@
 			import '@polymer/iron-demo-helpers/demo-pages-shared-styles';
 			import '@polymer/iron-demo-helpers/demo-snippet';
 			import 'd2l-typography/d2l-typography.js';
-			import '../../components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js';
```
AlexBedley: Oops I will undo this change. Copy-paste error

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -0,0 +1,241 @@
...
+		mappings[getHrefForUserId(i)] = formatUser(user);
+	});
+	activityNames.forEach((activityName, i) => {
+		mappings[getHrefForActivityNameId(i)] = formatActivityName(activityName);
+	});
+	courses.forEach((course, i) => {
+		mappings[getHrefForCourseId(i)] = formatCourse(course);
+	});
+
+	const pagedActivities = chunk(activities.map(a => formatActivity(a)), 3);
```
AlexBedley: Eventually `pageSize` should probably respect the query param, but for now this is just as good as it was before

AlexBedley: This is also where sorting and filtering will hook in to do their respective activities

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -0,0 +1,241 @@
...
+	const pages = pagedActivities.length;
+
+	pagedActivities.forEach((page, i) => {
+		mappings[getHrefForPageId(i)] = formatPage(page, "filters/", "sorts/", getHrefForNextPage(i, pages))
+	});
+
+	return mappings;
+}
+
+function chunk(array, count) {
```
AlexBedley: I ripped this from underscore.js and forgot to come back to this. I should probably just import underscore js unless we feel strongly against it

demo/d2l-evaluation-hub/table.js
```diff
@@ -0,0 +1,59 @@
+export default
```
AlexBedley: I hate that this is a js file. ES6 doesn't let you import JSON files via `import` so I made it a js file as a workaround. I'll probably fix this up tomorrow.

demo/d2l-evaluation-hub/d2l-evaluation-hub.html
```diff
@@ -16,6 +17,18 @@
...
 			import 'd2l-typography/d2l-typography.js';
 			import '../../components/d2l-evaluation-hub/d2l-evaluation-hub.js';
+			import intercept from './interceptor';
+			import mappingsParser from './demo-data-parser'
+			import data from './table'
+
+			const mappings = mappingsParser(data);
+
+			const oldFetch = window.d2lfetch.fetch.bind(window.d2lfetch);
+			window.d2lfetch.fetch = function(input, options) {
```
AlexBedley: I am intercepting calls to `window.d2lfetch.fetch`. If `mappings` (which is an object where the keys are urls and the values are siren JSON) contains an entry for `input` then `mappings[input]` is returned. Otherwise we just pass it along (so demos could potentially be half dynamically generated and half static JSON files)

AlexBedley:
Posting this early to get some feedback. I plan on adding some tests (and cleaning up some things I noticed doing my own review) tomorrow.

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -15,7 +16,19 @@
...
+			import '../../components/d2l-evaluation-hub/d2l-evaluation-hub.js';
+			import intercept from './interceptor';
+			import mappingsParser from './demo-data-parser'
+			import data from './table'
+
+			const mappings = mappingsParser(data);
+
+			const oldFetch = window.d2lfetch.fetch.bind(window.d2lfetch);
+			window.d2lfetch.fetch = function(input, options) {
```
AlexBedley: Hmm I just thought of something.. you won't be able to see what's going on in the Network tab anymore... I could add some logging here to mimic what you would get with the Network tab

demo/d2l-evaluation-hub/table.js
AlexBedley: Also, I made this data in 5 minutes just for this PR but we should probably look more carefully at our demo data and make sure it covers a large portion of the application, and matches the real API as closely as possible

AlexBedley:
Wow I really need to run the linter locally haha

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
svanherk: That'd be helpful! Also, should we just throw this into a shared js file that all demos can include?

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -33,13 +46,16 @@
...
+
+
 	</head>
 	<body class="d2l-typography">
 		<div class="vertical-section-container fixedSize">
 			<h3>Basic d2l-evaluation-hub-activities-list demo</h3>
 			<demo-snippet>
 				<template strip-whitespace>
-					<d2l-evaluation-hub-activities-list href="data/activities/activitiesList/unassessedActivities.json" token="token"></d2l-evaluation-hub-activities-list>
+					<d2l-evaluation-hub href="pages/0" token="whatever"></d2l-evaluation-hub>
```
svanherk: Another copy/paste error, should still be `d2l-evaluation-hub-activities-list`

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -0,0 +1,242 @@
+function formatName(firstName, lastName) {
+	return firstName + ' ' + lastName;
+}
+
+function parseUsers(data) {
+	const names = data.map(row => formatName(row.firstName, row.lastName));
+	const uniqueNames = [... new Set(names)];
```
svanherk: Does this work in IE11?

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
AlexBedley: I can modularize this but there will still be a bit of necessary glue because this code needs to run in the same module as the first import of the component otherwise you'll miss the initial load.

demo/d2l-evaluation-hub/demo-data-parser.js
AlexBedley: Wait we don't transpile our es6 code to es5?

svanherk: We do in BSI, but there's no compiling here.  `polymer serve` handles hosting and I think some polyfills, I don't know fit hey handle this polyfill

svanherk: Actually, I may be wrong about this.  Sounds like `polymer serve` may do some transpiling at runtime.  I just checked the demo though and it's definitely broken, not sure if that's the cause though

AlexBedley: If BSI deals with es6 compatibility, do we have a requirement that `polymer serve` needs to work in IE11? Does anyone dev in IE11?

svanherk: I don't do the dev work in IE11, but we should be testing everything in IE11 and if there was a bug it's way easier to debug/fix it using the demo page rather than having to host your own BSI and rebuilding (about 10 minutes) after each change

AlexBedley: IE11 is broken on master as well. I'll address IE11 compatability in a future PR then

svanherk: Hmm interesting, I just fixed it a few weeks ago.  I can take a look later as well

demo/d2l-evaluation-hub/demo-data-parser.js
AlexBedley: Oh I see we actually already import all of lodash into this repo so I'll just use lodash.chunk

package.json
```diff
@@ -31,11 +32,10 @@
 		"gulp-ejs": "^3.1.3",
 		"gulp-rename": "^1.2.3",
 		"lie": "^3.3.0",
+		"lodash-es": "^4.17.11",
 		"polymer-cli": "^1.9.2",
 		"wct-browser-legacy": "^1.0.1",
-		"whatwg-fetch": "^3.0.0",
-		"async": "^1.5.2",
-		"lodash": "^3.10.1"
```
AlexBedley: :( this diff is because package.json was previously manually edited. We should avoid manually editing this file (by using the command line tool `npm`)

AlexBedley: (Also nobody was using lodash - it doesn't work anyways in es6 - so I changed it to use the es6 version)

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
AlexBedley: done

demo/d2l-evaluation-hub/demo-data-parser.js
AlexBedley: done

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
AlexBedley: done

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
AlexBedley: done

package.json
svanherk: So this was being used to avoid error messages in the tests I believe, and I think `wct` needed a specific version.  All our other repos don't include this and just have the errors messages in the tests (they don't cause failures), so you're probably fine removing

demo/d2l-evaluation-hub/table.js
AlexBedley: > I hate that this is a js file. ES6 doesn't let you import JSON files via `import` so I made it a js file as a workaround. I'll probably fix this up tomorrow.

Turns out this is more complicated than it's worth so I'm just gonna leave it for now

AlexBedley:
image
I added some logging to replace what was lost in the network tab. It looks like the above. You can turn it off by setting `debug` to `false` but I left it on by default.

demo/d2l-evaluation-hub/demo-data-parser.js
svanherk: https://github.com/BrightspaceHypermediaComponents/activities/pull/46

AlexBedley:
Ok I resolved the merge conflict with the master teacher stuff. It has limitations (for example if multiple teachers are listed as the master teacher it will change them all to the last one it sees) but it works for most normal cases

svanherk:
> Ok I resolved the merge conflict with the master teacher stuff. It has limitations (for example if multiple teachers are listed as the master teacher it will change them all to the last one it sees) but it works for most normal cases.

You raise an interesting point... what is _supposed_ to happen if there's multiple master teachers?  And do we currently do that?

AlexBedley:
> You raise an interesting point... what is _supposed_ to happen if there's multiple master teachers? And do we currently do that?

We don't define the behaviour. What actually happens is quick eval will just take "the first one" in the list which, if it isn't sorted, could be any of the master teachers.

svanherk:
> We don't define the behaviour. What actually happens is quick eval will just take "the first one" in the list which, if it isn't sorted, could be any of the master teachers.

That feels wrong... @mtjandra do you know what the expectation is?  Should we check for all of them and list them all in the same cell?

svanherk:
Whoops!

demo/d2l-evaluation-hub/interceptor.js
```diff
@@ -0,0 +1,51 @@
...
+	const oldFetch = window.d2lfetch.fetch.bind(window.d2lfetch);
+
+	if (debug) {
+		console.group('All Mappings');
+		console.log(mappings);
+		console.groupEnd();
+	}
+
+	window.d2lfetch.fetch = function(input, options) {
+		const path = parsePathNoLeadingSlash(input);
```
AlexBedley: The reason I had to do this is because the siren action helper things (ie `performSirenAction`) create absolute urls which ruins the interceptor thing. A better fix might be to make the mock server only respond to absolute urls but I vote we defer that until it becomes a problem (if ever).

AlexBedley:
> That feels wrong... @mtjandra do you know what the expectation is? Should we check for all of them and list them all in the same cell?

Spoke to Surekha offline, she confirmed flvs will only have 1 master teacher per course and any other data will cause undefined behaviour (but we won't blow up)

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -0,0 +1,427 @@
...
+	if (currentId + 1 < pages) {
+		return getHrefForPageId(currentId + 1);
+	}
+}
+
+function getHrefForMasterTeacher(id) {
+	return `masterTeacher/${id}`;
+}
+
+function getMappings(data) {
```
AlexBedley: To review this massive file (sorry!) I would just read this function.

`parseX` functions operate on data directly from the table
`formatX` functions return siren representations of X
`getHrefForX` function generate a url for X that we use in our HM web/the interceptor

`mappings` is the return value that the interceptor operates on and it maps urls to siren entities.

AlexBedley: Reading this function makes me really wish js had named parameters :(

demo/d2l-evaluation-hub/interceptor.js
svanherk: Yeah Miles ran into issues with this when creating the static demo data as well

demo/d2l-evaluation-hub/demo-data-parser.js
svanherk: So I think you should put that above this function in a comment:
```
`parseX` functions operate on data directly from the table
`formatX` functions return siren representations of X
`getHrefForX` functions generate a url for X

`mappings` is the return value that the interceptor operates on and it maps urls to siren endpoints.
```
Was super helpful, and useful if anyone needs to expand this later.

demo/d2l-evaluation-hub/demo-data-parser.js
svanherk: So I merged master into this branch but this PR still breaks IE11 😞

AlexBedley:
Let it be known that I super promise to add ie11 support as soon as I can but merging this now just to unblock everyone else from adding demo data

mtjandra:
> > That feels wrong... @mtjandra do you know what the expectation is? Should we check for all of them and list them all in the same cell?
> 
> Spoke to Surekha offline, she confirmed flvs will only have 1 master teacher per course and any other data will cause undefined behaviour (but we won't blow up)

Yes this is correct, for now expectation is that we take any of the listed master teachers if multiple defined.

PR #47, opened 2019-02-27
Add loading shimmers and placeholders.

qwangd2l:


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -515,6 +515,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 		event.preventDefault();
 	}
 	_clampDescription(description) {
+		if (!description) return;
```
qwangd2l: Not too sure why the tests before this PR did not fail without this?
Undefined error: references to this parameter within the fastdom.mutate: ```const avgCharPerLine = description.length / ((currentLineNumber - 1));```

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -266,6 +366,27 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
...
+			imageShimmer: {
+				type: Boolean,
+				value: false
+			},
+			textPlaceholder: {
+				type: Boolean,
+				value: false,
+				observer: '_onTextPlaceholderChange'
+			},
+			_textPlaceholder: {
```
qwangd2l: The purpose of a second variable here (this one being private) is so that all resizes can be done before removing the placeholders. This avoids a flickering effect (especially with the description clamping).
There's not a second variable for the image shimmer because the image and the image shimmer share the same box (as opposed to text vs boxes that simulate text area in the textPlaceholder case which means they have different css).

components/d2l-activity-list-item/d2l-activity-list-item.js
m6jones: I think it was  made such that this didn't run unless there was a description. So I missed this edge case.

PR #49, opened 2019-02-28
Send events when list item has loaded.

m6jones:


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -597,7 +604,13 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 				this._fetchEntity(imageEntity.href)
 					.then(function(hydratedImageEntity) {
 						this._image = hydratedImageEntity;
-					}.bind(this));
+					}.bind(this))
+					.finally(() => {
+						this.dispatchEvent(new CustomEvent('d2l-activity-image-loaded', {
```
qwangd2l: I think we need to wait for d2l-course-image to fire its load event before firing this.
Most likely this event? https://github.com/Brightspace/course-image/blob/master/d2l-course-image.js#L170

m6jones: Oh!! thanks!

qwangd2l:
Hmm I'm having some issues listening to the text-loaded event here. Any ideas?
`d2l-activity-image-loaded` seems to be fine.
```
ready() {
    this.addEventListener('d2l-activity-text-loaded', this._removeTextPlaceholders); // not getting the event
    this.addEventListener('d2l-activity-image-loaded', this._removeImageShimmers); // good
}
```

m6jones:
My only guess is that for some reason it isn't getting to the code that fires it. Since I believe they are the same.

Tests seem to work.

qwangd2l:
> My only guess is that for some reason it isn't getting to the code that fires it. Since I believe they are the same.
> 
> Tests seem to work.

Will investigate further :).

PR #50, opened 2019-02-28
Fire the d2l-activity-text-loaded event after next render.

qwangd2l:
This seems to fix the issue of firing the event too early before the element is stamped (I experienced this in a dom-repeat situation).
`d2l-course-image` does this as well.
Any issues that anyone can think of?

### DIFF
--- a/components/d2l-activity-list-item/d2l-activity-list-item.js
+++ b/components/d2l-activity-list-item/d2l-activity-list-item.js
@@ -217,7 +217,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 					margin-right: 0.5rem;
 				}
 			</style>
-			<div class="d2l-activity-list-item-container">
+			<div class="d2l-activity-list-item-container" style="visibility:hidden;">
 				<hr class="d2l-activity-list-item-top-line" />
 				<a class="d2l-focusable" href$="[[_activityHomepage]]">
 					<span class="d2l-activity-list-item-link-text">[[_accessibilityDataToString(_accessibilityData)]]</span>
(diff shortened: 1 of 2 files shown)


test/d2l-activity-list-item/d2l-activity-list-item-test.js
```diff
@@ -137,8 +137,11 @@ describe('d2l-activity-list-item', () => {
 				expect(component._activityHomepage).to.equal('#');
 			});
 
-			it('should send text loaded event', () => {
-				expect(textLoadedSuccessfulSpy.calledOnce).to.be.true;
+			it(testCase.name + 'should send text loaded event', done => {
```
m6jones: You might want to bring this out to the same scope as the image loaded event. So the beforeEach doesn't run.

components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -217,7 +217,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 					margin-right: 0.5rem;
 				}
 			</style>
-			<div class="d2l-activity-list-item-container">
+			<div class="d2l-activity-list-item-container" style="visibility:hidden;">
```
qwangd2l: Should I leave this change for another PR?
It's addressing an issue that causes the image box to display its default css (which I changed in an earlier PR to match 842px+ sizes so it doesn't flicker too much) which may be too big (on smaller view dimensions). This results in visible flickering.
So I thought we'd just hide the whole thing and show it after the initial resize.
I tried doing hidden attribute but that also causes a slight flicker (from nothing to something).

m6jones:
What did your last change do?

components/d2l-activity-list-item/d2l-activity-list-item.js
m6jones: Yeah, this seems like a great change!!

qwangd2l:
> What did your last change do?

The organization name also has a load time which is addressed by moving the firing of the event to the listener of `d2l-organization-accessible` (which is fired by `d2l-organization-name`).

The issue that the visibility css change addresses is in that other comment.

PR #59, opened 2019-03-04
Add sorting to UI

AlexBedley:
Looking for some early feedback on this implementation before I commit to writing tests/demo data


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -111,13 +111,13 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
 			},
-			_sortHref: {
-				type: String,
-				value: ''
-			},
 			_pageNextHref: {
 				type: String,
 				value: ''
+			},
+			_currentEntity: {
```
AlexBedley: I figured it made more sense to just save the entire entity than having a specific href for sorting/filtering and updating both of them every single time.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -88,11 +88,11 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			_headers: {
 				type: Array,
 				value: [
-					{ key: 'displayName', sortKey: 'displayName', localizationKey: 'displayName', canLink: true },
-					{ key: 'activityName', sortKey: 'activityName', localizationKey: 'activityName', canLink: false },
-					{ key: 'courseName', sortKey: 'courseName', localizationKey: 'courseName', canLink: false },
-					{ key: 'submissionDate', sortKey: 'submissionDate', localizationKey: 'submissionDate', canLink: false },
-					{ key: 'masterTeacher', sortKey: 'masterTeacher', localizationKey: 'masterTeacher', canLink: false }
+					{ key: 'displayName', sortKey: 'displayName', sortClass: 'displayName', canSort: false, localizationKey: 'displayName', canLink: true },
```
AlexBedley: `sortKey` is just an id
`sortClass` allows us to find the specific sort in the lists of sorts
`canSort` is whether we can find a specific sort for this header (or not)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -140,24 +140,102 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		return window.D2L.Siren.EntityStore.fetch(url, this.token);
 	}
 
+	_loadSorts(entity) {
```
AlexBedley: This function runs on load and makes sure we have sorts for each column by setting `canSort` accordingly

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -140,24 +140,102 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+						console.log('found', header.sortClass);
+						header.canSort = true;
+					}
+				});
+				return Promise.resolve();
+			});
+	}
+
 	_sort(e) {
+		console.log(e.currentTarget.id);
```
AlexBedley: This function runs on click and does a ton of things.

1. it makes sure sorting is enabled for the column
2. it changes the header arrow things
3. it loads `/sorts`, finds the one we want, applies it in the correct direction, and navigates back to the collections route and reloads the data

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -171,8 +249,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		try {
 			var result = await this._parseActivities(entity);
 			this._data = result;
+			return this._loadSorts(entity);
```
AlexBedley: Made these changes so we actually log errors instead of swallowing them

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: Maybe `sortKey` should go away and we just have a universal id for headers

svanherk: If all of these are going to be the same, I think we can just make this an array of keys.  Josh is soon removing the "canLink" logic as well in his PR.  Instead of `canSort`, we could have a function that you pass the key into and it will tell you if you can sort or not, based on the info gathered on load.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: We already do save the current entity, you can get it by going `this.entity`.  We get it because we import and mixin this behaviour: https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L34

AlexBedley: awesome, thanks for the teaching moment! :)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Does it currently run on load, or you want it to?

svanherk: Nvm, I see what you mean.  In the load function

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -140,24 +140,102 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+					return Promise.reject('Could not load sorts endpoint after sort is applied');
+				}
+				const action = sortsEntity.entity.getActionByName('apply');
+				if (!action) {
+					return Promise.reject(`Could not find apply action in ${sortsEntity}`);
+				}
+				return action;
+			})
+			.then(collectionAction => {
+				this._loading = true;
```
svanherk: I think this should just call `_loadData`, it handles stuff like full table loading and setting the data for you.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -140,24 +140,102 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+			if (h !== e.currentTarget) {
+				h.removeAttribute('desc');
+				h.setAttribute('nosort', 'nosort');
 			}
-		}
+		});
+
+		const performSirenAction = this.performSirenAction;
+
+		return this._getHref(this._currentEntity, Rels.filters)
```
svanherk: I think we may already have this href in `_filterHref`

AlexBedley: Whoops this is a bug. It should be `Rels.sorts`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: I was treating `_loadData` as the initial load, as-in things we only want to happen once will be put in `_loadData`. There is probably some value in creating a function which just reloads the page though that can be used by `_loadData`, `_loadMore`, sorting and eventually filtering

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: Hmm not sure I understand this. `canSort` is "the info gathered on load" and since I only check canSort once I didn't write a helper function for it

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: does `this.entity` ever get updated? Is it kosher to manually update it? Won't that cause a total page reload? (I don't really know how `observers` work)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: `_loadData` is the initial load, but also a table reload.  Those look the exact same to us, since we want the table to hide and any errors to trigger a full page error

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -43,7 +43,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 						<dom-repeat items="[[_headers]]">
 							<template>
 								<template is="dom-if" if="[[_shouldDisplayColumn(item.key)]]">
-									<d2l-th><d2l-table-col-sort-button nosort on-click="_sort"><span>[[localize(item.localizationKey)]]</span></d2l-table-col-sort-button></d2l-th>
+									<d2l-th><d2l-table-col-sort-button nosort on-click="_sort" id="[[item.sortKey]]"><span>[[localize(item.localizationKey)]]</span></d2l-table-col-sort-button></d2l-th>
```
mtjandra: If I recall correctly, we need to be able to sort by First Name and Last Name separately, but right now there is only one button to handle sort by display name. Maybe something to think about in a later PR

AlexBedley: Purposefully deferring to future PR

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -148,24 +144,90 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+			});
+	}
+
 	_sort(e) {
+		const header = this._headers.find(h => h.key === e.currentTarget.id);
+
+		if (!header) {
+			return Promise.reject(`No matching header for ${e.currentTarget.id}`);
+		}
+		if (!header.canSort) {
```
svanherk: I don't think we want to reject if they try to sort a header that can't sort?  We actually need to change the header renderer above to not include the column button if the column cannot be sorted (https://github.com/BrightspaceHypermediaComponents/activities/pull/59/files#diff-c36f46a8beb24cb46d7a213237cccab8L47)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -148,24 +144,90 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+		headers.forEach(h => {
+			if (h !== e.currentTarget) {
+				h.removeAttribute('desc');
+				h.setAttribute('nosort', 'nosort');
 			}
-		}
+		});
+
+		return this._followLink(this.entity, Rels.sorts)
+			.then((sortsEntity => {
```
svanherk: Not urgent, but this file now has a mix of two different styles - `async + `await`, and `Promise` + `.then`. Do we want to pick one to make things more consistent adn readable?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: I think we should remove `loadSorts` here and add it to the `observer`: https://github.com/BrightspaceHypermediaComponents/activities/pull/59/files#diff-c36f46a8beb24cb46d7a213237cccab8L134

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -259,7 +323,6 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		}.bind(this));
 
 		this._filterHref = this._getHref(entity, Rels.filters);
```
svanherk: If you don't do here it here I'll do it in my PR, but this line and the property can also go, it won't be used here.

svanherk:
Approval based on the fact that tests are coming immediately

AlexBedley:
I hereby promise to add tests first thing tomorrow morning :) and then after that I'm gonna work on Stacey's filter demo data.

Merging it now to unblock other devs and prevent a massive merge conflict tomorrow morning.

PR #60, opened 2019-03-04
Hook up filter

svanherk:
This is based off a branch Carol has and still needs  demo updates.  I can't test the integration with sorting until either filter demo data is added, or the sort keys merge is in the LE.  Soon!


package.json
```diff
@@ -51,6 +51,7 @@
 		"@polymer/polymer": "^3.0.0",
 		"d2l-button": "BrightspaceUI/button#semver:^5",
 		"d2l-colors": "BrightspaceUI/colors#semver:^4",
+		"d2l-common": "BrightspaceHypermediaComponents/common#cbusman/d2l-hm-filter",
```
svanherk: This needs to change to master before we can merge this in.

components/d2l-evaluation-hub/d2l-evaluation-hub.js
```diff
@@ -14,23 +16,70 @@ class D2LEvaluationHub extends mixinBehaviors([D2L.PolymerBehaviors.Siren.Entity
...
+
+	_getHref(entity, rel) {
+		if (entity && entity.hasLinkByRel && entity.hasLinkByRel(rel)) {
+			return entity.getLinkByRel(rel).href;
+		}
+		return '';
+	}
+
+	_getFilterClasses(masterTeacher) {
+		var filters = [ 'activity-name', 'enrollments', 'completion-date' ];
```
AlexBedley: `const` 💪

components/d2l-evaluation-hub/d2l-evaluation-hub.js
```diff
@@ -14,23 +16,70 @@ class D2LEvaluationHub extends mixinBehaviors([D2L.PolymerBehaviors.Siren.Entity
...
-			'masterTeacher': {
+			masterTeacher: {
 				type: Boolean,
 				value: false,
 				reflectToAttribute: true
-			}
+			},
+			_filterHref: {
+				type: String,
+				computed: '_getFilterHref(entity)'
```
AlexBedley: when are `computed` things called? if `this.entity` is updated, is this called?

AlexBedley: This needs to be updated when `list.entity` is updated

components/d2l-evaluation-hub/d2l-evaluation-hub.js
```diff
@@ -14,23 +16,70 @@ class D2LEvaluationHub extends mixinBehaviors([D2L.PolymerBehaviors.Siren.Entity
...
+
+	_getFilterHref(entity) {
+		return this._getHref(entity, Rels.filters);
+	}
+
+	_getHref(entity, rel) {
+		if (entity && entity.hasLinkByRel && entity.hasLinkByRel(rel)) {
+			return entity.getLinkByRel(rel).href;
+		}
+		return '';
```
AlexBedley: We're fully expecting `Rels.filters` to be present on `entity`. if it isn't we need to display some sort of error to the user and like hide the filter control (or something). Presumably something will blow up downstream on an empty href but we do know it won't work at this point if we wanted to error out here.

svanherk: Yup, filter error state is next on my list with the filter empty state.  Basically, "handle all the filter scenarios"

components/d2l-evaluation-hub/d2l-evaluation-hub.js
svanherk: > when are computed things called? if this.entity is updated, is this called?
 Yup!  Very similar to the `listeners`

> This needs to be updated when list.entity is updated

Working on this now, will be part of this PR

components/d2l-evaluation-hub/d2l-evaluation-hub.js
AlexBedley: Ok, fine to defer to future PR then

components/d2l-evaluation-hub/d2l-evaluation-hub.js
svanherk: We add to this if we find `master-teacher` is on

AlexBedley: lol epic fail. `let`!

PR #64, opened 2019-03-05
Render columns individual

mtjandra:
* Rendering column individually to allow me to place the draft icon in more easily. Josh was doing something similar as well for his activity name component.

### DIFF
--- a/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -40,13 +40,10 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
 			for (var i = 0; i < expected.length; i++) {
 				var link = data[i].querySelector('d2l-link');
 				var span = data[i].querySelector('span');
-				var linkHidden = link.hasAttribute('hidden');
-				var spanHidden = span.hasAttribute('hidden');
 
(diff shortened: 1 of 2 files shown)


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -40,13 +40,10 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
-				var linkHidden = link.hasAttribute('hidden');
-				var spanHidden = span.hasAttribute('hidden');
 
-				assert.equal(expected[i].text, link.innerHTML);
-				assert.equal(expected[i].text, span.innerHTML);
-				assert.equal(!(expected[i].link), linkHidden);
-				assert.equal(!!(expected[i].link), spanHidden);
+				var item = link !== null ? link : span;
+
+				assert.equal(expected[i].text, item.innerHTML);
```
svanherk: This works for now, once we start using web components it won't.  I eventually see us just grabbing the `d2l-td` and checking the `innerHTML` of that using `assert.includes`

mtjandra: Yeah once we use components, you will not check innerHtml, since the component would handler that

mtjandra:
@jeroach let me know if this is beneficial to go while you work on demo data (unsure if that is what you are doing). It will help unblock my future work, but I can work around it. Your call!

PR #66, opened 2019-03-05
Us103573/draft icon eval

mtjandra:
image

* Adding draft icon to eval-hub
* May iterate on positioning/padding of the icon placement in future PR if needed

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -60,9 +60,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 							<d2l-tr>
 								<d2l-td>
 									<d2l-link href="[[s.activityLink]]">[[_getDataProperty(s, 'displayName')]]</d2l-link>
-									<template is="dom-if" if="[[s.isDraft]]">
-										<d2l-activity-evaluation-icon-base draft></d2l-activity-evaluation-icon-base>
-									</template>
+									<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
 								</d2l-td>
 								<d2l-td>
 									<d2l-activity-name href="[[_getDataProperty(s, 'activityNameHref')]]" token="[[token]]"></d2l-activity-name>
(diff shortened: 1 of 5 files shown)


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -8,42 +8,56 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
-		return expected;
+		return expectedActivities;
 	}
 
 	function createExpectedDataWithMasterTeacher(expectedArray) {
 		return createExpectedData(expectedArray, true);
 	}
 
-	function verifyData(expected, done) {
+	function verifyData(expectedActivities, done) {
```
mtjandra: I think we will want to re-visit this verification sometime in the future

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -56,6 +57,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 							<d2l-tr>
 								<d2l-td>
 									<d2l-link href="[[s.activityLink]]">[[_getDataProperty(s, 'displayName')]]</d2l-link>
+									<template is="dom-if" if="[[s.isDraft]]">
```
svanherk: Instead of wrapping this in a `dom-if`, could we just hook up the value straight to the `draft` attribute?
```
<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
```

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -266,6 +272,17 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		return result;
 	}
 
+	_determineIfActivityIsDraft(activity) {
+		if (activity.hasSubEntityByRel('https://api.brightspace.com/rels/evaluation')) {
```
svanherk: This should be added to the constants file

mtjandra: I'll make a PR on that repo

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: PR made https://github.com/Brightspace/d2l-hypermedia-constants/pull/102

svanherk: `dom-if` is a Polymer thing, so we try to avoid it if we can and it won't be needed with `Lit`.  The preference is `hidden`, which I believe you did use under the hood in the base component, and then here we can just set things directly.

mtjandra:
Okay I think this is good to go, tested within the LMS with firefox, chrome and ie11 seems to be working

PR #69, opened 2019-03-06
Get the description from the organization HM instead of the activity

m6jones:


m6jones:
I have to strip the HTML so I added that.

PR #73, opened 2019-03-06
US104018 - Show alert when failing to load more data

mtjandra:
alert_load_more

* Alert when failing to load data
* In future PR may add tryAgain functionality to the alert to try again from that dialog as well

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -29,6 +29,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				d2l-loading-spinner {
 					width: 100%;
 				}
+				d2l-alert {
+					margin: auto;
+				}
 				.d2l-evaluation-hub-activities-list-load-more-container {
 					padding-top: 1rem;
 					text-align: right;
@@ -38,9 +41,6 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					display: none;
 				}
 			</style>
-			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]" has-close-button>
-				[[localize(_health.errorMessage)]]
-			</d2l-alert>
 			<d2l-table hidden$="[[_fullListLoading]]" aria-colcount$="[[_headers.length]]" aria-rowcount$="[[_data.length]]">
 				<d2l-thead>
 					<d2l-tr>
@@ -85,6 +85,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					</dom-repeat>
 				</d2l-tbody>
 			</d2l-table>
+			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]" has-close-button>
+				[[localize(_health.errorMessage)]]
+			</d2l-alert>
 			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!_loading]]">[[localize('loading')]]</d2l-offscreen>
 			<d2l-loading-spinner size="80" hidden$="[[!_loading]]"></d2l-loading-spinner>
 			<template is="dom-if" if="[[_pageNextHref]]">
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -1,5 +1,6 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {EvaluationHubLocalize} from './EvaluationHubLocalize.js';
+import 'd2l-alert/d2l-alert.js';
 import 'd2l-table/d2l-table.js';
 import 'd2l-button/d2l-button.js';
 import 'd2l-loading-spinner/d2l-loading-spinner.js';
@@ -37,6 +38,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					display: none;
 				}
 			</style>
+			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]" has-close-button>
+				[[localize(_health.errorMessage)]]
+			</d2l-alert>
 			<d2l-table hidden$="[[_fullListLoading]]" aria-colcount$="[[_headers.length]]" aria-rowcount$="[[_data.length]]">
 				<d2l-thead>
 					<d2l-tr>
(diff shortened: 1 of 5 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -260,28 +272,40 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 	_loadMore() {
 		if (this._pageNextHref && !this._loading) {
 			this._loading = true;
-			this._followHref(this._pageNextHref).then(async function(u) {
```
mtjandra: Hide whitespace

svanherk:
I'd expect that alert would go below the table, centered.  And is the gif just to show what it looks like, it wouldn't actually show up when it does load more?

demo/d2l-evaluation-hub/page-handler.js
```diff
@@ -27,7 +27,9 @@ function addSortToHref(href, sortState) {
 	}
 }
 
-function createPageEndpoint(activities, sorts, pageNumber, filtersHref, sortsHref, nextPageHref) {
+function createPageEndpoint(activities, sorts, pageNumber, filtersHref, sortsHref, nextPageHref, failFirstTime) {
+	var shouldFail = failFirstTime;
```
AlexBedley: nit: `let`

mtjandra:
> I'd expect that alert would go below the table, centered. And is the gif just to show what it looks like, it wouldn't actually show up when it does load more?

That is not what I expected. I thought it would go on top, left justified. In demo page, the second time you load more we simulate failure and you get the alert. After clicking again it resolves itself so the alert is cleared.

Spec on where alert should go is not specified in AC, so unknown I can ask Erin

svanherk:
> In demo page, the second time you load more we simulate failure and you get the alert. After clicking again it resolves itself so the alert is cleared.

Ah ok, I didn't see the double-click.  Makes more sense!

>Spec on where alert should go is not specified in AC, so unknown I can ask Erin

Yeah I'd ask Erin, I think this is her last day before she goes on vacation.  I don't think I've ever seen this placement for an alert throughout our system

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -260,28 +272,40 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+							window.requestAnimationFrame(function() {
+								var newElementToFocus = D2L.Dom.Focus.getNextFocusable(lastFocusableTableElement, false);
+								newElementToFocus.focus();
+							});
+						}
 					}
-				}
-			}.bind(this));
+				}.bind(this))
+				.then(this._clearAlerts.bind(this))
```
AlexBedley: How come this is necessary? I would assume dismissing the error dialog would make the app "healthy" again (and cause the app to reload)

mtjandra:
> > In demo page, the second time you load more we simulate failure and you get the alert. After clicking again it resolves itself so the alert is cleared.
> 
> Ah ok, I didn't see the double-click. Makes more sense!
> 
> > Spec on where alert should go is not specified in AC, so unknown I can ask Erin
> 
> Yeah I'd ask Erin, I think this is her last day before she goes on vacation. I don't think I've ever seen this placement for an alert throughout our system

Nvm it should show up where submissions should be loaded, I will move it to the bottom. But will re-check for more info

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -258,5 +258,20 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
 				MockInteractions.tap(loadMore);
 			});
 		});
+		test('when handling load more failure, alert should pop up and alert should hide when alerts cleared', (done) => {
+			list._handleLoadMoreFailure();
+
+			flush(function() {
+				var alert = list.shadowRoot.querySelector('#list-alert');
```
AlexBedley: nit: `const`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: oh I get it. my incorrect assumption was the error dialog would make you unable to use the app until you dismissed it.

AlexBedley:
In general: should we just put this health thing at the "top level" (I think the app has 4 entry points: _loadData, _loadSorts, _sort and _loadMore) as a catch-all for any error?

mtjandra:
> In general: should we just put this health thing at the "top level" (I think the app has 4 entry points: _loadData, _loadSorts, _sort and _loadMore) as a catch-all for any error?

Can you expand what you mean when you say "top level".

For more context:
* `_loadMore`: is only run when the user clicks "Load More", and has its own error message. Will maybe introduce a "Try Again" button in the alert, but less likely than `_loadData`
* `_loadSorts` / `_sort`: will not generate an alert (not part of AC)
* `_loadData`: is ran once ideally (unless you update href). It has its own error message. Will maybe introduce a "Try Again" button in the alert.

svanherk:
If we aren't going to consider filters/sorts as part of this story, then we need another story

mtjandra:
> If we aren't going to consider filters/sorts as part of this story, then we need another story

Spoke to PM. We will get reqs on if it is needed from Erin. Right now it isn't something we know we want yet.

mtjandra:
Alright updated the alert to show up at the bottom, and centered. Updated the GIF as well

AlexBedley:
> * `_loadSorts` / `_sort`: will not generate an alert (not part of AC)

IMO `_loadSorts` should fail silently but `_sort` should display some kind of error to the user since their list won't be sorted

> * `_loadData`: is ran once ideally 

This is not true, sorting and filtering will call this function to reload the page

mtjandra:
> IMO _loadSorts should fail silently but _sort should display some kind of error to the user since their list won't be sorted

Yeah spoke to PM about this. Out of scope for now

> This is not true, sorting and filtering will call this function to reload the page

Yeah good catch! I will keep this in mind when handling failure for loadData

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -81,6 +86,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					</dom-repeat>
 				</d2l-tbody>
 			</d2l-table>
+			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]" has-close-button>
```
svanherk: Minor: Our id/class naming convention tends to append the component name for safety (`d2l-evaluation-hub-activities-list-alert`)

PR #74, opened 2019-03-06
Add proper spacing as per design; add nowrap to component

jeroach:
Screenshot of demo. Resizing works too - feel free to checkout locally and test demo page out.

nowrap_demo

### DIFF
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -16,7 +16,12 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 			<style>
 				:host {
 					display: block;
+					padding-right: 20px;
 					white-space: nowrap;
+				}
+				:host(:dir(rtl)) {
+					padding-right: 0px;
+					padding-left: 20px;
 				}
 				.d2l-activity-name-icon {
 					display: inline-block;
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -16,11 +16,23 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 			<style>
 				:host {
 					display: block;
-					font-weight: normal;
+					white-space: nowrap;
+				}
+				.d2l-activity-name-icon {
+					display: inline-block;
+					padding-right: 12px;
+					vertical-align: middle;
+				}
+				.d2l-activity-name-text {
+					display: inline-block;
+					overflow: hidden;
+					text-overflow: ellipsis;
+					vertical-align: middle;
+					max-width: 100%;
 				}
 			</style>
-			<d2l-icon icon="[[_activityIcon]]"></d2l-icon>
-			<span>[[_activityName]]</span>
+			<d2l-icon class="d2l-activity-name-icon" icon="[[_activityIcon]]"></d2l-icon>
+			<div class="d2l-activity-name-text">[[_activityName]]</div>
 		`;
 		activityNameTemplate.setAttribute('strip-whitespace', 'strip-whitespace');
 		return activityNameTemplate;
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -23,7 +23,7 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 					padding-right: 12px;
 					vertical-align: middle;
 				}
-				[dir="rtl"] .d2l-activity-name-icon {
+				:host(:dir(rtl)) .d2l-activity-name-icon {
 					padding-right: 0px;
 					padding-left: 12px;
 				}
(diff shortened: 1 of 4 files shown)


jeroach:
I still need the RTL support, but that can wait until tomorrow :)

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -16,11 +16,27 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
...
+					display: inline-block;
+					overflow: hidden;
+					text-overflow: ellipsis;
+					vertical-align: middle;
+					max-width: 100%;
 				}
 			</style>
-			<d2l-icon icon="[[_activityIcon]]"></d2l-icon>
-			<span>[[_activityName]]</span>
+			<d2l-icon class="d2l-activity-name-icon" icon="[[_activityIcon]]"></d2l-icon>
```
jeroach: I would assume that for the RTL case, these should be flipped. However, in looking at examples in our code base, this does not seem to be the case.

Any thoughts/suggestions? Thanks!

svanherk: The browser handles a lot of RTL for us, including flipping the order of elements

svanherk: For RTL, a few good things to watch for (that the browser doesn't handle) are:
1) Did I set padding or margins to only one side?
2) Did I `float` left or right?

There's definitely more, but those come up a lot

jeroach: Cool thanks!

jeroach: I'm pretty sure I've handled the padding, and I'm not making use of any `float`s

jeroach:
The `text-overflow: ellipsis;` is not playing nice with RTL; this is a common problem judging from my Google Fu. See image.

Working on this now.
text-overflow-ellipsis-rtl-bad

svanherk:
> The text-overflow: ellipsis; is not playing nice with RTL; this is a common problem judging from my Google Fu. See image.

What's wrong with it?  Browsers tend to move punctuation (ellipsis, question mark, etc) to the right for us.

jeroach:
You're right :)
https://d2l.slack.com/archives/C0PHG3QB0/p1551969105025300

jeroach:
Okay, so here's how to test the RTL:

Run the demo, and then change the URL to include: `?dir=rtl`

Like so: http://127.0.0.1:8081/components/d2l-activities/demo/d2l-activity-name/d2l-activity-name.html?dir=rtl

jeroach:
Thanks Carol!

I have a meeting with Daniel at 2 PM. Planning to merge this after if he's good with the approach.

Next step will likely be to implement the no wrap and column widths for the other table columns (if we decide to go that way).

fyi @svanherk @CarolBusman @AlexBedley

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -16,11 +16,32 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 			<style>
 				:host {
 					display: block;
-					font-weight: normal;
+					padding-right: 20px;
+					white-space: nowrap;
+				}
+				:host(:dir(rtl)) {
+					padding-right: 0px;
```
svanherk: Nit: If the value is `0` we usually remove the units.

AlexBedley:
I checked out your branch and noticed the table isn't doing the ellipses... is that coming in a future PR?

image

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -16,11 +16,32 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 			<style>
 				:host {
 					display: block;
-					font-weight: normal;
+					padding-right: 20px;
```
svanherk: Is this so it doesn't hit the side of the column?  The list view will probably handle that.  Until then, I feel like this padding should be decided by the consumer, not the web component

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -16,11 +16,32 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
...
+					white-space: nowrap;
+				}
+				:host(:dir(rtl)) {
+					padding-right: 0px;
+					padding-left: 20px;
+				}
+				.d2l-activity-name-icon {
+					display: inline-block;
+					padding-right: 12px;
```
svanherk: Nit: This should probably be `0.6rem` instead

jeroach:
> I checked out your branch and noticed the table isn't doing the ellipses... is that coming in a future PR?

This is what I was meeting with Daniel about - it's the story I'm picking up next.

components/d2l-activity-name/d2l-activity-name.js
jeroach: Yeah to avoid hitting the side. I can remove it from here and add it as part of the next story/PR.

jeroach:
Comments addressed - going to merge if tests pass.

PR #76, opened 2019-03-07
Vy/empty activities list2

vix633:
This is design-approved!

image


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -88,6 +106,14 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					<d2l-button class="d2l-evaluation-hub-activities-list-load-more" onclick="[[_loadMore]]">[[localize('loadMore')]]</d2l-button>
 				</div>
 			</template>
+			<template is="dom-if" if="[[!_data.length]]">
```
mtjandra: I don't think this is enough information to determine if you have finished all your work. If `_parseActivities` fails and throws, `_data.length` will be zero, but it could just be we failed to fetch the data.

Also if you are still loading data, we shouldn't display this image, I think you may need to ensure `_loading` is false as well

vix633: I feel like this is more about if there are activities to display, rather than if the user is done grading. I don't see a good way to handle not showing this image if the loading has failed. Do we have an error that we can show there instead?

mtjandra: Yeah I am adding an error for it. You can tell if loading failed in two places.
1. In `ready()` when `d2l-siren-entity-error` listener is triggered
2. When failing to parse activities in `_loadData`

I guess don't worry about failures for now, I'm working on failures now we can sync up later. I think handling the `_loading` case is still valid though

vix633: Sweet, i saw the error lang terms but didn't see the actual error handling yet. 

I've pushed the above suggestion!

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -88,6 +106,14 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					<d2l-button class="d2l-evaluation-hub-activities-list-load-more" onclick="[[_loadMore]]">[[localize('loadMore')]]</d2l-button>
 				</div>
 			</template>
+			<template is="dom-if" if="[[_shouldShowNoSubmissions]]">
```
mtjandra: I believe it needs to be `_shouldShowNoSubmissions()`, also it would be awesome if you added another demo snippet to the demo page to illustrate how the component looks when everything is complete

vix633: Still trying to add these tests with Stacey's help.

components/d2l-evaluation-hub/build/lang/ar.js
```diff
@@ -16,7 +18,8 @@ const LangArImpl = (superClass) => class extends superClass {
 			'loading': 'يتم الآن التحميل',
 			'masterTeacher': 'Master Teacher',
 			'submissionDate': 'تاريخ الإرسال',
-			'tryAgain': 'Try Again'
+			'tryAgain': 'Try Again',
+			'noSubmissions': 'There are no submissions that need your attention.'
```
svanherk: Why did this one not go to where it was supposed to alphabetically?

vix633: i was testing you,
you passed.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -167,6 +193,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		return window.D2L.Siren.EntityStore.fetch(url, this.token);
 	}
 
+	_shouldShowNoSubmissions(dateLength, isLoading, health) {
```
svanherk: *`dataLength`?

vix633: image

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -96,6 +114,14 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					<d2l-button class="d2l-evaluation-hub-activities-list-load-more" onclick="[[_loadMore]]">[[localize('loadMore')]]</d2l-button>
 				</div>
 			</template>
+			<template is="dom-if" if="[[_shouldShowNoSubmissions(_data.length, _loading, _health)]]">
```
svanherk: I think we want to pass in `health.isHealthy`, so we don't recalculate unless that specific property changes

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-empty.html
```diff
@@ -0,0 +1,53 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+
+		<title>d2l-evaluation-hub-activities-list demo</title>
```
svanherk: Minor: Maybe put "d2l-evaluation-hub-activities-list Empty State Demo" here, and update the `h3` below too

demo/index.html
```diff
@@ -39,6 +39,7 @@ <h4>Activity Name</h4>
 				<h4>Evaluation Hub</h4>
 				<li><a href="d2l-evaluation-hub/d2l-evaluation-hub.html">d2l-evaluation-hub</a></li>
 				<li><a href="d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html">d2l-evaluation-hub-activities-list</a></li>
+				<li><a href="d2l-evaluation-hub/d2l-evaluation-hub-activities-list-empty.html">d2l-evaluation-hub-activities-list-empty</a></li>
```
svanherk: I'd also change the test here to `d2l-evaluation-hub-activities-list (empty state)`, so people don't confuse this with a different component they can use

PR #77, opened 2019-03-07
Us104018/alert when initial load fails

mtjandra:
image

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -156,7 +156,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		this.addEventListener('d2l-siren-entity-error', function() {
 			this._fullListLoading = false;
 			this._loading = false;
-			this._handleInitialLoadFailure();
+			this._handleFullLoadFailure();
 		}.bind(this));
 		this._loadMore = this._loadMore.bind(this);
 	}
@@ -268,7 +268,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			this._clearAlerts();
 		} catch (e) {
 			// Unable to load activities from entity.
-			this._handleInitialLoadFailure().bind(this);
+			this._handleFullLoadFailure().bind(this);
 			return Promise.reject(e);
 		} finally {
 			this._fullListLoading = false;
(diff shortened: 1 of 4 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -86,7 +86,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					</dom-repeat>
 				</d2l-tbody>
 			</d2l-table>
-			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]" has-close-button>
```
mtjandra: Removed close button, does not provide much value either refresh page, or click load more again (if load more failed).

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -310,6 +313,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		this.set('_health', { isHealthy: false, errorMessage: 'failedToLoadMore' });
 	}
 
+	_handleInitialLoadFailure() {
+		this.set('_health', { isHealthy: false, errorMessage: 'failedToLoadData' });
```
mtjandra: @vix633 after this PR you should be able to use `_healthy.isHealthy` related to this comment https://github.com/BrightspaceHypermediaComponents/activities/pull/76#discussion_r263444313

demo/d2l-evaluation-hub/d2l-evaluation-hub.html
```diff
@@ -45,10 +45,20 @@
 		<div class="vertical-section-container fixedSize">
```
mtjandra: Updated demo pages to show error on load, and give context on expectation that clicking load more second time is suppose to fail

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -264,8 +265,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		try {
 			var result = await this._parseActivities(entity);
 			this._data = result;
+			this._clearAlerts();
 		} catch (e) {
 			// Unable to load activities from entity.
+			this._handleInitialLoadFailure().bind(this);
```
svanherk: I think we should call this `handleFullLoadFailure` to match `_fullListLoading`.  This won't just happen on the initial load, but also on sort and filter.

mtjandra: Renamed

PR #81, opened 2019-03-07
US104674 - Add sort by last name

mtjandra:
Adding sort by last name http://search.dev.d2l/source/xref/Lms/le/activities/D2L.LE.Activities.Extensibility/Api/Hypermedia/Sorts/ByName/SortByLastName.cs#37


mtjandra:
> Looks good, I'm assuming there is a chain of PRs after this to actually use this new demo data?

Yup there is one that depends on a few

PR #82, opened 2019-03-07
Us104674/name split sort

mtjandra:
namesplit

* Split sorting by firstName and LastName.
* Tried in LMS seems to be working fine

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -23,7 +23,7 @@ import './d2l-no-submissions-image.js';
 
 class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBehavior, D2L.PolymerBehaviors.Siren.SirenActionBehavior ], EvaluationHubLocalize(PolymerElement)) {
 	static get template() {
-		return html`
+		const evaluationHubActivitiesListTemplate = html`
 			<style include="d2l-table-style">
 				d2l-td {
 					font-weight: normal;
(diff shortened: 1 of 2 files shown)


mtjandra:
Test failing due to unmerged dependencies, will re-run once dependencies resolved

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -145,7 +149,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				value: [
 					{
 						key: 'displayName',
-						headers: [{ key: 'displayName', sortClass: 'first-name', canSort: false }]
+						headers: [
+							{ key: 'firstName', sortClass: 'first-name', suffix: ',', canSort: false },
```
mtjandra: In case we want to localize suffix to `/` or `-` or something. Not sure best solution for this

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -71,11 +71,15 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 											<template>
 												<template is="dom-if" if="[[header.canSort]]">
 													<d2l-table-col-sort-button nosort on-click="_sort" id="[[header.key]]">
-														<span>[[localize(header.key)]]</span>
+														<span>
+															[[localize(header.key)]]<template is="dom-if" if="[[header.suffix]]">[[header.suffix]]</template>
```
mtjandra: This is on one line, if not there is unwanted spacing around characters

svanherk: Ah yes, we want to `strip-whitespace` in all our components.  I just had to figure out how to do this for Josh's component: https://github.com/BrightspaceHypermediaComponents/activities/pull/68

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: We can ask Daniel but I suspect it's probably fine for now

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: I tired out using `strip-whitespace` but it isn't giving me the behaviour I want. Thinking to leave as is

svanherk: Ok, but we do want to add `strip-whitespace` anyways, we shouldn't rely on the way we space out the html for our layout.  Should you just add `padding-right` to the suffix?

PR #83, opened 2019-03-07
US104170 - truncate + nowrap for CourseName

jeroach:
As per design discussion, we are only adding truncate and nowrap for ActivityName (which already has it) and CourseName.

- First cut at truncate and no wrap for course name
- Specify padding for activity name

truncate_nowrap_coursename

This is as small as the ActivityName and CourseName columns would get:

image

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -38,7 +38,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					white-space: nowrap;
 				}
 				.d2l-activity-name-column {
-					padding-right: 2.4em;
+					padding-right: 2.4rem;
 				}
 				:host(:dir(rtl)) .d2l-activity-name-column {
 					padding-right: 0;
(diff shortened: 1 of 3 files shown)


jeroach:
As per offline chat with Daniel, we're not entirely happy with how the `First Name, Last Name` column looks with a draft icon - but fixing it is lower priority than other items.

I've added the following story to the backlog to capture the planned work:
`US104756`
https://rally1.rallydev.com/#/238990806412d/detail/userstory/292863321932?fdp=true

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -33,6 +33,21 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					text-align: right;
 					width: 100%;
 				}
+				.d2l-evaluation-hub-truncated-column {
+					max-width: 10rem;
+					white-space: nowrap;
+				}
+				.d2l-activity-name-column {
+					padding-right: 2.4em;
```
svanherk: We'll want this to be `rem` like the others

jeroach: oops - sorry - will fix

PR #85, opened 2019-03-08
US104672 - add filter and sort to eval link

amauryperez:


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -572,6 +575,42 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+			extraParams += '&sort=' + sortVal;
+		}
+
+		return extraParams;
+	}
+
+	_getQueryStringParam(name, url) {
+		if (!url) return null;
+
+		var regex = new RegExp('[?&]' + name + '(=([^&#]*)|&|#|$)');
```
AlexBedley: I would just use window.URL like:

```js
const parsedUrl = new window.URL(url, 'https://unfortunately-you-need-this-for-relative-urls.com');
const queryStringValue = parsed.searchParams.get(name); // handles decoding for you
```

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -572,6 +575,42 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+		var results = regex.exec(url);
+
+		if (!results) return null;
+		if (!results[2]) return null;
+		return decodeURIComponent(results[2].replace(/\+/g, ' '));
+	}
+
+	_buildRelativeUri(url, extraParams) {
```
AlexBedley: Similarly, just use `window.URL` to construct the url

```js
const parsedUrl = new window.URL(url, 'https://unfortunately-you-need-this-for-relative-urls.com');
parsedUrl.search = extraParams // handles escaping for you
return parsedUrl.pathname + parsedUrl.search;
```

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -572,6 +575,42 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		}
 		return true;
 	}
+
+	_getExtraParams(selfHref) {
+		var extraParams = '';
```
AlexBedley: nit: `let` for mutating variables, `const` for constants

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -184,7 +184,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				type: Boolean,
 				value: true
 			},
-			_filterHref: {
+			_selfHref: {
```
AlexBedley: Is this ever used? It gets set but never called?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
amauryperez: what is `https://unfortunately-you-need-this-for-relative-urls.com`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Does that work in IE11?  We polyfill for the demo, but Miles was seeing master IE11 completely broken because of `URLSearchParams`.  Which is not good, that's a bit more than a bug

amauryperez: When I run the tests in IE11, after making Alex suggested changes, all tests passed except `data displays correctly when master teacher toggled on` which might be for a different reason?

AlexBedley: Can we just polyfill `URLSearchParams`?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: Discussed offline: unfortunately if you supply a relative url, you also must supply a host. and it has to be a valid url otherwise `new window.URL` will throw.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
amauryperez: Stacey said she will

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
amauryperez: what is polyfill?

svanherk: I'm looking into this, since we also want to use it in our `performSirenAction` work around function.  I'm hoping we can add this at the BSI level

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: I forget what I thought I needed this for... 🤔

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
amauryperez: Looks like our url will not be relative so I think we can ignore the second param since it is optional and only needed in that case

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -598,6 +601,50 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+					value: sortVal
+				}
+			);
+		}
+
+		return extraParams;
+	}
+
+	_getQueryStringParam(name, url) {
+		const parsedUrl = new window.URL(url, 'https://notused.com');
```
AlexBedley: I saw Jamie and Stacey use `window.location.origin` instead of `'https://notused.com'`... Since `window.location.origin` will be the actual host of the relative url it might make more sense to use that. But on the other hand it really isn't used so maybe this is more obvious what's happening. 🤷‍♂️

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: IMO if it's not used in this PR we should just delete it

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: So we don't want to add the polyfill to BSI for everyone to have to download.  To support IE11 we'll need to rewrite this to parse the query params ourselves, but there are other usages of this that will break Quick Eval in IE11.  So this could be delayed until a pass is done to remove all IE11-unsupported things

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
amauryperez: done

PR #86, opened 2019-03-08
US104761 - Manage sort state through Internal Variables

mtjandra:
* Manage sort state through internal variable rather than html attributes
* Reason we do this is because we want to load sort state on load, but there are race conditions when it comes to when the `dom-if`s render
* By managing sort state internally, we can guarantee it will be reflected properly on the page, previously the view and data state were tied together too tightly
* This is a refactor to support loading sort state on initial load, not new functional changes expected here

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -70,7 +70,12 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 										<dom-repeat items="[[headerColumn.headers]]" as="header">
 											<template>
 												<template is="dom-if" if="[[header.canSort]]">
-													<d2l-table-col-sort-button nosort on-click="_sort" id="[[header.key]]">
+													<d2l-table-col-sort-button
+														nosort$="[[!header.sorted]]"
+														desc$="[[header.desc]]"
+														on-click="_updateSortState"
+														id="[[header.key]]"
+													>
 														<span>[[localize(header.key)]]</span>
 														<template is="dom-if" if="[[header.suffix]]">
 															<span>[[header.suffix]]&nbsp;</span>
@@ -155,21 +160,21 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					{
 						key: 'displayName',
 						headers: [
-							{ key: 'firstName', sortClass: 'first-name', suffix: ',', canSort: false },
-							{ key: 'lastName', sortClass: 'last-name', canSort: false }
+							{ key: 'firstName', sortClass: 'first-name', suffix: ',', canSort: false, sorted: false, desc: false  },
+							{ key: 'lastName', sortClass: 'last-name', canSort: false, sorted: false, desc: false  }
 						]
 					},
 					{
 						key: 'activityName',
-						headers: [{ key: 'activityName', sortClass: 'activity-name', canSort: false }]
+						headers: [{ key: 'activityName', sortClass: 'activity-name', canSort: false, sorted: false, desc: false }]
 					},
 					{
 						key: 'courseName',
-						headers: [{ key: 'courseName', sortClass: 'course-name', canSort: false }]
+						headers: [{ key: 'courseName', sortClass: 'course-name', canSort: false, sorted: false, desc: false }]
 					},
 					{
 						key: 'submissionDate',
-						headers: [{ key: 'submissionDate', sortClass: 'completion-date', canSort: false }]
+						headers: [{ key: 'submissionDate', sortClass: 'completion-date', canSort: false, sorted: false, desc: false }]
 					},
 					{
 						key: 'masterTeacher',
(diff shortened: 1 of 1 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+				}
+				else {
+					this.set(`_headerColumns.${i}.headers.${j}.sorted`, false);
+				}
+			});
 		});
+	}
 
-		return this._followLink(this.entity, Rels.sorts)
+	_fetchSortedData(sortClass, descending) {
```
mtjandra: This method is just what `_sort` was doing before when fetching sorting data

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			});
 	}
 
-	_sort(e) {
-		const headers = [].concat.apply([], this._headerColumns.map(headerColumn => headerColumn.headers));
-		const header = headers.filter(h => h.key === e.currentTarget.id)[0];
+	_updateSortState(event) {
```
mtjandra: Hard to tell from diff, but new logic here is basically

```
if (header.key === headerId) {
  var descending = header.sorted && !header.desc;
  this.set(`_headerColumns.${i}.headers.${j}.sorted`, true);
  this.set(`_headerColumns.${i}.headers.${j}.desc`, descending);

  this._fetchSortedData(header.sortClass, descending); // And this is just what was being done previously to fetch sorted data
}
else {
}
```

mtjandra:
Aside: Seems like we don't clear the sort state before applying new sort. I think we want to do that. Does this sound right? Also this is just refactor to prep for loading sorts on initial load, not functional changes expected here

AlexBedley:
> Aside: Seems like we don't clear the sort state before applying new sort. I think we want to do that. Does this sound right?

We shouldn't have to clear sort state. I know sorting by multiple columns isn't a required feature but it's the default behaviour so it should be easier than not having this feature (at this stage of the API)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -70,7 +70,12 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 										<dom-repeat items="[[headerColumn.headers]]" as="header">
 											<template>
 												<template is="dom-if" if="[[header.canSort]]">
-													<d2l-table-col-sort-button nosort on-click="_sort" id="[[header.key]]">
+													<d2l-table-col-sort-button
+														nosort$="[[!header.sorted]]"
+														desc$="[[header.desc]]"
```
AlexBedley: What does the $ suffix mean?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			});
 	}
 
-	_sort(e) {
-		const headers = [].concat.apply([], this._headerColumns.map(headerColumn => headerColumn.headers));
-		const header = headers.filter(h => h.key === e.currentTarget.id)[0];
+	_updateSortState(event) {
+		var headerId = event.currentTarget.id;
```
AlexBedley: nit: you wouldn't download a car, don't use `var`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+					this.set(`_headerColumns.${i}.headers.${j}.sorted`, true);
+					this.set(`_headerColumns.${i}.headers.${j}.desc`, descending);
 
-		const headerElements = this.shadowRoot.querySelectorAll('d2l-table-col-sort-button');
-		headerElements.forEach(h => {
-			if (h !== e.currentTarget) {
-				h.removeAttribute('desc');
-				h.setAttribute('nosort', 'nosort');
-			}
+					this._fetchSortedData(header.sortClass, descending);
```
AlexBedley: let's `return` this value to make testing easier (you'll have to save it as a temp until the forEach is done)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+				}
+				else {
+					this.set(`_headerColumns.${i}.headers.${j}.sorted`, false);
```
AlexBedley: Should you also reset `desc` to `false`? So the first click is always sorted ascending.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -254,48 +259,38 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
 
-				const sort = sortsEntity.entity.getSubEntityByClass(header.sortClass);
+				const sort = sortsEntity.entity.getSubEntityByClass(sortClass);
 				if (!sort) {
-					return Promise.reject(`Could not find sort class ${header.sortClass}`);
+					return Promise.reject(`Could not find sort class ${sortClass}`);
 				}
 
-				const actionName = ascending ? 'sort-ascending' : 'sort-descending';
+				const actionName = descending ? 'sort-descending' : 'sort-ascending';
```
AlexBedley: You would invert my logic!

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Haha what? I am using var because of supporting IE11, just using lowest common denominator

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: That should be the case: `var descending = header.sorted && !header.desc;`

If sorted is false, then descending is false, so you are always ascending on first click

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Haha, only because `d2l-table-col-sort-button` using attribute called `desc` made it easier. My bad!

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: ie11 supports `const` and `let`

* `const` - https://caniuse.com/#feat=const
* `let` - https://caniuse.com/#feat=let (just not in for-loops, but who uses for-loops anyways)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: But I don't think `_fetchSortedData` even returns anything. Do you want it to return the `entity in its promise? Right now it just ends with

```
.then((collection => {
  this.entity = collection;
}).bind(this));
```

AlexBedley: ah right that's only in my branch. it should `return Promise.resolve()`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: It is attribute binding https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: In that case I am so sorry for ignoring all your "nits" I will update!

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: > That should be the case: `var descending = header.sorted && !header.desc;`

Yeah you're right, although it was hard to see at first glance. Probably fine as-is.

> Also api side / query param, we never actually switch it, that is why I don't toggled, and is why I asked if we should `clear` sorts first

hmm I don't get this. we never actually switch what?

mtjandra: Steps:
0. Start with no sorts
1. Sort By First Name (defaults ascending)
2. Sort By First Name (now is sorting by descending)
3. Sort By Last Name (defaults ascending)

Now you have two sorts applied in priority:
0: Sort By Last Name Ascending
1: Sort By First name Descending

So I don't want to reset `desc` to false for first name, because it is really descending. That is why I didn't reset it

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
mtjandra: Ok updated to return the empty promise

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: yeah technically first name is sorted descending, although we shouldn't be displaying that to the user. As far as they are concerned (and the UI) there is only 1 sort. The API secretly knows there are multiple sorts applied though.

PR #87, opened 2019-03-08
Add sorting unit tests

AlexBedley:
First stab at some tests. Bring on the merge conflicts.


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+	return stubFollowLink(list, mappings);
+}
+
+suite('d2l-evaluation-hub-activities-list-sorting', () => {
+	let list;
+
+	setup(function() {
+		list = fixture('basic');
+		// huge hack, not sure why fixture is not resetting them
+		resetSortHeaders(list);
```
AlexBedley: Calling out this huge hack. Spend way too long trying to figure out why `fixture('basic')` would reset some state but not others. Rather than track this bug down I decided to just do this

test/d2l-evaluation-hub/data/sorting/initialEntity.json
```diff
@@ -0,0 +1,11 @@
+{
+  "links": [
+    {
+      "rel": [
+        "https://api.brightspace.com/rels/sorts"
+      ],
+      "href": "data/sorting/initialSorts.json"
```
AlexBedley: I added these stub files so the tests wouldn't freak out on initial load

AlexBedley:
This is ready for review now @svanherk @mtjandra

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -281,61 +287,68 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
 		const headerId = event.currentTarget.id;
+		let shouldSort = false;
+		let sortClass;
+		let descending;
 
 		this._headerColumns.forEach((headerColumn, i) => {
 			headerColumn.headers.forEach((header, j) => {
-				if (header.key === headerId) {
-					const descending = header.sorted && !header.desc;
+				if (header.key === headerId && header.canSort) {
```
AlexBedley: I know `header.canSort` is implied (since this function is only called from a sortable header which is also controlled by `header.canSort`) but I figured it couldn't hurt.

mtjandra: nit: I would prefer `( header.key === headerId ) && header.canSort`

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -281,61 +287,68 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+					descending = desc;
 				}
 				else {
 					this.set(`_headerColumns.${i}.headers.${j}.sorted`, false);
 				}
 			});
 		});
 
-		return result;
+		if (shouldSort) {
```
mtjandra: Was something actually broken, or is this a refactor?

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
mtjandra: Hmmm yeah fishy, this actually invalidates the tests `sorting is disabled on all columns by default` completely

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+	});
+
+	test('sorting is disabled on all columns by default', () => {
+		expect(list._headerColumns.filter(column => column.headers.some(h => h.canSort))).to.be.empty;
+	});
+
+	test('_loadSorts silently does nothing on null entity', () => {
+		return list._loadSorts(null);
```
mtjandra: I didn't know this. So the testing framework will fail if you return a rejection? It even knows it is asynchronous?

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+					return list._updateSortState(e).then(() => {
+						expect(list.entity).to.deep.equal(SirenParse(mappings['apply']));
+					});
+				});
+			});
+	});
+
+});
+
+function formatCollection(entities) {
```
mtjandra: Why even pass in `entities`? Seems you don't care about unassessed activities, since you always pass in empty array

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+
+		list._loadSorts(entity)
+			.then(() => {
+				done('_loadSorts should have rejected');
+			})
+			.catch((err) => {
+				expect(err.toString()).to.equal('Error: Could not load sorts endpoint');
+				done();
+			})
+			.catch((err) => {
```
mtjandra: Why do we need to catch the catch? I'm guessing the expect throws when false?

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+	test('_loadSorts rejects on bad sort entity', (done) => {
+		const mappings = {};
+		mappings[Rels.sorts] = null;
+		stubFollowLink(list, mappings);
+
+		const entity = formatCollection([]);
+
+		list._loadSorts(entity)
+			.then(() => {
+				done('_loadSorts should have rejected');
```
mtjandra: Cool I didn't know passing string to `done` counts as a `assert.fail` type thing

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+			.catch((err) => {
+				expect(err.toString()).to.equal('Error: Could not find sortable header for activityName');
+				done();
+			})
+			.catch((err) => {
+				done(err);
+			});
+	});
+
+	test('clicking a header sorts column ascending', () => {
```
mtjandra: If you want to check that the header is sorted ascending would you want to check the prop `list._headerColumns[1].headers[0].desc === false && list._headerColumns[1].headers[0].sorted == true` ? Not the best since basing of position instead of key but something like that.


Similar comment for descending test

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -0,0 +1,251 @@
...
+
+	list.performSirenAction = (action, fields) => {
+		if (mappings[action.name]) {
+			return Promise.resolve(SirenParse(mappings[action.name]));
+		}
+		return oldPerformSirenAction(action, fields);
+	};
+}
+
+function stubFollowLink(list, mappings) {
```
mtjandra: I think I would've preferred using test data instead of stubbing follow link and performSirenAction. Worried that it makes these tests brittle

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: yes `mocha` is pretty smart, if you return a Promise it will pass if it's fulfilled and fail if it's rejected

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: True, I probably started this with good intentions and then realized it doesn't really matter if the activities are sorted since I'm stubbing the method anyways. Will fix.

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: Correct

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: the magic of `mocha`!

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: I was seeing some weirdness so I refactored to see if it made a difference but I never unrefactored to see what the true issue was. If can see if this change is necessary (I do like rejecting if we didn't find the desired sort though so I'll keep that in)

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: I thought it was analogous to using `Moq` in C# unit tests... I don't want my tests to fail because of a bug in `performSirenAction` or `followLink` which I viewed as dependencies. Happy to chat further on this.

I don't like that these test operate different than the other tests (a learning curve to edit them) but I was hoping it wasn't too steep.

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: Hmm yes I should also assert state I think, good idea

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: Gonna just leave it in for now, hopefully we can come back and investigate further to remove the hack

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
AlexBedley: What are your thoughts on this @mtjandra ?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: OK I undid this change and it made no difference. Gonna go with what was there before.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -306,36 +311,40 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			});
 		});
 
-		return result;
+		if (result) {
```
mtjandra: `if(Promise.reject()) { console.log('hello'); }` this will print hello. I think you want 

`result.catch(function() { return Promise.reject(new Error(`Could not find sortable header for ${headerId}`)); })`

AlexBedley: If `result` rejects then I still want to return `result` because it holds information about exactly what went wrong. I only want the `else` to trigger if we don't find the header we were looking for (so when result stays `undefined`)

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
mtjandra: Okay makes sense, thanks!

PR #89, opened 2019-03-11
US104761 - Mock sort state prop

mtjandra:
* Updating demo data to reflect changes being applied in https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/12638/overview

PR #90, opened 2019-03-11
US104761 - Load sort state on load

mtjandra:
image

* Demo now shows sort state on initial load. I use `?sort=bydate-a` in demo data. LMS will need to be configured to use this sort by default as well.
* Demo will work correctly after https://github.com/BrightspaceHypermediaComponents/activities/pull/89

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -250,6 +250,11 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 							const sort = sortsEntity.entity.getSubEntityByClass(header.sortClass);
 							if (sort) {
 								this.set(`_headerColumns.${i}.headers.${j}.canSort`, true);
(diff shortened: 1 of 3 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -250,6 +250,11 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 							const sort = sortsEntity.entity.getSubEntityByClass(header.sortClass);
 							if (sort) {
 								this.set(`_headerColumns.${i}.headers.${j}.canSort`, true);
+								if (sort.properties && sort.properties.applied && (sort.properties.priority === 0)) {
```
AlexBedley: curious: why is the last clause in parenthesis? Doesn't `===` have priority over `&&`?

vix633: IMO more readable

mtjandra: Yeah I think you're right about priority. I put in the brackets for readability like Viktor said

PR #91, opened 2019-03-11
- Fixing issue where actions are losing query params

jmurray87:


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -588,6 +588,23 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+	_performSirenActionWithQueryParams(action) {
+		let url = new URL(action.href, window.location.origin);
+
+		if(!action.fields) {
+			action.fields = []
+		}
+
+		url.searchParams.forEach(function(value, key) {
+			if(!action.fields.find(x => x.name === key)){
+				action.fields.push({name: key, value: value, type: "hidden"});
```
svanherk: You're gonna have some linting errors with the spacing, missing comma in line 596 and the double quotes here

test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -308,5 +308,77 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
+				href : 'http://127.0.0.1/',
+				name: 'apply',
+				type: 'application/x-www-form-urlencoded',
+				method: 'GET'
+			};
+
+			const stub = sinon.stub(list, 'performSirenAction');
+			list._performSirenActionWithQueryParams(action);
+			sinon.assert.calledWith(stub, action, []);
+			done();
```
AlexBedley: Is `done` necessary here since we don't actually care what `performSirenAction` returns?

jmurray87: The test will timeout if done isn't called, but are you suggesting we don't pass through the lambda?

jmurray87: I went ahead and assumed you meant remove completely, I agree :)

AlexBedley: Really? That's surprising to me. I would expect this test to be fine without `done` because we're not doing any asserts after the promise resolves. I guess it depends how `sinon.stub` works.

AlexBedley: Ah ok I see what you meant by "pass through the lambda" now. Yes that's what I meant :P

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -588,6 +588,23 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+	_performSirenActionWithQueryParams(action) {
+		const url = new URL(action.href, window.location.origin);
+
+		if (!action.fields) {
+			action.fields = [];
+		}
+
+		url.searchParams.forEach(function(value, key) {
+			if (!action.fields.filter(x => x.name === key)[0]) {
+				action.fields.push({name: key, value: value, type: 'hidden'});
```
svanherk: Seems like now we're doing both (adding to `action.fields`, and sending it in as a parameter).  Does this cut down on a code path or add an additional one?

jmurray87: Just avoids running the broken `getSirenFields` here: https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js#L150
and here: https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js#L47.  T

svanherk: Gotcha, so we'll lose the weird `n=e` stuff until we fix it

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -588,6 +588,23 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
...
+			}
+		});
+
+		return this.performSirenAction(action, action.fields);
```
svanherk: Note: I did remove the second parameter for the filter PR because I was seeing weird things.  We should test and keep an eye out to make sure both paths work as we expect.

PR #93, opened 2019-03-11
The clamp on the description is done before skeleton is hidden

m6jones:


demo/d2l-activity-list-item/index.html
```diff
@@ -72,13 +72,13 @@ <h3>Basic d2l-activity-list-item demo</h3>
...
 
-						setTimeout(function() {
-							const shimmerPlaceholderTest = document.querySelector('#shimmer-placeholder-test');
-							if (shimmerPlaceholderTest) {
-								shimmerPlaceholderTest.removeAttribute('image-shimmer');
-								shimmerPlaceholderTest.removeAttribute('text-placeholder');
-							}
-						}, 2000);
+						const shimmerPlaceholderTest = document.querySelector('#shimmer-placeholder-test');
+						shimmerPlaceholderTest.addEventListener('d2l-activity-text-loaded', () => {
```
qwangd2l: I don't think there's enough time to add event listeners after the element is already created to see the loading well enough. Maybe if resources are not local, then there would be enough time.
Should we add the timeout back?

m6jones: Yeah, I can add the timeout back.

PR #94, opened 2019-03-11
dont interate when we have no entities

vix633:
This should address the console error we are getting, hopefully also the empty submission issue.


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -344,9 +344,11 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		this._fullListLoading = true;
 
 		try {
-			var result = await this._parseActivities(entity);
-			this._data = result;
-			this._clearAlerts();
+			if (entity.entities) {
+				var result = await this._parseActivities(entity);
+				this._data = result;
+				this._clearAlerts();
```
svanherk: Should we still be clearing alerts if the entity was changed but there just happened to be none in it? @mtjandra?

mtjandra: Yeah it should. We should be clearing the alerts anytime we successfully load data.

Both Responses:
```
{
	"links":[],
	"entities":[]	
}
```

and 

```
{
	"links":[]
}
```


Should be supported by our app

vix633: Pushed this.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -344,9 +344,12 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		this._fullListLoading = true;
 
 		try {
-			var result = await this._parseActivities(entity);
-			this._data = result;
+			if (entity.entities) {
+				var result = await this._parseActivities(entity);
+				this._data = result;
```
mtjandra: I think if `entity` is truthy, and `entity.entities` is falsy, you will want to set `this._data = []` to indicate there are no rows to show.

PR #95, opened 2019-03-12
- Reverting back to alternate use of perform action for consitency wi…

jmurray87:
Just going back to the other usage to be consistent with d2l-hm-filter


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -355,27 +355,9 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
-					value: 'testvalue'
-				},
-				{
-					type: 'hidden',
-					name : 'anothertestname',
-					value: 'anothertestvalue'
-				}
-			];
-
 			const stub = sinon.stub(list, 'performSirenAction');
```
svanherk: I found I had to add this to check that the fields were actually being set properly, otherwise this call just checked that the action reference was passed around: https://github.com/BrightspaceHypermediaComponents/common/blob/master/test/d2l-hm-filter/d2l-hm-filter.js#L259-L261

jmurray87: makes sense now that I am not passing the fields!

PR #96, opened 2019-03-12
Fix Some RTL Issues shown by the screenshot in Mile's PR

svanherk:
- ~Fix table RTL design~
- Fix Activity Name RTL
- Fix loading button RTL position
- Add the RTL file to the Quick Eval demos

image

### DIFF
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -28,8 +28,6 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 				}
 				.d2l-activity-name-text {
 					display: inline-block;
-					overflow: hidden;
-					text-overflow: ellipsis;
 					vertical-align: middle;
 					max-width: 100%;
 				}
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -40,6 +40,9 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					text-align: right;
 					width: 100%;
 				}
+				:host(:dir(rtl)) .d2l-evaluation-hub-activities-list-load-more-container {
+					text-align: left;
+				}
 				.d2l-evaluation-hub-truncated-column {
 					max-width: 10rem;
 					white-space: nowrap;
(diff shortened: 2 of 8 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -176,6 +181,10 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				value: false,
 				reflectToAttribute: true
 			},
+			dir: {
```
mtjandra: Ah so this is something you always have to pass down? Do we need this prop in evaluation hub as well?

svanherk: No, this seems to be specifically a `table` thing.  For the css styling, it uses `[dir="rtl"]` rather than `:host(:dir(rtl))`.  This may not be the "proper" way to fix it

mtjandra: I'm a bit confused why this prop is necessary. Even in the demo pages you set the dir on `document.documentElement` not on `d2l-evaluation-hub-activities-list`. Won't this attribute always be true?

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -28,8 +28,6 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 				}
 				.d2l-activity-name-text {
 					display: inline-block;
-					overflow: hidden;
```
jeroach: Removing these from here means that anyone who uses the `d2l-activity-name` component from now on will need to handle text overflow separately.

If we're cool with that, then I guess this is okay. Just feels a bit odd.

svanherk: I _think_ that's what we want, in case they want to use it in a table and have it wrap?  Or if Parent Portal wanted to use it in their card?

svanherk:
The table design is going to still look off in RTL.  This is actually a bug: https://github.com/BrightspaceUI/table/issues/205

When this is fixed, we'll get the fix for free.  I'm also about to add the list design, which makes this a lot less noticeable.  Only text-align will be off

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: We discussed this offline about why this works, but it isn't necessary.  Confirmed this is a bug, rubrics has the same problem in preview mode

PR #101, opened 2019-03-13
Light Table Style and Table Summary

svanherk:
image

### DIFF
--- a/components/d2l-evaluation-hub/build/lang/ar.js
+++ b/components/d2l-evaluation-hub/build/lang/ar.js
@@ -21,6 +21,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'masterTeacher': 'Master Teacher',
 			'noSubmissions': 'There are no submissions that need your attention.',
 			'submissionDate': 'تاريخ الإرسال',
+			'tableTitle': 'List of unevaluated Learner submissions from across courses and tools',
 			'tryAgain': 'Try Again'
 		};
 	}
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -87,7 +87,8 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					@apply --d2l-body-compact-text;
 				}
 			</style>
-			<d2l-table type="light" hidden$="[[_fullListLoading]]" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
+			<d2l-offscreen id="d2l-evaluation-hub-activities-list-table-summary">[[localize('tableTitle')]]</d2l-offscreen>
+			<d2l-table type="light" hidden$="[[_fullListLoading]]" title="[[localize('tableTitle')]]" aria-described-by$="d2l-evaluation-hub-activities-list-table-summary" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
 				<d2l-thead>
 					<d2l-tr>
 						<dom-repeat items="[[_headerColumns]]" as="headerColumn">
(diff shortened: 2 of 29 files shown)


components/d2l-evaluation-hub/build/lang/ar.js
```diff
@@ -21,6 +21,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'masterTeacher': 'Master Teacher',
 			'noSubmissions': 'There are no submissions that need your attention.',
 			'submissionDate': 'تاريخ الإرسال',
+			'tableTitle': 'List of unevaluated Learner submissions from across courses and tools',
```
mtjandra: Should `Learner` be lower case?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -87,7 +95,8 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 					@apply --d2l-body-compact-text;
 				}
 			</style>
-			<d2l-table hidden$="[[_fullListLoading]]" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
+			<d2l-offscreen id="d2l-evaluation-hub-activities-list-table-summary">[[localize('tableTitle')]]</d2l-offscreen>
+			<d2l-table type="light" hidden$="[[_fullListLoading]]" title="[[localize('tableTitle')]]" aria-describedby$="d2l-evaluation-hub-activities-list-table-summary" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
```
mtjandra: Does `aria-describedby` need that `$`? Seems like some hardcoded property similar to `type=light`

components/d2l-evaluation-hub/build/lang/ar.js
jeroach: nit: `noSubmissions` ends with a period. This term does not. Should we be consistent? Or are they used in different scenarios?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -26,7 +26,15 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 		const evaluationHubActivitiesListTemplate = html`
 			<style include="d2l-table-style">
 				d2l-td {
-					font-weight: normal;
+					font-size: 0.7rem;
+				}
+				d2l-td.d2l-username-column {
+					font-size: 0.8rem;
```
jeroach: Why do we need a larger font size for the username column? Design? :)

components/d2l-evaluation-hub/build/lang/ar.js
svanherk: This is the exact text Surekha had in the story, so I assumed there was reason for the uppercase.  You'll see this when you hover over the table, which is maybe why a period wasn't wanted.  @Surekh?

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: Yeah, this was Erin's request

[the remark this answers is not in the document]
svanherk: Ah yes thanks!  Missed this when I was selecting stuff to stage

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
svanherk: So this is an attribute that screenreaders look for, not something that `d2l-table` has as a property.  Without the `$`, it won't show up.  Whereas `type` is a property on `d2l-table`.

components/d2l-evaluation-hub/build/lang/ar.js
Surekh: Confirmed with Dan upper case for Learner is good.

PR #109, opened 2019-03-14
Display filter errors in Quick Eval

svanherk:
Branched off of https://github.com/BrightspaceHypermediaComponents/activities/pull/108.  Travis will fail until https://github.com/BrightspaceHypermediaComponents/common/pull/12 is in.
image

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub.js
@@ -45,6 +45,7 @@ class D2LEvaluationHub extends mixinBehaviors([D2L.PolymerBehaviors.Siren.Entity
 				}
 				d2l-alert {
 					margin: auto;
+					margin-bottom: 0.5rem;
 				}
 			</style>
 			<div class="d2l-evaluation-hub-top-bar">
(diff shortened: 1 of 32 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub.js
```diff
@@ -77,12 +93,14 @@ class D2LEvaluationHub extends mixinBehaviors([D2L.PolymerBehaviors.Siren.Entity
 	attached()  {
 		this.addEventListener('d2l-hm-filter-filters-loaded', this._filtersLoaded);
 		this.addEventListener('d2l-hm-filter-filters-updated', this._filtersChanged);
+		this.addEventListener('d2l-hm-filter-error', this._filterError);
```
mtjandra: Who is firing this event, I didn't find it in commons

svanherk: Here: https://github.com/BrightspaceHypermediaComponents/common/pull/12/files#diff-2b5d39b3c1c4daa2f9c10d89a213b39cR453
This is blocked on that PR

mtjandra:
I'm a bit confused on my understanding. Doesn't applying filter re-load the whole table. If the table fails to load data an error message already exists. I think the error message you are trying to add is when we cannot fetch the filters. In that case I would expect the error message to show up in the filter component not QE

PR #110, opened 2019-03-14
Hide the load more button when we are loading

svanherk:
This does look better and works well  I made sure the screenreader experience with this change was still smooth.


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -171,7 +171,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			</d2l-alert>
 			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!_loading]]">[[localize('loading')]]</d2l-offscreen>
 			<d2l-loading-spinner size="80" hidden$="[[!_loading]]"></d2l-loading-spinner>
-			<template is="dom-if" if="[[_pageNextHref]]">
+			<template is="dom-if" if="[[_shouldShowLoadMore(_pageNextHref, _loading)]]">
```
mtjandra: Looks fine to me. Only worry is loading is never set/changes using `this.set` so I am not 100% of reliability that this will always be captured correctly

svanherk: Discussed offline, `this.set` is only required for subproperties, but we could be using it for the properties as well

PR #111, opened 2019-03-14
US104780

jeroach:
- Add lang term, `title`, and `aria-describedby` to the username column

### DIFF
--- a/components/d2l-evaluation-hub/build/lang/ar.js
+++ b/components/d2l-evaluation-hub/build/lang/ar.js
@@ -12,7 +12,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'checkBackOften': 'Check back often for new submissions.',
 			'courseName': 'المقرر التعليمي',
 			'displayName': 'الاسم الأول، اسم العائلة',
-			'evaluate': 'Evaluate ',
+			'evaluate': 'Evaluate {displayName}',
 			'failedToLoadData': 'Unable to load submissions. Try again in a few minutes.',
 			'failedToLoadMore': 'Unable to load more submissions. Try again in a few minutes.',
 			'firstName': 'First Name',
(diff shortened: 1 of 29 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -144,7 +144,12 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 						<template>
 							<d2l-tr>
 								<d2l-td class="d2l-username-column">
-									<d2l-link href="[[s.activityLink]]">[[_getDataProperty(s, 'displayName')]]</d2l-link>
+									<d2l-offscreen id="d2l-evaluation-hub-activities-list-username">[[localize('evaluate')]][[_getDataProperty(s, 'displayName')]]</d2l-offscreen>
```
jeroach: Are ids relative? The highlight works, but I don't hear it read when using NVDA

components/d2l-evaluation-hub/build/lang/ar.js
```diff
@@ -12,6 +12,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'checkBackOften': 'Check back often for new submissions.',
 			'courseName': 'المقرر التعليمي',
 			'displayName': 'الاسم الأول، اسم العائلة',
+			'evaluate': 'Evaluate ',
```
svanherk: We don't want to just translate the word by itself, because different languages change the order.  I think you want this to be `Evaluate {username}`, and then when you use it I think you do `localize('evaluate', username, _getDataProperty(s, 'displayName'))`... I can track down an example if this doesn't work at all

jeroach: Ah, that makes sense. I was wondering how this would work.

Thank you, will fix.

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
jeroach: @svanherk Any concerns on this item? Is this just a known issue with NVDA?

svanherk: I'm going to pull the branch and try it, it seems correct to me from a spec POV.  I'd expect it to be read

PR #113, opened 2019-03-14
Just add the profile image

jeroach:
Just drop the `d2l-profile-image` in there.

**Updated** - Here's what it looks like:
just_the_profile_image

### DIFF
--- a/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
+++ b/components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
@@ -10,6 +10,7 @@ import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
 import 'd2l-polymer-siren-behaviors/store/siren-action-behavior.js';
 import 'd2l-polymer-behaviors/d2l-dom-focus.js';
 import 'd2l-link/d2l-link.js';
+import 'd2l-users/components/d2l-profile-image.js';
 import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
 import {Rels, Classes} from 'd2l-hypermedia-constants';
 import '../d2l-activity-name/d2l-activity-name.js';
@@ -31,6 +32,16 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 				}
 				d2l-td.d2l-username-column {
 					font-size: 0.8rem;
+				}
+
+				.d2l-user-badge-image {
+					display: inline-block;
+					padding-right: 0.6rem;
+					vertical-align: middle;
+				}
+				:host(:dir(rtl)) .d2l-user-badge-image {
+					padding-right: 0;
+					padding-left: 0.6rem;
 				}
 
 				/* Needed for Edge */
@@ -144,6 +155,7 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 						<template>
 							<d2l-tr>
 								<d2l-td class="d2l-username-column">
+									<d2l-profile-image class="d2l-user-badge-image" href="[[s.userHref]]" token="[[token]]" small=""></d2l-profile-image>
 									<d2l-link href="[[s.activityLink]]">[[_getDataProperty(s, 'displayName')]]</d2l-link>
 									<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
 								</d2l-td>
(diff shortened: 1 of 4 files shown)


components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
```diff
@@ -565,6 +578,14 @@ class D2LEvaluationHubActivitiesList extends mixinBehaviors([D2L.PolymerBehavior
 			});
 	}
 
+	_getUserHref(entity) {
+		if (entity.hasLinkByRel(Rels.user)) {
+			const link = entity.getLinkByRel(Rels.user);
+			return link.href;
+		}
+		return '';
```
AlexBedley: What happens when this branch is triggered (if they don't have a user rel)?

jeroach: The picture does not load, but the spacing holds, and the page loads correctly otherwise. See image:

picture_does_not_load

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -341,7 +383,7 @@ function getMappings(table) {
 				return formattedFilter;
 			});
 		mappings[getHrefForEnrollmentFilters(i)] = formatFilters(formattedFilters, getHrefForEnrollments(i));
-		mappings[getHrefForMasterTeacher(i)] = formatUser(teachersByCourse[course]);
+		mappings[getHrefForMasterTeacher(i)] = formatUser(teachersByCourse[course], 'first', 'last', 'href');
```
jeroach: These don't matter for master teacher so just putting dummy values in

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
AlexBedley: We probably should also remove the spacing, but it's also probably not worth worrying about in this PR (IMO the `d2l-profile-image` component should remove itself from the DOM)

demo/d2l-evaluation-hub/demo-data-parser.js
AlexBedley: Can you just omit them (implicitly make them `undefined`)?

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -16,21 +16,62 @@ function distinct_ie11_safe(array) {
 }
 
 function parseUsers(data) {
-	const names = data.map(row => formatName(row.firstName, row.lastName));
-	const uniqueNames = distinct_ie11_safe(names);
+	const parsedUsers = data.map((row, i) => {
```
AlexBedley: This change will cause each user to be unique... is that a requirement or should we still be deduping them? I know it doesn't really matter for the demo but I was interested in having multiple scenarios in the demo data (one user in multiple rows being one of those scenarios)

components/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.js
jeroach: Was easy enough so just did it:

some_displayed_some_hidden

jeroach: (wrapped with a `dom-if`)

demo/d2l-evaluation-hub/demo-data-parser.js
jeroach: No. Tried that - it doesn't render.

demo/d2l-evaluation-hub/demo-data-parser.js
jeroach: Oh, that's likely why one of the `Alex` users doesn't have a profile pic

jeroach: Preference to merge this as is, and fix the demo data in a future PR (if we want to indeed dedupe them)

mtjandra:
Trying out with less space I think this might still be acceptable:

image


Although I'm not sure if this is:
image

Trying on LMS now to see what it might look like

mtjandra:
Do you happen to remember the perm you need to view profile image. I have the following on, I thought it was enough, but maybe I forgot something.

image

It isn't showing the image for some reason

mtjandra:
I think there may be issue with profile image, I don't see the image loading, but I can see it in the evaluations screen for assignments

mtjandra:
Nvm had to use `http://kdx1-mtjandrawi.desire2learn.d2l` otherwise token doesn't work

PR #116, opened 2019-03-14
Reset all non primary headers to unsorted on load

AlexBedley:
This fixes `If I click between the columns too quickly, the UI will show multiple sort arrows at the same time`

There will be a follow up PR to add a test and fix the tests I gutted.

I've manually tested this change and it appears to work. You can reliably trigger the double sort arrows by spam clicking the same column a few times and then quickly spam clicking another column a few times. But after my change, you won't be able to get it to happen.

### DIFF
--- a/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
+++ b/test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
@@ -1,13 +1,6 @@
 import {Rels} from 'd2l-hypermedia-constants';
 import SirenParse from 'siren-parser';
 
-function findSortHeader(list, sortClass) {
-	return list._headerColumns
-		.map(column => column.headers)
-		.reduce((acc, val) => acc.concat(val), [])  // ie11 flatten
-		.reduce((acc, val) => val.sortClass === sortClass ? val : acc, undefined); // ie11 find
-}
-
 function resetSortHeaders(list) {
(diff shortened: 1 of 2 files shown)


test/d2l-evaluation-hub/d2l-evaluation-hub-activities-list-sorting.js
```diff
@@ -162,8 +162,6 @@ suite('d2l-evaluation-hub-activities-list-sorting', () => {
 			.then(() => {
 				return list._updateSortState(e).then(() => {
 					expect(list.entity).to.deep.equal(SirenParse(mappings['apply']));
-					expect(findSortHeader(list, enabledSort).desc).to.be.false;
```
AlexBedley: There's 2 things I don't like about the sort tests. The first is the fact that `fixture` does not reset them. I think this is related to our improper usages of `this.set` in the list component. According to the polymer docs we should be using `notifySplices`. The second is my janky interceptor that mocks out `_followLink` and `performSirenAction`. We should be using `sinon.stub` instead. I would like to refactor the tests in a separate PR so it does not block this change from our release tomorrow morning.

svanherk: Yeah, here's the official Polymer 3 version of the docs around all this for whenever you guys get to it: https://polymer-library.polymer-project.org/3.0/docs/devguide/model-data

We definitely have not been consistent

PR #121, opened 2019-03-19
DE33480 - Renaming 'Master Teacher' to 'Teacher'

mtjandra:


mtjandra:
> Any plans to remove the references to masterTeacher littered throughout the code? If we still call it that internally it might not be that bad

I didn't plan to. My personal preference is having internal names that map to one or many external names. But if you'd like to rename feel free to add a tech debt story!

PR #123, opened 2019-03-19
US104726 - Support page size in query param of demo

mtjandra:
pagesize

This is step one:
* Allow configuring default page size mock server uses with query parameter

### DIFF
--- a/demo/d2l-quick-eval/d2l-quick-eval-activities-list.html
+++ b/demo/d2l-quick-eval/d2l-quick-eval-activities-list.html
@@ -22,8 +22,8 @@
 			import data from './table';
 
 			const url = new URL(window.location);
-			const pageSize = url.searchParams.get('pageSize') || 3;
-			const mappings = mappingsParser(data, pageSize);
+			const pageSize = url.searchParams.get('pageSize');
+			const mappings = mappingsParser(data, { pageSize: pageSize });
 			// Parameters: data, network logging, api request delay
 			startMockServer(mappings, true, 100);
 
--- a/demo/d2l-quick-eval/d2l-quick-eval-activities-list.html
+++ b/demo/d2l-quick-eval/d2l-quick-eval-activities-list.html
@@ -21,7 +21,9 @@
 			import mappingsParser from './demo-data-parser';
 			import data from './table';
 
-			const mappings = mappingsParser(data);
+			const url = new URL(window.location);
+			const pageSize = url.searchParams.get('pageSize') || 3;
+			const mappings = mappingsParser(data, pageSize);
 			// Parameters: data, network logging, api request delay
 			startMockServer(mappings, true, 100);
 
--- a/demo/d2l-quick-eval/d2l-quick-eval.html
+++ b/demo/d2l-quick-eval/d2l-quick-eval.html
@@ -22,8 +22,8 @@
 			import data from './table';
 
 			const url = new URL(window.location);
-			const pageSize = url.searchParams.get('pageSize') || 3;
-			const mappings = mappingsParser(data, pageSize);
+			const pageSize = url.searchParams.get('pageSize');
+			const mappings = mappingsParser(data, { pageSize: pageSize });
 			// Parameters: data, network logging, api request delay
 			startMockServer(mappings, true, 100);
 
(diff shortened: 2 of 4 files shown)


demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
```diff
@@ -21,7 +21,9 @@
 			import mappingsParser from './demo-data-parser';
 			import data from './table';
 
-			const mappings = mappingsParser(data);
+			const url = new URL(window.location);
```
svanherk: This is making the IE11 problem worse.  But we should know in the next two weeks if we have to continue supporting it.

mtjandra: This is for demo data only, and demo data has polyfill. Also IE demo doesn't work at all right now, I'm told this is a known issue

jmurray87:
Won't affect the change we made for the filter options if its just in the demo,  the change I made still respects the pageSize query param, but sets it to max int if its not present

mtjandra:
> Won't affect the change we made for the filter options if its just in the demo, the change I made still respects the pageSize query param, but sets it to max int if its not present

That makes the default use case not ideal, since you would no longer see load more. This is for demo purposes only

jmurray87:
The change I made only applies to the filters/{guid} route

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
svanherk: Yup, it's currently broken and tracked in a defect, but this is just making things worse (if prod ever was fixed, we'd want our demos to work.  There's a polyfill, but that's just hiding problems and should be removed).

svanherk:
> The change I made only applies to the filters/{guid} route

But won't the query param be passed down to that route?  Like if the my-unaccessed-activities route has a `pageSize`, does the filter link also have that? (this is what I was unsure about, it may be totally fine)

demo/d2l-evaluation-hub/d2l-evaluation-hub-activities-list.html
mtjandra: Yes IE and fixing filtering demo data should be handled seperatley from what I'm doing here. Thanks for the heads up!

mtjandra:
> But won't the query param be passed down to that route? Like if the my-unaccessed-activities route has a pageSize, does the filter link also have that? (this is what I was unsure about, it may be totally fin

Good call out guys. I will definitely be thinking of this once it comes to the implementation!

demo/d2l-evaluation-hub/demo-data-parser.js
```diff
@@ -349,7 +349,11 @@ function getHrefForMasterTeacher(id) {
 *
 * `mappings` (which is the return value) maps urls to siren endpoints to be consumed by the interceptor
 */
-function getMappings(table) {
+function getMappings(table, pageSize) {
+	if (!pageSize) {
```
AlexBedley: IMO we should make the second param to `getMappings` something like `options` and `pageSize` should be a member of that object. Just so when someone adds another config thing they don't add another parameter

demo/d2l-evaluation-hub/d2l-evaluation-hub.html
```diff
@@ -21,7 +21,9 @@
 			import mappingsParser from './demo-data-parser';
 			import data from './table';
 
-			const mappings = mappingsParser(data);
+			const url = new URL(window.location);
+			const pageSize = url.searchParams.get('pageSize') || 3;
```
AlexBedley: I find it weird that we have 2 places with a default value. I would remove one of them

mtjandra: Alright updated when fixing merge conflicts

demo/d2l-evaluation-hub/demo-data-parser.js
mtjandra: Alright updated when fixing merge conflicts

mtjandra:
> It doesn't handle the cases where you want a dynamic page size but it's better than it was before! 

Correct, next PR will handle dynamic page sizes, this is step one

PR #125, opened 2019-03-19
Fix sort demo bug

mtjandra:
* Looks like I made a mistake at some point
* This should fix the demo for sorting

### DIFF
--- a/demo/d2l-quick-eval/sort-handler.js
+++ b/demo/d2l-quick-eval/sort-handler.js
@@ -97,10 +97,11 @@ function formatSort(sort, sortsHref, ascState, descState, sortState) {
 		]
 	};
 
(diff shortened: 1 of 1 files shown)


demo/d2l-evaluation-hub/sort-handler.js
```diff
@@ -97,10 +97,11 @@ function formatSort(sort, sortsHref, ascState, descState, sortState) {
 		]
 	};
 
-	if (sortState[0] && sortState[0].id === sort.id) {
+	var lastSort = sortState.length - 1;
```
AlexBedley: nit `const`

mtjandra: Updated while fixing merge conflicts

PR #124, opened 2019-03-19
Us104726/demo data bookmark

mtjandra:
* Builds on top of https://github.com/BrightspaceHypermediaComponents/activities/pull/123
* Use bookmark query parameter to know what activity to start from
* Use pageSize dynamically

PR #127, opened 2019-03-20
US104726 - Retain number of shown activities on sort

mtjandra:
**Note:** Demo only works after https://github.com/BrightspaceHypermediaComponents/activities/pull/124

retainNumActivities

* Retain number of activities shown when sorting

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -426,7 +426,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				return action;
 			}).bind(this))
 			.then((collectionAction => {
-				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : {};
+				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : undefined;
 				const collection = this._performSirenActionWithQueryParams(collectionAction, customParams);
 				return collection;
 			}).bind(this))
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -261,6 +261,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				type: Array,
 				value: [ ]
 			},
+			_numberOfCurrentlyShownActivities: {
+				type: Number,
+				computed: '_computeNumberOfCurrentlyShownActivities(_data)'
+			},
 			_fullListLoading: {
 				type: Boolean,
 				value: true
@@ -307,6 +311,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 	setLoadingState(state) {
 		this.set('_fullListLoading', state);
 		this.set('_loading', state);
+	}
+
+	_computeNumberOfCurrentlyShownActivities(data) {
+		return data.length;
 	}
 
 	_myEntityStoreFetch(url) {
@@ -418,7 +426,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				return action;
 			}).bind(this))
 			.then((collectionAction => {
-				const collection = this._performSirenActionWithQueryParams(collectionAction);
+				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : {};
+				const collection = this._performSirenActionWithQueryParams(collectionAction, customParams);
 				return collection;
 			}).bind(this))
 			.then((collection => {
@@ -682,7 +691,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		);
 	}
 
-	_performSirenActionWithQueryParams(action) {
+	_performSirenActionWithQueryParams(action, customParams) {
 		const url = new URL(action.href, window.location.origin);
 
 		if (!action.fields) {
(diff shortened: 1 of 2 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -418,7 +426,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				return action;
 			}).bind(this))
 			.then((collectionAction => {
-				const collection = this._performSirenActionWithQueryParams(collectionAction);
+				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : {};
+				const collection = this._performSirenActionWithQueryParams(collectionAction, customParams);
```
mtjandra: HREF on web-component never gets updated. So this `pageSize` param should not be propagated to filters or main WC. I tested on LMS as well.

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -65,6 +65,15 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
 		done();
 	}
 
+	function arrayContainsWhere(arr, f) {
```
AlexBedley: I think even ie supports [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) (although ironically it does not support `includes`)

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -410,6 +419,27 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
+
+			const action = {
+				href : 'http://127.0.0.1/',
+				name: 'apply',
+				type: 'application/x-www-form-urlencoded',
+				method: 'GET'
+			};
+
+			const customParams = { customParam1: 'custom', customParam2: 'custom2' };
+			sinon.stub(list, 'performSirenAction', function(passedAction) {
```
AlexBedley: nit: I don't mind the current implementation but `sinon` has some cool built-ins you could use. You seem to be familiar with the [stub API](https://sinonjs.org/releases/latest/stubs/) but there's also the [spy API](https://sinonjs.org/releases/latest/spies/)

For example:

```js
const expectedAction = ...; // whatever
const spy = sinon.spy(list, 'performSirenAction');

list._performSirenActionWithQueryParams(action, customParams);

assert(spy.withArgs(expectedAction).calledOnce);
```

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -418,7 +426,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				return action;
 			}).bind(this))
 			.then((collectionAction => {
-				const collection = this._performSirenActionWithQueryParams(collectionAction);
+				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : {};
```
AlexBedley: What case does the `> 0` case handle? If we were sorting from an error state?

nit: Also `{}` evaluates to true so we are entering that `if (customParams)` block in the function below

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -261,6 +261,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				type: Array,
 				value: [ ]
 			},
+			_numberOfCurrentlyShownActivities: {
+				type: Number,
+				computed: '_computeNumberOfCurrentlyShownActivities(_data)'
```
AlexBedley: Curious as to why you went this way vs just using `data.length` everywhere (I don't know anything about polymer `computed` properties)

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Thank you!!! I will this out

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Thanks for the reference! Although I'm not sure what this buys us. Though I do notice now that I probably want to add a `done()` within the stub.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: > What case does the > 0 case handle? If we were sorting from an error state?

Yes I was thinking if say we received an error on load, so there is no information in the data. Then someone clicks one of the header, and the data loads correctly. I wouldn't want to maintain showing nothing, but instead whatever the default is.

> nit: Also {} evaluates to true so we are entering that if (customParams) block in the function below

 This is why I dislike JS. I will change it to undefined thanks

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: I just wanted to name `data.length` as `_numberOfCurrentlyShownActivities`. I doubt the meaning of data.length will ever change, but just incase it does. It seemed to make sense in my head that our component knows how many activities are shown, so a property seemed good. Might just be personal preference though

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -455,6 +476,44 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
...
+				fields : defaultParams
+			};
+
+			const customParams = { customParam1: 'custom', customParam2: 'custom2' };
+			sinon.stub(list, 'performSirenAction', function(passedAction) {
+				const fields = passedAction.fields;
+				assert.equal(4, fields.length);
+
+				assert.isTrue(fields.some(function(elm) { return elm.name === 'testname' && elm.value === 'testvalue'; }));
+				assert.isTrue(fields.some(function(elm) { return elm.name === 'anothertestname' && elm.value === 'anothertestvalue'; }));
```
mtjandra: `[1, 2, 3].some((e) => e === 2)` throws in IE11 console. I'm not even sure how the test suites run.... but that is why I didn't use anonymous functions

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Hmm yeah it's not a revolutionary refactor - I just find using the built-ins more readable. One equality assertion is likely easier to reason with than a series of custom assertions. But like I said, I'm ok with the current implementation.

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Super weird, I wonder if some transpiling is going on at a certain level. the magic of `polymer test`

PR #128, opened 2019-03-20
US105009 - Switch from using displayName to concatenation of first na…

mtjandra:
* This is step one to achieving the swap between first name and last name on sort
* Instead of using display name, we will get the firstName and lastName individually and then concat them together
* When you can't see either first name or last name, you will see "Anonymous User"
* This change shouldn't be introducing any new behavior, should be acting the same way as before
* Tested against LMS as well, seems like things are working as intended

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -653,8 +653,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		return this._followLink(entity, Rels.user)
 			.then(function(u) {
 				if (u && u.entity) {
-					const firstName = this._tryGetName(u.entity, Rels.firstName, '');
-					const lastName = this._tryGetName(u.entity, Rels.lastName, '');
+					const firstName = this._tryGetName(u.entity, Rels.firstName, null);
+					const lastName = this._tryGetName(u.entity, Rels.lastName, null);
 					const defaultDisplayName = this._tryGetName(u.entity, Rels.displayName, '');
 
 					const displayName = {
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -160,12 +160,12 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 									<template is="dom-if" if="[[s.userHref]]">
 										<d2l-profile-image class="d2l-user-badge-image" href="[[s.userHref]]" token="[[token]]" small=""></d2l-profile-image>
 									</template>
-									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[localize('evaluate', 'displayName', s.displayName)]]</d2l-offscreen>
+									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[_localizeEvaluationText(s)]]</d2l-offscreen>
 									<d2l-link
-										title="[[localize('evaluate', 'displayName', s.displayName)]]"
+										title="[[_localizeEvaluationText(s)]]"
 										aria-describedby$="d2l-quick-eval-activities-list-username"
 										href="[[s.activityLink]]"
-									>[[_getDataProperty(s, 'displayName')]]</d2l-link>
+									>[[_formatDisplayName(s)]]</d2l-link>
 									<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
 								</d2l-td>
 								<d2l-td class="d2l-quick-eval-truncated-column d2l-activity-name-column">
@@ -606,11 +606,20 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 			});
 	}
 
-	_formatDisplayName(
-		firstName,
-		lastName,
-		defaultDisplayName
+	_localizeEvaluationText(
+		data
 	) {
+		const formattedDisplayName = this._formatDisplayName(data);
+		return this.localize('evaluate', 'displayName', formattedDisplayName);
+	}
+
+	_formatDisplayName(
+		data
+	) {
+		const firstName = data.displayName.firstName;
+		const lastName = data.displayName.lastName;
+		const defaultDisplayName = data.displayName.defaultDisplayName;
+
 		if (!lastName && !firstName) {
 			return defaultDisplayName;
 		}
@@ -648,11 +657,11 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 					const lastName = this._tryGetName(u.entity, Rels.lastName, '');
 					const defaultDisplayName = this._tryGetName(u.entity, Rels.displayName, '');
 
-					const displayName = this._formatDisplayName(
-						firstName,
-						lastName,
-						defaultDisplayName
-					);
+					const displayName = {
+						'firstName': firstName,
+						'lastName': lastName,
+						'defaultDisplayName': defaultDisplayName
+					};
 
 					item.displayName = displayName;
 				}
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -606,13 +606,57 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 			});
 	}
 
+	_formatDisplayName(
+		firstName,
+		lastName,
+		defaultDisplayName
+	) {
+		if (!lastName && !firstName) {
+			return defaultDisplayName;
+		}
+		if (!lastName) {
+			return firstName;
+		}
+		if (!firstName) {
+			return lastName;
+		}
+		return firstName + ' ' + lastName;
+	}
+
+	_tryGetName(
+		entity,
+		rel,
+		defaultValue
+	) {
+		if (!entity || !entity.hasSubEntityByRel(rel)) {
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,57 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+		entity,
+		rel,
+		defaultValue
+	) {
+		if (!entity || !entity.hasSubEntityByRel(rel)) {
+			return defaultValue;
+		}
+
+		const subEntity =  entity.getSubEntityByRel(rel);
+		if (!subEntity || !subEntity.properties || subEntity.hasClass('default-name')) {
```
mtjandra: class `default-name` is applied to either firstName or lastName when the user does not have permissions to view the name. It is never applied to display name

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,57 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
 		return this._followLink(entity, Rels.user)
 			.then(function(u) {
-				if (u && u.entity && u.entity.hasSubEntityByRel(Rels.displayName)) {
-					item.displayName = u.entity.getSubEntityByRel(Rels.displayName).properties.name;
+				if (u && u.entity) {
+					const firstName = this._tryGetName(u.entity, Rels.firstName, '');
+					const lastName = this._tryGetName(u.entity, Rels.lastName, '');
+					const defaultDisplayName = this._tryGetName(u.entity, Rels.displayName, '');
+
+					const displayName = this._formatDisplayName(
```
mtjandra: I'm going to follow up with one more commit, I think I need to do the formatting when rendering to html rather than in here.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,66 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+					const lastName = this._tryGetName(u.entity, Rels.lastName, '');
+					const defaultDisplayName = this._tryGetName(u.entity, Rels.displayName, '');
+
+					const displayName = {
+						'firstName': firstName,
+						'lastName': lastName,
+						'defaultDisplayName': defaultDisplayName
+					};
+
+					item.displayName = displayName;
```
mtjandra: I should probably rename `item.displayName` to `item.nameInfo` or something. But will do that separate. Diff is getting kind of large.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,66 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+		if (!lastName && !firstName) {
+			return defaultDisplayName;
+		}
+		if (!lastName) {
+			return firstName;
+		}
+		if (!firstName) {
+			return lastName;
+		}
+		return firstName + ' ' + lastName;
```
AlexBedley: Does this assume `displayName === firstName + ' ' + lastName`? Is this just a north american thing? For example I know all my asian cousins write their name `lastName + ' ' + firstName`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,66 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+		return subEntity.properties.name;
+	}
+
 	_getUserPromise(entity, item) {
 		return this._followLink(entity, Rels.user)
 			.then(function(u) {
-				if (u && u.entity && u.entity.hasSubEntityByRel(Rels.displayName)) {
-					item.displayName = u.entity.getSubEntityByRel(Rels.displayName).properties.name;
+				if (u && u.entity) {
+					const firstName = this._tryGetName(u.entity, Rels.firstName, '');
```
AlexBedley: I'm not a huge fan of relying on the fact that `''` evaluates to `false` in terms of truthiness - could be confusing to certain readers

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,66 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+					const lastName = this._tryGetName(u.entity, Rels.lastName, '');
+					const defaultDisplayName = this._tryGetName(u.entity, Rels.displayName, '');
+
+					const displayName = {
+						'firstName': firstName,
```
AlexBedley: nit: should we update the eslint to treat optionally quoted members as an error?

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -553,5 +554,100 @@ import '@polymer/iron-test-helpers/mock-interactions.js';
 			assert.equal(evalLink, expectedEvalLink);
 		});
 
+		test('_formatDisplayName return firstName when firstName defined and lastName undefined', () => {
```
AlexBedley: I love seeing function level unit tests! You could wrap each function's tests in a sub `suite` so the top level isn't so polluted

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -606,13 +606,66 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 			});
 	}
 
+	_localizeEvaluationText(
```
AlexBedley: in general: should we be creating helper files and importing them instead of creating this huge monolith? It might make the file easier to read if we choose good interfaces.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Do we need to change the sorts? I think they use the firstName and lastName separately already but just checking.

I'm super worried about getting this right with UIP and everything but if you've already tested on a live LMS that should be good enough

mtjandra: Nope shouldn't need to. Sorting functionality is independent of frontend, that is all handled server side. With UIP if you cannot see the firstName or lastName, then it is treated as the lowest thing when comparing against other firstName or lastNames.

I'm sure some more e2e testing around edge cases you are concerned of is appropriate to do sometime.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: I'll need to double check how LMS handles it today. I know if profile image this kind of localization was not supported. I'm not sure if we can support it "nicely" in our app. That was the purpose for using displayName in the first place, because we should rely on the LMS to format it correctly.

mtjandra: I'm almost 100% sure our naming doesn't capture all cases in our ideal world. Names are pretty hard, and having to support this flip front-end side right now is just asking for trouble

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Sounds good will use `null` instead

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: I'm indifferent, feel free to put up a PR for this if you like!

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: I think we have tech debt stories to clean up unit tests. There it a  bit of house keeping we can def do!

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Yes I want to do some house keeping here as well! File is pretty large...., I'm thinking we can do with a tech debt story

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: From what I can tell LMS doesn't take localization into effect for firstname or lastName. http://search.dev.d2l/source/xref/Lms/lp/framework/core/D2L.LP/LP/Users/Privacy/Domain/Default/UserPrivacyFormatter.cs#82

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: I created [US105219](https://rally1.rallydev.com/#/42960320374d/detail/userstory/295709733836?fdp=true)

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: It looks like there is some "localization" logic in the sense that you can request different formats (`DisplayNameFormat`). Ideally our app should be able to tell the name component we want "LastName FirstName". That component should be able to request `DisplayNameFormat.LastNameFirstName` from the API and display that instead (some kind of hm endpoint representing name with formatting actions)

I know this is just one step on the stairs. What's the future plan for this column's data?

mtjandra: Plan for column data is to keep the same, but to format different based on sort, I'll have you on that one soon.

And yeah I like the idea of requesting different name formats from API. Maybe I can look into switching to that!

PR #130, opened 2019-03-21
Us105009/NameSwap

mtjandra:
nameSwap

* Depends on https://github.com/BrightspaceHypermediaComponents/activities/pull/128
* Swap firstName and lastName on sort
* Test in LMS that the proper ordering is maintained when going back to QE from an evaluation page

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -162,12 +162,12 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 									<template is="dom-if" if="[[s.userHref]]">
 										<d2l-profile-image class="d2l-user-badge-image" href="[[s.userHref]]" token="[[token]]" small=""></d2l-profile-image>
 									</template>
-									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[_localizeEvaluationText(s)]]</d2l-offscreen>
+									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]</d2l-offscreen>
 									<d2l-link
-										title="[[_localizeEvaluationText(s)]]"
+										title="[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]"
 										aria-describedby$="d2l-quick-eval-activities-list-username"
 										href="[[s.activityLink]]"
-									>[[_formatDisplayName(s)]]</d2l-link>
+									>[[_formatDisplayName(s, _headerColumns.0.meta.firstThenLast)]]</d2l-link>
 									<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
 								</d2l-td>
 								<d2l-td class="d2l-quick-eval-truncated-column d2l-activity-name-column">
@@ -236,6 +236,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				value: [
 					{
 						key: 'displayName',
+						meta: { firstThenLast: true },
 						headers: [
 							{ key: 'firstName', sortClass: 'first-name', suffix: ',', canSort: false, sorted: false, desc: false  },
 							{ key: 'lastName', sortClass: 'last-name', canSort: false, sorted: false, desc: false  }
@@ -316,7 +317,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		this.set('_loading', state);
 	}
 
-
 	_computeNumberOfCurrentlyShownActivities(data) {
 		return data.length;
 	}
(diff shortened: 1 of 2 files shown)


mtjandra:
I like the idea referenced at https://github.com/BrightspaceHypermediaComponents/activities/pull/128#discussion_r268286795

I will need to think about it a bit more, but it does seem more appropriate. I may transition to that approach afterward.

mtjandra:
Separate concern: Right now there is no comma between the first and last name in the design. I suspect design will want to change that, but will confirm with them.

AlexBedley:
> Separate concern: Right now there is no comma between the first and last name in the design. I suspect design will want to change that, but will confirm with them. The problem with not having a column, is you don't know when the firstName/lastName ends/starts if there are spaces

I would just put the comma, it looks super weird without it. I agree that design will probably want to change it anyways. The assignments page (which is where the dual name sorting comes from) also has commas:

image

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -295,7 +296,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 	static get observers() {
 		return [
 			'_loadData(entity)',
-			'_loadSorts(entity)'
+			'_loadSorts(entity)',
+			'_handleNameSwap(_headerColumns.0.headers.*)'
```
AlexBedley: while I do like the approach of listening for changes to the actual header state, is there a more robust way we can listen? This approach heavily relies on not re-ordering the headers. Or at the very least add a unit test ensuring that your assumptions must be true, ie: _headerColumn.0 is the name column etc.

mtjandra:
>I would just put the comma, it looks super weird without it. I agree that design will probably want to change it anyways. The assignments page (which is where the dual name sorting comes from) also has commas:

Spoke with Erin. Looks like we only want comma when it is lastName then firstName, but not when it is firstName than lastName.

I will make that change

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: A few suggestions:

* we could change the data type from Array to Object and fetch headers using their keys (enumerate using `Object.keys` or make a `forEachHeader` helper function)
* You could use a `dom-if` and render the name column differently based on header state

AlexBedley: Actually I changed my mind again, the ordering of `_headerColumn` is not gonna change since that's the order they are rendered. I would just add a test ensuring the order and move on

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -319,6 +321,17 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		return data.length;
 	}
 
+	_handleNameSwap(entry) {
```
AlexBedley: So you could unit test this function by stubbing out `this.set`. It might be worth an integration level test (with simulated clicks etc.) for the entire behaviour.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -236,6 +236,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				value: [
 					{
 						key: 'displayName',
+						meta: { firstThenLast: true },
```
AlexBedley: nit: I'm not a huge fan of the `meta` name, it feels redundant - it belongs at the top level (alongside `key`) IMO

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Yeah exactly. I believe it is an array to do the `dom-repeat`, and the positions do matter so they shouldn't change. But you are right about the unit test. I'll look into adding them if they don't exist already

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Yeah let me look into integration level test

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Yeah I wasn't sure what to do with this. I didn't have meta before. But `firstThenLast` is specific to the  header, and I wanted all keys to be same across headers.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: TestAdded

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -708,5 +725,43 @@ import SirenParse from 'siren-parser';
 			assert.equal(name, expectedName);
 		});
 
+		test('firstName begins before lastName, clicking lastName puts it before firstName and clicking firstName puts it before lastName', (done) => {
+
+			var nameHeaders = list._headerColumns[0].headers;
```
AlexBedley: nit: `const`

mtjandra: Since this is the only thing, I will try to remember to clean this up next time around

PR #133, opened 2019-03-22
US105014 - Set relative column widths based on design; account for teacher column

jeroach:
**US105014**

Had a chat with Erin re: desired column widths. Details in the story: https://rally1.rallydev.com/#/238990806412d/detail/userstory/294463117992

TL;DR
With `master-teacher` on: 25 / 20 / 20 / 15 / 20
With `master-teacher` off: 30 / 25 / 25 / 20

Pics:
**List with items**
column_widths_items

**List with items and master teacher on**
column_widths_items_with_teacher

**List without items**
column_widths_no_items


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -662,6 +674,41 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		return result;
 	}
 
+	_getWidthCssClass(columnKey) {
+		if (this.masterTeacher) {
+			switch (columnKey) {
+				case 'displayName':
+					return 'd2l-quick-eval-25-column';
+					break;
```
AlexBedley: nit: don't need `break` because of the `return` but I'm not sure what good styling looks like for `switch` statements anymore

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -662,6 +674,41 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		return result;
 	}
 
+	_getWidthCssClass(columnKey) {
```
AlexBedley: This function needs unit tests

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -662,6 +674,41 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+		if (this.masterTeacher) {
+			switch (columnKey) {
+				case 'displayName':
```
AlexBedley: It might be time to start saving these headers keys as constants - maybe in a follow up PR

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -662,6 +674,41 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+				case 'activityName':
+				case 'courseName':
+				case 'masterTeacher':
+					return 'd2l-quick-eval-20-column';
+					break;
+				case 'submissionDate':
+					return 'd2l-quick-eval-15-column';
+					break;
+				default:
+					return '';
```
AlexBedley: We should `throw` or otherwise signal an error IMO because something is really wrong if we reach here

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Oh I see you fixed this while I was reviewing. All good.

jeroach: yeah this blew up the linter - fixed already

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: yep, will do

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: True

vix633:
Adding myself so i can adjust the skeleton when this ships

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: Plan to defer to a follow up PR; should chat with team if we want a global constants file or specific to each file

AlexBedley: I think we should limit the constants to the smallest scope possible - I think just this file needs them.

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -612,7 +612,42 @@ import SirenParse from 'siren-parser';
...
+				assert.equal(err.message, 'Invalid column key');
+			}
+		});
+		test('_getWidthCssClass throws an error when passed an invalid column key (with master teacher on)', () => {
+			list.masterTeacher = true;
+
+			try {
+				list._getWidthCssClass('notARealColumnKey');
+			} catch (err) {
+				assert.equal(err.message, 'Invalid column key');
```
AlexBedley: tip: you can do `assert.throws(() => list._getWidthCssClass('notARealColumnKey'), 'Invalid column key');`

CarolBusman: I would like to see this added. I think it makes it more obvious that we're testing to make sure that an error explicitly _does_ get thrown in this instance.

jeroach:
Needed to commit what I had in order to merge with master and tidy up.

Fixing lint issues now

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -612,7 +612,42 @@ import SirenParse from 'siren-parser';
 			var evalLink = list._buildRelativeUri(url, params);
 			assert.equal(evalLink, expectedEvalLink);
 		});
+		test('_getWidthCssClass returns correct value when passed column key (with master teacher off)', () => {
+			const validColumnKeys = ['displayName', 'activityName', 'courseName', 'submissionDate'];
+			const expectedCssClasses = ['d2l-quick-eval-30-column', 'd2l-quick-eval-25-column', 'd2l-quick-eval-25-column', 'd2l-quick-eval-20-column'];
 
+			for (let i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
jeroach: Tried this - the `masterTeacher` property within the call to `_getWidthCssClass()` was undefined.

Going to fiddle with it a bit more, but shouldn't hold up this PR.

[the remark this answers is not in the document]
AlexBedley: Ah you might have to .bind it like:
```js
const actualCssClasses = validColumnKeys.map(list._getWidthCssClass.bind(list));
```

PR #135, opened 2019-03-26
Update readme so tests can run again!

m6jones:


m6jones:
I need better page for how to setup Artifactory. I already set it up on both my computers. Not sure how I did it anymore.

jeroach:
Can this happen automatically via `npm i` - e.g. saved as a dev dependency?

jeroach:
FYI this does work locally :) Just wondering if we can skip a step

m6jones:
Sadly it doesn't seem to be working for everyone. Will look into make it better.

m6jones:
Disabling attest right now seems to be the best option :(

PR #136, opened 2019-03-26
sorting unit test refactor

AlexBedley:
https://rally1.rallydev.com/#/238990806412d/detail/userstory/294463131584

This PR wants to accomplish 2 things:

1. put back in sorting tests after they were gutted to make the shipping deadline
2. create an example of tests we think are good so we can link back to them and spread knowledge through our team of good testing practices in polymer

It definitely accomplishes `1.` but I'd like your ruthless review to ensure `2.`. Thanks!


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -448,17 +447,13 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				if (!action) {
 					return Promise.reject(new Error(`Could not find apply action in ${sortsEntity}`));
 				}
-				return action;
-			}).bind(this))
-			.then((collectionAction => {
```
AlexBedley: This `then` separation that I originally put in didnt make sense. I just got rid of the extra `then`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -448,17 +447,13 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
 				const customParams = this._numberOfCurrentlyShownActivities > 0 ? {pageSize: this._numberOfCurrentlyShownActivities} : undefined;
-				const collection = this._performSirenActionWithQueryParams(collectionAction, customParams);
-				return collection;
+				return this._performSirenActionWithQueryParams(action, customParams);
 			}).bind(this))
 			.then((collection => {
 				this.entity = collection;
 				this._dispatchSortUpdatedEvent(collection);
-				return Promise.resolve();
+				return Promise.resolve(collection);
```
AlexBedley: Our functions should return things that make sense for testability purposes

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.html
```diff
@@ -17,7 +17,7 @@
 	<body>
 		<test-fixture id="basic">
 			<template strip-whitespace>
-				<d2l-quick-eval-activities-list href="data/sorting/initialEntity.json" token="testtoken"></d2l-quick-eval-activities-list>
+				<d2l-quick-eval-activities-list></d2l-quick-eval-activities-list>
```
AlexBedley: There was some weird timing issues with loading the fake data so I just got rid of it and the timing issues went away. I just noticed I should delete those unused fake data files now.

- [ ] delete unused test files

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -11,37 +11,16 @@ function resetSortHeaders(list) {
 	});
```
AlexBedley: This file is unreviewable with the unified diff. I recommend opening the file in a new window

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
+				}
+			];
+			testData.forEach(testCase => {
+				test(testCase.name, (done) => {
+					var stub = sinon.stub(list, '_fetchSortedData');
+					const e = {
+						currentTarget: testCase.currentTarget
+					};
+
+					list._updateSortState(e)
```
AlexBedley: I tried for probably way too long to get `chai-as-promised` working with `polymer test` and I could not (it came down to the fact that it did not have an es6 wrapper so you can't `import` it)

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
 
-	test('_loadSorts silently does nothing on null entity', () => {
-		return list._loadSorts(null);
-	});
-
-	test('_loadSorts determines which headers are sortable', () => {
-		const enabledSortClasses = ['activity-name', 'course-name'];
-		enableSorts(list, enabledSortClasses);
+	suite('_loadSorts', () => {
+		test('_loadSorts silently does nothing on null entity', () => {
```
mtjandra: super nit: only mentioning because you are focusing on testing. Technically this doesn't verify _loadSorts does nothing. Not sure how to effectively test absence of "everything". This just verifies that sending in a null empty does not result in a promise rejection.

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
+						.returns(testCase.sortEntity);
+
+					list._loadSorts(entity)
+						.then(() => {
+							done('_loadSorts should have rejected');
+						})
+						.catch(() => {
+							done();
+						})
+						.catch((err) => {
```
mtjandra: I don't think this catch will ever be used. Unless there is some condition where `done()` would throw

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
-				done('_updateSortState should have rejected');
-			})
-			.catch((err) => {
-				expect(err.toString()).to.equal('Error: Could not find sortable header for undefined');
-				done();
-			})
-			.catch((err) => {
-				done(err);
-			});
+					expect(enabledSorts).to.have.same.members(enabledSortClasses.map(x => x.className));
```
mtjandra: What a sentence! 😲

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
+				{
+					name: 'unsortable header',
+					currentTarget: 'activityName'
+				}
+			];
+			testData.forEach(testCase => {
+				test(testCase.name, (done) => {
+					var stub = sinon.stub(list, '_fetchSortedData');
+					const e = {
+						currentTarget: testCase.currentTarget
```
mtjandra: I think you intended to have `currentTarget: {id: testCase.currentTarget}`

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
```diff
@@ -57,148 +36,203 @@ suite('d2l-quick-eval-activities-list-sorting', () => {
...
+			const loadDataStub = sinon.stub(list, '_loadData');
+			const loadSortsStub = sinon.stub(list, '_loadSorts');
+
+			followLinkStub.withArgs(list.entity, Rels.sorts).returns(Promise.resolve({ entity: sorts }));
+			performActionStub.withArgs(sortAction).returns(sorts);
+			performActionStub.withArgs(applyAction, undefined).returns(collection);
+
+			return list._fetchSortedData('activity-name', false)
+				.then(actual => {
+					expect(sortUpdatedStub.withArgs(collection).calledOnce).to.be.true;
```
mtjandra: Nice

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: I love this approach, I think it will make things way easier to test. I want to take it to the next step - the code that doesn't need to know about polymer (the "pure" functions) should ideally be in a different file completely.

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.html
AlexBedley: >  I wonder if this was the cause of you needing to use resetSortHeaders.

I thought so too but unfortunately that wasn't it. Still tracking this one down.

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
AlexBedley: Hmm true I can fix this up by ensuring this._followLink is never called and change the test name to be less confident we do nothing

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
AlexBedley: oops yes I deleted an assert on the error name but not the associated catch. will fix

- [ ] delete unused `catch`

test/d2l-quick-eval/d2l-quick-eval-activities-list-sorting.js
AlexBedley: oops good catch

- [ ] use `id`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Yup makes sense to me. I think all of "polymer" is side-effects, I'm not sure what is "polymer" that isn't a side-effect.

PR #137, opened 2019-03-26
Enforce usage of `const`/`let` over `var`

AlexBedley:
We shouldn't use `var` because it causes surprises due to it being function-scoped (vs block-scoped which is what C# developers are used to).

`const` allows developers to catch runtime errors at "compile-time"
`let` has block-scope, which many developers are used to and can avoid hard to catch bugs such as the one below (stolen from the eslint docs)

```js
var count = people.length;
var enoughFood = count > sandwiches.length;

if (enoughFood) {
    var count = sandwiches.length; // accidentally overriding the count variable
    console.log("We have " + count + " sandwiches for everyone. Plenty for all!");
}

// our count variable is no longer accurate
console.log("We have " + count + " people and " + sandwiches.length + " sandwiches!");
```

Some environments don't fully support it but I think we are shipping to the ones that do:
* https://caniuse.com/#feat=const
* https://caniuse.com/#feat=let

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,10 @@
   "parserOptions": {
       "sourceType": "module"
   },
+  "rules": {
+    "no-var": "error",
+    "prefer-const": "error"
+  },
   "globals": {
     "fastdom": false
   }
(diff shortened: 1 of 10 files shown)


AlexBedley:
@m6jones if this doesn't work for your team, I'm happy to explore alternative options (the simplest being to just create an `.eslintrc.json` in our specific subdirectories)

This change is definitely not necessary or blocking anything, so I'm happy to decline it in the face of a real issue, but I think it encourages good development practices and I would like to get it in.

m6jones:
The only issue I see is that it should have already been in. 
`"extends": "brightspace/polymer-3-config"`
Should include this in the linting rules. I actually thought it was already in anyways nice!

PR #138, opened 2019-03-27
Cbusman/add search box

CarolBusman:
The search box and message area will not be shown unless the `search-enabled` attribute is added to the quick eval component, so we should be safe to add this in without the search stuff showing up before we're ready since the renderer does not add this attribute currently.

The visibility of the search will be controlled by the `quick-eval-enable-search` feature flag.


jeroach:
Can we get some screenshots of before and after (for the various cases)?

CarolBusman:
Quick Eval without the `search-enabled` attribute:
image
Quick Eval with the `search-enabled` attribute, and `_showSearchResultSummary` set to `false`:
image
Quick Eval with the `search-enabled` attribute and `_showSearchResultSummary` set to `true`:
image

jeroach:
Screenshots look good! One question - there looks to be a block of whitespace between the bottom of the search results summary and the top of the activity list. Is this intentional?

I ask because the spacing seems tighter when the search results summary is not displayed.

package.json
```diff
@@ -63,6 +63,7 @@
 		"d2l-organizations": "BrightspaceHypermediaComponents/organizations#semver:^1",
 		"d2l-polymer-behaviors": "Brightspace/d2l-polymer-behaviors-ui#semver:^2",
 		"d2l-polymer-siren-behaviors": "Brightspace/polymer-siren-behaviors#semver:^1",
+		"d2l-search-widget": "Brightspace/d2l-search-widget-ui#semver:^4",
```
svanherk: This is one of those "black market" components that will be deprecated soon, since it doesn't even use `d2l-input-search` or the `entity-store` under the hood.  What this should actually be is a `d2l-hm-search` component in https://github.com/BrightspaceHypermediaComponents/common (exactly like the filter, using the UI component (https://github.com/BrightspaceUI/inputs/blob/master/d2l-input-search.js) and the siren behaviours (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js))

CarolBusman:
Actually, I think it should be farther. [Rally](https://rally1.rallydev.com/#/238990806412d/detail/userstory/294844044608) shows a distance of 54px from the bottom of the message container to the top of the text in the list. As there's 12px of padding around the list headers, the message container was given a `margin-bottom` of 32px (because I can't math: that should be 42px).

package.json
svanherk: I'd talked to @AlexBedley about being able to use the exact same component here and in the filter, not sure if that's in scope for the first slice but I think you'd be better off directly using `d2l-input-search` here and then pulling it out into a component, rather than using a non-entity-store component with a different UI

CarolBusman: I can make a basic hm wrapper for now that doesn't do anything, since we don't know what the api will look like.

svanherk: That'd be awesome, then we can at least point people there if they need this.  Maybe they'll even help build it 😛.  `my-courses` is the main user of the other one, hoping to get that switched soon as part of the refactor

jmurray87: I have a PR up for the api implementation, reach out if you need to know what the api looks like!

jeroach:
> Actually, I think it should be farther. Rally shows a distance of 54px from the bottom of the message container to the top of the text in the list. 

Thanks Carol! I had not seen the design before :)

package.json
CarolBusman: Will do @jmurray87; that'll be the next step. First we're just making sure that it looks prettiful :)

jeroach: Can we remove this reference now? 😄

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -155,6 +194,13 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this.entity = e.detail.sortedActivities;
 	}
 
+	_searchResultsChanged(/*e*/) {
+
```
jeroach: When is it expected that these two methods will be implemented? If it's going to be a bit, perhaps we should add a comment re: expected behaviour?

CarolBusman: They're going to be implemented after we get the `d2l-hm-filter` hooked up to the api and we know what the event comes back looking like. That's the next step once the api search gets done (which sounds like it's close to being done?), so it's not going to stay like this for long.

jeroach: Sounds good - no action needed 😄

components/d2l-quick-eval/d2l-quick-eval-search-results-container.js
```diff
@@ -0,0 +1,69 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import {QuickEvalLocalize} from './QuickEvalLocalize.js';
+import 'd2l-link/d2l-link.js';
+
+class D2LQuickEvalSearchResultsContainer extends QuickEvalLocalize(PolymerElement) {
```
AlexBedley: nit: I would toss the word `Summary` in there somewhere. My initial read of the name made me think this was where the search results would be displayed.

components/d2l-quick-eval/d2l-quick-eval-search-results-container.js
```diff
@@ -0,0 +1,69 @@
...
+			}
+			span {
+				margin-right: 2.3em;
+			}
+			:host(:dir(rtl)) span {
+				margin-right: 0;
+				margin-left: 2.3em;
+			}
+		</style>
+		<div class="d2l-msg-container" hidden$="[[!_showSearchResultSummary]]">
```
AlexBedley: Where does `_showSearchResultSummary` come from? Is it missing from `properties`?

Also it feels weird that this component is responsible for whether it's displayed or not. I would expect that to be a responsibility of the parent. Can we refactor this logic out of this component?

edit: It looks like we actually are doing this in the parent already. `hidden$="[[!_searchResultsMessageEnabled()]]`. We should just delete this code then

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -84,6 +107,14 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 			_showFilterError: {
 				type: Boolean,
 				value: false
+			},
+			_showSearchResultSummary: {
```
AlexBedley: Does this variable need to exist? We set it to `true` and never change it

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: This might just be a difference of opinion but I would delete these empty functions and the event listener hook ups. They are trivial to add back when you need them but could just languish here for a long time if we don't implement search right away (or things change and the person that picks up the story just blindly follows the original plan)

components/d2l-quick-eval/d2l-quick-eval-search-results-container.js
CarolBusman: Yup; I'd originally written this directly in quick-eval, and must have missed this when I moved it into its own component.

components/d2l-quick-eval/d2l-quick-eval.js
CarolBusman: Once we hook up the search results from the api, it will be set to `false` when we don't have a search applied to the activity list (when first pulling up the list, or if the search has been cleared). For now, it's just set to `true` for testing.

AlexBedley: ah makes sense thanks for the explanation

PR #140, opened 2019-03-27
Build translations

mtjandra:
Building translations from https://github.com/BrightspaceHypermediaComponents/activities/pull/139

### DIFF
--- a/components/d2l-activity-evaluation-icon/build/lang/ar.js
+++ b/components/d2l-activity-evaluation-icon/build/lang/ar.js
@@ -7,7 +7,7 @@ const LangArImpl = (superClass) => class extends superClass {
 	constructor() {
 		super();
 		this.ar = {
-			'draftInfo': 'Evaluation saved as draft'
+			'draftInfo': 'تم حفظ التقييم كمسودة'
 		};
 	}
 };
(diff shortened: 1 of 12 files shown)


mtjandra:
I think maybe renaming to quick-eval might have resulting in failure to translate all. https://github.com/BrightspaceHypermediaComponents/activities/pull/139#discussion_r269593721

Serge bot says
> `d2l-activities_evaluationHub` Strings: 14 -- Words: 81
> `d2l-activities_evaluationIcon` Strings: 1 -- Words: 4

Maybe because it thought is was evaluationHub.

Risk you take one with current implemented process

jeroach:
@mtjandra Is there anything I can dig into re: the issues due to the rename?

Sorry about that. I admit I don't fully understand how Serge works.

mtjandra:
@jeroach yeah not your fault at all! This is because we get terms during cert, but then applies changes on master branch. TBH I still don't fully understand how it works either. I was more calling out the risk of using serge how we are at the moment

The change you made was good. It should be picked up next release from what I understand.

m6jones:
The manual building at least gives us a chance to not change anything until translations come in. So we can change the translations in the `json` while the build files don't change.

I believe serge works by automating the process of sending out terms to get translated then getting them back. The person in the middle might not translate things? 

I agree, having the translation come in during cert makes things difficult because they go into master not the build where they came from. We can get cert to act on a branch so maybe having a production branch and have it run off that?

svanherk:
I'd check with Jeff S.  If the translations did go out, but can't be able to be added back in automatically because of the name change, he may actually have them ready to go somewhere

PR #141, opened 2019-03-28
Us105390 update activity to use updated organization

ciss1995:


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -249,7 +249,7 @@ class D2lActivityListItem extends mixinBehaviors([IronResizableBehavior, D2L.Pol
 							</div>
 							<div hidden$="[[_textPlaceholder]]">
 								<h2 class="d2l-activity-list-item-title">
-									<d2l-organization-name href="[[_organizationUrl]]"></d2l-organization-name>
+									<d2l-organization-name href="[[_organizationUrl]]" token="[[token]]"></d2l-organization-name>
```
svanherk: I think we need to either temporarily add a `token` property, or get this element using `entity-behaviour`.  @m6jones Which would you prefer?

ciss1995: Yup agreed! The demo was only working when I had token as a property.

ciss1995: We could follow the behavior that other components are doing in that repo. quick-eval is getting it from `entity-behaviour`

svanherk: Yeah if you're willing to set that up here like we've been doing in the `organization` and `enrollments`, that'd be best

demo/d2l-activity-list-item/index.html
```diff
@@ -53,17 +53,17 @@ <h3>Basic d2l-activity-list-item demo</h3>
 				<template>
 					<d2l-activity-list-item image-shimmer text-placeholder></d2l-activity-list-item>
 					<d2l-activity-list-item href="../data/base/activity.json" id="shimmer-placeholder-test" image-shimmer text-placeholder></d2l-activity-list-item>
```
ciss1995: Forgot to pass a token here.

components/d2l-activity-list-item/d2l-activity-list-item.js
m6jones: So we can use the entity-behaviour for this element. I will need to update our fra when we make the change. So make sure you do a major update.

m6jones:
Is this something we need to merge now?

ciss1995:
@m6jones maybe? We are in the process of releasing bsi with my-courses updates. @svanherk

svanherk:
Was another activity component added here that we need to update as well?  Or did it not use any of the organization entities?

PR #142, opened 2019-03-29
US105524 - Computed number of activities to show should never go down

mtjandra:
* This is prep work for filters to try and maintain the number of activities to show
* This is needed because filters may reduce the number of activities shown, and I don't think we ever want to reduce the number of activities shown if we can avoid it

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -282,9 +282,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				type: Array,
 				value: [ ]
 			},
-			_numberOfCurrentlyShownActivities: {
+			_numberOfActivitiesToShow: {
 				type: Number,
-				computed: '_computeNumberOfCurrentlyShownActivities(_data)'
+				computed: '_computerNumberOfActivitiesToShow(_data, _numberOfActivitiesToShow)',
+				value: 0
 			},
 			_fullListLoading: {
 				type: Boolean,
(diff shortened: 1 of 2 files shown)


test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -792,5 +792,13 @@ import SirenParse from 'siren-parser';
 
 			});
 		});
+
+		test('_computerNumberOfActivitiesToShow returns max of data lenght, and previously shown number of activities', function() {
```
AlexBedley: I would be more interested in an integration level test (to make sure polymer properly handles calling computer properties with the old value) but maybe we attach that to your next PR.

nit: spelling mistake `lenght`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -339,8 +340,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 		return !fullListLoading && isLoading;
 	}
 
-	_computeNumberOfCurrentlyShownActivities(data) {
-		return data.length;
+	_computerNumberOfActivitiesToShow(data, currentNumberOfActivitiesShown) {
```
AlexBedley: nit: grammar `computer` -> `compute`

mtjandra: Updated

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Updated, spelling mistake.

PR #143, opened 2019-04-02
US105524 - Maintain number of activities shown when filters are modified

mtjandra:
* Anytime `_numberOfActivitiesToShow` is updated in quick-eval-activities-list, we fire that event
* quick-eval listens for that event, and tells the filter component, "next time you request for activities, try to give me X amount"
* this way the number of activities should persist as best as possible through sort and filter
* Tested in LMS seems to be working fine

filter_size

#### Dependencies
* Requires changes from https://github.com/BrightspaceHypermediaComponents/common/pull/19 to work

### DIFF
--- a/test/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/test/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -157,13 +157,38 @@ import SirenParse from 'siren-parser';
 		test('instantiating the element works', function() {
 			assert.equal(list.tagName.toLowerCase(), 'd2l-quick-eval-activities-list');
 		});
+		test('_numberOfActivitiesToShow starts with default value of 20', function() {
+			assert.equal(20, list._numberOfActivitiesToShow);
+		});
 		test('attributes are set correctly', function() {
 			assert.equal(list.href, 'blah');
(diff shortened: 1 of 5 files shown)


test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -157,14 +157,40 @@ import SirenParse from 'siren-parser';
...
+			list._numberOfActivitiesToShow = expectedNumberOfActivitiesToShow;
+		});
+		test('when data size increased, _numberOfActivitiesToShow matches size and event is triggered', function(done) {
+			const expectedNumberOfActivitiesToShow = 100;
+			const fakeData = new Array(expectedNumberOfActivitiesToShow);
+
+			list.addEventListener('d2l-quick-eval-activities-list-activities-shown-number-updated', function(e) {
+				assert.equal(expectedNumberOfActivitiesToShow, list._numberOfActivitiesToShow);
+				assert.equal(expectedNumberOfActivitiesToShow, e.detail.count);
+				list._data = [];
```
mtjandra: Seems like a bug in text-fixture. It isn't resetting this, so I had to add this myself

PR #146, opened 2019-04-03
US105617 - Remove repetition of name in screen reader when navigating with arrows

mtjandra:
… row with arrow keys

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -187,6 +187,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 									<d2l-link
 										title="[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]"
 										href="[[s.activityLink]]"
+										aria-label="[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]"
 									>[[_formatDisplayName(s, _headerColumns.0.meta.firstThenLast)]]</d2l-link>
 									<d2l-activity-evaluation-icon-base draft$="[[s.isDraft]]"></d2l-activity-evaluation-icon-base>
 								</d2l-td>
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -184,10 +184,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 											aria-hidden="true">
 										</d2l-profile-image>
 									</template>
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -178,10 +178,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 									<template is="dom-if" if="[[s.userHref]]">
 										<d2l-profile-image class="d2l-user-badge-image" href="[[s.userHref]]" token="[[token]]" small=""></d2l-profile-image>
 									</template>
-									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]</d2l-offscreen>
```
mtjandra: Right now it may say something like "Evaluate Josh Roach, Josh Roach Link", feedback received that this was too much information. With this change it will now say "Josh Roach Link" instead when navigating with arrows using screen reader

jeroach: Are we sure this is desired? I understand that getting rid of the duplication is ideal, but I would have expected that the remaining screen reader speech would now be "Evaluate Josh Roach"

As it stands, there is no context given to a screen reader user as to what the link does (other than it has something to do with Josh Roach).

Perhaps just double check with Carin :)

mtjandra: Yes I sent an e-mail as well, this is my current best guess, will be waiting on that

mtjandra: Yeah so I think ideally she want to hear "Evaluate {student name}". Not sure if that is possible though. I think the visual link will always be picked up by screen-readers.

I tried adding "aria-hidden", but then using tabs the screen-reader says "blank". I tried removing focus, but then the links are not clickable. I will need to investigate further and see if it is possible to only say "Evaluate {student name}" while still visually showing "{student name}"

jeroach: Thanks Miles :) Appreciate the effort! Ping me when this is ready for final review

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -184,11 +184,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 											aria-hidden="true">
 										</d2l-profile-image>
 									</template>
-									<d2l-offscreen id="d2l-quick-eval-activities-list-username">[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]</d2l-offscreen>
 									<d2l-link
 										title="[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]"
-										aria-describedby$="d2l-quick-eval-activities-list-username"
 										href="[[s.activityLink]]"
+										aria-label="[[_localizeEvaluationText(s, _headerColumns.0.meta.firstThenLast)]]"
```
mtjandra: Alright this should work as we intend after this PR gets in https://github.com/BrightspaceUI/link/pull/90

PR #147, opened 2019-04-04
Activity card

qwangd2l:


qwangd2l:
> We will want to make this use the entity store at some point. However, we are not using it yet for the activity list item. So I think when we switch that one we will switch this one.

Does this involve using `BrightspaceHypermediaComponents/organizations#ssemver:^3` where a token is needed but not a presentation href?

m6jones:
Yes, and you would extend the behaviour to get the fetches for free.

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -0,0 +1,293 @@
...
+	}
+	connectedCallback() {
+		super.connectedCallback();
+		const image = this.shadowRoot.querySelector('d2l-course-image');
+		if (image) {
+			image.addEventListener('course-image-loaded', this._activityImageLoaded.bind(this));
+		}
+		this.addEventListener('d2l-organization-accessible', this._onD2lOrganizationAccessible);
+	}
+	disconnectedCalledback() {
```
AdmiralCoco: Typo: `disconnectedCallback`
I wonder if was this working as expected? 😛

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -0,0 +1,293 @@
...
+				this._image = imageEntity;
+			}
+		}
+
+		return Promise.resolve();
+	}
+	_onD2lOrganizationAccessible(e) {
+		if (e && e.detail && e.detail.organization) {
+			if (e.detail.organization.name) {
+				this._accessibilityData.organizationName = e.detail.organization && e.detail.organization.name;
```
AdmiralCoco: Extra check here for `e.detail.organization` isn't needed

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -0,0 +1,293 @@
...
+			}
+			if (e.detail.organization.code) {
+				this._accessibilityData.organizationCode = e.detail.organization.code;
+			}
+		}
+		if (e.detail.semesterName) {
+			this._accessibilityData.semesterName = e.detail.semesterName && e.detail.semesterName;
```
AdmiralCoco: Unnecessary check here also

components/d2l-activity-card/d2l-activity-card.js
qwangd2l: Did not test disconnected :(
Good catch!

PR #148, opened 2019-04-09
Jroach/us104901

jeroach:
As per https://rally1.rallydev.com/#/238990806412d/detail/userstory/293707273844

Original scope of story has been split - remaining work as part of the following defect:
https://rally1.rallydev.com/#/238990806412d/detail/defect/299881504144?fdp=true

(as per discussion in standup today)

This PR also adds back lang terms that were missing from the `lang` folder - I mentioned this last week in Slack.


mtjandra:
When tabbing across the column headers I now hear "Column X Clickable Button", but previously it used to say "Column X Clickable {name} Button".

Seems there is less information given when tabbing now, I was testing on demo page with NVDA

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -418,10 +431,17 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
...
+						this.set(`_headerColumns.${i}.ariaSort`, 'descending');
+					} else {
+						this.set(`_headerColumns.${i}.ariaSort`, 'ascending');
+					}
+
 					result = this._fetchSortedData(header.sortClass, descending);
 				}
 				else {
 					this.set(`_headerColumns.${i}.headers.${j}.sorted`, false);
+					this.set(`_headerColumns.${i}.ariaSort`, 'none');
```
mtjandra: I think you need to update `ariaSort`  in `_followLink` method as well. I see sorted and ascending/descending being manipulated in there as well

mtjandra: I think we would also want to test ariaSort similar to how we do the property of sorted/descending/ascending

jeroach:
> When tabbing across the column headers I now hear "Column X Clickable Button", but previously it used to say "Column X Clickable {name} Button".

> Seems there is less information given when tabbing now, I was testing on demo page with NVDA.

> Also I noticed that before when using arrows, the column names would actually highlight, now they don't seem to do so

Hmmm, I'll dig a bit. Might have something to do with the `aria-hidden=true` I'm setting on the `span` but that was in an effort to reduce spoken duplication.

As for the highlighting - that seems odd. Not sure what could have caused that, but I'll take a look :)

Thanks!

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: As an FYI, the `aria-sort` is mostly to satisfy the goal of the future defect linked in the description, which was why I didn't take the time to harden it. It's not working as intended, but it's not causing undesirable behaviour either.

Figured I'd leave it in since it should make the defect easier to fix. Not sure when would be the ideal time to add tests, given the discussion at standup re: relative priority of stories.

jeroach:
@mtjandra More investigation to come, but on first listen, the demo page is not as full featured as the LMS page. Will be double checking exact wording, but it's clearer when wired up in the LMS.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: My personal preference would be to leave out `aria-sort` if it isn't being used, that way we avoid all the headache

I think we can always point back to this comment in the story if you think its useful

jeroach: I think you're right :)

I'll rip it out of this PR, and open up a branch with the `aria-sort` specific stuff, and link that WIP branch to the defect.

jeroach:
@mtjandra Okay, cleaned up all the `aria-sort` stuff. PR should be much cleaner now.

Will link WIP branch for reference soon

jeroach:
@mtjandra Circling back on this:

> When tabbing across the column headers I now hear "Column X Clickable Button", but previously it used to say "Column X Clickable {name} Button".

In the LMS, the reading is "Sort by X Column Header Button"

e.g. "Sort by Activity Name Column Header Button"

components/d2l-quick-eval/lang/zh.json
```diff
@@ -2,6 +2,7 @@
    "activityName" : "活动名称",
    "caughtUp" : "You're all caught up!",
    "checkBackOften" : "Check back often for new submissions.",
+   "clearSearch": "清除搜索",
```
mtjandra: Out of curiosity, do you happen to know how/when these terms got in without it being in its associated `lang` file?

jeroach: Because we keep them both in source control, the workflow was either:

- Manually added terms to file in `build` folder (likely)
OR
- Added terms to lang folder, built locally (which would update the files in the `build` folder, and forgot to commit changes to files in `lang` folder (only committing changes in `build` folder

PR #150, opened 2019-04-10
US105617 - Use attribute variable 

mtjandra:
* Noticed aria-label wasn't used when using activities-list within quick-eval
* This should fix it


jeroach:
Good catch! I've been in many a pickle, trying to figure out why something isn't working, only to finally realize that I forgot the `$`

PR #151, opened 2019-04-10
Fixing margins.

CarolBusman:
The summary container should be 18px below the `d2l-hm-search`, and 54px above the headers in the activities list.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -55,8 +55,8 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 					margin-bottom: 0.5rem;
 				}
 				d2l-quick-eval-search-results-summary-container {
-					margin-bottom: 42px;
-					margin-top: 18px;
+					margin-bottom: 18px;
+					margin-top: 12.25px;
 					display: block;
 				}
 				[hidden] {
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -23,17 +23,15 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 				}
 				h1 {
 					@apply --d2l-heading-1;
-					float: left;
 					margin: 0;
 				}
-				:host(:dir(rtl)) h1 {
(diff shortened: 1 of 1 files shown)


jeroach:
Can you please link to the design doc for reference? Thanks!

CarolBusman:
Required margins are in the description of rally story [US104385](https://rally1.rallydev.com/#/238990806412d/detail/userstory/290539106444).

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -55,8 +55,8 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 					margin-bottom: 0.5rem;
 				}
 				d2l-quick-eval-search-results-summary-container {
-					margin-bottom: 42px;
-					margin-top: 18px;
+					margin-bottom: 18px;
+					margin-top: 12.25px;
```
jeroach: 2 questions/thoughts:
- first, these should be in `rem` instead of `px` (if memory the standard conversion for D2L usages is 1 rem = 20 px)
- potentially dumb question, but why is margin top not 18px and bottom 54 px? I'm not taking the size of the component into account am I? ;)

jeroach:
If you could post a screenshot of the before and after in the LMS, that would really help with review :) Thanks!

components/d2l-quick-eval/d2l-quick-eval.js
CarolBusman: The top margin should now be 18px (0.9rem). The search bar was previously aligned with the top of the header instead of the bottom. I've corrected that, and adjusted the top margin.

The bottom margin is going to be less than 54px (2.7rem) since it is to be 54px above the headers in the list, and there are two sets of paddings pushing the top of the span down: top padding between the `d2l-quick-eval-activities-list` and its content (`d2l-table`), and padding between the `d2l-th` and its content (the header `span`s).

CarolBusman:
Here are the distances between elements in the LMS currently: 
image
23.75px (24) between the bottom of the search box to the top of the summary container, and 78 px between the bottom of the summary container and the top of the table headers.

This is the current state of the demo after the changes in this PR: 
image
18.25px (18) between the bottom of the search box to the top of the summary container, and 54 px between the bottom of the summary container and the top of the table headers.

CarolBusman:
@mtjandra Any issues with the margin changes?

PR #152, opened 2019-04-10
Jroach/us105893

jeroach:
AC linked in story:
https://rally1.rallydev.com/#/238990806412d/detail/userstory/299882752556

- Updated lang terms to remove the word "filter" from the message shown along with the "hook" image
- Hook image now displays when there are 0 search results

Getting screenshots now - will post once LMS is setup.


mtjandra:
There seems to be a regression based of the demo pages for empty state. The one with the filters applied shows a cat and not the hook

jeroach:
> There seems to be a regression based of the demo pages for empty state. The one with the filters applied shows a cat and not the hook

Hmmm, I'll take a look. I think it's bad demo data as opposed to the actual logic since I've been testing on my local LMS with Amaury's data generation

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -249,10 +249,13 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this.entity = e.detail.results;
 		this._showSearchResultSummary = !e.detail.searchIsCleared;
 		this._searchResultsCount = this.entity.entities && this.entity.entities.length ? this.entity.entities.length : 0;
+		list.searchApplied = true;
 		this._clearErrors();
 	}
 
 	_clearSearchResults() {
```
mtjandra: Do you happen to know the repo that will fire off the `d2l-quick-eval-search-results-summary-container-clear-search` event? Want to double check this event is fired when the search is empty or just whitespace

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -249,10 +249,13 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this.entity = e.detail.results;
 		this._showSearchResultSummary = !e.detail.searchIsCleared;
 		this._searchResultsCount = this.entity.entities && this.entity.entities.length ? this.entity.entities.length : 0;
+		list.searchApplied = true;
```
CarolBusman: `searchApplied` should be set to the same thing as `showSearchResultSummary` (`!e.detail.searchIsCleared`). We hit this handler both when the search box applies and clears the search results.
Then, we can also remove the setting of `searchApplied` from `_clearSearchResults()`.

components/d2l-quick-eval/d2l-quick-eval.js
CarolBusman: This event comes from the [d2l-quick-eval-search-results-summary-container](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.js#L59), which is in this repo. The event is fired when the user clicks the Clear Search link in the summary container:
image and is used to let the quick eval component know that the user wants to clear the search criteria.

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Hmmm, tried this, and it broke this assert:
https://github.com/BrightspaceHypermediaComponents/activities/pull/152/files#diff-263ce507b36690f295dffa4766b0d63eR299

Was `false` instead of `true` - are we sure it's fired? Maybe a test data issue?

components/d2l-quick-eval/d2l-quick-eval.js
mtjandra: Thanks Carol. In that case Josh I don't think we have all the bases covered here. Seems like a user can clear out their search term manually (deleting character rather than clicking that link), and QE wouldn't know that search is no longer applied

CarolBusman: If the user clears out the search box and clicks search, or clicks the clear button in the search box, that should get handled by the `d2l-hm-search-results-loaded` handler, `_searchResultsLoaded()`.

jeroach:
> There seems to be a regression based of the demo pages for empty state. The one with the filters applied shows a cat and not the hook

Fixed. Good catch. The demo page was still setting the old `criteria-applied` variable, and thus was doing nothing

components/d2l-quick-eval/d2l-quick-eval.js
CarolBusman: Ah! There's an issue in `performSearch` in the test file: `search.value = searchString;` should be `searchBar.value = searchString;` as the `d2l-hm-search` doesn't have a `value` property, and doesn't pass it down to the `d2l-input-search`.

CarolBusman: Actually, we can just get rid of `_showSearchResultSummary` and control the `hidden` state on the summary container with `searchApplied`.

jeroach: Fixed!

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: @mtjandra Did @CarolBusman 's last comment clear up this issue? If not, what specifically do you think we need to fix/add? Thanks!

mtjandra: Yes I think it captured what I was concerned about. Only thing is I'm not sure if whitespace gets trimmed at https://github.com/BrightspaceHypermediaComponents/common/blob/5f98f39cf6b8740873f59358761fe3b5ec1de9ac/components/d2l-hm-search/d2l-hm-search.js#L74

For example `if('  ') { console.log("HI") }` prints out "HI" for me, but that is a separate concern

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -249,6 +249,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this.entity = e.detail.results;
 		this._showSearchResultSummary = !e.detail.searchIsCleared;
 		this._searchResultsCount = this.entity.entities && this.entity.entities.length ? this.entity.entities.length : 0;
+		list.searchApplied = !e.detail.searchIsCleared;
```
mtjandra: I think in the future we would want to pass these attributes into `d2l-quick-eval-activities-list` just like you did in the demo

test/d2l-quick-eval/d2l-quick-eval.html
```diff
@@ -82,7 +82,7 @@
 			function performSearch(searchString) {
 				const searchBar = search.shadowRoot.querySelector('d2l-input-search');
 				if (searchString) {
-					search.value = searchString;
+					searchBar.value = searchString;
```
mtjandra: Was this always previously referring to undefined? Surprised tests passed

jeroach: Yeah my thoughts exactly

jeroach:
Going to do one final LMS test, with all of the latest changes. If that goes well, then I'll merge and begin the BSI dance 🕺

PR #155, opened 2019-04-11
Moving filter dropdown closer to search (was 40px, now 5px).

CarolBusman:
Before change: 
image
-

After change:
image
-

To verify distance between elements (should be 5):
```
var t = document.querySelector('d2l-quick-eval').shadowRoot.querySelector('d2l-hm-filter').getBoundingClientRect();
var u = document.querySelector('d2l-quick-eval').shadowRoot.querySelector('d2l-hm-search').getBoundingClientRect();
u.left - t.right
```

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -36,7 +36,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 				d2l-hm-search {
 					display: inline-block;
 					width: 250px;
-					margin-left: 2rem;
+					margin-left: .25rem;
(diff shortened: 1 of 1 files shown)


jeroach:
@AlexBedley I think this is part of the reason we use `rem` units. These scale better with different font sizes.

The conversion of 20 px = 1 rem is for the default font size.

But designs are usually done in px.

AlexBedley:
> @AlexBedley I think this is part of the reason we use `rem` units. These scale better with different font sizes.
> 
> The conversion of 20 px = 1 rem is for the default font size.
> 
> But designs are usually done in px.

Agreed, I just wanted to make sure that design is ok with the margins scaling along with the text (is it a hard 5px or just 5px at normal zoom)

PR #157, opened 2019-04-15
Ensure to reset nextHref back to its default value when no data is pr…

mtjandra:
* Last sprint planning we mentioned we would add this bugfix during cert together with langterms
* Putting up PR against last cert branch now so we don't forget


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -497,6 +497,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 				this._data = result;
 			} else {
 				this._data = [];
+				this._pageNextHref = '';
```
AlexBedley: I hate that we rely on the fact that "empty string evaluates to false" everywhere. But probably an issue for another PR

PR #158, opened 2019-04-16
Moving siren entity interactions to separate behavior

CarolBusman:
Also adding a couple helper methods so that the activities list doesn't need to import Rels.

Tests have been moved from the activities list to the new behaviour so there should be the same coverage as previously, just organized differently.


svanherk:
So something that may be of interest to you guys - a new pattern is being developed around storing logic specific to the HM route, but not really specific to a certain component.  Each repo would have one or more `Entity`s in charge of keeping this logic in one place, for the entire repo to use.  It uses https://github.com/BrightspaceHypermediaComponents/siren-sdk to help (developed by @m6jones).  It pulls in the `siren-polymer-behaviours` and uses the entity store, so the list wouldn't have to.  It's already been added to `organizations`: https://github.com/BrightspaceHypermediaComponents/organizations/blob/master/OrganizationEntity.js

`enrollments` is next, and then I'd expect `activities` soon (@m6jones do you guys have plans to add it?  If not I expect Gaudi needs to for your components, to get rid of the event dependency).  The `ActivitiesEntity` would handle getting an exposing all activities info, and would also know how to call an `OrganizationEntity` to get course info and a `UserEntity` to get user info, etc.

m6jones:
So I have plans to add it when you launch my courses. Note that our dependencies in this repo are based on a FRA so they don't have to be in sync.

PR #159, opened 2019-04-16
Add lang terms for UI search results hotfix

jeroach:
Just getting the lang terms in now, so they can get to translators sooner. 

Still working on the actual fix - that will be a follow up PR


CarolBusman:
In @AlexBedley 's testing of the search, he identified that we would need to change the placeholder text on the search bar to "Search For" instead of "Search". Would it make sense to get that into this PR so that we can get that to translations as well?

mtjandra:
If the langterms are required for hotfix, I think you will want to sync with Jeff. I believe translations have already gone out.

mtjandra:
Also if needed to go into cert, you will want to follow up with PR to `release/20.19.5` as well

jeroach:
> In @AlexBedley 's testing of the search, he identified that we would need to change the placeholder text on the search bar to "Search For" instead of "Search". Would it make sense to get that into this PR so that we can get that to translations as well?

Since these are for a hotfix, and I assume this would be an update to an existing lang term, instead of a new term, then the "Search For" text would also get hotfixed. If we're good with that then cool. But want to confirm

PR #160, opened 2019-04-16
DE33386 - Demo working in IE11

mtjandra:
* Progress to support IE11
* This gets demo to work with fake data but not with real data yet


mtjandra:
Okay Alex, will update the name on that file soon

PR #162, opened 2019-04-16
De33386/remove search param get ie11

mtjandra:
* Allows loading QE on load, but still cannot apply filters yet

### DIFF
--- a/components/d2l-quick-eval/compatability/universal-methods.js
+++ b/components/d2l-quick-eval/compatability/universal-methods.js
@@ -9,4 +9,34 @@ const StringEndsWith = (str, search) => {
 	return str.substring(str.length - search.length, str.length) === search;
 };
 
-export { StringEndsWith };
+const GetQueryStringParams = (queryString) => {
+	const query = {};
+	if (queryString) {
+		const trimmedQueryString = queryString[0] === '?' ? queryString.substr(1) : queryString;
+		const queryStrings = trimmedQueryString.split('&');
+		const validQueryStrings = queryStrings.filter(p => p.indexOf('=') >= 0);
+
+		validQueryStrings.forEach(function(validQueryString) {
+			const keyValuePair = validQueryString.split('=');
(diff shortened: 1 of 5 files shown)


components/d2l-quick-eval/compatability/universal-methods.js
```diff
@@ -0,0 +1,39 @@
+const StringEndsWith = (str, search) => {
```
AlexBedley: I would just call out that this file only exists because of ie11 and comment what built-in function you are trying to mimic so when we remove ie11 support we can easily revert this change.

components/d2l-quick-eval/compatability/universal-methods.js
```diff
@@ -0,0 +1,39 @@
...
+		return false;
+	}
+
+	return str.substring(str.length - search.length, str.length) === search;
+};
+
+const GetQueryStringParams = (queryString) => {
+	const query = {};
+	if (queryString) {
+		const pairs = (queryString[0] === '?' ? queryString.substr(1) : queryString).split('&');
```
AlexBedley: nit: I feel like this deserves 2 lines

components/d2l-quick-eval/compatability/universal-methods.js
```diff
@@ -0,0 +1,43 @@
...
+		const validPairs = pairs.filter(p => p.indexOf('=') >= 0);
+
+		for (let i = 0; i < validPairs.length; i++) {
```
AlexBedley: nit: foreach

components/d2l-quick-eval/compatability/universal-methods.js
mtjandra: I can add a comment to the top of the file in the earlier PR https://github.com/BrightspaceHypermediaComponents/activities/pull/160

This is still WIP

AlexBedley: woops got my PRs mixed up. I was wondering why you kept force-pushing haha!

PR #163, opened 2019-04-16
DE33386 - Allow QE to load in IE11 with real LMS hookup

mtjandra:
* On initial load QE should now load in IE11
* Filtering still doesn't work

### DIFF
--- a/components/d2l-quick-eval/compatability/universal-methods.js
+++ b/components/d2l-quick-eval/compatability/universal-methods.js
@@ -14,4 +14,34 @@ const StringEndsWith = (str, search) => {
 	return str.substring(str.length - search.length, str.length) === search;
 };
 
-export { StringEndsWith };
+const GetQueryStringParams = (queryString) => {
+	const query = {};
+	if (queryString) {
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/compatability/universal-methods.js
```diff
@@ -14,4 +14,34 @@ const StringEndsWith = (str, search) => {
...
 
-export { StringEndsWith };
+const GetQueryStringParams = (queryString) => {
+	const query = {};
+	if (queryString) {
+		const trimmedQueryString = queryString[0] === '?' ? queryString.substr(1) : queryString;
+		const queryStrings = trimmedQueryString.split('&');
+		const validQueryStrings = queryStrings.filter(p => p.indexOf('=') >= 0);
+
+		validQueryStrings.forEach(function(validQueryString) {
```
AlexBedley: nit: mixed function syntax outer function is lambda, but this one is not

mtjandra: Yeah :(, maybe an eslint rule we can add? I'd approve for either `() =>` or `function(x)`

PR #164, opened 2019-04-17
DE33386 - Remove all usages for url.searchParams

mtjandra:
* QE component should now be comptable with IE11
* Filtering still doesn't work
* Reason why QE actions will still not work in IE11 is because of https://github.com/Brightspace/polymer-siren-behaviors/issues/41 but this is a platform issue not a QE issue

### DIFF
--- a/components/d2l-quick-eval/compatability/universal-methods.js
+++ b/components/d2l-quick-eval/compatability/universal-methods.js
@@ -14,6 +14,22 @@ const StringEndsWith = (str, search) => {
 	return str.substring(str.length - search.length, str.length) === search;
 };
 
+const DictToQueryString = (queryParams) => {
+	if (queryParams === null || queryParams === undefined) {
+		return '';
+	}
+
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/compatability/universal-methods.js
```diff
@@ -44,4 +60,4 @@ const GetQueryStringParam = (name, url) => {
 	return null;
 };
 
-export { StringEndsWith, GetQueryStringParam, GetQueryStringParams };
+export { DictToQueryString, StringEndsWith, GetQueryStringParam, GetQueryStringParams };
```
CarolBusman: nit: We're starting to get a lot of exports here, should we start using line breaks?

mtjandra: Yeah sounds good, I'll update it when I rename `universal-methods` to the proposed `ie11shim`

PR #169, opened 2019-04-23
Jroach/us106128

jeroach:
`[20.19.05][Cert Fix][UI] Fix X Search Results`

Mostly done - following items are still WIP
- more unit tests
- minor issue with search count when user has loaded a _lot_ more


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -253,8 +263,11 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		list.entity = e.detail.results;
 		this.entity = e.detail.results;
 		this._showSearchResultSummary = !e.detail.searchIsCleared;
-		this._searchResultsCount = this.entity.entities && this.entity.entities.length ? this.entity.entities.length : 0;
 		list.searchApplied = !e.detail.searchIsCleared;
+
+		if (this.entity && this.entity.entities) {
```
CarolBusman: `else { this._updateSearchResultsCount(0); }` ?
If we didn't get any entity(ies) back, I'm wondering if we shouldn't clear out the count.

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -273,6 +286,16 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this._showSearchError = false;
 		this._showFilterError = false;
 	}
+
+	_updateSearchResultsCount(count) {
+		this._searchResultsCount = count;
+
+		this._moreSearchResults = false;
+		const next = 'next';
+		if (this.entity && this.entity.hasLinkByRel && this.entity.hasLinkByRel(next) && this.entity.getLinkByRel(next)) {
```
CarolBusman: Is there ever a situation where `hasLinkByRel` will be true but `getLinkByRel` will end up being falsey? If so, should we `if (this._getHref(entity, next))` ?

components/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.js
```diff
@@ -42,15 +42,21 @@ class D2LQuickEvalSearchResultsSummaryContainer extends QuickEvalLocalize(Polyme
...
+				type: Boolean,
+				value: false
 			}
 		};
 	}
 
-	_getSummaryString(num) {
-		if (num === 1) {
+	_getSummaryString(searchResults, moreResults) {
+		if (searchResults === 1) {
```
CarolBusman: Should the `moreResults` check come first before `searchResults === 1`? In the unlikely event that page size should be set to 1, should we be displaying `1+ search results`?

test/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.html
```diff
@@ -0,0 +1,53 @@
...
+		<test-fixture id="moreResults">
+			<template strip-whitespace>
+				<d2l-quick-eval-search-results-summary-container search-results-count="25" more-results></d2l-quick-eval-search-results-summary-container>
+			</template>
+		</test-fixture>
+
+		<script type="module">
+			var searchResultsSummary;
+
+			suite('d2l-quick-eval-search-results-summary-container', function() {
```
CarolBusman: Would like to see some tests on `_getSummaryString` for 0, 1, 1+, multiple, and multiple+. I'm making the assumption that we will never have a count of 0 with `more-results` set.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -545,8 +545,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Si
 						}
 					}
 				}.bind(this))
-				.then(this._clearAlerts.bind(this))
-				.catch(function() {
+				.then(x => {
+					this._clearAlerts.bind(x);
+					this._dispatchPageSizeEvent(this._data.length);
```
mtjandra: I don't think this is necessary. `_dispatchPageSizeEvent` is already fired every time `_numberOfActivitiesToShow` is updated. And `_numberOfActivitiesToShow` is updated anytime `_data` is updated. This might actually cause a bug where we might set a number lower than a previous number. This should only go up or stay the same. Don't think it should ever go down

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -273,6 +286,16 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this._showSearchError = false;
 		this._showFilterError = false;
 	}
+
+	_updateSearchResultsCount(count) {
```
mtjandra: Instead of figuring out when to call this. I would recommend using an observer on `entity`. If I understand right, you will always want to mirror the length of its entitites

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -186,6 +191,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 	_updateNumberOfActivitiesToShow(e) {
```
mtjandra: Actually I'm a little confused on this method. Does both QE and QE-list both share the same entity? Not sure why the count is coming from the QE-list, but then later we get the count from QE's `entity`

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -273,6 +288,16 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this._showSearchError = false;
 		this._showFilterError = false;
 	}
+
+	_updateSearchResultsCount(count) {
+		this._searchResultsCount = count;
+
+		this._moreSearchResults = false;
+		const nextHref = this._getHref(this.entity, 'next');
```
mtjandra: Hmm also since we already calculate if we should load more in activities list, maybe we should only calculate it once

components/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.js
```diff
@@ -42,15 +42,21 @@ class D2LQuickEvalSearchResultsSummaryContainer extends QuickEvalLocalize(Polyme
...
+				value: false
 			}
 		};
 	}
 
-	_getSummaryString(num) {
-		if (num === 1) {
+	_getSummaryString(searchResults, moreResults) {
+		if (moreResults) {
+			return this.localize('searchResultsMore', 'num', searchResults);
```
AlexBedley: what does this do when `searchResults` is 1?

test/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.html
```diff
@@ -0,0 +1,53 @@
...
+		</test-fixture>
+
+		<test-fixture id="moreResults">
+			<template strip-whitespace>
+				<d2l-quick-eval-search-results-summary-container search-results-count="25" more-results></d2l-quick-eval-search-results-summary-container>
+			</template>
+		</test-fixture>
+
+		<script type="module">
+			var searchResultsSummary;
```
AlexBedley: The linter should have failed on this - so this file probably isn't being linted. Might want to separate this code out to a separate file if you want the linter safety.

components/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.js
jeroach: Good call - by addressing Carol's concern re: if page size is 1, this logic is now flawed.

Fixing with a check on:
`if (searchResults === 1 && !moreResults) {`

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Thinking about this and in regards to your other comment:

> Actually I'm a little confused on this method. Does both QE and QE-list both share the same entity? Not sure why the count is coming from the QE-list, but then later we get the count from QE's entity

An observer on `entity` would work from the perspective of `d2l-quick-eval`, but we also want to update the results count when the `d2l-quick-eval-activities-list-activities-shown-number-updated` event is fired from `d2l-quick-eval-activities-list`

And the extracted value is different in the 2 cases:
- Coming from `d2l-quick-eval` the value is: `this.entity.entities.length`
- Coming from `d2l-quick-eval-activities-list` the value is: `e.detail.count` where `e` is the thrown event

It does seem a bit odd, but I don't have context for the original rationale. Thoughts on all of this? @CarolBusman @mtjandra @AlexBedley 

If this requires a bit of a refactor, I'm not sure if that would fall under the cert fix (depends on the size/effort of such a potential refactor).

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Resolving this comment to reduce noise as the question is being discussed above.

jeroach:
Build failure looks to be related to `d2l-activity-list-item-test.js`

`npm test` is passing locally for me, so it's likely a recent commit to `master`

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Fixing this now - actually solved another issue I was having :)

Thanks!

test/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.html
jeroach: I swear I tried `let` before and it caused a test failure/issue. But just tried again and it worked. Must have been a different issue last week that confused me,

jeroach: Actually, not sure why the linter didn't catch it (based on the rules):

```
		"lint:js": "eslint . --ext .js,.html test/**/*.js test/**/*.html demo/**/*.html",
		"lint:wc": "polymer lint -i \"./*\" \"components\" \"demo\" \"test\"",
```

components/d2l-quick-eval/d2l-quick-eval.js
CarolBusman: I believe the reason we're not sharing the entity between QE and QE-list is because we want to avoid 2-way binding in anticipation of Lit.
We could add an observer to `entity` in QE that just calls `_updateSearchResultsCount` with the current count on the entity, and also call it from the `d2l-quick-eval-activities-list-activities-shown-number-updated` handler? That way we should be covered in all situations where the entity is updated in QE, and the case where it's updated in the list.

test/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.html
jeroach: Working on these right now :)

test/d2l-quick-eval/d2l-quick-eval-search-results-summary-container.html
jeroach: Changed to `let` - still not sure why the file isn't being linted

components/d2l-quick-eval/d2l-quick-eval.js
mtjandra: I see, thanks for the input. Yeah I think the reason I was confused was because both `d2l-quick-eval` and `d2l-quick-eval-activities-list` shared the same `href`. But I see what's going on now. You're right this is probably a separate concern.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Sorry Josh, I think maybe a new variable / event is needed. I noticed if say I load all 100+ student after clicking load more several times. After adding a search term that matches them all. After clicking load more it stays at `20+` instead of incrementing `20+`, `40+` etc.

I think this is what you brought up in standup where the search is not respecting page size. This is why the problem occurs

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -268,6 +283,16 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		this._showSearchError = false;
 		this._showFilterError = false;
 	}
+
+	_updateSearchResultsCount(count) {
+		this._searchResultsCount = count;
+		this._moreSearchResults = false;
```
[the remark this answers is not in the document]
jeroach: Will update to use an `else`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: Fixed with a different event

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Resolving this comment thread - can consider this separate from the cert fix if desired

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -534,8 +534,10 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Qu
 						}
 					}
 				}.bind(this))
-				.then(this._clearAlerts.bind(this))
-				.catch(function() {
+				.then(x => {
+					this._clearAlerts.bind(x);
+					this._dispatchSearchResultsCountUpdatedLoadMoreEvent(this._data.length);
```
mtjandra: I would recommend using the same pattern as `_numberOfActivitiesToShow` for some property called `_numberOfActivitiesShown` and just follow that number to populate the search result. That way you are guaranteed to always have the right value without risk of missing it now or in future.

I have found a bug where the number does not apply correctly when toggling filters.

filter_bug

mtjandra: Hmm I see a bug when applying a filter  where nothing matches.

nothign_matches

jeroach: Thanks Miles - sorry for all the back and forth on this - seems when I fix one thing I break something else :(

Working on it

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -298,6 +298,11 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Qu
 				computed: '_computeNumberOfActivitiesToShow(_data, _numberOfActivitiesToShow)',
 				value: 20
 			},
+			_numberOfActivitiesShownInSearchResults: {
+				type: Number,
+				computed: '_computeNumberOfActivitiesShownInSearchResults(_data)',
+				value: 20
```
mtjandra: I don't think you want a default here. Now I also do not see the number updating at all it is always 20. Feel free to test yourself. If you use `go/backdoor` user tenant `8bff9875-a67d-4cfe-ae45-aaeee13dcae0` and userId `30220` using `auth-dev` to generate token, and for href use https://8bff9875-a67d-4cfe-ae45-aaeee13dcae0.activities.api.proddev.d2l/my-unassessed-activities

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -184,6 +191,13 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 		}
 	}
 
+	_updateNumberOfActivitiesShownInSearchResults(e) {
+		if (e && e.detail) {
+			this.set('_numberOfActivitiesShownInSearchResults', e.detail.count);
```
mtjandra: No property `_numberOfActivitiesShownInSearchResults` exists in QE

jeroach: It's on QE-list

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Was on master sorry

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Miles and I had a Zoom call offline - pretty sure this line can go. Verifying now.

Otherwise, we're confident the code is working as intended.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
jeroach: I'll test out with out a default (zero as the backup case)

jeroach: All good with no default value!

PR #179, opened 2019-04-26
Translation updates

jeroach:


jeroach:
I forgot to build the translations before merging the auto PR against master - will fix that once this is merged and in the BSI dance

PR #180, opened 2019-04-29
US106434 Pass token to organization-name and organization-info

ciss1995:


svanherk:
I think we need to add the `entity-behavior` so that `token` is a property

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -1,5 +1,7 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {Classes, Rels} from 'd2l-hypermedia-constants';
+import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
```
edwardg8: very minor about the spacing

edwardg8:
Don't we have similar PR before?

svanherk:
> Don't we have similar PR before?

This one was for `d2l-activity-list-item`: https://github.com/BrightspaceHypermediaComponents/activities/pull/141

edwardg8:
Oh that make sense, thanks

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -10,12 +12,14 @@ import 'd2l-organizations/components/d2l-organization-info/d2l-organization-info
 import 'd2l-organizations/components/d2l-organization-name/d2l-organization-name.js';
 import 'd2l-typography/d2l-typography.js';
 import SirenParse from 'siren-parser';
+import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
```
svanherk: I think we don't need this line if we have `siren-sdk/mixin/entity-mixin.j`, or vice versa?  @m6jones?

PR #182, opened 2019-04-30
Us106434 update activity card

ciss1995:


components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -139,9 +154,74 @@ class D2lActivityCard extends PolymerElement {
 				type: String,
 				value: 'javascript:void(0)',
 				computed: '_cardHrefComputed(sendEventOnClick, _activityHomepage)'
+			},
+			/*
+			* Presentation Attributes
```
m6jones: Oops, this is my bad. The notifications shouldn't be part of an activity card since they don't have a user attached. I will fix it but if any of it causes problems for this PR delete it.

ciss1995: Actually, it's my bad. organization-info only cares about Course Code and Semester. I will remove the rest.

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -69,12 +73,15 @@ class D2lActivityCard extends PolymerElement {
 				</div>
 
 				<div class="d2l-activity-card-content-container" slot="content">
-					<d2l-organization-name href="[[_organizationUrl]]"></d2l-organization-name>
+					<d2l-organization-name href="[[_organizationUrl]]" token="[[token]]"></d2l-organization-name>
 					<d2l-card-content-meta>
 						<d2l-organization-info
 							class="d2l-activity-card-content-organization-info"
 							href="[[_organizationUrl]]"
 							presentation-href="[[presentationHref]]"
```
svanherk: I think we can remove `presentation-href` here, right?  Since instead we pass in `show-organization-code` and `show-semester-name`, this is no longer a property on `d2l-organization-info`

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -10,12 +12,14 @@ import 'd2l-organizations/components/d2l-organization-info/d2l-organization-info
 import 'd2l-organizations/components/d2l-organization-name/d2l-organization-name.js';
 import 'd2l-typography/d2l-typography.js';
 import SirenParse from 'siren-parser';
+import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
```
svanherk: Do we need this if we use the siren sdk?

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -85,7 +92,8 @@ class D2lActivityCard extends PolymerElement {
 	static get properties() {
 		return {
 			presentationHref: {
```
svanherk: @m6jones, how is this used in the FRA?  Shouldn't this look more like the other components we've been moving, and instead of passing in a `presentationHref` you instead pass in `showOrganizationCode` and `showSemesterName` and we just pass that on through to `d2l-organization-info`?

components/d2l-activity-card/d2l-activity-card.js
ciss1995: It wasn't working without it, so I think so?

components/d2l-activity-card/d2l-activity-card.js
ciss1995: Yes, we can! Thanks for the catch.

components/d2l-activity-card/d2l-activity-card.js
svanherk: I'm guessing because of this line: https://github.com/BrightspaceHypermediaComponents/activities/pull/182/files#diff-34495308e74092ec7a7852c4d8976d4aR165

Depending on what Maya says though, I think maybe we want to delete that section and just pass through the booleans instead of call the api to get them

ciss1995: I think that it is complaining about `D2L.PolymerBehaviors.Siren.EntityBehavior`
`Uncaught TypeError: Cannot read property 'EntityBehavior' of undefined`. 
But you're right the fetch requires that line too.

svanherk: Ah right, but couldn't we remove that behaviour too?  Cause the entity mixin gives us the same stuff? (If we are able to remove that code that uses the fetch)

ciss1995: Yup if we remove that fetch, we could

components/d2l-activity-card/d2l-activity-card.js
m6jones: Yes, I 100% agree. This was made before those changes.

components/d2l-activity-card/d2l-activity-card.js
m6jones: Yeah, this shouldn't be required. Delete anything that needs this.

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -139,7 +150,7 @@ class D2lActivityCard extends PolymerElement {
 				type: String,
 				value: 'javascript:void(0)',
 				computed: '_cardHrefComputed(sendEventOnClick, _activityHomepage)'
-			}
```
svanherk: I don't think we want this change

ciss1995: It actually deleted `}` and added `},` ! Do you mean removing `,`?

svanherk: Yeah, looks like we accidentally added a trailing comma

svanherk: I'm kinda surprised the linter didn't care 🤔

ciss1995: Haha yeah! I will remove it.

PR #183, opened 2019-05-03
Sync Lang

mtjandra:
Following up from https://github.com/BrightspaceHypermediaComponents/activities/pull/179#issuecomment-487105062

Changes were not ported to master


mtjandra:
> Looks good. What language is the fi.js file for? I notice it's not updated here.

Not sure maybe Finnish? Yeah nothing in that files seems to be translated. Perhaps it translates to the same thing, or maybe it isn't supported since translators are not making adjustments

PR #184, opened 2019-05-06
US105214 - Add ability to log error

mtjandra:
* Log errors for QE

### DIFF
--- a/components/d2l-quick-eval/QuickEvalLogging.js
+++ b/components/d2l-quick-eval/QuickEvalLogging.js
@@ -1,0 +1,56 @@
+import {dedupingMixin} from '@polymer/polymer/lib/utils/mixin.js';
+
+/*
+ * @polymer
+ * @mixinFunction
+ */
+const QuickEvalLoggingImpl = (superClass) => {
+	return class extends superClass {
+		constructor() {
+			super();
+		}
+
+		static get properties() {
+			return {
+				logger: {
+					type: String
+				}
+			};
+		}
+
+		_logError(error, additionalContext) {
+
+			if (!this.logger) {
+				return;
+			}
+
+			const message = additionalContext || {};
+
+			if (error) {
+				if (!error.message && !error.stack) {
+					message.errorMessage = error;
+				}
+				if (error.message) {
+					message.errorMessage = error.message;
+				}
+				if (error.stack) {
+					message.stackTrace = error.stack;
+				}
+			}
+
+			window.fetch(
+				this.logger,
+				{
+					method: 'POST',
+					mode: 'no-cors',
+					headers: {
+						'Content-Type': 'application/json'
+					},
+					body: JSON.stringify([message])
+				}
+			);
+		}
+	};
+};
+
+export const QuickEvalLogging = dedupingMixin(QuickEvalLoggingImpl);
--- a/components/d2l-quick-eval/QuickEvalLogging.js
+++ b/components/d2l-quick-eval/QuickEvalLogging.js
@@ -6,13 +6,10 @@ import {dedupingMixin} from '@polymer/polymer/lib/utils/mixin.js';
  */
 const QuickEvalLoggingImpl = (superClass) => {
 	return class extends superClass {
-		constructor() {
-			super();
-		}
 
 		static get properties() {
 			return {
-				logger: {
+				loggingEndpoint: {
 					type: String
 				}
 			};
@@ -20,11 +17,12 @@ const QuickEvalLoggingImpl = (superClass) => {
 
 		_logError(error, additionalContext) {
 
-			if (!this.logger) {
+			if (!this.loggingEndpoint) {
 				return;
 			}
 
-			const message = additionalContext || {};
+			const message = {};
+			message.additionalContext = additionalContext || {};
 
 			if (error) {
 				if (!error.message && !error.stack) {
@@ -39,7 +37,7 @@ const QuickEvalLoggingImpl = (superClass) => {
 			}
 
 			window.fetch(
-				this.logger,
+				this.loggingEndpoint,
 				{
 					method: 'POST',
 					mode: 'no-cors',
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -25,7 +25,6 @@ import {StringEndsWith} from './compatability/ie11shims.js';
  * @polymer
  */
 
-
 class D2LQuickEvalActivitiesList extends mixinBehaviors(
 	[D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior, D2L.PolymerBehaviors.QuickEval.D2LHMSortBehaviour],
 	QuickEvalLogging(QuickEvalLocalize(PolymerElement))
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -1,5 +1,6 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {QuickEvalLocalize} from './QuickEvalLocalize.js';
+import {QuickEvalLogging} from './QuickEvalLogging.js';
 import 'd2l-alert/d2l-alert.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import 'd2l-table/d2l-table.js';
@@ -24,7 +25,11 @@ import {StringEndsWith} from './compatability/ie11shims.js';
  * @polymer
  */
 
-class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior, D2L.PolymerBehaviors.QuickEval.D2LHMSortBehaviour], QuickEvalLocalize(PolymerElement)) {
+
+class D2LQuickEvalActivitiesList extends mixinBehaviors(
+	[D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior, D2L.PolymerBehaviors.QuickEval.D2LHMSortBehaviour],
+	QuickEvalLogging(QuickEvalLocalize(PolymerElement))
+) {
 	static get template() {
 		const quickEvalActivitiesListTemplate = html`
 			<style include="d2l-table-style">
@@ -411,7 +416,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Qu
 			this._clearAlerts();
 
 		} catch (e) {
-			// Unable to load activities from entity.
+			this._logError(e, {developerMessage: 'Unable to load activities from entity.'});
 			this._handleFullLoadFailure();
 			return Promise.reject(e);
 		} finally {
@@ -449,7 +454,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Qu
 					}
 				}.bind(this))
 				.then(this._clearAlerts.bind(this))
-				.catch(function() {
+				.catch(function(e) {
+					this._logError(e);
 					this._loading = false;
 					this._handleLoadMoreFailure();
 				}.bind(this));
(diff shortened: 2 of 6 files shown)


components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,49 @@
+import {dedupingMixin} from '@polymer/polymer/lib/utils/mixin.js';
+
+/*
+ * @polymer
+ * @mixinFunction
+ */
+const QuickEvalLoggingImpl = (superClass) => {
```
mtjandra: Nothing QE specific, probably should be moved somewhere common once adoption for logging in front-end takes off. Keeping in just QE for now

components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,49 @@
...
+
+			return window.fetch(
+				this.logger,
+				{
+					method: 'POST',
+					mode: 'no-cors',
+					headers: {
+						'Content-Type': 'application/json'
+					},
+					body: JSON.stringify([message])
```
mtjandra: Endpoint expects it to be in an array, in case you want to batch messages

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -23,7 +24,10 @@ import {StringEndsWith} from './compatability/ie11shims.js';
  * @polymer
  */
 
-class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior], QuickEvalLocalize(PolymerElement)) {
+class D2LQuickEvalActivitiesList extends mixinBehaviors(
```
mtjandra: Added logs in our catches

components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+const QuickEvalLoggingImpl = (superClass) => {
+	return class extends superClass {
+		constructor() {
+			super();
+		}
+
+		static get properties() {
+			return {
+				logger: {
```
mtjandra: Right now we are writing `logger` but we should probably change it to `loggingEndpoint`, http://search.dev.d2l/source/xref/Lms/le/evaluationhub/D2L.LE.EvaluationHub/Web/Desktop/Views/EvaluationHubView.Renderer.cs#104

components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,56 @@
...
+				}
+				if (error.message) {
+					message.errorMessage = error.message;
+				}
+				if (error.stack) {
+					message.stackTrace = error.stack;
+				}
+			}
+
+			window.fetch(
```
mtjandra: We don't return anything on purpose, logging is purely for side-effect, should have avoid any impact on client application

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -544,7 +548,8 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.Qu
 					}
 				}.bind(this))
 				.then(this._clearAlerts.bind(this))
-				.catch(function() {
+				.catch(function(e) {
+					this._logError(e);
```
AlexBedley: Should we add some context like `Unable to load more`

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -86,7 +90,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 				more-results="[[_moreSearchResults]]"
 				hidden$="[[!_searchResultsMessageEnabled(_showSearchResultSummary, searchEnabled)]]">
 			</d2l-quick-eval-search-results-summary-container>
-			<d2l-quick-eval-activities-list href="[[href]]" token="[[token]]" master-teacher="[[masterTeacher]]"></d2l-quick-eval-activities-list>
+			<d2l-quick-eval-activities-list href="[[href]]" token="[[token]]" logger="[[logger]]" master-teacher="[[masterTeacher]]"></d2l-quick-eval-activities-list>
```
AlexBedley: I think there are things we want to log in this file as well. I'd like to cover the filtering and search errors at least

components/d2l-quick-eval/QuickEvalLogging.js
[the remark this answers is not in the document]
mtjandra: We have a polyfill https://github.com/Brightspace/brightspace-integration/blob/73fc587011a9f72961276aa33fa80e80b598d367/web-components/bsi.js#L1

I did double check by hitting quad site with IE

components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,56 @@
...
+			return {
+				logger: {
+					type: String
+				}
+			};
+		}
+
+		_logError(error, additionalContext) {
+
+			if (!this.logger) {
```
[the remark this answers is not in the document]
mtjandra: What would the caller do if logging fails? Since it is transparent to the end-user I don't think there is much they can do

components/d2l-quick-eval/QuickEvalLogging.js
```diff
@@ -0,0 +1,56 @@
+import {dedupingMixin} from '@polymer/polymer/lib/utils/mixin.js';
+
+/*
+ * @polymer
+ * @mixinFunction
+ */
+const QuickEvalLoggingImpl = (superClass) => {
+	return class extends superClass {
+		constructor() {
```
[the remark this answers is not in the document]
mtjandra: Thanks I'll try to remove and see what happens

components/d2l-quick-eval/QuickEvalLogging.js
mtjandra: I'm not expecting a lot of errors, let me know if my assumption is wrong.

To my understanding, any perf issues should not effect QE, but maybe logs come in late, or get stalled. This would be handled by the infrastructure ingesting the events (firehose/lambda I believe).

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -23,7 +24,10 @@ import {StringEndsWith} from './compatability/ie11shims.js';
  * @polymer
  */
 
-class D2LQuickEvalActivitiesList extends mixinBehaviors([D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior], QuickEvalLocalize(PolymerElement)) {
+class D2LQuickEvalActivitiesList extends mixinBehaviors(
+	[D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior],
+	QuickEvalLogging(QuickEvalLocalize(PolymerElement))
```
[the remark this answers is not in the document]
mtjandra: From what I can tell `D2LQuickEvalSirenHelperBehavior` is a behaviour. In polymer 2.x and 3.x behaviors are no longer documented and mixins are preferred. I believe we are just supporting legacy code with behaviors

test/d2l-quick-eval/d2l-quick-eval-logging.html
```diff
@@ -0,0 +1,117 @@
...
+				sandbox.stub(window, 'fetch', function(logger, options) {
+
+					assert.equal('expectedLogger', logger);
+
+					assert.equal('POST', options.method);
+					assert.equal('no-cors', options.mode);
+					assert.equal('application/json', options.headers['Content-Type']);
+
+					const messageBody = JSON.parse(options.body);
+					assert.equal(Array, messageBody.constructor);
```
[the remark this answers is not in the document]
mtjandra: Nice! Will update

test/d2l-quick-eval/d2l-quick-eval-logging.html
```diff
@@ -0,0 +1,117 @@
...
+				logger = fixture('logger');
+				noLogger = fixture('no-logger');
+				sandbox = sinon.sandbox.create();
+			});
+
+			teardown(function() {
+				sandbox.restore();
+			});
+
+			test('sends POST request with expected options formatted in the correct way', function(done) {
```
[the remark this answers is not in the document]
mtjandra: Sounds good

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Sounds good

components/d2l-quick-eval/d2l-quick-eval.js
mtjandra: I thought those were other components in different repos, but yes we should log when filter/search fails as well

components/d2l-quick-eval/QuickEvalLogging.js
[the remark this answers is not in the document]
AlexBedley: They could retry if they need accurate client-side metrics. We don't really have a use-case for it, I'm ok if you don't make this change.

components/d2l-quick-eval/QuickEvalLogging.js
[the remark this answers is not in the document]
AlexBedley: Seems a bit fragile that it only works when packaged with BSI. Is this a common pattern?

In any case, probably not worth holding up this PR

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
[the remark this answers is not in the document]
AlexBedley: Thanks for the explanation

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: > I was to log at the source, not by its events

I want to challenge that statement. I don't think we should make those components know about logging. Logging is usually app-specific (just like the endpoints) and every app might not want to hear about all the filtering errors. If the events are set up correctly, the apps should be able to listen to the filtering component and log what they want.

My mind might be changed if we start seeing everyone listening to the same filtering things. But for now I think it makes sense to keep the logging as part of the QE controller (Which is unfortunately spread out between quick-eval and quick-eval-list right now, but after the refactor I'm working on it will just be quick-eval)

components/d2l-quick-eval/QuickEvalLogging.js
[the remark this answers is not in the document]
mtjandra: Yeah nice thought! I think if we want to support retry it should be within the logger itself, definitely something to look into

test/d2l-quick-eval/d2l-quick-eval-logging.html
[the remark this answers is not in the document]
mtjandra: I just put any additional context into `message.additionalContext` to avoid collision

components/d2l-quick-eval/d2l-quick-eval.js
mtjandra: Okay spoke offline. I believe in the short term we will listen to the events and log, but once we get a common logging modules, this will change.

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -281,6 +286,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 	}
 
 	_searchError() {
+		this._logError("Failed to retrieve search results.");
```
mtjandra: Unfortunately we don't know the search term, or the filters selected, and we have no stack trace. This is because for the time being we are listening to events rather than sourcing it from the root

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -259,6 +263,7 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
 	}
 
 	_filterError() {
+		this._logError('Failed to retrieve filter results');
```
AlexBedley: For this and the search errors we should be logging the actual errors. Add the parameter to `_filterError` and `_searchError` and add it to the logs.

https://github.com/BrightspaceHypermediaComponents/common/blob/9b4593a1721696f9549e1936c2ca5cc8d2563aad/components/d2l-hm-filter/d2l-hm-filter.js#L493
https://github.com/BrightspaceHypermediaComponents/common/blob/5f98f39cf6b8740873f59358761fe3b5ec1de9ac/components/d2l-hm-search/d2l-hm-search.js#L115

mtjandra: Oh nice thanks will do

PR #188, opened 2019-05-08
Change loading animation in activity-card and activity-list-item from… 

qwangd2l:
… shimmers to pulsing (even for text placeholders).

Rename shimmer to placeholer so it's more generic.
Change the href of activity-card to always be activity homepage.

Story: https://rally1.rallydev.com/#/39602890252ud/detail/userstory/304121020968

### DIFF
--- a/components/d2l-activity-card/d2l-activity-card.js
+++ b/components/d2l-activity-card/d2l-activity-card.js
@@ -267,7 +267,7 @@ class D2lActivityCard extends EntityMixin(PolymerElement) {
 	_activityImageLoaded() {
 		this._imageLoading = false;
 	}
-	_sendClickEvent() {
+	_sendClickEvent(event) {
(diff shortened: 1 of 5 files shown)


components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -287,6 +280,7 @@ class D2lActivityCard extends EntityMixin(PolymerElement) {
 			bubbles: true,
 			composed: true
 		}));
+		event.preventDefault();
```
AdmiralCoco: Is event supposed to be an argument here?

m6jones: event isn't defined either.

qwangd2l: Ah yes! Good catch. I was testing this in the FRA where I had this in the parameters but not here :(

PR #189, opened 2019-05-09
Adding toggle button to main quick-eval component, defaulting to off.

CarolBusman:
There should be no change when `toggleEnabled` is `false`:
image
Toggle and filter/search should be on the next line below the QE title when `toggleEnabled` is `true`:
image

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -17,6 +17,12 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 					border-bottom-left-radius: 0.3em;
 					border-width: 1px;
 				}
+				:host button.d2l-quick-eval-view-toggle-left {
+					margin-left: 0.9rem;
+				}
+				:host(:dir(rtl)) button.d2l-quick-eval-view-toggle-left {
+					margin-right: 0.9rem;
+				}
 				.d2l-quick-eval-view-toggle-right,
 				:host(:dir(rtl)) .d2l-quick-eval-view-toggle-left {
 					border-top-right-radius: 0.3em;
@@ -55,6 +61,7 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 				}
 			</style>
 			<div>
+				[[localize('viewBy')]]
 				<button class="d2l-quick-eval-view-toggle-left" on-click="_selectSubmissions" selected$="[[_isSelected(_viewTypes.submissions, currentSelected)]]">[[localize('submissions')]]</button>
 				<button class="d2l-quick-eval-view-toggle-right" on-click="_selectActivities" selected$="[[_isSelected(_viewTypes.activities, currentSelected)]]">[[localize('activities')]]</button>
 			<div>
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -37,6 +37,7 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 					color: var(--d2l-color-ferrite);
 					cursor: pointer;
 					display: inline;
+					font-family: inherit;
 					margin: 0;
 					min-height: calc(2rem + 2px);
 					outline: none;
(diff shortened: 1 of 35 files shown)


CarolBusman:
Not yet setting the `current-selected` attribute on the toggles as this will almost definitely need to be changed once the second view is added.

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -55,6 +61,7 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 				}
 			</style>
 			<div>
+				[[localize('viewBy')]]
```
jeroach: This is good for now, but we may want to consider making this a low level heading (e.g. `h5`) or something to aid navigation for screen readers

test/d2l-quick-eval/d2l-quick-eval.html
```diff
@@ -126,7 +127,8 @@
 						)
 					);
 				});
-				test('the filter, search, and list exist', function() {
+				test('the toggle, filter, search, and list exist', function() {
+					assert.isNotNull(toggle);
```
jeroach: Could add a simple true/false test that the toggle is visible (or not) if the variable is set (or not)

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
jeroach: Or maybe a label for the button? Just so we get all the typography/resizing stuff for free :)

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -55,8 +62,19 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 				}
 			</style>
 			<div>
-				<button class="d2l-quick-eval-view-toggle-left" on-click="_selectSubmissions" selected$="[[_isSelected(_viewTypes.submissions, currentSelected)]]">[[localize('submissions')]]</button>
-				<button class="d2l-quick-eval-view-toggle-right" on-click="_selectActivities" selected$="[[_isSelected(_viewTypes.activities, currentSelected)]]">[[localize('activities')]]</button>
+				<label id="d2l-quick-eval-view-toggle-label">[[localize('viewBy')]]</label>
```
AlexBedley: Semantically, I feel like the label doesn't belong to the buttons. I know the design isn't finalized for the real buttons, but my guess is they will be separate from the label (I could be totally wrong here though). My point is just the harder we make it to incorporate the shared component, the less likely it will happen.

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -53,11 +54,19 @@ class D2LQuickEval extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityBeha
...
+					width: 0px;
+					height: 36px;
+					overflow: hidden;
+				}
 			</style>
 			<div class="d2l-quick-eval-top-bar">
 				<template is="dom-if" if="[[headerText]]">
 					<h1>[[headerText]]</h1>
 				</template>
+				<div class="flex-break" hidden$="[[!toggleEnabled]]"></div>
```
AlexBedley: Hey can you explain this? My CSS isn't that great. Are we just resetting some kind of styling above? Or is this for padding?

CarolBusman: When we don't have the toggle, the title and the filter/search are on the same line. When the toggle is there, the toggle and filter/search are on the same line under the title.

There is not an easy way to put a line break in a `flex-box`, so what this does is places an empty div the full width of the `flex-box` to force a line break.

jeroach: Cool, I was wondering this too

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
CarolBusman: I can just have the label there for now and remove `aria-labelledby`, and let design let us know exactly how they want the `h`/`label`/`labelledby` to be handled?

AlexBedley: Yeah not sure what the correct pattern is if we need to make `aria-labelledby` work. I just wanted to call out that this component is quickly becoming only useful to us (which could be ok as well)

jeroach: I think for now we can be totally okay with components that are only useful for us. Given the discussion from the last front end retro :)

PR #190, opened 2019-05-09
US106353 - extract sorting

AlexBedley:
The first of many.

This PR just extracts the 3 sorting methods and drops them in a behaviour. My approach to this refactor is to move as much functionality out of d2l-quick-eval-activities-list to reduce the diff in further PRs and to make it easier to extract the core hm functionality.

Next PR: cleaning up the interface between d2l-hm-sort-behaviour and d2l-quick-eval-activities-list and separating the sorting tests from the d2l-quick-eval-activities-list tests

PR #195, opened 2019-05-14
US106353 - clean up sorting interface

AlexBedley:
Previous PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/190

This PR: Tried to get the quick-eval specific things into `list` and keep the generic sorting stuff in the behaviour. I originally tried to have a complete sorting solution, but I couldn't figure out a good way to bind the sort headers to the sort behaviour. In the interest of time, I went with this smaller refactor.

Next PR: Going to create a new layer, `d2l-quick-eval-submission` that will serve as the controller for the submissions view. The new hierarchy will be `d2l-quick-eval` > `d2l-quick-eval-submission` > `d2l-quick-eval-activities-list`


AlexBedley:
> Again to confirm, no logic was changed right? We're just moving the code around.

Correct! I slightly changed `_loadSorts` to actually return the sorts. But you'll notice I didn't change the tests (apart from the rename).

PR #197, opened 2019-05-14
Changes for RTL support for d2l-activity-list-item and d2l-activity-card

qwangd2l:
d2l-activity-list-item: add support for styles for RTL and also reverse tags order in RTL.

d2l-activity-card: remove a right margin that's too small to notice anyway (0.05rem/1px) so that left and right margin are the same. No changes needed for RTL after this.

### DIFF
--- a/components/d2l-activity-card/d2l-activity-card.js
+++ b/components/d2l-activity-card/d2l-activity-card.js
@@ -36,7 +36,7 @@ class D2lActivityCard extends EntityMixin(PolymerElement) {
 				.d2l-activity-card-content-container {
 					display: flex;
 					flex-direction: column;
-					margin: -0.35rem 0 -0.1rem -0.05rem;
+					margin: -0.35rem 0 -0.1rem;
 					overflow-wrap: break-word; /* replaces 'word-wrap' in Firefox, Chrome, Safari */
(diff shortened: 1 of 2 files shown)


qwangd2l:
> How does it look in IE11/Edge?
> I've had some issues where RTL was applying without RTL on the page - doesn't look like there should be any issues here but just wondering

The demo page is a bit weird in Edge/IE11:
Edge: the placeholders don't show up while loading but the rest of it is fine in RTL
IE11: the tags do not show up after loading but everything else is fine

I've tested these 2 browsers when I pasted these changes into the FRA's node modules and they seem fine there (except footer items since they're not an actual thing yet I don't think?).

AdmiralCoco:
>Edge: the placeholders don't show up while loading but the rest of it is fine in RTL
IE11: the tags do not show up after loading but everything else is fine 

I see - these are expected behaviours?

qwangd2l:
> > Edge: the placeholders don't show up while loading but the rest of it is fine in RTL
> > IE11: the tags do not show up after loading but everything else is fine
> 
> I see - these are expected behaviours?

Probably needs fixing but is not introduced with these changes since I also observe them in master branch.

PR #198, opened 2019-05-16
US106353 - add submissions controller

AlexBedley:
Previous PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/196

This PR: Adding a middle class called `d2l-quick-eval-submissions` that right now does nothing. This will be the controller for the submissions portion of the application. The plan is to slowly move functionality from above (quick-eval) and below (quick-eval-activities-list) until this follows the flux pattern.

Next PR: Since Josh is doing some search/filter stuff, I'm going to focus on moving things up from `list` next.

PR #199, opened 2019-05-16
US106353 - move last bit of list-specific code out of sort

AlexBedley:
Previous PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/198

This PR: Just noticed one last list-specific thing in sort that should be moved to the outside

Next PR: Since Josh is doing some search/filter stuff, I'm going to focus on moving things up from `list` next.


AlexBedley:
I noticed this while I was speaking to Josh and reading over his shoulder. I did one final sweep of that sort file, it _should_ be ok now :)

AlexBedley:
Hmm I noticed the diff didn't shorten when I merged the PR before this. I wonder if that's a bitbucket-only thing. Anyways gonna try and remake this PR with the real diff

PR #202, opened 2019-05-23
Cbusman/activity card component

CarolBusman:
Data binding and basic functionality.


jeroach:
Screenshot/gif of demo page? :)

jeroach:
General thought/question - this seems to be a lot more implemented than what our slice plan was. Any thoughts/concerns with this?

CarolBusman:
Default view:
image
Hover view:
image

CarolBusman:
> General thought/question - this seems to be a lot more implemented than what our slice plan was. Any thoughts/concerns with this?

Given that the work/POC was already done and that this is FAR from the final version of this component and not yet ready to be used, I don't see any downside to having this checked in to be expanded on in future work. I imaging there will be many future changes to compartmentalize and account for layout/styling requirements, but it's useful as a base to build off of.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,174 @@
...
+				button:focus d2l-icon {
+					color: var(--d2l-color-celestine-plus-1);
+				}
+				circle {
+					stroke: var(--d2l-color-tungsten);
+					fill: transparent;
+				}
+			</style>
+			<div class="d2l-quick-eval-card" on-click="_clicked" tabindex="-1">
+				<div class="d2l-quick-eval-card-left">
```
AlexBedley: nit: is this class unused? I would remove all unused classes

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	_clicked() {
+		if (this._focused) {
+			this.blur();
+		}
+		this._focused = !this._focused;
+	}
+
+	_getNewSubmissionsText() {
```
AlexBedley: How come some of these get their own function, but others are just put in-line?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,174 @@
...
+			},
+			_focused: {
+				type: Boolean,
+				value: false
+			}
+		};
+	}
+
+	_clicked() {
+		if (this._focused) {
```
AlexBedley: Do we need to maintain our own `_focused`? Could we not just check focus? (maybe with `document.activeElement`?)

package.json
```diff
@@ -65,7 +65,7 @@
 		"d2l-polymer-behaviors": "Brightspace/d2l-polymer-behaviors-ui#semver:^2",
 		"d2l-polymer-siren-behaviors": "Brightspace/polymer-siren-behaviors#semver:^1",
 		"d2l-table": "BrightspaceUI/table#semver:^2",
-		"d2l-tooltip": "BrightspaceUI/tooltip#semver:^3",
+		"d2l-tooltip": "github:BrightspaceUI/tooltip#semver:^3",
```
AlexBedley: Why was this change necessary?

test/d2l-quick-eval/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,12 @@
+(function() {
```
AlexBedley: Is it necessary to wrap these tests in an iife?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
CarolBusman: `_getSubmissionTooltipText()` because passing 5 arguments to `localize` was looking long and unwieldy in html binding and was hard to read.
`_getNewSubmissionsText()` because the parser doesn't like `unread+resubmitted` as one of the variables. I considered using a computed property and binding to localize directly, but since I was already planning on using a wrapper for the tooltip text, I decided to be consistent.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
CarolBusman: This click handler runs after the div has already been focused from the click, so `document.activeElement` is always our container div inside this handler.
I really would prefer to do away with tracking our own `_focused` though, if there's a method I've missed.

package.json
CarolBusman: Missed that one! I'll revert.

test/d2l-quick-eval/d2l-quick-eval-activity-card.js
CarolBusman: Probably not necessary, but it's a habit I picked up a while ago to keep variables from leaking everywhere. Some of the other tests are also wrapped. Could it cause issues with the tests?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,174 @@
...
+				button:hover d2l-icon,
+				button:focus d2l-icon {
+					color: var(--d2l-color-celestine-plus-1);
+				}
+				circle {
+					stroke: var(--d2l-color-tungsten);
+					fill: transparent;
+				}
+			</style>
+			<div class="d2l-quick-eval-card" on-click="_clicked" tabindex="-1">
```
dbatiste: What gets announced by screenreader when focus lands here?  Does the screenreader user know they can "click" to perform some action, and do they know what it's going to do?  I may be inclined to use `role="button"`.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
AlexBedley: Sounds good, I was mostly just curious

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,174 @@
...
+				}
+				.d2l-quick-eval-card-right,
+				.d2l-quick-eval-activity-card-items-meters,
+				.d2l-quick-eval-activity-card-items-buttons {
+					display: flex;
+					align-items: center;
+				}
+				.d2l-quick-eval-card:hover .d2l-quick-eval-activity-card-items-meters,
+				.d2l-quick-eval-card:focus-within .d2l-quick-eval-activity-card-items-meters,
+				.d2l-quick-eval-card:not(:hover):not(:focus-within) .d2l-quick-eval-activity-card-items-buttons {
```
dbatiste: Usually I try to structure the CSS in a way that avoid `:not`.  It's typically an extra selector that can be avoided, and in the past, has been know to have poor performance, though it may not be noticeable depending on the circumstances.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
CarolBusman: This click is to allow for mobile users that can't hover a way to toggle the state to see the buttons. When using a screenreader/keyboard navigation the intended use is to tab past this div entirely and hit the buttons directly.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
AlexBedley: Yeah the elements themselves don't seem to track focus unfortunately =/

test/d2l-quick-eval/d2l-quick-eval-activity-card.js
AlexBedley: Nope, I was just curious, I hadn't seen an iife in years!

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
dbatiste: Is this purely for mouse users...? I'm not seeing how a keyboard user would get to here.

CarolBusman: It is for mobile users with the side effect that it is also usable with a mouse. Since the flipping of the stats/buttons visibility is based on `:focus-within` and not `:focus`, focusing on the child buttons will have the same effect visually as having the div focused.

dbatiste: I see.  Thanks.  :)

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
dbatiste: Something to be aware of: with shadow-DOM enabled in a browser that supports it, `document.activeElement` will return the element in the **document's DOM context** that contains the focus.  If this component was not wrapped inside of another shadow-DOM and focus was on the `div`, `document.activeElement` would return the `d2l-quick-eval-activity-card` element.  To enable getting at the element that really has the focus, the shadow-DOM spec provides for `activeElement` on hosts that have shadow-DOM - and in cases where we want to know the actual element focused, we have this [getComposedActiveElement](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-focus.js#L20) helper function that drills in.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -0,0 +1,169 @@
...
+			</style>
+			<div class="d2l-quick-eval-card d2l-visible-on-ancestor-target" on-click="_clicked" tabindex="-1">
+				<div>
+					<d2l-activity-name href="[[activityNameHref]]" token="[[token]]"></d2l-activity-name>
+					<div>[[activityType]] &bull; [[localize('due', 'date', dueDate)]]</div>
+				</div>
+				<div class="d2l-quick-eval-card-right">
+					<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText()]]</div>
+					<d2l-tooltip for="d2l-quick-eval-activity-card-submissions" position="bottom">[[_getSubmissionTooltipText()]]</d2l-tooltip>
+					<div class="d2l-quick-eval-activity-card-items-container">
```
jeroach: This looks so much cleaner and easier to read/understand :)

PR #203, opened 2019-05-27
Remove store from activities.

m6jones:
The store doesn't work nicely with the way I allowed embedded HM entities to be added. The siren-sdk will fix this issue! So YAY!

The issue that was caused was because the href was changed when adding the entity that entity would be fetched again. This would cause a fetch call to our outside service and fail it since there was too many.  Eek.

So this PR is here to fix this for cert and isn't a long term fix.

### DIFF
--- a/components/d2l-activity-card/d2l-activity-card.js
+++ b/components/d2l-activity-card/d2l-activity-card.js
@@ -91,6 +91,7 @@ class D2lActivityCard extends PolymerElement {
 				type: String,
 				observer: '_onHrefChange'
 			},
+			token: String,
 			entity: {
 				type: Object,
 				value: function() {
--- a/components/d2l-activity-card/d2l-activity-card.js
+++ b/components/d2l-activity-card/d2l-activity-card.js
@@ -1,6 +1,5 @@
 import { html, PolymerElement } from '@polymer/polymer/polymer-element.js';
 import { Classes, Rels } from 'd2l-hypermedia-constants';
-import { EntityMixin } from 'siren-sdk/mixin/entity-mixin.js';
 import 'fastdom/fastdom.min.js';
 import 'd2l-card/d2l-card.js';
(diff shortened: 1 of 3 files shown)


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -301,6 +299,7 @@ class D2lActivityListItem extends mixinBehaviors([
 				type: String,
 				observer: '_onHrefChange'
 			},
+			token: String,
```
jkehres: Was this previously inherited from the mixin?

AKobets:
What would be a longterm fix ?

components/d2l-activity-list-item/d2l-activity-list-item.js
m6jones: Yes, the mixin also included the `href`

m6jones:
> What would be a longterm fix ?

Switching this over to use the `siren-sdk`

components/d2l-activity-card/d2l-activity-card.js
```diff
@@ -15,7 +14,7 @@ import SirenParse from 'siren-parser';
  * @customElement
  * @polymer
  */
-class D2lActivityCard extends EntityMixin(PolymerElement) {
+class D2lActivityCard extends PolymerElement {
```
qwangd2l: Do we need a token property added to this component?
Token is used in this component:
```
<d2l-organization-name href="[[_organizationUrl]]" token="[[token]]"></d2l-organization-name>
<d2l-organization-info
    class="d2l-activity-card-content-organization-info"
    href="[[_organizationUrl]]"
    token="[[token]]"
    show-organization-code="[[showOrganizationCode]]"
    show-semester-name="[[showSemesterName]]"
></d2l-organization-info>

m6jones: Yup! will do.

PR #204, opened 2019-05-28
US106353 - move loading to controller

AlexBedley:
This change is larger than I wanted it to be but I had lots of trouble decoupling the loading from the rest of the application. In the end, this change is:

* I moved all the entity things into `submissions` and now `submissions` passes `_data` into `list`
  * Effectively that means moving `_loadData`, `_loadMore`, and `_parseActivities` up one level
  * `list` now emits a `load-more` event which the controller listens for
  * sorting still lives in `list` (for now) but it is triggered by the controller (see `_handleSorts`). Sorted data is emitted via events and listened to by the controller. (This will change in the next PR)
* I didn't modify any of the functions except when we acquired things via `querySelector` I just changed it to look in the correct shadowDom
  * The change is to make `this.shadowDom.querySelector` -> `this.list.shadowDom.querySelector`
* I got part way through fixing the tests before I realized half of them don't make sense anymore. I'm marking them as `skip` in this PR, with the promise to migrate them to the `submissions` tests or modify them to make sense

Be happy to stop by your desks to answer any questions/clear anything up.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -339,9 +339,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 			'_dispatchActivitiesShownInSearchResultsEvent(_numberOfActivitiesShownInSearchResults)'
 		];
 	}
-	ready() {
-		super.ready();
-	}
 
 	_handleSorts(entity) {
 		// entity is null on initial load
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -226,7 +226,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 
 			<template is="dom-if" if="[[_shouldShowLoadMore(_pageNextHref, _loading)]]">
 				<div class="d2l-quick-eval-activities-list-load-more-container">
-					<d2l-button class="d2l-quick-eval-activities-list-load-more" onclick="[[_loadMore]]">[[localize('loadMore')]]</d2l-button>
+					<d2l-button class="d2l-quick-eval-activities-list-load-more" onclick="[[_dispatchLoadMore]]">[[localize('loadMore')]]</d2l-button>
 				</div>
 			</template>
 			<template is="dom-if" if="[[_shouldShowNoSubmissions(_data.length, _loading, _health.isHealthy, filterApplied, searchApplied)]]">
@@ -334,8 +334,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 	}
 	static get observers() {
 		return [
-			'_loadData(entity)',
-			'_handleSorts(entity)',
 			'_handleNameSwap(_headerColumns.0.headers.*)',
 			'_dispatchPageSizeEvent(_numberOfActivitiesToShow)',
 			'_dispatchActivitiesShownInSearchResultsEvent(_numberOfActivitiesShownInSearchResults)'
@@ -343,15 +341,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 	}
 	ready() {
 		super.ready();
-		this.addEventListener('d2l-siren-entity-error', function() {
-			this._fullListLoading = false;
-			this._loading = false;
-			this._handleFullLoadFailure();
-		}.bind(this));
-		this._loadMore = this._loadMore.bind(this);
 	}
-
-	constructor() { super(); }
 
 	_handleSorts(entity) {
 		// entity is null on initial load
@@ -405,7 +395,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 
 		if (result) {
 			return result.then(sortedCollection => {
-				this.entity = sortedCollection;
 				this._dispatchSortUpdatedEvent(sortedCollection);
 			});
 		} else {
@@ -453,70 +442,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 		return !dataLength && !isLoading && isHealthy && (filterApplied || searchApplied);
 	}
 
-	async _loadData(entity) {
-		if (!entity) {
-			return Promise.resolve();
-		}
-		this._loading = true;
-		this._fullListLoading = true;
-
-		try {
-			if (entity.entities) {
-				const result = await this._parseActivities(entity);
-				this._data = result;
-			} else {
-				this._data = [];
-				this._pageNextHref = '';
-			}
-			this._clearAlerts();
-
-		} catch (e) {
-			this._logError(e, {developerMessage: 'Unable to load activities from entity.'});
-			this._handleFullLoadFailure();
-			return Promise.reject(e);
-		} finally {
-			this._fullListLoading = false;
-			this._loading = false;
-		}
-	}
-
-	_loadMore() {
-		if (this._pageNextHref && !this._loading) {
-			this._loading = true;
-			this._followHref(this._pageNextHref)
-				.then(async function(u) {
-					if (u && u.entity) {
-						const tbody = this.shadowRoot.querySelector('d2l-tbody');
-						const lastFocusableTableElement = D2L.Dom.Focus.getLastFocusableDescendant(tbody, false);
-
-						try {
-							if (u.entity.entities) {
-								const result = await this._parseActivities(u.entity);
-								this._data = this._data.concat(result);
-							}
-						} catch (e) {
-						// Unable to load more activities from entity.
-							throw e;
-						} finally {
-							this._loading = false;
-							window.requestAnimationFrame(function() {
-								const newElementToFocus = D2L.Dom.Focus.getNextFocusable(lastFocusableTableElement, false);
-								if (newElementToFocus) {
-									newElementToFocus.focus();
-								}
-							});
-						}
-					}
-				}.bind(this))
-				.then(this._clearAlerts.bind(this))
-				.catch(function(e) {
-					this._logError(e, {developerMessage: 'Unable to load more.'});
-					this._loading = false;
-					this._handleLoadMoreFailure();
-				}.bind(this));
-		}
-	}
-
 	_clearAlerts() {
 		this.set('_health', { isHealthy: true, errorMessage: '' });
 	}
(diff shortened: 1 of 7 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -226,7 +226,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 
 			<template is="dom-if" if="[[_shouldShowLoadMore(_pageNextHref, _loading)]]">
 				<div class="d2l-quick-eval-activities-list-load-more-container">
-					<d2l-button class="d2l-quick-eval-activities-list-load-more" onclick="[[_loadMore]]">[[localize('loadMore')]]</d2l-button>
+					<d2l-button class="d2l-quick-eval-activities-list-load-more" onclick="[[_dispatchLoadMore]]">[[localize('loadMore')]]</d2l-button>
```
AlexBedley: pressing Load More now emits an event to tell the controller to load more

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -334,25 +334,15 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
...
 		super.ready();
-		this.addEventListener('d2l-siren-entity-error', function() {
-			this._fullListLoading = false;
-			this._loading = false;
-			this._handleFullLoadFailure();
-		}.bind(this));
-		this._loadMore = this._loadMore.bind(this);
 	}
 
-	constructor() { super(); }
```
AlexBedley: this was redundant so I deleted it

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -405,7 +395,6 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 
 		if (result) {
 			return result.then(sortedCollection => {
-				this.entity = sortedCollection;
```
AlexBedley: Since we're receiving `_data` from the controller, we no longer use `this.entity` in `list`. Instead we emit an event which the controller listens for

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -13,6 +19,148 @@ class D2LQuickEvalSubmissions extends PolymerElement {
...
+			'_loadData(entity)',
+			'_handleSorts(entity)'
+		];
+	}
+
+	get list() {
+		return this.shadowRoot.querySelector('d2l-quick-eval-activities-list');
+	}
+
+	_handleSorts(entity) {
```
AlexBedley: This is a placeholder. In the next PR, all the sorting will be in this class. For now, observe here and do the work in `list`.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -13,6 +19,148 @@ class D2LQuickEvalSubmissions extends PolymerElement {
...
+		this.list._handleSorts(entity);
+	}
+
+	async _loadData(entity) {
+		if (!entity) {
+			return Promise.resolve();
+		}
+		this.list._loading = true;
+		this.list._fullListLoading = true;
```
AlexBedley: In order to not modify the functions, whenever they used something inside of `list` I just changed it to `list.theThing`

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -265,7 +265,9 @@ class D2LQuickEval extends mixinBehaviors(
 	}
 
 	_filtersChanged(e) {
-		const list = this.shadowRoot.querySelector('d2l-quick-eval-submissions').shadowRoot.querySelector('d2l-quick-eval-activities-list');
+		const submissions = this.shadowRoot.querySelector('d2l-quick-eval-submissions');
+		const list = submissions.shadowRoot.querySelector('d2l-quick-eval-activities-list');
+		submissions.entity = e.detail.filteredActivities;
```
AlexBedley: Since submissions contains the source of truth now, we need to tell it about filter things

test/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -0,0 +1,12 @@
+suite('d2l-quick-eval-submissions', function() {
```
AlexBedley: Most of the skipped tests will eventually end up here. But I wanted to finish the refactor before moving them.

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -334,25 +334,15 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
...
 		return [
-			'_loadData(entity)',
-			'_handleSorts(entity)',
 			'_handleNameSwap(_headerColumns.0.headers.*)',
 			'_dispatchPageSizeEvent(_numberOfActivitiesToShow)',
 			'_dispatchActivitiesShownInSearchResultsEvent(_numberOfActivitiesShownInSearchResults)'
 		];
 	}
 	ready() {
 		super.ready();
```
jeroach: Is `ready` now redundant too, since all it does it call the `super`?

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -226,7 +226,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 
 			<template is="dom-if" if="[[_shouldShowLoadMore(_pageNextHref, _loading)]]">
```
jeroach: Might want to double check the imports - might be able to tidy them up since we're deleting a lot of code.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -13,6 +19,148 @@ class D2LQuickEvalSubmissions extends PolymerElement {
...
+	}
+
+	static get observers() {
+		return [
+			'_loadData(entity)',
+			'_handleSorts(entity)'
+		];
+	}
+
+	get list() {
```
jeroach: Curiosity - I've never seen `get` used on a non-standard method - how does this work? :)

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -13,6 +19,148 @@ class D2LQuickEvalSubmissions extends PolymerElement {
...
+
+	ready() {
+		super.ready();
+		this.addEventListener('d2l-siren-entity-error', function() {
+			this.list._fullListLoading = false;
+			this.list._loading = false;
+			this.list._handleFullLoadFailure();
+		}.bind(this));
+
+		this.addEventListener('d2l-quick-eval-submission-table-load-more', this._loadMore.bind(this));
```
jeroach: Why are these in `ready` instead of `attached`? Don't we also need to remove the event listeners in `detached`?

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -265,7 +265,9 @@ class D2LQuickEval extends mixinBehaviors(
...
 		list.entity = e.detail.filteredActivities;
```
jeroach: Is this line needed?

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -293,7 +295,10 @@ class D2LQuickEval extends mixinBehaviors(
 	}
 
 	_searchResultsLoaded(e) {
-		const list = this.shadowRoot.querySelector('d2l-quick-eval-submissions').shadowRoot.querySelector('d2l-quick-eval-activities-list');
+		const submissions = this.shadowRoot.querySelector('d2l-quick-eval-submissions');
+		const list = submissions.shadowRoot.querySelector('d2l-quick-eval-activities-list');
+
+		submissions.entity = e.detail.results;
 		list.entity = e.detail.results;
```
jeroach: Same here - I thought list is using `_data` now instead of `entity`

test/d2l-quick-eval/d2l-quick-eval-submissions.js
jeroach: Sounds good

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Good catch!

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Will look at

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: You can make any function a getter. It's just syntactic sugar so I can do `this.list` instead of `this.list()`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: Hmm yes I think I misread https://polymer-library.polymer-project.org/3.0/docs/devguide/events#imperative-listeners. I will move it to attached/detached

components/d2l-quick-eval/d2l-quick-eval-submissions.js
jeroach: Cool beans :)

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: yes because it's how the filter state is passed to the sort stuff (without this line, sorting would remove your filter state). It won't survive the next PR though once sorting is in `submissions`

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
AlexBedley: Couldn't find any unused imports, but thanks for the reminder!

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: Turns out we were both slightly wrong. polymer recommends you use annotated event listeners to listen for events from children. I've made that change

PR #206, opened 2019-05-29
Cbusman/activity page empty message

CarolBusman:
No submissions and no criteria matches pages added to second view.

`_data` and `filterApplied` properties added as currently-non-functional placeholders for when this is hooked up to the api. These (along with `searchCleared` which is included in the search behaviour) are the minimum properties required for the different no results states. `_loading` and `_health` were not included as properties in order to reduce future rework.


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -62,11 +86,42 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				search-results-count$="[[searchResultsCount]]"
 				hidden$="[[searchCleared]]">
 			</d2l-quick-eval-search-results-summary-container>
+			<div class="d2l-quick-eval-no-submissions" hidden$="[[!_shouldShowNoSubmissions(_data, filterApplied, searchCleared)]]">
```
AlexBedley: This design is fine, and it might even be better, but I want to call out that `submissions` has the empty states in the child display class (which doesn't exist for `activities` yet). I can't think of a use-case where one is better than the other, but I do think we should be consistent. Totally happy to change `submissions` to match what's done here though.

test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -0,0 +1,49 @@
+import '@polymer/iron-test-helpers/mock-interactions.js';
+
+(function() {
+	var list;
```
AlexBedley: Hmm I thought we changed the eslint rules to not allow `var` anymore. Oh well.

test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -0,0 +1,49 @@
...
+			list = fixture('basic');
+		});
+		test('instantiating the element works', function() {
+			assert.equal(list.tagName.toLowerCase(), 'd2l-quick-eval-activities');
+		});
+		test('attributes are set correctly', function() {
+			assert.equal(list.href, 'blah');
+			assert.equal(list.token, 't');
+		});
+		test('if there is no data in the list, d2l-quick-eval-no-submissions-image is shown', () => {
```
AlexBedley: These tests are all extremely similar and you might consider parameterizing them to clean it up a bit.

https://mochajs.org/#dynamically-generating-tests

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -62,11 +86,42 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+			},
+			filterApplied: {
+				type: Boolean,
+				value: false
+			}
+		};
+	}
+
+	_shouldShowNoSubmissions() {
+		return !(this._data.length) && !(this.filterApplied || !this.searchCleared);
```
AlexBedley: You might consider applying [DeMorgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to make it all ANDs (debatable on what's easier to read, but I find mixing ANDs + ORs to be hard to read)

PR #212, opened 2019-06-04
US107199 - Use user preference for date time format

mtjandra:
Needs to go in together with
* https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/124/overview

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
@@ -110,7 +110,7 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 	_getSubmissionDate: function(entity) {
 		if (entity.hasSubEntityByClass('localized-formatted-date')) {
 			const i = entity.getSubEntityByClass('localized-formatted-date');
-			return i.properties.text;
+			return i.properties.localizedDate;
 		}
 		return '';
 	},
--- a/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities-list.js
@@ -205,7 +205,7 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 									<span>[[_getDataProperty(s, 'courseName')]]</span>
 								</d2l-td>
 								<d2l-td>
-									<span>[[_getDataProperty(s, 'submissionDate')]]</span>
+									<span>[[_localizeDateTimeFormat(s, 'submissionDate')]]</span>
 								</d2l-td>
 								<template is="dom-if" if="[[_shouldDisplayColumn('masterTeacher')]]">
 									<d2l-td>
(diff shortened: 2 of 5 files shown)


components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -110,7 +110,7 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 	_getSubmissionDate: function(entity) {
 		if (entity.hasSubEntityByClass('localized-formatted-date')) {
 			const i = entity.getSubEntityByClass('localized-formatted-date');
-			return i.properties.text;
+			return i.properties.localizedDate;
```
mtjandra: No tests failed for this change here. What is the current recommend way to test these things. Perhaps it was just missing coverage. Should I make a unit test for this one and `_localizeDateTimeFormat`?

AlexBedley: There's a new pattern going around to handle fetching data from hm entities which is why these tests got deprioritized (basically a js class representing the hm entity with a bunch of helper methods). It probably isn't wrong to unit test this function though. If you have time to add tests that would be nice.

demo/d2l-quick-eval/d2l-quick-eval.html
```diff
@@ -1,5 +1,15 @@
 <!doctype html>
-<html lang="en">
+<html lang="en"
+	data-lang-default="en-us"
+	data-intl-overrides="{
```
AlexBedley: So if I'm understanding this correctly, the LMS will set these attributes on the `html` element in production?

AlexBedley: What happens when these attributes aren't present?

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -498,6 +498,11 @@ class D2LQuickEvalActivitiesList extends mixinBehaviors(
 		return lastName + ', ' + firstName;
 	}
 
+	_localizeDateTimeFormat(item, prop) {
```
AlexBedley: I'm not a huge fan of `_getDataProperty` (polymer seems to normalize encoding the path to the data as a string which I guess is a js thing but still feels wrong. If you know your data is at `x.y.z` then just call `x.y.z` and cut out the complicated middle man.)

Just take `item` and call `item.submissionDate` yourself IMO. At the very least, it would testing this easier.

demo/d2l-quick-eval/d2l-quick-eval.html
mtjandra: Yes LMS will always set these attributes. If attributes are not present they fallback to defaults. Which is the `short` format

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
mtjandra: Will leave this for now, but feel free to update during refactor. Mostly this is to work if say we had a completion date column as well, then that date would be localized as well

PR #213, opened 2019-06-04
Jroach/us106343

jeroach:
This PR accomplishes the story of extracting the filter behavior. Here's what's in here:

- Add `d2l-hm-filter-behavior.js` and extract filter logic to it
- Update `d2l-hm-filter-behavior.js` to remove entity behavior inherit, as well as listeners
- Update `d2l-quick-eval-activities.js` to use the filter and search behaviors
- Add tests for `d2l-hm-filter-behavior.js`
- Update tests for `d2l-hm-search-behavior.js`

NB:
- Removed `d2l-quick-eval-activities.js` from being referenced in `d2l-quick-eval.js`

The above is only temporary until the event logic in `d2l-quick-eval` is refactored down into `d2l-quick-eval-submissions.js`. Tests for `d2l-quick-eval.js` will fail otherwise. Could also just skip the tests that fail, but I'm hoping this is all a mute point since Alex should have the refactor piece soon 😄

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
@@ -1,5 +1,3 @@
-import './d2l-siren-helper-behavior.js';
-
 window.D2L = window.D2L || {};
 window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
 window.D2L.PolymerBehaviors.QuickEval = window.D2L.PolymerBehaviors.QuickEval || {};
@@ -14,9 +12,9 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		searchAction: {
 			type: Object
 		},
-		searchCleared: {
+		searchApplied: {
 			type: Boolean,
-			value: true
+			value: false
 		},
 		searchError: {
 			type: Boolean,
@@ -26,28 +24,6 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 			type: Number,
 			value: 0
 		}
-	},
-
-	observers: [
-		'_setSearchAction(entity)'
-	],
-
-	attached: function() {
-		this.addEventListener('d2l-hm-search-results-loading', this._clearSearchError);
-		this.addEventListener('d2l-hm-search-results-loaded', this._searchResultsLoaded);
-		this.addEventListener('d2l-hm-search-error', this._errorOnSearch);
-		this.addEventListener('d2l-quick-eval-search-results-summary-container-clear-search', this._clearSearchResults);
-		this.addEventListener('d2l-hm-filter-filters-updating', this._clearSearchError);
-		this.addEventListener('d2l-hm-filter-filters-updated', this._clearSearchError);
-	},
-
-	detached: function() {
-		this.removeEventListener('d2l-hm-search-results-loading', this._clearSearchError);
-		this.removeEventListener('d2l-hm-search-results-loaded', this._searchResultsLoaded);
-		this.removeEventListener('d2l-hm-search-error', this._errorOnSearch);
-		this.removeEventListener('d2l-quick-eval-search-results-summary-container-clear-search', this._clearSearchResults);
-		this.removeEventListener('d2l-hm-filter-filters-updating', this._clearSearchError);
-		this.removeEventListener('d2l-hm-filter-filters-updated', this._clearSearchError);
 	},
 
 	_setSearchAction: function(entity) {
@@ -65,7 +41,7 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 
 	_searchResultsLoaded: function(e) {
 		this.entity = e.detail.results;
-		this.searchCleared = e.detail.searchIsCleared;
+		this.searchApplied = !e.detail.searchIsCleared;
 
 		if (this.entity && this.entity.entities) {
 			this.searchResultsCount = this.entity.entities.length;
@@ -80,12 +56,11 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 	},
 
 	_clearSearchResults: function() {
-		this.searchCleared = true;
+		this.searchApplied = false;
 	}
 };
 
 /** @polymerBehavior */
 D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviour = [
-	D2L.PolymerBehaviors.Siren.D2LSirenHelperBehavior,
 	D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl
 ];
--- a/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
@@ -40,11 +40,11 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 	},
 
 	_searchResultsLoaded: function(e) {
-		this.entity = e.detail.results;
+		const entity = e.detail.results;
 		this.searchApplied = !e.detail.searchIsCleared;
 
-		if (this.entity && this.entity.entities) {
-			this.searchResultsCount = this.entity.entities.length;
+		if (entity && entity.entities) {
+			this.searchResultsCount = entity.entities.length;
 		} else {
 			this.searchResultsCount = 0;
 		}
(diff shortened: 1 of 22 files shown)


components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -28,24 +26,6 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		}
 	},
 
-	observers: [
-		'_setSearchAction(entity)'
-	],
-
-	attached: function() {
```
jeroach: Event listening has moved to `d2l-quick-eval-activities.js`

components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -14,9 +12,9 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		searchAction: {
 			type: Object
 		},
-		searchCleared: {
+		searchApplied: {
```
jeroach: Renamed this to `searchApplied` and updated the logic.

Now it matches `filterApplied` and is less confusing IMO

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -69,30 +71,44 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				}
 			</style>
 			<div class="d2l-quick-eval-activity-list-modifiers">
-				<d2l-hm-filter>
+				<d2l-hm-filter
+					href="[[filterHref]]"
+					token="[[token]]"
+					category-whitelist="[[filterIds]]"
+					on-d2l-hm-filter-filters-loaded="_filtersLoaded"
```
jeroach: Event listening is now here. This is the Polymer recommend way of defining listeners. Alex has done this already over in `d2l-quick-eval-submissions` 😄

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -102,26 +118,48 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+		];
+	}
+
+	get filterIds() {
+		// [ 'activity-name', 'enrollments' ]
+		const filters = [ 'c806bbc6-cfb3-4b6b-ae74-d5e4e319183d', 'f2b32f03-556a-4368-945a-2614b9f41f76' ];
+		return filters;
+	}
+
+	_loadFilterAndSearch(entity) {
```
jeroach: Call this in the `observers` to ensure the search and filters are good to go

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -102,26 +118,48 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
-		return !(this._data.length) && !(this.filterApplied || !this.searchCleared);
+		return !(this._data.length) && !(this.filterApplied || this.searchApplied);
 	}
 
 	_shouldShowNoCriteriaResults() {
-		return !(this._data.length) && (this.filterApplied || !this.searchCleared);
+		return !(this._data.length) && (this.filterApplied || this.searchApplied);
+	}
+
+	_activitiesSearchLoaded() {
```
jeroach: Just added this to show the pattern. `activities` and `submissions` can define their own methods to handle events from their `d2l-hm-search` and `d2l-hm-filter`

These methods:
- Do any DOM operations specific to the component
- Call into the common behavior to do the data logic

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -106,8 +106,9 @@ class D2LQuickEval extends mixinBehaviors(
 				</d2l-quick-eval-search-results-summary-container>
 				<d2l-quick-eval-submissions href="[[href]]" token="[[token]]" logging-endpoint="[[loggingEndpoint]]" master-teacher="[[masterTeacher]]"></d2l-quick-eval-submissions>
 			</div>
-			<d2l-quick-eval-activities href="[[href]]" token="[[token]]" logging-endpoint="[[loggingEndpoint]]" hidden$="[[!_showActivitiesView]]"></d2l-quick-eval-activities>
 		`;
+		// TODO: put this back in the html text - breaks QE tests due to event listeners still present at this level
```
jeroach: Very temporary - can go back in once the event listener logic is moved out of `d2l-quick-eval`

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -102,26 +118,48 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
 
+	static get observers() {
+		return [
+			'_loadFilterAndSearch(entity)'
+		];
+	}
+
+	get filterIds() {
+		// [ 'activity-name', 'enrollments' ]
+		const filters = [ 'c806bbc6-cfb3-4b6b-ae74-d5e4e319183d', 'f2b32f03-556a-4368-945a-2614b9f41f76' ];
```
AlexBedley: Where did these GUIDs come from? Has Jamie already implemented filtering?

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: Hmm thinking about this more... if you add in `e.stopPropagation()` to the behaviours, does that make the tests pass? That's probably a better short-term solution than commenting out the component. We might even want to leave in the stop propagation.

test/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -0,0 +1,86 @@
+import SirenParse from 'siren-parser';
+
+(function() {
```
AlexBedley: nit: This is not necessary

test/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -0,0 +1,86 @@
...
+	var filterBehavior;
```
AlexBedley: nit: should be `let` and also this can be scoped to the `suite` instead of the whole file

components/d2l-quick-eval/d2l-quick-eval-activities.js
jeroach: Hmmm, I may have made an incorrect assumption. I took them from here:
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-quick-eval/d2l-quick-eval.js#L244

The assumption was that each guid is mapped to a given `global` filter type. If it is instead one guid per filter implemented for a given usage, then these are wrong.

Since the activities view cannot filter by date, I removed that filter.

AlexBedley: Since the new API is returning ActivityUsage instead of UserActivityUsage, the type is different so the filter is different, so they'll have different GUIDs. We'll have to wait for Jamie to implement before we can fill this in.

In the meantime, we can remove the whitelist completely (so we'll get filters as Jamie adds them) and we just have to remember to add the whitelist back in before we ship.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -69,30 +71,44 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				}
 			</style>
 			<div class="d2l-quick-eval-activity-list-modifiers">
-				<d2l-hm-filter>
+				<d2l-hm-filter
+					href="[[filterHref]]"
+					token="[[token]]"
+					category-whitelist="[[filterIds]]"
```
AlexBedley: You'll need to remove this line as well (as well as the `filterIds` getter) otherwise we won't see any filters

jeroach: Ah, whitelist, yes :)

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: `e.stopPropagation()` exploded the tests when I tried it quickly. Likely would need to refactor/add logic to a lot of the tests to support this. Personally, since it's a temporary thing, I'm not inclined to put a lot of effort into test fixing.

AlexBedley: I'm surprised that it blew up the tests. I didn't think we were relying on propagation.

CarolBusman: Don't we add custom event listeners in the tests to verify states after events are thrown? If we stop propagation in the component, then the event won't reach the tests, no?

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -102,26 +117,47 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
 		};
 	}
 
+	static get observers() {
+		return [
+			'_loadFilterAndSearch(entity)'
+		];
+	}
+
+	get filterIds() {
```
AlexBedley: We should ruthlessly delete unused code

jeroach:
CI failed for an unrelated reason - kicked again

PR #214, opened 2019-06-05
Abedley/us106353/refactor8

AlexBedley:
Turns out we need the last bit of the refactor to make filter/search work in a 2-application world.

* Remove EntityBehaviour from quick-eval
  * move search/filter to `submissions`
  * delete a ton of useless code now that the data only lives in `submissions`
  * I tried to mostly move the functions, but I couldn't resist doing some cleanup. I'll call it out via comments
* Ignore quick-eval tests for now
  *  Just like the skipped list tests, these tests will heavily change after the next PR
  * Most of them are written as hybrid unit/integration tests, making a lift and shift difficult
  * We should approach our second testing attempt holistically and really try and nail down what we want tests to look like.

TODO:

- [x] fix tests

Next PR: gonna take a breather and rename `d2l-quick-eval-activities-list` to `d2l-quick-eval-submissions-table`


components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
```diff
@@ -101,6 +101,13 @@ D2L.PolymerBehaviors.Siren.D2LSirenHelperBehaviorImpl = {
 			searchParams[param.name] = param.value;
 		});
 		return parsedUrl.pathname + DictToQueryString(searchParams);
+	},
+
+	_getAction: function(entity, name) {
```
AlexBedley: Figured this function belonged here

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -267,10 +267,6 @@ class D2LQuickEvalActivitiesList extends QuickEvalLogging(QuickEvalLocalize(Poly
 				type: Array,
 				value: [ ]
 			},
-			_numberOfActivitiesShownInSearchResults: {
```
AlexBedley: We don't need to track this in 3 places anymore. Yay refactor :)

components/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -460,21 +451,6 @@ class D2LQuickEvalActivitiesList extends QuickEvalLogging(QuickEvalLocalize(Poly
 		);
 	}
 
-	_dispatchActivitiesShownInSearchResultsEvent(countOfSearchResults) {
```
AlexBedley: This is an example of "communication code" that will die because everything lives in `submissions` now

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -1,19 +1,90 @@
...
+				[hidden] {
+					display: none;
+				}
+				.d2l-quick-eval-activity-list-modifiers {
+					float: right;
+				}
+				.clear {
+					clear: both;
+				}
+				d2l-quick-eval-activities-list {
```
AlexBedley: This is the only new css. I had to change the selector to the list to make the top padding work again.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -74,14 +144,50 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
+				value: false
+			},
 		};
 	}
 
 	static get observers() {
 		return [
 			'_loadData(entity)',
-			'_handleSorts(entity)',
-			'_dispatchPageSizeEvent(_numberOfActivitiesToShow)'
```
AlexBedley: No longer needed now that everything that needs to know about pagesize lives here

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -129,6 +235,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 							if (u.entity.entities) {
 								const result = await this._parseActivities(u.entity);
 								this._data = this._data.concat(result);
+								this._updateSearchResultsCount(this._data.length);
```
AlexBedley: Before we were observing `_data` changes to set this. I decided to just add this call in the 3 places we know it changes (search, filter, load more) for now. Hopefully going to have a simpler solution after I add the search behaviour.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,23 +372,99 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		);
 	}
 
+	_filtersLoaded(e) {
+		this.list.filterApplied = e.detail.totalSelectedFilters > 0;
```
AlexBedley: There's still a bunch of garbage in here like this because `list` still knows about things :( but I'm trying to make the most of the time since the refactor has gone on long enough

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,23 +372,99 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
+		this._logError(e.detail.error, { developerMessage: 'Failed to retrieve search results.' });
+		this.list.setLoadingState(false);
+		this._showSearchError = true;
+	}
+
+	_searchResultsMessageEnabled() {
+		return this._showSearchResultSummary && this.searchEnabled;
+	}
+
+	// move
```
AlexBedley: oops

- [ ] delete this

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,23 +372,99 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
+	}
+
 	_computeNumberOfActivitiesToShow(data, currentNumberOfActivitiesShown) {
 		return Math.max(data.length, currentNumberOfActivitiesShown);
 	}
 
-	_dispatchPageSizeEvent(numberOfActivitiesToShow) {
-		this.dispatchEvent(
-			new CustomEvent(
-				'd2l-quick-eval-activities-list-activities-shown-number-updated',
```
AlexBedley: no longer needed :)

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -121,223 +54,34 @@ class D2LQuickEval extends mixinBehaviors(
...
-			_searchResultsCount: {
-				type: Number,
-				value: 0
+			loggingEndpoint: {
+				type: String
 			},
-			_moreSearchResults: {
-				type: Boolean,
-				value: false
+			href: {
```
AlexBedley: Since I removed the behaviours, I had to add in these properties explictly, otherwise we would not be able to pass them to the children.

This kind of highlights an interesting design pattern we might try. If we have a behaviour that a child implements but the parent does not, it means we have to explicitly add these pass-through properties. However, if we split our behaviours into 2, we could accomplish this with less work from developers (and make our intentions with the element clearer). For example, if we split `EntityBehaviour` into `EntityBehaviourProperties` and `EntityBehaviour` (with`EntityBehaviour` implementing  `EntityBehaviourProperties`) then quick-eval could implement `EntityBehaviourProperties` and get the pass-through properties for free. Just a thought!

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -121,223 +54,34 @@ class D2LQuickEval extends mixinBehaviors(
...
-				value: false
+			token: {
+				type: Object,
 			}
 		};
 	}
 
 	static get is() { return 'd2l-quick-eval'; }
 
-	attached()  {
```
AlexBedley: Instead of using attached/detached, I'm using the polymer recommended annotated event listeners. https://polymer-library.polymer-project.org/3.0/docs/devguide/events#annotated-listeners

Also, as an aside, we should stop using `attached` and `detached` because they were deprecated in Polymer 2.0. Instead we should be using `connectedCallback` and `disconnectedCallback` which is Polymer 3.0 and Lit compliant. The biggest thing to know is to not forget to call `super` in these methods, otherwise they are exactly the same as their polymer 1.0 versions. https://polymer-library.polymer-project.org/3.0/docs/devguide/custom-elements#element-lifecycle

The only reason `attached` and `detached` are working is because we're using `mixinBehaviours` which provides the backwards compatible behaviour.

test/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -21,31 +21,6 @@ suite('d2l-quick-eval-submissions', function() {
 		assert.equal(submissions.token, 't');
 	});
 
-	test('when _updateNumberOfActivitiesToShow updated, event "d2l-quick-eval-activities-list-activities-shown-number-updated" fires', function(done) {
```
AlexBedley: These tests were deleted because we no longer need `d2l-quick-eval-activities-list-activities-shown-number-updated`

test/d2l-quick-eval/d2l-quick-eval.html
```diff
@@ -98,7 +98,8 @@
 			};
 
 			suite('d2l-quick-eval', function() {
-				setup(function() {
+				setup(/** @this Mocha */function() {
```
AlexBedley: This comment is to prevent eslint from blowing up about the risky `this` (which is fine in this case because it's a Mocha global)

components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
jeroach: Actually, will it be used anymore? I know the search behavior doesn't use it. What was the other usage?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
jeroach: This will also get cleaned up when `submissions` is updated to use the filter behavior, so I'm cool with a bit of garbage kicking around for a bit longer.

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: good point - I was not aware - thanks for the information!

components/d2l-quick-eval/d2l-quick-eval.js
jeroach: Could we not just keep `EntityBehavior` on `d2l-quick-eval`? That would give us the `href` and `token`

Either approach is cool though (six one, half dozen the other) :)

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -17,10 +8,7 @@ import './d2l-quick-eval-submissions.js';
  * @customElement
  * @polymer
  */
-class D2LQuickEval extends mixinBehaviors(
-	[D2L.PolymerBehaviors.Siren.EntityBehavior],
-	QuickEvalLogging(QuickEvalLocalize(PolymerElement))
-) {
+class D2LQuickEval extends PolymerElement {
```
jeroach: This class is so clean now :)

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,23 +372,98 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
-				{
-					detail: {
-						count: numberOfActivitiesToShow
-					},
-					composed: true,
-					bubbles: true
-				}
-			)
-		);
+	_getSearchAction(entity) {
```
jeroach: Yeah I think this is the only usage of `getAction` so it can go when `submissions` starts to use the `searchBehavior`

components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
jeroach: No other usages - this can go when `submissions` uses `searchBehavior`

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: We could, but it would fall apart once we add the second href for the activities app. And this would encourage the poor design we just refactored out of - two controllers, each with the same copy of entity that needs to be kept in sync.

PR #217, opened 2019-06-07
 Get rid of complicated _getDataProperty

AlexBedley:
It probably made sense when we made it, but now it's just a function that replaces basic functionality (namely, `.`)

Killing it with fire!


AlexBedley:
The above "changed the base branch" message was to force github to reset the diff (Josh I think you had this problem earlier and we just remade the PR)

PR #218, opened 2019-06-07
Move loading to controller

AlexBedley:
Further cleanup before sort/filter behaviours. Let's get loading into the controller (move decision making out of `table`).

This change also contains what I would consider a bugfix around loading. We weren't setting the load states before sending the sort request so you would see some weird data, then the loading screen would flash (during the time it takes to parse the new data), then you'd see your sorted data. This change just turns the loading screen on before we send the sort request (as you'd expect).

Previously:
1. Click sort header to sort, send sort request
2. See some wonky rearranged data
3. Data is received
4. Loading screen flashes on then off
5. Sorted data is displayed

This PR:
1. Click sort header to sort, send sort request
2. Loading screen appears
3. Data is received
4. Loading screen disappears
5. Sorted data is displayed

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
@@ -294,11 +294,6 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 		];
 	}
 
-	setLoadingState(state) {
-		this.set('_fullListLoading', state);
-		this.set('_loading', state);
-	}
-
 	_isLoadingMore(fullListLoading, isLoading) {
 		return !fullListLoading && isLoading;
 	}
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -92,6 +92,8 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				master-teacher="[[masterTeacher]]"
 				_data="[[_data]]"
 				_header-columns="[[_headerColumns]]"
+				_loading="[[_loading]]"
+				_full-list-loading="[[_fullListLoading]]"
 				on-d2l-quick-eval-submissions-table-load-more="_loadMore"
 				on-d2l-quick-eval-submissions-table-sort-requested="_handleSortRequested">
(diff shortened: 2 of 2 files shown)


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -347,6 +357,8 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		});
 
 		if (result) {
+			this._loading = true;
+			this._fullListLoading = true;
```
AlexBedley: This is the bugfix

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -294,11 +294,6 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 		];
 	}
 
-	setLoadingState(state) {
```
AlexBedley: I considered just moving `setLoadingState` to the controller but it was hiding the fact that we only need to set one of the loading states to get what we want. For now just set both of them explicitly, hopefully I find time in the future to refactor our loading booleans.

AlexBedley: To expand further:

`_loading` controls the loading spinner present at the bottom of the page
`_fullListLoading` controls the visibility of the table and the flashy skeleton loading graphic

My thoughts are we don't need to set `_loading` everytime we also set `_fullListLoading`. We only need to set it when we request another page. I wanna refactor `_fullListLoading` to just `loading` and `_loading` to `_loadingMore`.

AlexBedley:
The failure was some saucelabs connection failure

components/d2l-quick-eval/d2l-quick-eval-submissions.js
mtjandra: Seems to make sense, I don't remember design being against this during planning

AlexBedley: Yeah someone called it out last demo, figured I could just put the fix in. Unlike the rest of these PRs, I created a story specifically for this bugfix so it woudln't get lost: https://rally1.rallydev.com/#/238990806412d/detail/userstory/312908634608

PR #220, opened 2019-06-07
Move health into controller

AlexBedley:
Moving more decision making to the controller part 3. `table` should just be told what the health is, rather than keep track of it itself.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -95,6 +95,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				_loading="[[_loading]]"
 				_full-list-loading="[[_fullListLoading]]"
 				show-load-more="[[_showLoadMore]]"
+				_health="[[_health]]"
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -95,6 +95,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				_loading="[[_loading]]"
 				_full-list-loading="[[_fullListLoading]]"
 				show-load-more="[[_showLoadMore]]"
+				_health="[[_health]]"
```
mtjandra: Oh one interesting thing to note, which I didn't realize earlier. I think the `_UnderscoreSomething` typically refers to a private member. I think the ideal would've been to make it public. Since components/modules outside know about it and is driving it

AlexBedley: Agreed. I'm making all new members "public" (no underscore) but leaving whatever was there before until I do a final pass and remove the underscore from all of the passed in params to `table`

PR #222, opened 2019-06-07
Clean up usages of `list` in submissions controller

AlexBedley:
I'm preferring the pattern of passing information to children via attributes, rather than searching the DOM for the element and calling functions on it directly

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -225,6 +225,10 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				type: Boolean,
 				value: false
 			},
+			masterTeacher: {
+				type: Boolean,
+				value: false
+			}
 		};
 	}
 
@@ -233,10 +237,6 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			'_loadData(entity)',
 			'_handleSorts(entity)'
 		];
-	}
-
-	get list() {
-		return this.shadowRoot.querySelector('d2l-quick-eval-submissions-table');
 	}
 
 	async _loadData(entity) {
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -225,6 +225,10 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				type: Boolean,
 				value: false
 			},
+			masterTeacher: {
```
AlexBedley: This was forgotten, just added it to be explicit (I actually though it was mandatory but I guess not...?)

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -235,10 +239,6 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		];
 	}
 
-	get list() {
```
AlexBedley: deleting the getter to discourage usage of it

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -271,7 +271,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			this._followHref(this._pageNextHref)
 				.then(async function(u) {
 					if (u && u.entity) {
-						const tbody = this.list.shadowRoot.querySelector('d2l-tbody');
+						const tbody = this.shadowRoot.querySelector('d2l-quick-eval-submissions-table').shadowRoot.querySelector('d2l-tbody');
```
AlexBedley: couldn't find another way to do this, so leaving it in for now. I think it's fine, in this case we literally need to call HTMLElement functions so fetching it via the DOM seems right.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -324,7 +324,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 				const getUserName = this._getUserPromise(activity, item);
 				const getCourseName = this._getCoursePromise(activity, item);
 				const getMasterTeacherName =
-					this.list._shouldDisplayColumn('masterTeacher')
+					this.masterTeacher
```
AlexBedley: This is the same check as `_shouldDisplayColumn`, no functional change

PR #224, opened 2019-06-11
Cbusman/basic activities list

CarolBusman:
Hooking up the activities controller and list to populate with the grouped activities. There has been no additional css/formatting added.
Does NOT contain search or filter functionality yet. Though the rel and action exist in the HM call, the actual functionality has not yet been implemented. The data/search/filter checkboxes in the demo have been left in for now until we have a search and filter available. Note that the data checkbox populates the list with dummy data, but the titles of the activities were not hooked up.
Note also that `_loading`, etc. has also not yet been added - this is the next step. Currently the kitty is displayed while the HM data is being retrieved, which disappears once we populate `_data`.

image


components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -90,7 +90,7 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 	async _getActivityPromise(activityEntity) {
 		let rel;
 		let activityIcon;
-		if (activityEntity.hasClass(Classes.activities.userQuizAttemptActivity)) {
+		if (activityEntity.hasClass(Classes.activities.userQuizAttemptActivity) || activityEntity.hasClass(Classes.activities.userQuizActivity)) {
```
AlexBedley: Hmm this seems suspicious... we're relying on the convenient fact that they both access activity name via the same rels. I guess rels shouldn't change much so maybe we're ok.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,73 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import {QuickEvalLocalize} from '../QuickEvalLocalize.js';
+import '../activity-card/d2l-quick-eval-activity-card.js';
+
+class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
+	static get is() { return 'd2l-quick-eval-activities-list'; }
+	static get template() {
+		return html`
+			<div>
```
AlexBedley: Is this `div` required?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,73 @@
...
+				<dom-repeat items="[[_courses]]" as="c">
+					<template>
+						<h3>[[c.name]]</h3>
```
AlexBedley: Hmm can we be sure that `h3` is right here? Semantically, is this an `h3` (as in, is there guaranteed to be an h2 and h1 above it) or do we just want the styling of an `h3`?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,73 @@
...
+					// }
+				]
+			},
+			_courses: {
+				type: Array,
+				computed: '_groupByCourse(activities)'
+			},
+			token: {
+				type: String,
+				value: ''
```
AlexBedley: does this default value serve a purpose?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,73 @@
...
+			}
+		};
+	}
+
+	_groupByCourse() {
```
AlexBedley: I think this function should be moved to the controller, so we don't maintain a separate copy of the data here. I know it's bound to `activities` so it's probably fine but I think it'll make testing easier and make this class easier to reason about

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,73 @@
...
+		const grouped = this.activities.reduce((acts, a) => {
```
AlexBedley: nit: this function is very hard to verify. I think a less efficient, but more readable groupBy would be preferred (or import a library with a groupBy like underscore etc.)

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -107,6 +122,26 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
+		if (entity) {
+			if (entity.hasClass(Classes.activities.quizActivity)) {
+				return 'quiz';
+			} else if (entity.hasClass(Classes.activities.assignmentActivity)) {
+				return 'assignment';
+			} else if (entity.hasClass(Classes.activities.discussionActivity)) {
+				return 'discussion';
+			}
+		}
+		return '';
```
AlexBedley: does this default value serve a purpose? I think omitting this line makes the code more clear (unless we want empty string, it seems like we just want a falsey value). I know this occurs many times in our codebase but I feel like it makes the code harder to read (it makes me think we're relying on this being empty string, but we really don't care).

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -112,6 +113,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				<h2 class="d2l-quick-eval-no-criteria-results-heading">[[localize('noResults')]]</h2>
 				<p class="d2l-body-standard">[[localize('noCriteriaMatch')]]</p>
 			</div>
+			<d2l-quick-eval-activities-list 
+				hidden$="[[!_shouldShowActivitiesList(_data, filterApplied, searchCleared)]]"
```
AlexBedley: `_shouldShowActivitiesList` takes 0 parameters but is called with 3 here

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -130,13 +135,62 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+
+		} catch (e) {
+			this._logError(e, {developerMessage: 'Unable to load activities from entity.'});
+			return Promise.reject(e);
+		}
+	}
+
+	async _parseActivities(entity) {
+		const promises = [];
+		entity.entities.forEach(function(activity) {
```
AlexBedley: nit: I'm not a huge fan of this style of bringing multiple promises together via `item`, side effects make it hard to see what's going on and makes the functions harder to test.

I think if we use `async` more liberally we can make parsing a bit cleaner:

```js
async _parseActivities(entity) {
	return Promise.all(entity.entities.map(async function(activity) {
		const evaluationStatus  = await this._getEvaluationStatusPromise(activity);
		const courseName = await this._getCoursePromise(activity);

		return {
			courseName: courseName,
			assigned: evaluationStatus.assigned,
			completed: evaluationStatus.completed,
			published: evaluationStatus.published,
			evaluated: evaluationStatus.evaluated,
			unread: evaluationStatus.unread,
			resubmitted: evaluationStatus.resubmitted,
			dueDate: this._getActivityDueDate(activity),
			activityType: this._getActivityType(activity),
			activityNameHref: this._getActivityNameHref(activity)
		};
	}.bind(this)));
}
```

This way seems cleaner, and makes the tests for `_getCoursePromise` and `_getEvaluationStatusPromise` easier. I know `_getCoursePromise` is shared with the other class, but I'd be in favour of refactoring that as well.

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,222 @@
+import '@polymer/iron-test-helpers/mock-interactions.js';
+
+(function() {
```
AlexBedley: This wrapper is not necessary and should be deleted

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,222 @@
...
+	let list;
```
AlexBedley: This variable should be defined at the smallest scope possible

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,222 @@
...
+
+	const activitiesSource = [
```
AlexBedley: nit: exporting this to a json file might make this file more readable

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,222 @@
...
+			published: 3,
+			evaluated: 4,
+			unread: 5,
+			resubmitted: 6,
+			dueDate: '2019-01-01T01:01:01.001Z',
+			activityType: 'discussion'
+		}
+	];
+
+	var expectedCourses = [
```
AlexBedley: `const`

test/d2l-quick-eval/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,222 @@
...
+						assert.equal(card.resubmitted, a.resubmitted);
+						assert.equal(card.dueDate, a.dueDate);
+						assert.equal(card.activityType, localize(a.activityType));
+						assert.equal(card.activityNameHref, a.activityNameHref);
+						assert.equal(card.token, list.token);
+						element++;
+					}
+				}
+			});
+			done();
```
AlexBedley: it's hard to tell, but is this `done` being called in the right place (inside the `requestAnimationFrame`)?

test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -1,49 +1,105 @@
 import '@polymer/iron-test-helpers/mock-interactions.js';
 
 (function() {
```
AlexBedley: delete unnecessary code

test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -1,49 +1,105 @@
...
-	var list;
+	let act, list, noSubmissionComponent, noCriteriaResultsComponent;
```
AlexBedley: smallest possible scope

test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -1,49 +1,105 @@
...
+			unread: 5,
+			resubmitted: 6,
+			dueDate: '2019-01-01T01:01:01.001Z',
+			activityType: 'discussion'
+		}
+	];
+
+	async function loadPromise(url) {
+		var entity = await window.D2L.Siren.EntityStore.fetch(url, '');
+		await act._loadData(entity.entity);
```
AlexBedley: I would pass in `act`, or more broadly, can we replace this function and just set `href`?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
CarolBusman: Not at the moment I don't believe (I think you can have a `dom-repeat` as the base for a component?), but we will need it for styling later on.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
CarolBusman: The design spec specified `h3`:
image

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
CarolBusman: It's a habit to define an explicit "blank" value so that I don't run up against weird `undefined` and/or `null` behaviours down the line. Should it be removed?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
AlexBedley: could we not style `:host` instead of this div in the future?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
AlexBedley: I'm wondering if the design literally wants an `h3` though. Won't it mess up tab order and screen readers if the headers are improperly nested? I'm wondering if the design just wants look and feel similar to a "Heading 3" and not necessarily an `h3`.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
AlexBedley: I think so. Either we expect weird behaviour, and we should handle that expected behaviour, or we're not sure what will happen, and in that case I think we want to fail fast and loudly. Unless there's a specific case we're guarding against I think we should just remove the defaults.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
CarolBusman: The thing I'm concerned about is padding. If that works with `:host`, then yeah we can probably do away with the `div`.

components/d2l-quick-eval/d2l-quick-eval-activities.js
CarolBusman: The parameters in the binding are so that it triggers a recalculation when those properties change, but inside the function we're accessing the properties directly off of `this` so we'd get unused parameter errors.

test/d2l-quick-eval/d2l-quick-eval-activities.js
CarolBusman: If we call `_loadData` directly like this, it lets us wait until it's done so we can verify the results. I don't know that there's a way to wait for the data to be loaded if we just set the href outside of looping a timeout until it's ready (which I don't think we want to do).

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
CarolBusman: Checked with design, and they do want this to be an `h3`.

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: ah nice! I still think we should use the passed-in parameters (add and use `data` instead of `this._data`) but I'm ok omitting unused parameters to make the linter happy

test/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: We can listen for `d2l-siren-entity-changed`

https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L105

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -0,0 +1,57 @@
...
+			</dom-repeat>
+		`;
+	}
+	static get properties() {
+		return {
+			courses: {
+				type: Array,
+				value: [
+					// {
+					// 	courseName: '',
```
AlexBedley: Is this comment still accurate?

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -38,6 +38,17 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 			}.bind(this));
 	},
 
+	_getCourseName: async function(entity) {
+		let result;
+		await this._followLink(entity, Rels.organization)
```
AlexBedley: Writing this without `async/await` is probably more succint: `return this._followLink(...).then(o => o.entity.properties.name)`. If you REALLY want to use `await` then I would suggest a version without side-effects like:

```js
let result = await this._followLink(...).then(o => o.entity.properties.name);
return result;
```

Always prefer no side effects IMO, it makes code easier to reason about. This applies for the function below as well.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -130,13 +135,62 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
 	static get observers() {
 		return [
+			'_loadData(entity)',
 			'_loadFilterAndSearch(entity)'
 		];
 	}
 
+	async _loadData(entity) {
+		if (!entity) {
+			return Promise.resolve();
```
AlexBedley: After looking into the async/await stuff this line I originally wrote is a bit weird. We could just `return;` which might make our intentions more clear

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -130,13 +135,62 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+				const result = await this._parseActivities(entity);
+				this._data = result;
+			} else {
+				this._data = [];
+				this.list._pageNextHref = '';
+			}
+
+		} catch (e) {
+			this._logError(e, {developerMessage: 'Unable to load activities from entity.'});
+			return Promise.reject(e);
```
AlexBedley: similarly, we can just `throw e;` here which is probably more idiomatic with `async`

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -130,13 +135,62 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+				resubmitted: evalStatus.resubmitted,
+				dueDate: this._getActivityDueDate(activity),
+				activityType: this._getActivityType(activity),
+				activityNameHref: this._getActivityNameHref(activity)
+			};
+		}.bind(this)));
+		return this._groupByCourse(result);
+	}
+
+	_groupByCourse(act) {
```
AlexBedley: I would like to see some unit tests specifically targeting this helper function

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
AlexBedley: Hmm to me this is a technical problem (to get the screen readers to work, and to keep our html semantically correct), not really a design problem, but I'm ok playing the waiting game until we go over this with a screen reader

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -38,6 +38,19 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
+	_getCourseNamePromise: async function(entity) {
+		return this._followLink(entity, Rels.organization)
+			.then(function(o) {
+				if (o && o.entity && o.entity.properties) {
+					return o.entity.properties.name;
+				}
+			});
+	},
+
 	_getCoursePromise: function(entity, item) {
```
AlexBedley: One thing I just remembered is we need to clean this duplicate function up after the merge

CarolBusman: Yup! But that's used by the other controller, and I didn't want to touch yet another thing in this PR, so that's going in after. And I'll probably change that parsing function all at once to follow this pattern.

PR #225, opened 2019-06-11
Refactor loading to prepare for behaviours

AlexBedley:
This PR originally came from this comment (I was hoping we didn't have to do it but it makes sense to do it now): https://github.com/BrightspaceHypermediaComponents/activities/pull/218#discussion_r291725739

I got halfway through adding the filter/search behaviours but I encountered a problem around loading, specifically the default values. The default value for `_loading` is `true`, we assume the initial state will be to fetch the data. However the initial value for `filtersLoading` and `searchLoading` is `false`, because at first we are neither filtering or searching. But since they share a single loading state, `_loading`, there's no good way to differentiate "on the first load" and while it could be bandaided with more loading booleans, it felt like we were heading down the wrong design path.

This PR includes a bit of cleanup in the table: I changed the variables names to denote their exact purpose (`showLoadingSkeleton`, `showLoadingSpinner`). I also made the loading variables distinct (so we aren't setting both of them at the same time all the time). Lastly, I left some scaffolding for my behaviour change (the compute methods).

**Before:**

* `_loading` - denotes if the app is currently waiting for an http request to return
* `_fullListLoading` - exactly the same as `_loading` except in the case where we load more, this wasn't used.

Essentially the absence of `_fullListLoading` was used to determine if we should display the spinner or the skeleton and I found it a bit confusing that we had 2 variables that do pretty much the exact same thing.

**This PR:**

* `_loading` - denotes if the app is currently loading via initial load, search, sort, or filtering. Controls `showLoadingSkeleton`.
* `_loadingMore` - denotes if the app is currently loading via load more. Controls `showLoadingSpinner`.

Note that there is no case where these will both be `true` anymore (since we only allow 1 request at the same time).

**In the future:**

Each behaviour (sort, filter, search) will maintain it's own loading state. These will be coalesced in the controller under `showLoadingSkeleton`. This will allow us to write behaviours more naturally, but still provide a convenient interface to work with.

As usual, happy to swing by your desk to give a better walkthrough

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
@@ -132,7 +132,7 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 				}
 			</style>
 			<d2l-offscreen id="d2l-quick-eval-submissions-table-table-summary">[[localize('tableTitle')]]</d2l-offscreen>
-			<d2l-table class="d2l-quick-eval-table" type="light" hidden$="[[_fullListLoading]]" aria-describedby$="d2l-quick-eval-submissions-table-table-summary" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
+			<d2l-table class="d2l-quick-eval-table" type="light" hidden$="[[showLoadingSkeleton]]" aria-describedby$="d2l-quick-eval-submissions-table-table-summary" aria-colcount$="[[_headerColumns.length]]" aria-rowcount$="[[_data.length]]">
 				<d2l-thead>
 					<d2l-tr>
 						<dom-repeat items="[[_headerColumns]]" as="headerColumn">
@@ -214,9 +214,9 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]">
 				[[localize(_health.errorMessage)]]
 			</d2l-alert>
-			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!_loading]]">[[localize('loading')]]</d2l-offscreen>
-			<d2l-quick-eval-skeleton hidden$="[[!_fullListLoading]]"></d2l-quick-eval-skeleton>
-	     	<d2l-loading-spinner size="80" hidden$="[[!_isLoadingMore(_fullListLoading,_loading)]]"></d2l-loading-spinner>
+			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!isLoading]]">[[localize('loading')]]</d2l-offscreen>
+			<d2l-quick-eval-skeleton hidden$="[[!showLoadingSkeleton]]"></d2l-quick-eval-skeleton>
+	     	<d2l-loading-spinner size="80" hidden$="[[!showLoadingSpinner]]"></d2l-loading-spinner>
 
 			<template is="dom-if" if="[[showLoadMore]]">
 				<div class="d2l-quick-eval-submissions-table-load-more-container">
@@ -259,7 +259,7 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 				type: Array,
 				value: [ ]
 			},
-			_fullListLoading: {
+			showLoadingSkeleton: {
 				type: Boolean,
 				value: true
 			},
@@ -270,9 +270,9 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 					errorMessage: ''
 				}
 			},
-			_loading: {
+			showLoadingSpinner: {
 				type: Boolean,
-				value: true
+				value: false
 			},
 			showLoadMore: {
 				type: Boolean,
@@ -286,6 +286,10 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 				type: Boolean,
 				value: false
 			},
+			isLoading: {
+				type: Boolean,
+				computed: '_computeIsLoading(showLoadingSpinner, showLoadingSkeleton)'
+			}
 		};
 	}
 	static get observers() {
@@ -294,8 +298,8 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 		];
 	}
 
-	_isLoadingMore(fullListLoading, isLoading) {
-		return !fullListLoading && isLoading;
+	_computeIsLoading(showLoadingSpinner, showLoadingSkeleton) {
+		return showLoadingSpinner || showLoadingSkeleton;
 	}
 
 	_handleNameSwap(entry) {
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -214,9 +214,9 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 			<d2l-alert id="list-alert" type="critical" hidden$="[[_health.isHealthy]]">
 				[[localize(_health.errorMessage)]]
 			</d2l-alert>
-			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!_loading]]">[[localize('loading')]]</d2l-offscreen>
-			<d2l-quick-eval-skeleton hidden$="[[!_fullListLoading]]"></d2l-quick-eval-skeleton>
-	     	<d2l-loading-spinner size="80" hidden$="[[!_isLoadingMore(_fullListLoading,_loading)]]"></d2l-loading-spinner>
+			<d2l-offscreen role="alert" aria-live="aggressive" hidden$="[[!isLoading]]">[[localize('loading')]]</d2l-offscreen>
```
AlexBedley: This is the only component that cares if we're loading (and not what we're loading) so I left it in the table as a computed value. But now I'm thinking I should make it `readOnly` to prevent accidental misuse.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -187,20 +187,12 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
-				type: Boolean,
-				value: true
-			},
 			_pageNextHref: {
 				type: String,
 			},
 			_showLoadMore: {
 				type: Boolean,
-				computed: '_computeShowLoadMore(_pageNextHref, _loading)'
+				computed: '_computeShowLoadMore(_pageNextHref, _showLoadingSpinner, _showLoadingSkeleton)'
```
AlexBedley: Aside: You guys might have noticed I'm pseudo-adopting a naming convention:

* `_compute` methods are `computed` properties
* `_show` properties are booleans that mainly control visibility
* `_on` functions are event listeners
*`_onXChanged` functions are observers

etc.

I'm finding it much easier to read the code this way

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -260,14 +267,13 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
 		} finally {
-			this._fullListLoading = false;
 			this._loading = false;
 		}
 	}
 
 	_loadMore() {
-		if (this._pageNextHref && !this._loading) {
-			this._loading = true;
+		if (this._pageNextHref) {
```
AlexBedley: I got rid of the `loading` check. I couldn't figure out how to trigger it multiple times accidentally so I just deleted it as unused code 🤷‍♂ I can put it back in if we feel strongly

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -514,18 +515,25 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
-		return !_data.length && !_loading && _health.isHealthy && !(_filterApplied || _searchApplied);
+	_computeShowNoSubmissions(_data, _loading, _loadingMore, _health, _filterApplied, _searchApplied) {
+		return !_data.length && !_loading && !_loadingMore && _health.isHealthy && !(_filterApplied || _searchApplied);
+	}
+
+	_computeShowNoCriteria(_data, _loading, _loadingMore, _health, _filterApplied, _searchApplied) {
+		return !_data.length && !_loading && !_loadingMore && _health.isHealthy && (_filterApplied || _searchApplied);
+	}
+
+	_computeShowLoadingSpinner(_loadingMore) {
```
AlexBedley: This function and it's sibling are stubs right now. But this illustrates how the behaviour PR will make use of these and add more loading properties.

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
AlexBedley: Oh TIL computed properties are inherently read-only. Polymer for the win!

https://polymer-library.polymer-project.org/3.0/docs/devguide/properties

mtjandra: What is this thing anyways? Is this for screen readers?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
mtjandra: I think you want `if (this._pageNextHref && !this._loadingMore) `

I think this is to prevent someone spamming the button. Once the button is pressed it should disappear. I think. I could be wrong

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -391,7 +397,6 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 
 		if (result) {
```
mtjandra: This isn't part of your PR but I don't get this check. I think both `Promise.reject()` and `Promise.resolve()` are both `truthy`. Unless I guess `_handleSortRequested` can be called with some event that doesn't match any header ID. Maybe that's the safegaurd

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -228,6 +220,22 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			masterTeacher: {
 				type: Boolean,
 				value: false
+			},
+			_loading: {
```
mtjandra: Isn't this redundant? Why not just use `_showLoadingSkeleton`?

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
AlexBedley: > What is this thing anyways? Is this for screen readers?

yes

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: right now it's redundant but soon it will be one of many loading properties (along with filtersLoading, searchLoading, etc.) that will make up `showLoadingSkeleton`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
mtjandra: This is my only real concern I'd say, unsure if this allow pressing loading more multiple times, where the behaviour is different from before

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: > Unless I guess `_handleSortRequested` can be called with some event that doesn't match any header ID. Maybe that's the safegaurd

It's protecting against this scenario. The check exists so we can give a better error message than `TypeError: Cannot read property 'then' of undefined` which I got a lot during testing and didn't find helpful.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: > unsure if this allow pressing loading more multiple times

I tried a bunch and couldn't get it to work. I don't mind putting it back in if you feel strongly since it's been there since its inception.

mtjandra: I only feel strongly if it will cause a problem, sound like we are okay :P

AlexBedley: meh I put it back in. I also noticed another spot where we might potentially not reset the loading state (if the entity fails to load)

PR #226, opened 2019-06-13
Add filter behaviour to submissions

AlexBedley:
FYI: @jeroach - I know you're gone but you'll be interested in this when you come back as you wrote the original implementation

This PR adds the filter behaviour to submissions. This involved:

* Refactor the filter behaviour itself to do more
  * it implements EntityBehaviour
  * it now registers event listeners
  * it maintains its own loading state 
* Refactor the submissions controller
  * implement FilterBehaviour
  * Remove event listeners (that now live in behaviour)
  * use new `filtersLoading` property to determine if we show the loading skeleton


components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -12,41 +13,69 @@ D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviourImpl = {
 
 	properties: {
 		filterHref: {
-			type: String
+			type: String,
+			computed: '_computeFilterHref(entity)'
```
AlexBedley: Instead of manually calling `_setFilterHref`, it now just automatically updates when `entity` changes

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -12,41 +13,69 @@ D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviourImpl = {
...
 		},
 		filterApplied: {
 			type: Boolean,
 			value: false
 		},
 		filterError: {
-			type: Boolean,
+			type: Object,
```
AlexBedley: It's now an object that contains the actual error rather than a boolean signalling the presence of an error

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -12,41 +13,69 @@ D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviourImpl = {
...
+		}
 	},
 
-	_setFilterHref: function(entity) {
-		if (entity && entity.hasLinkByRel && entity.hasLinkByRel(Rels.filters)) {
-			this.filterHref = entity.getLinkByRel(Rels.filters).href;
-		} else {
-			this.filterHref = '';
-		}
+	attached: function() {
```
AlexBedley: I had to use the Polymer 1.0 `attached` and `detached` because legacy behaviours don't support `connectedCallback`. I had to make it a legacy behaviour because EntityBehaviour is a legacy behaviour.

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -12,41 +13,69 @@ D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviourImpl = {
...
 	},
 
-	_errorOnFilter: function() {
-		this.filterError = true;
+	_onFilterError: function(e) {
+		this.filtersLoading = false;
+		this.filterError = e.detail;
+	},
+
+	_clearErrors: function() {
```
AlexBedley: Probably the most controversial part of this change is my strategy to deal with clearing errors.

`_clearErrors` needs to clear all the errors (otherwise a search error would prevent filtering to clear the error). Since the behaviours don't know about each other they need a way to communicate. I went with this strategy where `submissions` just overrides `_clearErrors` and clears both the error states itself. I wrote many tests to assert this behaviour.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -235,15 +219,16 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			},
 			_showLoadingSkeleton: {
 				type: Boolean,
-				computed: '_computeShowLoadingSkeleton(_loading)'
+				computed: '_computeShowLoadingSkeleton(_loading, filtersLoading)'
```
AlexBedley: The presence of the loading skeleton depends on `filtersLoading` now as well as the regular `_loading`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -257,6 +242,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			if (entity.entities) {
 				const result = await this._parseActivities(entity);
 				this._data = result;
+				this._updateSearchResultsCount(this._data.length);
```
AlexBedley: Previously we were doing this manually when searching and filtering, but since those both hook into the entity stuff, we can just do it here once.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -439,13 +400,10 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		return filters;
 	}
 
-	_getFilterHref(entity) {
-		return this._getHref(entity, Rels.filters);
-	}
-
+	// @override - do not change the name
```
AlexBedley: I left this comment here to denote that the correctness of this element depends on this function name matching the one in the behaviour

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -523,12 +481,18 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		return !_data.length && !_loading && !_loadingMore && _health.isHealthy && (_filterApplied || _searchApplied);
 	}
 
+	_onFilterErrorChange(filterError) {
```
AlexBedley: I couldn't get logging into the behaviour, so listening for filterErrors this way to still achieve logging

test/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -8,8 +8,11 @@ suite('d2l-hm-filter-behavior', function() {
 	test('default state is correct', () => {
 		assert.isNotOk(filterBehavior.filterHref);
 		assert.isFalse(filterBehavior.filterError);
+		assert.isFalse(filterBehavior.filterApplied);
+		assert.isFalse(filterBehavior.filtersLoading);
```
AlexBedley: This file is basically a brand new file, and it might be easier to review it that way. I tried to keep one assertion per test for simplicity

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
```diff
@@ -12,41 +13,69 @@ D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviourImpl = {
...
+		this.filterError = false;
```
CarolBusman: If `filterError` is an object now, should we be setting this to `null`?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
CarolBusman: So we're going to show the skeleton if we've all the data, but the filters are still loading? Why wait for the filters?

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
mtjandra: Maybe default value should be something like

```
{
   isError: false,
   details: undefined
}
```

Not the greatest names, but I'm sure you can come up with something. Then instead of using false, you can check that `isError` prop

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: Ah maybe it doesn't have the greatest name. `filtersLoading` means we're waiting on the actual filtered data to come back, not the filters themselves. `_loading` covers sort, search, and the initial load right now, but eventually it will only cover the initial load (sort and search will have their own loading variables in their own behaviour)

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
AlexBedley: I didn't expect much pushback on this type coercion but since both you and Miles brought it up I guess I should change it.

My initial thought is I want `filterError` to be truthy and contain error information when there's an error, and falsey when there is not an error. I thought setting it to `false` made my intentions more clear than setting it to `null` or `undefined`.

I'm not a huge fan of having `filterError` use an `isError` property (as Miles suggested) because of all the polymer guff that comes with detecting changes to object properties. If setting it to `null` here is more clear then I would prefer that.

components/d2l-quick-eval/behaviors/d2l-hm-filter-behavior.js
AlexBedley: I thought of this, but decided against it for the reason you mentioned (having to do weird polymer things to get changes detected). It's not a blocker, just an annoyance I wanted to avoid.

AlexBedley:
- [ ] import filter behaviour in `submissions`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
CarolBusman: `filteredDataLoading` maybe? It's a little longer, but would help avoid confusion?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
jeroach: I too had many issues with my attempts to get logging into the behaviour, so this is cool

mtjandra: Hopefully I will get logging into the filter during inspiration, haven't touched it yet though

PR #232, opened 2019-06-26
Update Org to V5

m6jones:
Update this repo in version 20197 to use the new translations from organization repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
 		"d2l-loading-spinner": "BrightspaceUI/loading-spinner#semver:^7",
 		"d2l-localize-behavior": "BrightspaceUI/localize-behavior#semver:^2",
 		"d2l-offscreen": "BrightspaceUI/offscreen#semver:^4",
-		"d2l-organizations": "BrightspaceHypermediaComponents/organizations#semver:^4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
 		"d2l-loading-spinner": "BrightspaceUI/loading-spinner#semver:^7",
 		"d2l-localize-behavior": "BrightspaceUI/localize-behavior#semver:^2",
 		"d2l-offscreen": "BrightspaceUI/offscreen#semver:^4",
-		"d2l-organizations": "BrightspaceHypermediaComponents/organizations#semver:^4",
+		"d2l-organizations": "BrightspaceHypermediaComponents/organizations#semver:^5",
```
mtjandra: What's the breaking/incompatible change between v4 and v5? I skimmed through it a bit looks like mostly serge/localization setup. Just double checking that nothing needs to be adjusted in activities, otherwise should be good!

m6jones: The breaking change was for the enrollment repo.

m6jones: https://github.com/BrightspaceHypermediaComponents/organizations/blob/master/components/d2l-organization-updates/OrganizationUpdatesMixin.js

m6jones: This used to be an behaviour

PR #233, opened 2019-06-26
DE34932 - Treat plus as space

mtjandra:
> I believe this should be correct. Query params should always be encoded. So if we see a + symbol we know it should be a space. To actually represent a plus symbol it should be represented as %2B.
>We receive a + symbol instead of %20 is because of http://search.dev.d2l/source/xref/Lms/lp/framework/core/D2L/LP/Text/Html/UrlUtility.cs#125

Related:
* https://github.com/BrightspaceHypermediaComponents/common/pull/27

We should fix the polymer behavior at some point instead of maintaining these hacks in three places. Same fix follows here as the related PRs above

### DIFF
--- a/components/d2l-quick-eval/compatability/ie11shims.js
+++ b/components/d2l-quick-eval/compatability/ie11shims.js
@@ -3,6 +3,15 @@ All functions in this file should work across all browsers.
 Initially intended for enabling compatability with IE11, but applies to any browser missing some functionality.
 */
 
+const _urlDecodePlusAsSpace = (str) => {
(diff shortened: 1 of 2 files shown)


mtjandra:
>  I wonder if we should be using this if it's not IE11, so only IE11 is using our custom code here

All browsers use this code. I initially named it universal-methods but changed it to indicate it only exists because of ie11 in https://github.com/BrightspaceHypermediaComponents/activities/pull/165

> once IE11 support is dropped, we can just remove all this

Yup!

PR #235, opened 2019-07-02
Cbusman/pass url to activities controller

CarolBusman:
Leaving `href` and the new `submissions-href` properties both in quick eval as the LMS is passing the url into the `href` property currently.

Next steps are to update BSI version, push new BSI version to LMS and update the quick eval renderer to use the new `submissions-href` attribute (and `activities-href`), then return here to remove the old href property and helper function that will no longer be needed.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -39,8 +39,8 @@ class D2LQuickEval extends PolymerElement {
 				<h1 class="d2l-quick-eval-header" hidden$="[[activitiesViewEnabled]]"">[[headerText]]</h1>
 			</template>
 			<d2l-quick-eval-view-toggle current-selected="submissions" hidden$="[[!activitiesViewEnabled]]" on-d2l-quick-eval-view-toggle-changed="_toggleView"></d2l-quick-eval-view-toggle>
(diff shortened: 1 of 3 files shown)


mtjandra:
Is the backend already implemented? I originally was thinking that `d2l-quick-eval` still only takes in one href, then it later follows links that splits into `activitesHref` and `submissionHref` for `d2l-quick-eval-activities-list` and `d2l-quick-eval-submissions` respectivley

CarolBusman:
@mtjandra Yes, the href attribute is being assigned [here](http://search.dev.d2l/source/xref/Lms/le/evaluationhub/D2L.LE.EvaluationHub/Web/Desktop/Views/EvaluationHubView.Renderer.cs#72). Unfortunately, the links for the activities and submissions are both directly on `whoami`. There's a mini-discussion [here](https://d2l.slack.com/archives/CK602V4HG/p1560872224013700) where we discussed how we want to get both links into the component.

m6jones:
This component takes in multiple `href`
If having multiple, I think both need names other than `href` and it shouldn't extend the `entity` behaviour.

mtjandra:
> Unfortunately, the links for the activities and submissions are both directly on whoami

Okay sounds good. I think because Maya's suggestion is satisfied, I'm fine with having two hrefs. Although my preference would be one href. Whoami doesn't seem like a good fit, but something like

* whoami
    * my-unassessed-activities //Then use this link
        * my-unassessed-activities-list
        * my-unassessed-activity-collections

m6jones:
https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-behavior.js I forgot to add the link :/

PR #237, opened 2019-07-05
Separating unread items into its own component.

CarolBusman:
Looking ahead at the responsive layouts, it will be easier if this section and its styling is in its own component, since it moves around and changes.

Also updated the action buttons with the new icons.

There should be no visual change to the # of unread section:

image

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
@@ -1,0 +1,41 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import {QuickEvalLocalize} from '../QuickEvalLocalize.js';
+import 'd2l-tooltip/d2l-tooltip.js';
+
+class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(PolymerElement) {
+	static get is() { return 'd2l-quick-eval-activity-card-unread-submissions'; }
+	static get template() {
+		return html`
+			<style>
+				#d2l-quick-eval-activity-card-submissions {
+					text-align: center;
+					align-items: center;
+				}
+			</style>
+			<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText(unread, resubmitted)]]</div>
+			<d2l-tooltip for="d2l-quick-eval-activity-card-submissions" position="bottom">[[_getSubmissionTooltipText(unread, resubmitted)]]</d2l-tooltip>
+		`;
+	}
+	static get properties() {
+		return {
+			unread: {
+				type: Number,
+				value: 0
+			},
+			resubmitted: {
+				type: Number,
+				value: 0
+			}
+		};
+	}
+
+	_getNewSubmissionsText() {
+		return this.localize('unreadSubmissions', 'num', this.unread + this.resubmitted);
+	}
+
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -157,12 +154,8 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
 	}
 
-	_getNewSubmissionsText() {
-		return this.localize('unreadSubmissions', 'num', this.unread + this.resubmitted);
-	}
-
-	_getSubmissionTooltipText() {
-		return this.localize('unreadSubmissionsDetail', 'unread', this.unread, 'resub', this.resubmitted);
+	_showUnreadSubmissions() {
+		return this.unread || this.resubmitted;
```
mtjandra: I think you want `(this.unread > 0) || (this.resubmitted > 0)`. We'll want to set `hidden` to a boolean. For example `return -1 || 0` would return `-1`, which seems to be truthy

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -0,0 +1,43 @@
...
+	static get template() {
+		return html`
+			<style>
+				#d2l-quick-eval-activity-card-submissions {
+					text-align: center;
+					align-items: center;
+					display: flex;
+					justify-content: space-around;
+				}
+			</style>
```
mtjandra: Does the styling also need to include

```
[hidden] {
    display: none;
}
```

Like in d2l-quick-eval-activity-card.js  ? I'm not sure

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
CarolBusman: If we're getting -1 as a value for either of these, I think we have a much bigger problem :)
I'll update.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
CarolBusman: We don't use hidden in the markup inside this component, so we shouldn't need it in the styling at this level either.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -0,0 +1,43 @@
...
+				value: 0
+			},
+			resubmitted: {
+				type: Number,
+				value: 0
+			}
+		};
+	}
+
+	_getNewSubmissionsText() {
```
mtjandra: Since you are passing variables in, why not use them instead of referring to `this`?

CarolBusman: Habit and personal style I think mostly. Is there an argument for one style over the other?

mtjandra: One is functional, meaning I can always expect the same output based on inputs. The other is based on side-effects, so is dependent on state, calling the method each time can produce different outputs. Also just worried if polymer does some odd things and we get something unexpected

PR #239, opened 2019-07-08
Vy/qe sort telemetry

vix633:
Adding sort telemetry


components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+ * Behavior for sending telemetry messages to the telemetry service
+ * @polymerBehavior
+ */
+D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
+	logSortEvent: function(telemetryData) {
+		if (!telemetryData || !telemetryData.telemetryEndpoint) {
+			return;
+		}
+
+		const client = new window.d2lTelemetryBrowserClient.Client({ endpoint: telemetryData.telemetryEndpoint });
```
jeroach: nit: could move this closer to it's usage - as far as I can tell it isn't needed to build out the event

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+window.D2L.PolymerBehaviors.QuickEval = window.D2L.PolymerBehaviors.QuickEval || {};
+
+/**
+ * Behavior for sending telemetry messages to the telemetry service
+ * @polymerBehavior
+ */
+D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
+	logSortEvent: function(telemetryData) {
+		if (!telemetryData || !telemetryData.telemetryEndpoint) {
+			return;
```
jeroach: Do we want to error/log in this case? Should this ever happen?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -365,7 +378,8 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 	_handleSortRequested(evt) {
 		let result;
 		const headerId = evt.detail.headerId;
-
+		this._telemetryData.columnName = headerId;
+		this._telemetryData.telemetryEndpoint = this.dataTelemetryEndpoint;
```
mtjandra: I don't think I'd ever expect the endpoint to change. And it seems odd to keep track of the same piece of data in two places. There could potentially be a synchronization error in the future. Also the `_telemetryData` doesn't seem to need to be a property just a local variable should suffice.

I would recommend doing something like 

```
const telemetryData = {};
telemetryData.columnName  = headerId;

...

telemetryData.sortDirection = descending ? 'desc' : 'asc';

...

this.logSortEvent(this.dataTelemetryEndpoint, telemetryData);
```

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+window.D2L = window.D2L || {};
+window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+window.D2L.PolymerBehaviors.QuickEval = window.D2L.PolymerBehaviors.QuickEval || {};
+
+/**
+ * Behavior for sending telemetry messages to the telemetry service
+ * @polymerBehavior
+ */
+D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
+	logSortEvent: function(telemetryData) {
```
mtjandra: May want to write some tests for this guy

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -375,13 +389,16 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
 					result = this._applySortAndFetchData(header.sortClass, descending, customParams);
+					this._telemetryData.sortDirection = descending ? 'desc' : 'asc';
 				}
 				else {
 					this.set(`_headerColumns.${i}.headers.${j}.sorted`, false);
 				}
 			});
 		});
 
+		this.logSortEvent(this._telemetryData);
```
jeroach: Should this be in the `if(result)`? I suppose it depends if we want to fire if and only if the sort is successful

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -71,6 +71,9 @@ class D2LQuickEval extends PolymerElement {
 			loggingEndpoint: {
 				type: String
 			},
+			dataTelemetryEndpoint: {
```
jeroach: Is there another PR where this is set for the actual usage? Or is that still to come?

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!telemetryData || !telemetryData.telemetryEndpoint) {
+			return;
+		}
+
+		const client = new window.d2lTelemetryBrowserClient.Client({ endpoint: telemetryData.telemetryEndpoint });
+		const eventBody = new window.d2lTelemetryBrowserClient.EventBody();
+
+		eventBody.setAction('Sort')
+			.addCustom('columnName', telemetryData.columnName || 'unknown')
+			.addCustom('sortDirection', telemetryData.sortDirection || 'unknown');
```
mtjandra: nit: I think I would actually just do

```
eventBody.setAction('Sort')
			.addCustom('columnName', telemetryData.columnName )
			.addCustom('sortDirection', telemetryData.sortDirection ');
```

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: This just means telemetry is off

components/d2l-quick-eval/d2l-quick-eval-submissions.js
vix633: I'm leaning towards firing this anyway. I think we want to fire this on user actions, independent of their success.

jeroach: Will the underlying data indicate a status code for success/fail/reason for failure?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
vix633: Are you suggesting taking the telemetry endpoint as a second param for the log method?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
vix633: You'll see stuff in the network tab if you're looking for it. This is specifically to log user behaviour, though. Not application behaviour.

components/d2l-quick-eval/d2l-quick-eval.js
vix633: There isn't, I kind of thought a prop needed to exist for the binding to work. Happy to remove if that's not the case.

jeroach: It's needed :) Just wanted to see the big picture - sounds like it's still to come.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
jeroach: Sounds good.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
mtjandra: Yeah I would recommend that rather than maintaining the endpoint in two places

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: So, i talked to rubrics guys, and they don't generally hook up telemetry all the way up to Athena for testing. Network tab is good enough to see the payload being sent. So, you'll have to wait for quad to see the data

vix633: Unless you want me to share a captured payload for you

mtjandra: I was just curious on how things are looking when we consume these events. Waiting for quad works for me

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -364,7 +369,9 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 
 	_handleSortRequested(evt) {
 		let result;
+		let telemetryData;
```
AlexBedley: Did you mean to initialize this to {}? Also, it should be `const`

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+window.D2L = window.D2L || {};
+window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+window.D2L.PolymerBehaviors.QuickEval = window.D2L.PolymerBehaviors.QuickEval || {};
+
+/**
+ * Behavior for sending telemetry messages to the telemetry service
+ * @polymerBehavior
+ */
+D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
+	logSortEvent: function(telemetryData, telemetryEndpoint) {
```
AlexBedley: I would make `telemetryEndpoint` a property of this behaviour, unless we're planning on sending different events to different endpoints

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!telemetryData || !telemetryEndpoint) {
+			return;
+		}
+
+		const client = new window.d2lTelemetryBrowserClient.Client({ endpoint: telemetryEndpoint });
+		const eventBody = new window.d2lTelemetryBrowserClient.EventBody();
+
+		eventBody.setAction('Sort')
+			.addCustom('columnName', telemetryData.columnName || 'unknown')
+			.addCustom('sortDirection', telemetryData.sortDirection || 'unknown');
```
AlexBedley: nit: A cleaner pattern to denote `sortDirection` would be to just have a boolean property called `descending`. We use this pattern throughout the sorting framework and the HM API.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
AlexBedley: nit: Not sure what our future plans are, but I'd separate this function into two: something like `logEvent` which is more general purpose, and then _maybe_ `logSortEvent` if the boilerplate is really that onerous on the consumers.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
AlexBedley: What's the purpose of the `unknown` value?

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+
+		const event = new window.d2lTelemetryBrowserClient.TelemetryEvent()
+			.setDate(new Date())
+			.setType('TelemetryEvent')
+			.setSourceId('quick-eval')
+			.setBody(eventBody);
+
+		client.logUserEvent(event);
```
AlexBedley: We should `return` this value just for testing purposes

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -0,0 +1,36 @@
...
+		const client = new window.d2lTelemetryBrowserClient.Client({ endpoint: telemetryEndpoint });
+		const eventBody = new window.d2lTelemetryBrowserClient.EventBody();
+
+		eventBody.setAction('Sort')
+			.addCustom('columnName', telemetryData.columnName || 'unknown')
+			.addCustom('sortDirection', telemetryData.sortDirection || 'unknown');
+
+		const event = new window.d2lTelemetryBrowserClient.TelemetryEvent()
+			.setDate(new Date())
+			.setType('TelemetryEvent')
```
AlexBedley: Out of curiousity, what other types are there?

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
mtjandra: I actually don't mind not returning. Wouldn't want to have someone accidentally depend on the telemetry succeeding/failing

AlexBedley: I mean I'm all for not giving people a rope to hang themselves with but in this case I think it's pretty harmless. Plus it would be annoying to test without it.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: Yeah, this should be done when we add new events. its really easy to tease apart. Right now i want to fix tests and get this merged.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: I preferred to set this to unknown rather than undefined, or null. If those are to get set somehow.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: When there are tests for new events, we can tease this apart. Also, Ii don't intend to test to ensure that telemetry client properly forms events here.

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
vix633: There is also performance event.

vix633:
Hey guys, I'd like to merge as is. Are there any glaring things i should change?

vix633:
I've actually identified another good place to write some tests, but don't want to block this change on the tests, since it fails gracefully. I'd like to add some tests in another PR along with a small refactor.

What I want to do is split out the log event logic from the logsortevent method, to make a generic log method that paves the way for the next user events. I'd also like to write tests to ensure that we get a failure of the telemetryClient api changes.

PR #244, opened 2019-07-10
Cbusman/responsive layout for qe cards

CarolBusman:
What is included in this PR: 
- relative positioning and sizing of elements
- changes to relative positioning and sizing based on screen width/media queries

What is not included:
- formatting of elements: font size, internal positioning/layout (e.g. buttons), etc.
- responsive positioning between activity cards (next step).

Design: https://files.slack.com/files-pri/T025EV233-FJDFXLJ02/activity_card___ui_specs___draft_1.png

Screenshots will be added in comments.

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
@@ -11,20 +11,44 @@ class D2LQuickEvalActivityCardItems extends mixinBehaviors([D2L.PolymerBehaviors
 				.d2l-quick-eval-activity-card-items-container {
 					display: flex;
 					align-items: stretch;
+					justify-content: space-between;
 				}
 				::slotted(*) {
-					border-width: 0 1px 0 0;
-					border-style: solid;
-					border-color: var(--d2l-color-tungsten);
 					width: 7.5rem;
 					height: 3rem;
 					text-align: center;
 					display: flex;
 					align-items: center;
 					justify-content: space-around;
+					flex: 1 1 0;
 				}
-				::slotted(*:first-child) {
-					border-left-width: 1px;
+				@media (min-width: 525px) {
+					::slotted(*) {
+						border-width: 0 1px 0 0;
+						border-style: solid;
+						border-color: var(--d2l-color-tungsten);
+					}
+					::slotted(*:last-child) {
+						border-right-width: 0;
+					}
+					:host([small]) ::slotted(*) {
+						border: none;
+					}
+				}
+				@media (min-width: 900px) {
+					:host([small]) ::slotted(*) {
+						border-width: 0 1px 0 0;
+						border-style: solid;
+						border-color: var(--d2l-color-tungsten);
+					}
+					:host([small]) ::slotted(*:first-child),
+					::slotted(*:first-child) {
+						border-left-width: 1px;
+					}
+					:host([small]) ::slotted(*:last-child),
+					::slotted(*:last-child) {
+						border-right-width: 1px;
+					}
 				}
 			</style>
 			<div class="d2l-quick-eval-activity-card-items-container">
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -15,7 +15,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 		return html`
 			<style include="d2l-visible-on-ancestor-styles">
 				d2l-activity-name {
-					height: .9rem;
+					min-height: .9rem;
 				}
 				.d2l-quick-eval-card {
 					padding-bottom: .9rem;
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -13,36 +13,150 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 	static get template() {
 		return html`
 			<style include="d2l-visible-on-ancestor-styles">
-				.d2l-quick-eval-card {
-					border: 3px solid var(--d2l-color-ferrite);
-					border-radius: 10px;
-					padding: 2.2rem;
-					display: flex;
-					justify-content: space-between;
+				d2l-activity-name {
+					height: .9rem;
 				}
-				.d2l-quick-eval-card:hover,
-				.d2l-quick-eval-card:focus-within {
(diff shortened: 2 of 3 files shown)


CarolBusman:
The red box shows where the activity name is positioned.
900px+:
image

525-899px:
image

524px-:
image

jeroach:
In all of the views except the full size, the design does not indicate the type of activity. E.g. should the text highlighted in yellow here be removed? (as well as the other smaller screen widths)

Or is just a placeholder for the icon? I was thinking the icon was part of the activity name component.
60975511-205cff00-a2fa-11e9-92fe-3bb63c821ebe

CarolBusman:
@jeroach The "subtitle" (activity type and due date) is going to be separated out into its own sub-component, where it will handle the showing (and not) of the activity type/due date. So all the text is currently showing all the time, but that is on the list for future responsive changes.

jeroach:
Awesome, thanks @CarolBusman

CarolBusman:
@AlexBedley The buttons were showing up for me in the small modes: is that in FF again?
The text definitely shouldn't be overlapping. I'll have to check to see if the activity name is already set up to truncate with ... if the text is too long. If it is, then there'll be some css I'll need to set at this level that should be included in this PR. If not, then that will be a separate PR for adding that to the activity name.

jeroach:
Once the merge conflicts are resolved, I'll check this out and confirm locally

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
```diff
@@ -11,20 +11,44 @@ class D2LQuickEvalActivityCardItems extends mixinBehaviors([D2L.PolymerBehaviors
...
 					align-items: stretch;
+					justify-content: space-between;
 				}
 				::slotted(*) {
-					border-width: 0 1px 0 0;
-					border-style: solid;
-					border-color: var(--d2l-color-tungsten);
 					width: 7.5rem;
 					height: 3rem;
 					text-align: center;
```
maboelsoudd2l: nit: should we can generalize the design of this? `::slotted(*)` would not make assumptions about the content (where `text-align: center;` assumes that the content is text) and if you want styling specific to the content, it would be put where the slot is used (which you did in `d2l-quick-eval-activity-card.js`). So that now that we changed the content to a meter, it is no longer applicable or might interfere with the meter component (it doesn't affect the meter because the text is supposed to be center-aligned anyways).

maboelsoudd2l:
So I was playing around with the activities page in the browser too, and I think a found the same text-overlap problem that @AlexBedley pointed out.
Screen Shot 2019-07-17 at 1 47 53 PM
I felt that the subtitle is too close to the activity-name component. When I went to investigate, I think I found the cause of the problem. I think this css:
```
d2l-activity-name {
    height: .9rem;
}
```
Might be the culprit, it constricts the size of the activity, smaller than it really is so it overflows. When i disable it, it makes size to auto and fixes the problem. Not sure if this is the best solution, but just sharing what I found.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
CarolBusman: This is set as a default for if the content is not set. I believe it should be able to be overridden by anything placed in the slot, should its `text-align` need to be different.

maboelsoudd2l: fair enough

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -151,6 +286,12 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 		};
 	}
 
+	ready() {
+		super.ready();
+		window.addEventListener('resize', this._onWindowResize.bind(this));
```
dbatiste: Looks like this never gets removed.  Usually I would update the function here (`this._onWindowResize = this._onWindowResize.bind(this)`), but delay event wire-up until the element is connected to the DOM, and of course remove the handler when disconnected.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -177,6 +318,17 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
 	}
 
+	_onWindowResize() {
+		const width = Math.max(document.documentElement.clientWidth, window.innerWidth || 0);
+		const actions = this.shadowRoot.querySelector('.d2l-quick-eval-card-actions d2l-quick-eval-activity-card-items');
+		if (width >= 900) {
+			actions.setAttribute('visible-on-ancestor', '');
+		} else {
+			actions.removeAttribute('visible-on-ancestor');
+			actions.removeAttribute('d2l-visible-on-ancestor-hide');
```
dbatiste: Note: this attribute is not really part of the public API for the behavior - it is maintained internally by the behavior and subject to change.  Removing the `visible-on-ancestor` attribute should be enough, and the behavior should be smart enough to remove `d2l-visible-on-ancestor-hide`.  For instance, the new mixin version (for use with Lit elements) of this behavior, [core/mixins/visible-on-ancestor-mixin.js](https://github.com/BrightspaceUI/core/blob/master/mixins/visible-on-ancestor-mixin.js) does not have this attribute - it still uses an attribute for CSS selector, but it is more clearly named as private.   If we need a fix, we should check and fix it too if necessary.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -177,6 +318,17 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 		return (unread > 0) || (resubmitted > 0);
 	}
 
+	_onWindowResize() {
+		const width = Math.max(document.documentElement.clientWidth, window.innerWidth || 0);
+		const actions = this.shadowRoot.querySelector('.d2l-quick-eval-card-actions d2l-quick-eval-activity-card-items');
+		if (width >= 900) {
```
dbatiste: What is the rationale for disabling the behavior when the view-port is less than 900?  The reason I ask is that this is inconsistent with other uses.  Perhaps ok, but the behavior already has [a way to deal with lack of hover on touch devices](https://github.com/BrightspaceUI/core/blob/master/mixins/visible-on-ancestor-mixin.js#L16).  So, do we want this to be different from the existing usage?  

If there are a lot of cards, then there will be a lot of resize event handlers (making measurements) which can result in janky resize, especially since `resize` is typically dispatched many times.  A media-query (by the behavior, if a consistent break-point could be determined), or resize-observer would be more performant since they avoid measurements (i.e. `documentElement.clientWidth` and `window.innerWidth`).

PR #248, opened 2019-07-11
US107861 - Use and save toggle state

mtjandra:
Since toggle state isn't part of HM we will need to pass down relative url to update the config, as well as its current state

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -1,6 +1,7 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {QuickEvalLocalize} from './QuickEvalLocalize.js';
 import 'd2l-colors/d2l-colors.js';
+import 'd2l-fetch/d2l-fetch.js';
 
 /**
  * @customElement
@@ -87,7 +88,10 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 		return {
 			currentSelected: {
 				type: String,
-				observer: '_dispatchSelectionChanged'
+				observer: '_handleSelectionChange'
+			},
+			toggleHref: {
+				type: String
 			},
 			_viewTypes: {
 				type: Object,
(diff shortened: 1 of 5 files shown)


components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -111,7 +126,29 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 	_isSelected(view) {
 		return this.currentSelected === view;
 	}
-	_dispatchSelectionChanged(view) {
+	_handleSelectionChange(view, previousView) {
+
+		if (
+			!!this.toggleHref
+			&& previousView !== null
+			&& previousView !== undefined
```
mtjandra: When rendering QE for the first time, this method is triggered. Starts off as undefined. No need to update config until user actually toggles

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -87,7 +88,10 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 		return {
 			currentSelected: {
 				type: String,
-				observer: '_dispatchSelectionChanged'
+				observer: '_handleSelectionChange'
```
AlexBedley: Could you not just set a `default` to `_viewType.submissions`? Or you could put the default one level up in quick-eval.

https://polymer-library.polymer-project.org/3.0/docs/devguide/properties#configure-values

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -111,7 +126,29 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 	_isSelected(view) {
 		return this.currentSelected === view;
 	}
-	_dispatchSelectionChanged(view) {
+	_handleSelectionChange(view, previousView) {
+
+		if (
+			!!this.toggleHref
```
AlexBedley: nit: what specific cases are you worried about here? Can we omit the `!!`?

mtjandra: Yeah I can omit

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
mtjandra: Yeah I can set a default. But I will still keep the logic in the ready method just in case something unexpected comes from the backend

mtjandra: I opted out of using default. I couldn't really reference the constant in the properties within properties itself. Also since I'm doing sanitation in ready method, there was really no point.

mtjandra:
> BTW, it's hard to verify your fixes when you squash the commits. I mean I know the thing I asked for was fixed, but not what else was changed unless I do the review over again.

Ah sorry. Usually clicking the `force-pushed` link will show a nice diff. But because I rebased as well it shows a lot more than that. I will keep this in mind for next time

AlexBedley:
> Ah sorry. Usually clicking the `force-pushed` link will show a nice diff. But because I rebased as well it shows a lot more than that. I will keep this in mind for next time

no worries! :)

PR #251, opened 2019-07-15
Abedley/us107658 search

maboelsoudd2l:
implemented search in activities using `d2l-hm-search-behavior`.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -20,7 +20,11 @@ import './activities-list/d2l-quick-eval-activities-list.js';
  */
 
 class D2LQuickEvalActivities extends mixinBehaviors(
-	[D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior, D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviour],
+	[
+		D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehavior,
+		D2L.PolymerBehaviors.QuickEval.D2LHMFilterBehaviour,
+		D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviour
+	],
 	QuickEvalLogging(QuickEvalLocalize(PolymerElement))
 ) {
 	static get template() {
@@ -81,9 +85,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 					search-action="[[searchAction]]"
 					placeholder="[[localize('search')]]"
 					aria-label$="[[localize('search')]]"
-					on-d2l-hm-search-results-loading="_clearSearchError"
-					on-d2l-hm-search-results-loaded="_activitiesSearchLoaded"
-					on-d2l-hm-search-error="_errorOnSearch">
+					>
 				</d2l-hm-search>
 			</div>
 			<div class="clear"></div>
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -191,8 +194,20 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+		this.filterError = null;
+	}
+
+	_clearSearchResults() {
+		const search = this.shadowRoot.querySelector('d2l-hm-search');
+		search.clearSearch();
+	}
+
+	_updateSearchResultsCount(courses) {
+		this.searchResultsCount = courses.reduce(
```
jeroach: Where is the `this.searchResultsCount` defined? Since it was removed from the search behavior

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -191,8 +194,20 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		}
 	}
 
-	_loadSearch(entity) {
-		this._setSearchAction(entity);
+	// @override - do not change the name
+	_clearErrors() {
+		this.searchError = false;
```
jeroach: What is the reason behind the `searchError` being a bool VS the `filterError` being an object? Seems inconsistent

test/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -23,82 +23,78 @@ import SirenParse from 'siren-parser';
 				]
 			};
 
-			assert.isNotOk(searchBehavior.searchAction);
```
jeroach: I think we should keep the pre-test sets and asserts for the various cases.

Otherwise the test isn't proving anything, since the value could have always been the value were asserting for.

By explicitly setting a value we don't expect before we execute our method under test, we can be sure that it's doing what we want it to do :)

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: good catch

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: Good point, I'll change it to an object.

test/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
maboelsoudd2l: Isn't that what the `'default state is correct'` test case for? It checks once for all the initial states of our variables, so that we don't have to do this again.

jeroach: You're totally right! I didn't think about it that way. Good call.

test/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -23,82 +23,78 @@ import SirenParse from 'siren-parser';
...
-					results: null,
-					searchIsCleared: false
-				}
-			};
 
-			searchBehavior._searchResultsLoaded(e);
-			assert.isTrue(searchBehavior.searchApplied);
-			assert.equal(0, searchBehavior.searchResultsCount);
-			assert.isFalse(searchBehavior.searchError);
+		test('_clearErrors sets searchError to false', () => {
```
jeroach: nit: names of tests should reflect change to `null`

PR #255, opened 2019-07-17
robust switcher rtl styling and tabbing support

maboelsoudd2l:
Fixes some of the issues noted in testing.

this is what it looks like in in RTL:
Screen Shot 2019-07-17 at 6 44 02 PM
and this is what it looks like when the submissions is tabbed (You can barely see a difference):
Screen Shot 2019-07-17 at 6 45 14 PM 1

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -58,11 +58,8 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 					-moz-user-select: none;
 					-ms-user-select: none;
 				}
-				:host button:hover {
+				:host button:hover, :host button:focus {
 					border: 1px solid var(--d2l-color-celestine) !important;
-				}
-				:host button:focus {
-					background-color: var(--d2l-color-gypsum);
 				}
 				:host button.d2l-quick-eval-view-toggle-left[selected], :host button.d2l-quick-eval-view-toggle-right[selected] {
 					background-color: var(--d2l-color-tungsten);
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -12,22 +12,25 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 		const toggleTemplate = html`
 			<style>
 				:host button.d2l-quick-eval-view-toggle-left,
-				:host(:dir(rtl)) :host button.d2l-quick-eval-view-toggle-right {
-					border-top-left-radius: 0.3em;
-					border-bottom-left-radius: 0.3em;
+				:host(:dir(rtl)) button.d2l-quick-eval-view-toggle-right {
+					border-top-left-radius: 0.3rem;
+					border-bottom-left-radius: 0.3rem;
 					border-right-color: transparent;
+					border-top-right-radius: 0rem;
+					border-bottom-right-radius: 0;
+					border-left-color: var(--d2l-color-mica);
 				}
 				:host button.d2l-quick-eval-view-toggle-left {
-					margin-left: 0.9rem;
-				}
-				:host(:dir(rtl)) button.d2l-quick-eval-view-toggle-left {
-					margin-right: 0.9rem;
+					margin-inline-start: 0.9rem;
 				}
 				:host button.d2l-quick-eval-view-toggle-right,
-				:host(:dir(rtl)) :host button.d2l-quick-eval-view-toggle-left {
-					border-top-right-radius: 0.3em;
-					border-bottom-right-radius: 0.3em;
+				:host(:dir(rtl)) button.d2l-quick-eval-view-toggle-left {
+					border-top-right-radius: 0.3rem;
+					border-bottom-right-radius: 0.3rem;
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -58,7 +61,10 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 				:host button:hover {
 					border: 1px solid var(--d2l-color-celestine) !important;
 				}
-				:host button[selected] {
+				:host button:focus {
+					background-color: #e6eaf0;
```
CarolBusman: I believe this should be `--d2l-color-gypsum`. http://design.d2l/style/colour/#Change_History

jeroach:
Not strictly in the context of this PR, but is the intention that the toggle button follow the same style guidelines as `Standard` buttons? http://design.d2l/components/buttons/

e.g. in our case we have 2 `Standard` buttons lined up side by side

maboelsoudd2l:
> Not strictly in the context of this PR, but is the intention that the toggle button follow the same style guidelines as `Standard` buttons? http://design.d2l/components/buttons/
> 
> e.g. in our case we have 2 `Standard` buttons lined up side by side. And in the context of this PR, I'd expect to see a blue outline on the tabbed button (when it enters a `focused` state)
> 
> Image of standard button styling:
> button_styles

My thoughts exactly. I was a bit conflicted when I saw this, so i contacted @Surekh last night about whether to do blue outline vs button-dimming. Hopefully once she gives more direction on this, I can make the necessary changes.

svanherk:
design.d2l is not up-to-date with the new WCAG 2.1 styles.  The screenshot above is the old style and isn't compliant.  A lot of the new specs can be found in this powerpoint: https://d2lmail-my.sharepoint.com/:p:/g/personal/sbracewell_desire2learn_com/Efr5ihjeO4xDobT9Yk8hORABGtpHrDKA6wbhmJYiBR-MZQ?rtime=8y5iyY0L10g
I'm not sure when design plans to update the website, some things may still be a bit in flux.

Surekh:
@maboelsoudd2l - I shared design.d2l with you, so that you know where the design specs for the platform are. What Stacey said is true, design.d2l is not up to date, yet. The specs in our user stories are provided by Erin and Dan. That should be always considered to be true and if in doubt check with Erin or Dan.

jeroach:
Thanks @svanherk - looking at that deck, it seems that the `View Switchers` doesn't have a style shown for `focus` - does this mean that no style is applied when the button is focused? or is it that `hover` === `focus`

svanherk:
I believe the monolith version got the updates Design wanted, @margaree did those updates and talked to design about hover vs focus vs selected.  Marg, what was the outcome again?

margaree:
The monolith did get the WCAG design updates, as seen in the slide mentioned (though the background colour in the monolith in the latest 20.19.9 site isn't appearing as `sylvite` right now so I'm going to take a look into that, since I believe it used to be). 
See view switcher in the calendar in https://qa2019717129g.bspc.com/d2l/le/calendar/6606 for an example. The focus behaviour is different from hover and has the `background-color` be gypsum (`#e3e9f1`), which is the background color a button is when focused.

jeroach:
Thank you both for clearing that up! @svanherk @margaree

margaree:
So I have talked more with Tara about this, and the hover and focus state should behave the same (as seen in the slides as "hover").

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
maboelsoudd2l: good catch

AlexBedley:
this branch has rotted a bit. Do you mind merging master in to see how it interacts with the css changes that have been merged in the meantime?

AlexBedley:
Thanks for merging master in.

1. With RTL on, what is the correct tab order? Right now the tab order is the exact same in RTL and LTR. Which feels weird that the "end" of the line is receiving focus before the beginning of the line. In the current layout, I would expect the tab order to be `filter -> search -> toggle`. Right now it's `toggle -> filter -> search`.
2. Kind of related, but should the toggle be the rightmost element? And also should we move the title to the right side? Hacked together screenshot below to show my meaning. This would solve my first concern as well, by making the UI line up with the tab order.
image

dbatiste:
I think for RTL, it's correct... `toggle -> filter -> search`.  Tab order should be the same in LTR and RTL.

AlexBedley:
> I think for RTL, it's correct... `toggle -> filter -> search`. Tab order should be the same in LTR and RTL.

This makes sense to me. But in that case I would suggest we implement my 2nd suggestion

m6jones:
> > I think for RTL, it's correct... `toggle -> filter -> search`. Tab order should be the same in LTR and RTL.
> 
> This makes sense to me. But in that case I'd like us to implement my 2nd suggestion

I agree with your second suggestion.

AlexBedley:
There's a bug somewhere in RTL mode, when you toggle to activities the filter/search elements are no longer left-aligned.

activities:
image

submissions:
image

AlexBedley:
The other thing I noticed is that the padding between View By and the toggle is wrong in Edge

image

maboelsoudd2l:
> The other thing I noticed is that the padding between View By and the toggle is wrong in Edge
> 
> image

Can we just create a story for this? I can't test edge on my computer.

PR #257, opened 2019-07-18
Adding responsive layout to activities list.

CarolBusman:
Adding the responsive css to the activities list - vertical spacing and course name size.

900px+:
image

525-899px:
image

524px-:
image

Also included a change to the activity card demo as the property name changed.

### DIFF
--- a/components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
@@ -16,11 +16,19 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 			.d2l-quick-eval-activities-list-card-spacer {
 				background: var(--d2l-color-sylvite);
 				height: .3rem;
+				width: 100vw;
+				position: relative;
+				left: 50%;
+				right: 50%;
+				margin-left: -50vw;
+				margin-right: -50vw;
+			}
+			.d2l-quick-eval-activities-list-card-spacer-border {
+				border-top: 1px solid var(--d2l-color-mica);
 			}
 			h2 {
 				margin-bottom: .6rem;
 				margin-top: .9rem;
-				margin-inline-start: .9rem;
 				min-height: .6rem;
 				line-height: .6rem;
 				font-size: .8rem;
@@ -30,7 +38,6 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 					font-size: 1rem;
 					margin-top: 1.2rem;
 					margin-bottom: .6rem;
-					margin-inline-start: 0;
 					min-height: 1rem;
 					line-height: 1rem;
 				}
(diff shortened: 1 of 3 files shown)


jeroach:
I noticed something odd with an assignment with a due date - when resizing from full size down, the switcher button overflows the card itself. See attached capture:

We can chat offline if you have questions :)
resize_card_overflow

jeroach:
Another thing - I find the bullet point a bit too close to the type when the page is full size:
full_size_too_close

When the page is in the 525-899px rangem, the spacing between the bullet and the type seems better:
smaller_size_spacing_looks_good

maboelsoudd2l:
@jeroach I think it's for the following reasons:
1. witchcraft
2. changing the parent container to display flex. it makes it look weird
without flex:
Screen Shot 2019-07-18 at 3 28 49 PM
and with `display: flex`:
Screen Shot 2019-07-18 at 3 29 02 PM
Not a conclusive study but trying to help with my best guess :(

CarolBusman:
@jeroach That is odd. Can you verify that it was caused by the changes in this PR? If not, then it should probably be handled in the story for separating out the activity type/due date into its own component since that will change all the formatting and layout for that section all over again. I suspect it was introduced with the on-demand due date display interacting with the responsive card.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -12,6 +13,35 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
...
+				background: var(--d2l-color-sylvite);
+				height: .3rem;
+			}
+			h2 {
+				margin-bottom: .6rem;
+				margin-top: .9rem;
+				margin-inline-start: .9rem;
+				min-height: .6rem;
+				line-height: .6rem;
+				font-size: 16px;
```
dbatiste: Are we sure we want fixed `16px`?  Normally we'd convert this to `0.8rem` so that it scales with user's account setting.  It's a little odd to set the line-height less than the font-size - hopefully there aren't any overflow issues.

Note: both [body-compact](https://github.com/BrightspaceUI/core/blob/master/components/typography/styles.js#L17) and [heading4](https://github.com/BrightspaceUI/core/blob/master/components/typography/styles.js#L87) use the `0.8rem` size.  For heading-4, the `font-weight` is also different.

AlexBedley:
I checked chrome, firefox and edge on windows. Looks good! The inner parts of the card are still broken on ff and edge but that's out of scope for this PR.

The only weirdness I noticed was on the smallest responsive size:

1. I'm sure it follows the spec but the course title looks weirdly small compared to the activity name. The other responsive levels have the course name about equal or bigger than the activity name.

image

2. Once again, I'm sure it follows the spec but the separator bars on the smallest responsive size look wrong.

image

I'm no designer but it feels like the course name should be horizontally centered or something. The different separator shadows/heights look weird as well.

CarolBusman:
@AlexBedley Pretty sure I'm reading the spec right: 6px of grey in between the course name and the first activity, 12 pixels of grey between two activities, and 6 pixels of grey after the last activity and before the next course name, with darker grey 1px borders underneath each activity.
The activity name font size should probably be handled by the activity card, so that would be a separate PR.

AlexBedley:
> @AlexBedley Pretty sure I'm reading the spec right: 6px of grey in between the course name and the first activity, 12 pixels of grey between two activities, and 6 pixels of grey after the last activity and before the next course name, with darker grey 1px borders underneath each activity.

Yeah it looked like it was following the spec, my thought is do we think the spec should change? If its just me then maybe it's just a difference of style. But if we all feel the same way then maybe it's worth bringing to the designers?

> The activity name font size should probably be handled by the activity card, so that would be a separate PR.

Sounds good.

CarolBusman:
Got a LGTM from Daniel after the latest checkin.

PR #259, opened 2019-07-19
Maboelsoudd2l/us108432

maboelsoudd2l:
Screenshot of activities
Screen Shot 2019-07-19 at 6 45 16 PM
Tooltips:
Discussion
Screen Shot 2019-07-19 at 6 46 06 PM
Assignment
Screen Shot 2019-07-19 at 6 46 33 PM
Quiz
Screen Shot 2019-07-19 at 6 47 08 PM

- [ ] designer signoff

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
@@ -14,8 +14,8 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 					justify-content: space-around;
 				}
 			</style>
-			<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText(unread,resubmitted,activityType)]]</div>
-			<d2l-tooltip for="d2l-quick-eval-activity-card-submissions" position="bottom">[[_getSubmissionTooltipText(unread,resubmitted,activityType)]]</d2l-tooltip>
+			<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText(unread, resubmitted, activityType)]]</div>
+			<d2l-tooltip for="d2l-quick-eval-activity-card-submissions" position="bottom">[[_getSubmissionTooltipText(unread, resubmitted, activityType)]]</d2l-tooltip>
 		`;
 	}
 	static get properties() {
@@ -52,11 +52,11 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 		};
 	}
 
-	_getNewSubmissionsText(unread,resubmitted,activityType) {
+	_getNewSubmissionsText(unread, resubmitted, activityType) {
 		return this.localize(this.activityTypeLocalizeNew[activityType], 'num', this.unread + this.resubmitted);
 	}
 
-	_getSubmissionTooltipText(unread,resubmitted,activityType) {
+	_getSubmissionTooltipText(unread, resubmitted, activityType) {
 		return this.localize(this.activityTypeLocalizeDetail[activityType], 'newNum', this.unread, 'resub', this.resubmitted);
 	}
 }
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
@@ -47,7 +47,7 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 			</style>
 			<div id="d2l-quick-eval-activity-card-submissions-container" class="d2l-quick-eval-activity-card-submissions-container">
 				<div class="d2l-quick-eval-activity-card-submissions-number">[[_getNewSubmissionsNumber(unread, resubmitted)]]</div>
-				<div class="d2l-quick-eval-activity-card-submissions-subtitle">[[_getNewSubmissionsSubtitle(unread, resubmitted, activityType)]]</div>
+				<div class="d2l-quick-eval-activity-card-submissions-subtitle">[[_getNewSubmissionsSubtitle(activityType)]]</div>
 			</div>
 			<d2l-tooltip for="d2l-quick-eval-activity-card-submissions-container" position="bottom">[[_getSubmissionTooltipText(unread, resubmitted, activityType)]]</d2l-tooltip>
 		`;
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
@@ -14,8 +14,8 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 					justify-content: space-around;
 				}
 			</style>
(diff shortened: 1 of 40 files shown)


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -14,8 +14,8 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 					justify-content: space-around;
 				}
 			</style>
-			<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText(unread, resubmitted)]]</div>
-			<d2l-tooltip for="d2l-quick-eval-activity-card-submissions" position="bottom">[[_getSubmissionTooltipText(unread, resubmitted)]]</d2l-tooltip>
+			<div id="d2l-quick-eval-activity-card-submissions">[[_getNewSubmissionsText(activityType)]]</div>
```
CarolBusman: We should still take in `unread` and `resubmitted` here so that it triggers the binding to update if those values ever change.

maboelsoudd2l: ok, didn't know that's how bindings worked in polymer.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -27,16 +27,37 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 			resubmitted: {
 				type: Number,
 				value: 0
+			},
+			activityType: {
+				type: String,
+				value: ''
```
AlexBedley: nit: what's the purpose of this default value?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -27,16 +27,37 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
...
+			},
+			activityTypeLocalizeNew: {
+				type: Object,
+				value: {
+					'Assignment': 'newSubmissions',
```
AlexBedley: nit: these json keys don't need to be quoted. Maybe I'll add [this rule](https://eslint.org/docs/rules/quote-props) to our linter if I find the time

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -27,16 +27,37 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
...
+					'Discussion': 'newPosts',
+					'Quiz': 'newAttempts'
+				}
+			},
+			activityTypeLocalizeDetail: {
```
AlexBedley: nit: we don't really want consumers to customize this, does it need to be a property?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
maboelsoudd2l: i thought we shouldn't let values be undefined. But i can remove it, no worries.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
maboelsoudd2l: sure

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
maboelsoudd2l: I agree, we can set it as a constant at the top.

AlexBedley:
On the smallest responsive size some padding is missing between the number and the text.

image

Compare that to the spec here: https://d2lmail.sharepoint.com/sites/uxteam/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fuxteam%2FShared%20Documents%2FAssessment%20%2D%20Quick%20Eval%20and%20More%2FQuick%20Eval%20%2D%20Design%20Resources%2FSpecs%2FActivity%20Card%20%2D%20specs%20%2D%20responsive%2Epng&parent=%2Fsites%2Fuxteam%2FShared%20Documents%2FAssessment%20%2D%20Quick%20Eval%20and%20More%2FQuick%20Eval%20%2D%20Design%20Resources%2FSpecs&p=true&cid=d61719b2-7990-452d-921d-9301d0d92caa

AlexBedley:
In RTL mode, the submissions should probably appear on the other side of the buttons. Not sure if things you did in your other PR will fix this or if more work needs to be done.

image

jeroach:
When the page width hits `556px` there seems to be an odd overlap issue with "new submissions"
edge_case

jeroach:
I might have an older version of the design (and I've requested access to the link Alex mentioned), but in the spec I've been using as a reference: https://files.slack.com/files-pri/T025EV233-FJDFXLJ02/activity_card___ui_specs___draft_1.png

the lang terms for "new submissions", "new attempts", and "new posts" are capitalized - though the terms are different, so it might just be an outdated spec

jeroach:
I think the tooltip lang term is slightly off when the page is rtl:

tooltip_in_rtl

maboelsoudd2l:
@jeroach 
So I was able to replicate this behaviour in rtl:

Screen Shot 2019-07-26 at 11 48 59 AM

but I inserted some arabic lang terms and it doesn't seem to show the same problem anymore:
Screen Shot 2019-07-26 at 1 00 29 PM
The tooltip says "25 new, 0 repeat-submissions" which is what is expected.

jeroach:
> The tooltip says "25 new, 0 repeat-submissions" which is what is expected.

Hmmm, okay - no concerns for that one then.

maboelsoudd2l:
> EDIT: checked the new spec, terms are indeed capitalized. Also, the wording looks to be be "ungraded attempts" and "unread submissions" - worth a double check with Dan/Erin to confirm which is right.

Thanks for the heads-up, I chatted with Dan and I was in-fact using an old spec. Will capitalize in the next commits.

maboelsoudd2l:
> When the page width hits `556px` there seems to be an odd overlap issue with "new submissions"

I spoke to Dan about this and investigated this edge-case. This seems to be more of an english issue than arabic. Because the english word "submissions" is long and the arabic "الإرسالات" or "الطلبات" are short (these are my layman translations). Regardless it has to be fixed in english anyways, but after speaking with Dan he asserts that this doesn't seem to be the case in the mockup because the action buttons in the mockup are less wide than ones currently in quad. So he feels that we should look into this once the action buttons are fixed.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -24,6 +24,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 				}
 				.d2l-quick-eval-card-actions {
 					padding-top: .6rem;
+					order: 3;
```
CarolBusman: I don't think we need to set the order in the default styling, since they're in the order we want in the html already. 

Can we set the order on the unread component to `-1` in the 900px media query, and it should move it to before the other two (default is 0)?

maboelsoudd2l: That actually works much better than what I did. Thanks!

PR #260, opened 2019-07-22
Fix translator errors

mtjandra:
* Fixup translations


mtjandra:
> Human mistake by the translator? 

Yeah this is likely. I think often times they do come back ill formed. Not all translators know how to translate ICU formatting properly. I think tests/validation shouldn't be handled by individual projects, but at the translation process level

PR #262, opened 2019-07-25
Cbusman/action button component

CarolBusman:
After talking with design, it looks like we won't be able to take advantage of `d2l-button-subtle` or `d2l-button-icon` as they do not appear customizable in the way we need to match the specs for this component.

Big buttons (black border is for demonstration only, to show where the icon and text align to the button):
image

Small buttons:
image

Got an LGTM from Daniel.


dbatiste:
I'd question whether we really need this variant, and whether this is a one-off or a gap in our existing button designs.

jeroach:
> I'd question whether we really need this variant, and whether this is a one-off or a gap in our existing button designs.

@dbatiste Are you thinking that we would be better served to tweak/add logic to one of `d2l-button-subtle` or `d2l-button-icon` to meet our use case?

CarolBusman:
@dbatiste @jeroach 
`d2l-button-icon` doesn't have the right icon colour (and I couldn't see that we could change that from the outside).
`d2l-button-subtle` assumes that there will be text, and therefore doesn't align the icon properly.
We also want to align the icon to the top of its parent instead of the top of the shaded hover area, which is possible but... awkward with our already existing buttons from my initial analysis.

I understand that we want to use this design on at least one other page that is currently using a temporary button format:
image

svanherk:
Interesting... that looks somewhat like image-based nav's buttons, if they weren't images but instead were icons.  Hover/focus effects are slightly different, they treat the whole thing as a button.
Hover:
image
Focus:
image

dbatiste:
@jeroach : If this is something that design indeed requires, and it's something that's going to come up more than once, then I think incorporating it into the design patterns and our core button components should on the table - whether it be a modification to `d2l-button-subtle`, or a different button component in `core`.  If it's not going to be used elsewhere, then I think we really need to weigh the cost vs value.

@CarolBusman : what's the required color?

@svanherk : I was thinking the same thing when I saw the screenshot.  ;)  With image-based nav, the height and width are uniform to ensure that the "tiles" in the tray are laid out nicely.  The text is a single line with a width chosen to meet most cases, and is truncated with ellipsis using standard browser ellipsis.  In the Activities case, the text is allowed to wrap, though I am not convinced it should.

CarolBusman:
@dbatiste The icon should be `--d2l-color-celestine-minus-1` (#005694) on hover and focus, and the `d2l-button-icon` doesn't change (iirc it stays ferrite in all states).

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -0,0 +1,118 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import 'd2l-icons/d2l-icon.js';
+import 'd2l-icons/tier3-icons.js';
+import 'd2l-icons/tier2-icons.js';
+
+class D2LQuickEvalActivityCardActionButton extends PolymerElement {
+	static get is() { return 'd2l-quick-eval-activity-card-action-button'; }
```
AlexBedley: As we learned this morning, this is no longer necessary

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -0,0 +1,118 @@
...
+				:host {
+					--d2l-quick-eval-card-button-icon-size: 1.2rem;
+					--d2l-quick-eval-card-button-icon-background-size: 1.5rem;
+					--d2l-quick-eval-card-button-font-size: .6rem;
+					--d2l-quick-eval-card-button-icon-padding: calc(calc(var(--d2l-quick-eval-card-button-icon-background-size) - var(--d2l-quick-eval-card-button-icon-size)) / 2);
+					--d2l-quick-eval-card-button-icon-hover: 0 0 0 var(--d2l-quick-eval-card-button-icon-padding) var(--d2l-color-gypsum);
+					--d2l-quick-eval-card-button-icon-focus-inner: 0 0 0 calc(var(--d2l-quick-eval-card-button-icon-padding) + 2px) white;
+					--d2l-quick-eval-card-button-icon-focus-outer: 0 0 0 calc(var(--d2l-quick-eval-card-button-icon-padding) + 4px) var(--d2l-color-celestine);
+				}
+				:host button {
```
AlexBedley: Also learned from this morning, `:host` is redundant

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -0,0 +1,118 @@
...
+					}
+					:host {
+						--d2l-quick-eval-card-button-icon-size: 1.5rem;
+						--d2l-quick-eval-card-button-icon-background-size: 2.1rem;
+						--d2l-quick-eval-card-button-font-size: .7rem;
+					}
+				}
+			</style>
+			<button 
+				aria-labelledby$="d2l-quick-eval-activity-card-button-text"
```
AlexBedley: Do these need to be $ attributes?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -0,0 +1,118 @@
...
+				formnovalidate$="[[formnovalidate]]"
+				formtarget$="[[formtarget]]"
+				name$="[[name]]"
+				title$="[[text]]"
+				type$="[[type]]">
+				<div class="d2l-quick-eval-activity-card-button-icon">
+					<d2l-icon icon="[[_computeIcon(3)]]" class="d2l-quick-eval-card-button-icon-large"></d2l-icon>
+					<d2l-icon icon="[[_computeIcon(2)]]" class="d2l-quick-eval-card-button-icon-small"></d2l-icon>
+				</div>
+				<span id="d2l-quick-eval-activity-card-button-text">[[text]]</span>
```
AlexBedley: Since there will be many of these on the page, should we be using the unique id service thing that Dave mentioned this morning? For edge support

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
CarolBusman: I believe the `aria-` attributes are in the category of those that do need `$`, since they are attributes and not properties.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -0,0 +1,126 @@
...
+					}
+				}
+			</style>
+			<button 
+				aria-labelledby$="[[_labelledbyId]]"
+				aria-expanded$="[[ariaExpanded]]"
+				aria-haspopup$="[[ariaHaspopup]]"
+				aria-label$="[[_ariaLabel]]"
+				form$="[[form]]"
+				formaction$="[[formaction]]"
```
maboelsoudd2l: Why do we need `form$` buttons?

CarolBusman:
@jeroach Yup, but that was the easiest way to set up to show the different sizes. The small and large are dependent on the media queries/screen width, so the small one's not going to look good when you're >900, and the large one's going to not look right when youre <900px.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
CarolBusman: These attributes were taken from the `d2l-button-subtle`, to be sure we covered everything. I don't believe it's needed for quick eval but it's useful if the component is ever made shared.

jeroach:
> @jeroach Yup, but that was the easiest way to set up to show the different sizes. The small and large are dependent on the media queries/screen width, so the small one's not going to look good when you're >900, and the large one's going to not look right when youre <900px.

Ahhh, thanks for clearing that up :)

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
```diff
@@ -0,0 +1,62 @@
...
+				</style>
+			`;
+			document.body.appendChild($_documentContainer.content);
+		</script>
+	</head>
+	<body class="d2l-typography">
+		<style>
+			.button-wrapper-big {
+				border: 1px solid black;
+				width: 150px;
```
maboelsoudd2l: why doesn't the button have it's own width and height? or at least min-height and min width?

maboelsoudd2l:
I just tried to implement the button on quad (with minimal changes):

Screen Shot 2019-07-31 at 10 57 16 AM

Why are we squishing the text to the icon?
Also why doesn't the button have a white background to cover the meters?

CarolBusman:
@maboelsoudd2l The button size is made to be configurable from the outside, so it stretches to fit its container however big that happens to be (as in the demo). The icon snaps to the top middle of the container, and the text snaps to the bottom middle. 
Since we anticipated this being a shared component maybe eventually, I made the size configurable and the background transparent. These variables will need to be set at the card level so that everything lines up.

PR #263, opened 2019-07-26
fixing extra quote

dlockhart:
Not sure if linting either isn't enabled or isn't able to catch this kind of error...


jeroach:
Thanks Dave! We definitely have linting enabled - must be an odd edge case

PR #264, opened 2019-07-29
Add submission list

vix633:


components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -83,6 +83,7 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 				if (e && e.entity && e.entity.properties) {
 					const p = e.entity.properties;
 					const publishAll = this._getAction(e.entity, 'publish-all-feedback');
+					const submissionListHref = this._getRelativeUriProperty(entity);
```
svanherk: Slack is down, so not sure what part you had a question about specifically.  But should this be `e.entity`?  Or do you want to get it from the parent entity?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -298,6 +302,21 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 		this._onWindowResize();
 	}
 
+	_navigateToSubmissionListPath() {
```
jeroach: nit: ideally we keep the naming consistent, e.g. `_dispatchViewSubmissionListEvent`

test/d2l-quick-eval/d2l-quick-eval-activity-card.js
```diff
@@ -9,6 +9,15 @@
 			assert.equal(qeActivityCard.tagName.toLowerCase(), 'd2l-quick-eval-activity-card');
 		});
 
+		test('allsubmissions event fires with submissionListHref', function(done) {
```
jeroach: Can you update this test to use the mock style that I sent you the example of?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -298,6 +302,21 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
+		this.dispatchEvent(
+			new CustomEvent(
+				'd2l-quick-eval-activity-view-submissionlist',
```
jeroach: nit: Also likely want to add a hyphen between `submission` and `list` for consistency (`submission-list`)

test/d2l-quick-eval/d2l-quick-eval-activity-card.js
vix633: Not sure I see the point, we fire the event currently and that tests the workflow really well

jeroach: Currently you're just verifying that the `href` was empty - ideally we'd want to set a "valid looking" url and ensure that it was called with the url we set.

If that can be accomplished without a mock, then go for it :)

vix633: As discussed offline, I'll leave this as is because the mock will only add a non-empty string, not necessarily adding higher coverage since we wouldn't be mocking the siren response.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
jeroach: This still needs renaming

PR #266, opened 2019-07-29
Suggestions for accessibility fixes

maboelsoudd2l:
Just a number of things that I noticed in my accessibility pass that I wanted to try out/fix. Feel free to be as critical as you want.

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -209,6 +209,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 							</d2l-quick-eval-activity-card-items>
 						</div>
 						<d2l-quick-eval-activity-card-unread-submissions
+							tabindex="1"
 							unread="[[unread]]"
 							resubmitted="[[resubmitted]]"
 							activity-type="[[activityType]]"
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -188,7 +188,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 					stroke: var(--d2l-color-tungsten);
 				}
 			</style>
-			<div class="d2l-quick-eval-card d2l-visible-on-ancestor-target">
+			<div class="d2l-quick-eval-card">
 				<div class="d2l-quick-eval-card-titles">
 					<d2l-activity-name href="[[activityNameHref]]" token="[[token]]"></d2l-activity-name>
 					<div class="d2l-quick-eval-card-subtitle"><span>[[localize(activityType)]]</span> <span hidden$="[[!formattedDueDate]]"> &bull; [[localize('due', 'date', formattedDueDate)]]</span></div>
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -188,7 +188,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 					stroke: var(--d2l-color-tungsten);
 				}
 			</style>
-			<div class="d2l-quick-eval-card d2l-visible-on-ancestor-target">
```
CarolBusman: Moving this to the action buttons container means that the buttons will only show up when the speedometer/action button area is moused over instead of anywhere in the card. Is this the behaviour that we want?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -198,22 +198,23 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 						<div class="d2l-quick-eval-card-meters">
 							<d2l-quick-eval-activity-card-items>
 								<div>
-							<d2l-meter-radial value="[[completed]]" max="[[assigned]]" percent$="[[_denominatorOver99(assigned)]]" text="[[localize('completed')]]"></d2l-meter-radial>
+							<d2l-meter-radial tabindex="0" value="[[completed]]" max="[[assigned]]" percent$="[[_denominatorOver99(assigned)]]" text="[[localize('completed')]]"></d2l-meter-radial>
```
CarolBusman: Why do we need to be able to tab to these (and the unread submissions)? Are they not being read by the screen reader?

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -71,16 +71,16 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 			<div>
 				<label id="d2l-quick-eval-view-toggle-label">[[localize('viewBy')]]</label>
 				<button
+					aria-expanded$="[[_isSelected(_viewTypes.submissions, currentSelected)]]"
```
dlockhart: I would use `aria-pressed` here instead of `aria-expanded` -- since this is acting like a toggle.

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -71,16 +71,16 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
...
 					class="d2l-quick-eval-view-toggle-left"
 					on-click="_selectSubmissions"
 					selected$="[[_isSelected(_viewTypes.submissions, currentSelected)]]"
-					aria-labelledby="d2l-quick-eval-view-toggle-label"
```
dlockhart: Good catch! I'm going to mention this particular bug in our review later this morning.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -198,22 +198,23 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
+							<d2l-meter-radial tabindex="0" value="[[evaluated]]" max="[[assigned]]" percent$="[[_denominatorOver99(assigned)]]" text="[[localize('evaluated')]]"></d2l-meter-radial>
 								</div>
 								<div>
-							<d2l-meter-radial value="[[published]]" max="[[assigned]]" percent$="[[_denominatorOver99(assigned)]]" text="[[localize('published')]]"></d2l-meter-radial>
+							<d2l-meter-radial tabindex="0" value="[[published]]" max="[[assigned]]" percent$="[[_denominatorOver99(assigned)]]" text="[[localize('published')]]"></d2l-meter-radial>
 								</div>
 							</d2l-quick-eval-activity-card-items>
 						</div>
 						<d2l-quick-eval-activity-card-unread-submissions
+							tabindex="1"
```
dlockhart: You shouldn't need any of these `tabindex`es. You only need to make things tabbable for keyboard uses if they need to interact with the component -- which for these they don't. We can chat more about this in the review this morning.

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -71,16 +71,16 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 			<div>
 				<label id="d2l-quick-eval-view-toggle-label">[[localize('viewBy')]]</label>
```
dlockhart: I don't think you need the `id` on the label anymore.

maboelsoudd2l: oh yeah

PR #269, opened 2019-07-31
Using new action button in activity card.

CarolBusman:
image

Dropping in the new action buttons into the activity card. Verified that click handlers (`_dispatchViewSubmissionListEvent` and `_dispatchPublishAllEvent`) still get fired on click.

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -25,7 +25,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 				.d2l-quick-eval-card-actions {
 					padding-top: .6rem;
 				}
-				.d2l-quick-eval-card-actions div {
+				.d2l-quick-eval-card-actions d2l-quick-eval-activity-card-action-button {
 					height: 2.1rem;
 					background: white;
 					display: block;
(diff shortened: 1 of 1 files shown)


CarolBusman:
medium size:
image
small size:
image

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -216,19 +213,21 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 						<div class="d2l-quick-eval-card-actions">
 							<d2l-quick-eval-activity-card-items visible-on-ancestor small>
 								<div>
```
maboelsoudd2l: is this `<div>` necessary? Based on Lockhart's suggestions, we should avoid unnecessary DOM nodes.
```

Comment (maboelsoudd2l) [reply components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js]: I remember you said the action buttons take up the size of their container. Instead of wrapping it in a container `div`, maybe there a way we can style in the css the `d2l-quick-eval-activity-card-action-button` directly?
something like:
```
d2l-quick-eval-activity-card-action-button {
width: 150px;
height: 60px;
}
```

CarolBusman: Ah! This was a holdover from using straight buttons as a placeholder. The buttons calculated width differently than other elements with regards to borders, so it wasn't matching up with the speedometers underneath. Looks like the component handles it just fine though.

PR #271, opened 2019-08-01
Adding disabled styling to action button.

CarolBusman:
Disabled button in the middle, looks the same in all three states (default, hovered, focused).

image

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
@@ -48,11 +48,11 @@ class D2LQuickEvalActivityCardActionButton extends mixinBehaviors(
 					align-items: center;
 				}
 				:host(:hover) button,
-				:host(:hover) .d2l-quick-eval-activity-card-button-icon,
+				:host(:hover) d2l-icon,
 				:host(:focus-within) button,
-				:host(:focus-within) .d2l-quick-eval-activity-card-button-icon,
+				:host(:focus-within) d2l-icon,
 				:host(:focus) button,
-				:host(:focus) .d2l-quick-eval-activity-card-button-icon {
+				:host(:focus) d2l-icon {
 					text-decoration: underline;
 					color: var(--d2l-color-celestine-minus-1);
 				}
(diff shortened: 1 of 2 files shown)


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -1,10 +1,15 @@
...
 import 'd2l-icons/d2l-icon.js';
 import 'd2l-icons/tier3-icons.js';
 import 'd2l-icons/tier2-icons.js';
 import 'd2l-polymer-behaviors/d2l-id.js';
+import 'd2l-button/d2l-button-behavior.js';
 
-class D2LQuickEvalActivityCardActionButton extends PolymerElement {
+class D2LQuickEvalActivityCardActionButton extends mixinBehaviors(
+	[
+		D2L.PolymerBehaviors.Button.Behavior
```
AlexBedley: It makes sense to me that our button extends the button behaviour, but what does the button behaviour do?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -1,10 +1,15 @@
...
+	], PolymerElement) {
 	static get template() {
 		return html`
 			<style>
```
maboelsoudd2l: not sure if this is the best place to ask this, but how come the action buttons doesn't change the cursor when hovered over? How will people know that this is a button? most links and buttons have a `cursor: pointer;` in the css. furthermore most disabled buttons might have something like `cursor: not-allowed;` or a 
```
button[disabled] {
  cursor: not-allowed;
}
```

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
CarolBusman: In this case, it makes it so that the click event on the component gets squashed if it's disabled, since consumers would add `onclick` to the component, not the button inside.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
CarolBusman: I don't believe cursors were covered in the design, so they are just the default for button (which is pointer). I think users are expected to know that these are buttons because:
1. Training
2. They change when you hover over them
3. The text associated with the button describes an action.

PR #272, opened 2019-08-01
US108493 - confirm dialog for publish all

jeroach:
[US108493](https://rally1.rallydev.com/#/detail/userstory/320625537956?fdp=true): [UI] Dialog to confirm Publish All

- Using legacy MVC dialog as a temporary measure
- The dialog will NOT work in demo pages
- No real way to write tests for the dialog either

- [x] Get sign off from design

I've attached a screen cap of this working on my local instance via local BSI pointing to my branch:

publish_all_confirm_dialog


jeroach:
I talked to Dan - the only outstanding item is that we want to have different lang terms in the dialog for  assignments and discussions (more details in the story).

I'm working on that now.

Otherwise, he's happy with it :)

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -321,12 +321,18 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 	}
 
 	_dispatchPublishAllEvent() {
+		let confirmMessage = this.localize('publishAllConfirmDialogMessageForDiscussion');
```
CarolBusman: How expensive is `localize` and does it matter in this case? Should this be an if/else only setting `confirmMessage` once?

jeroach: Good point, but this logic is actually going to change given that we only need one lang term now.

Miles is making excellent progress on standardizing the existing dialogs for Assignment and Discussion, so we can just use the standardized term here.

It's on my list for today.

jeroach:
Lang terms updated (both to put the capitalized terms in the `lang` folder files, and to reduce to a single confirmation dialog term:

capitalized_terms


publish_all_confirm_dialog_single_lang_term

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -259,9 +265,29 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+			'',
+			this.localize('yes'),
+			this.localize('no'),
+			this.localize('close'),
+			this._activitiesListId,
+			() => {}
+		);
+
+		confirmEvent.AddListener(
+			Function.prototype.construct = function(result) {
```
maboelsoudd2l: Few questions, though take with a grain of salt since this code is temporary and subject to change:
- Instead of `.bind(this)` why not use the es6 function syntax? `(result)=> {...}`
- Extending native prototypes is bad practice, is there another way to do this? 
- Why are we altering the `Function.prototype.construct` every single time `_publishAll ` is called?

jeroach: Good point - I was not aware of this :)

ES6 seems to work

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -259,9 +265,29 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+			this.localize('yes'),
+			this.localize('no'),
+			this.localize('close'),
+			this._activitiesListId,
+			() => {}
+		);
+
+		confirmEvent.AddListener(
+			(result) => {
+				if (result && evt.detail.publishAll) {
```
AlexBedley: Should we even display the confirmation dialog if `evt.detail.publishAll` is falsey?

jeroach: Maybe not, but then we'd want to likely disable the button. Lest the user try to click a button and it doing nothing.

jeroach: Chatted offline - going to not show the dialog for now. In future we should also have the button disabled.

PR #273, opened 2019-08-01
Maboelsoudd2l/skeleton activity

maboelsoudd2l:
This is what the skeleton looks like:

Screen Shot 2019-08-01 at 2 04 10 PM
Screen Shot 2019-08-01 at 2 03 55 PM
Screen Shot 2019-08-01 at 2 03 38 PM


components/d2l-quick-eval/d2l-quick-eval-activities-skeleton.js
```diff
@@ -0,0 +1,119 @@
...
+				.large, .medium, .small {
+					width: 100%;
+				}
+				.medium {
+					display: none;
+				}
+				.small {
+					display: none;
+				}
+				@media (max-width: 900px) {
```
CarolBusman: Can we switch these queries around so that `small` is shown by default, then the `medium` and `large` are shown based on `min-width`?
Mobile first css isn't too big of a deal in this specific situation because the css is minimal, but it would be a good idea to keep things consistent.

maboelsoudd2l: sure

PR #275, opened 2019-08-01
use D2L typography styles

Surekh:
I think it is better and simple to use d2l-typography styles - http://design.d2l/style/typography/.
Confirmed with Erin that heading is h2 with d2l Heading 3 - Lato Bold 20 style.

### DIFF
--- a/components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
+++ b/components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
@@ -2,6 +2,7 @@ import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 import {QuickEvalLocalize} from '../QuickEvalLocalize.js';
 import '../activity-card/d2l-quick-eval-activity-card.js';
 import 'd2l-colors/d2l-colors.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
 
 class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 	static get is() { return 'd2l-quick-eval-activities-list'; }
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -26,21 +27,10 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
...
 				border-top: 1px solid var(--d2l-color-mica);
 			}
-			h2 {
-				margin-bottom: .6rem;
-				margin-top: .9rem;
-				min-height: .6rem;
-				line-height: .6rem;
-				font-size: .8rem;
+			.d2l-quick-eval-activities-course-name-heading {
+				@apply --d2l-heading-3;
```
AlexBedley: Hmm should this be `--d2l-heading-2`? Or should we change the html to `<h2>`? Or are the CSS and the HTML actually disjoint and the name `--d2l-heading-2` is just a suggestion?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -26,21 +27,10 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
...
 			}
 			@media (min-width: 525px) {
-				h2 {
```
AlexBedley: How come this isn't necessary anymore?

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
Surekh: It should be an H2 for A11Y with H3 for styling.

> It is important to note that Heading styles are independent of semantic Heading tag order and meaning in HTML. Most of the time they will match, but they don’t have to.

It is acceptable, for example, to use the Heading 3 style on an <h2> tag, as in this example, where the Heading 3 style is used as the title of a widget which has Heading 2 styles within it. In code, this example may or may not be implemented as an actual <h3> tag, but from a visual hierarchy standpoint the styles make sense.

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
Surekh: I think (but you all can correct me), d2l-typography does that.

PR #276, opened 2019-08-02
Added a wrapper for activity-name

maboelsoudd2l:
Should look exactly the same, but now the screen reader can tab through it, which i was able to do successfully.

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
@@ -141,6 +141,11 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 					.d2l-quick-eval-card-titles {
 						min-height: 3rem;
 					}
+					.d2l-activity-name-wrapper {
+						font-size: 0.95rem;
+						font-weight: 400;
+						margin: 0;
+					}
 					.d2l-quick-eval-card-subtitle {
 						font-size: .7rem;
 						line-height: .7rem;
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -187,7 +192,9 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 			</style>
 			<div class="d2l-quick-eval-card d2l-visible-on-ancestor-target">
 				<div class="d2l-quick-eval-card-titles">
-					<d2l-activity-name href="[[activityNameHref]]" token="[[token]]"></d2l-activity-name>
+					<h3 class="d2l-activity-name-wrapper">
```
AlexBedley: Nice, I'm glad this works

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -141,6 +141,11 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 					.d2l-quick-eval-card-titles {
 						min-height: 3rem;
 					}
+					.d2l-activity-name-wrapper {
+						font-size: 0.95rem;
+						font-weight: 400;
```
AlexBedley: Is there some d2l-typography css class we could extend here instead of rolling our own?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
maboelsoudd2l: On mac, I tested Firefox, safari and chrome. They work well in safari and chrome, but Firefox doesn't have any headers loaded. It's likely a problem with firefox and not a problem with this h3 tag.

With screen-readers i only got to test Voiceover, because NVDA is windows only and JAWS not free and windows only.

Maybe someone would like to test the behaviour on windows?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
maboelsoudd2l: Good point!

maboelsoudd2l:
I just spoke to the designers concerning this. @AlexBedley @jeroach I know we should verify on windows, but the only person who has the screen reader software are Peggy and Carin. I recommend we land it in quad and if it leads to issues, we'll put it in the backlog.

AlexBedley:
> the only people who have the screen reader software are Peggy and Carin

NVDA is free and I have it installed. But how did you get this on the QE page in the LMS? Presumably running it on the demo page wouldn't be the same

maboelsoudd2l:
Update: i tested the behaviour on @Surekh computer using NVDA to confirm it works:
NVDA

PR #277, opened 2019-08-02
US108701 - update count after publish all

AlexBedley:
Problem: After clicking publish all, the activity view displays incorrect information about the evaluation status
Solution: Update the activity view to display current data

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -200,6 +200,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const result = await Promise.all(entity.entities.map(async function(activity) {
 			const evalStatus = await this._getEvaluationStatusPromise(activity);
 			const courseName = await this._getCourseNamePromise(activity);
+			const id = this.getEvaluationStatusHref(activity);
 			return {
 				courseName: courseName,
 				assigned: evalStatus.assigned,
@@ -213,7 +214,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				key: this._getOrgHref(activity),
 				dueDate: this._getActivityDueDate(activity),
 				activityType: this._getActivityType(activity),
-				activityNameHref: this._getActivityNameHref(activity)
+				activityNameHref: this._getActivityNameHref(activity),
+				id: id
 			};
 		}.bind(this)));
 		return this._groupByCourse(result);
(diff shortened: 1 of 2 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -200,6 +200,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const result = await Promise.all(entity.entities.map(async function(activity) {
 			const evalStatus = await this._getEvaluationStatusPromise(activity);
 			const courseName = await this._getCourseNamePromise(activity);
+			const id = this.getEvaluationStatusHref(activity);
```
AlexBedley: I needed a way to link a publish-all event to the `_data` we have stored. I decided to create an `id` field, using the evaluation status rel href since those are unique. I wanted something that would not require another http call.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -299,5 +305,23 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+			for (let j = 0; j < this._data[i].activities.length; j++) {
+
+				if (this._data[i].activities[j].id === id) {
+					this.set(`_data.${i}.activities.${j}.assigned`, evalStatus.assigned);
+					this.set(`_data.${i}.activities.${j}.completed`, evalStatus.completed);
+					this.set(`_data.${i}.activities.${j}.published`, evalStatus.published);
+					this.set(`_data.${i}.activities.${j}.evaluated`, evalStatus.evaluated);
+					this.set(`_data.${i}.activities.${j}.unread`, evalStatus.newsubmissions);
+					this.set(`_data.${i}.activities.${j}.resubmitted`, evalStatus.resubmissions);
+					return true;
```
AlexBedley: returning `true` here so when we get to displaying publishAll errors we know if we found the activity or not. I could remove this if you think YAGNI

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -213,7 +214,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				key: this._getOrgHref(activity),
 				dueDate: this._getActivityDueDate(activity),
 				activityType: this._getActivityType(activity),
-				activityNameHref: this._getActivityNameHref(activity)
+				activityNameHref: this._getActivityNameHref(activity),
+				id: id
```
vix633: id is evaluationstatus id? Can the id have a way better name that describes what we're identifying?

AlexBedley: hmmm I could rename this to `evaluationStatusHref` which is what it is. My original thinking is that it doesn't really matter what `id` is (it was originally just an integer) just that it's unique. But I could see how that's confusing

AlexBedley: done

PR #281, opened 2019-08-06
Maboelsoudd2l/toggle different sizes

maboelsoudd2l:
Checked rtl and different browsers (chrome, safari, firefox, osx-edge).
This is what it looks like in different sizes:
Screen Shot 2019-08-06 at 2 47 35 PM
Screen Shot 2019-08-06 at 2 47 16 PM
Screen Shot 2019-08-06 at 2 46 56 PM

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -32,10 +32,9 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const quickEvalActivitiesTemplate = html`
 			<style>
 				.d2l-quick-eval-activity-list-modifiers {
-					float: right;
-				}
-				:host(:dir(rtl)) .d2l-quick-eval-activity-list-modifiers {
-					float: left;
+					display: flex;
+					margin-top: 18px;
+					width: 100%;
 				}
 				.clear {
 					clear: both;
@@ -44,6 +43,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 					display: inline-block;
 					width: 250px;
 					margin-left: .25rem;
+					flex: 1;
 				}
 				.d2l-quick-eval-no-submissions,
 				.d2l-quick-eval-no-criteria-results {
@@ -85,9 +85,9 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				.d2l-body-standard {
 					@apply --d2l-body-compact-text;
 				}
-				@media (max-width: 900px) {
+				@media (min-width: 525px) {
 					.d2l-quick-eval-activity-list-modifiers {
-						margin-top: 18px;
+						width: auto;
 						float: left;
 						clear: both;
 					}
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -12,6 +12,12 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 	static get template() {
 		const toggleTemplate = html`
 			<style>
+				:host {
+					margin: 0 -0.9rem;
+				}
+				label {
+					margin: 0 0.9rem;
```
maboelsoudd2l: A fix for https://rally1.rallydev.com/#/detail/userstory/323678156208?fdp=true

components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -12,6 +12,12 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 	static get template() {
 		const toggleTemplate = html`
 			<style>
+				:host {
+					margin: 0 -0.9rem;
```
maboelsoudd2l: Magical css inspired by bootstrap.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -85,6 +85,26 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				.d2l-body-standard {
 					@apply --d2l-body-compact-text;
 				}
+				@media (min-width: 525px) {
+					.d2l-quick-eval-activity-list-modifiers {
+						width: auto;
+						float: left;
```
AlexBedley: Does this change support RTL?

AlexBedley: oh I see the rtl selectors. Looks good

AlexBedley:
In the medium size (when the viewby and filter/search are left aligned) it looks a bit weird because they are offset a little. Screenshot: 
image


I'd expect their left edge to match (right now the filter/search is indented more). What do the designers think?

AlexBedley:
Another thing I found - in medium mode the filter panel opens above the button, whereas in big and small mode it opens below. I think we want the filter panel to always open below. On the demo page the filter panel is super small because it opens above which makes it less usable.

image

AlexBedley:
I tested this on windows edge and it looks good. Nice!

maboelsoudd2l:
> In the medium size (when the viewby and filter/search are left aligned) it looks a bit weird because they are offset a little. Screenshot:
> image
> 
> I'd expect their left edge to match (right now the filter/search is indented more). What do the designers think?

Oops, completely forgot to update the filter in Activities. Just force-pushed the changes to fix this

maboelsoudd2l:
> Another thing I found - in medium mode the filter panel opens above the button, whereas in big and small mode it opens below. I think we want the filter panel to always open below. On the demo page the filter panel is super small because it opens above which makes it less usable.
> 
> image

So this is not something in quick-eval, the dropdown is coded up to be below if there is sufficient space. If there is no space, the dropdown is above. https://github.com/BrightspaceUI/dropdown/blob/f28df5196e10fabc05cec4b33bca324b4b5c9ebf/d2l-dropdown-content-behavior.js#L630 I assume that behaviour makes sense.

I found that if you scroll down so that space below > space above, the dropdown automatically moves down as expected.

AlexBedley:
I verified that your push fixes the indenting issue. Nice.

I'm cool with the dropdown up/down thing.

PR #282, opened 2019-08-07
Vy/count and evaluate all

vix633:
Adds Evaluate all wireup and new submissions hover and wireup
image


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -213,6 +214,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 							</d2l-quick-eval-activity-card-items>
 						</div>
 						<d2l-quick-eval-activity-card-unread-submissions
+							on-click="_dispatchViewEvaluateNewEvent"
```
mtjandra: Looks good to me. Only thing is I would've though `d2l-quick-eval-activity-card-unread-submissions` and `d2l-quick-eval-activity-card-action-button` would handle their events themselves, but here  `d2l-quick-eval-activity-card` is managing the events they fire.

For example say HM doesn't give you the appropriate link (no link available). I don't think you would want to fire the event on-click in this case

vix633: I just pulled these guys in via the master merge and didn't consider that. Is it worth refactoring?

mtjandra: My only main concern is if it is possible that you show one of those two icons, but you don't get a link for it. I'm not sure what would happen if you do `window.location.href = ''`

vix633: I think is worth changing eventually. I've added this to a to;do list for later.

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -86,17 +86,35 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
 					const publishAll = this._getAction(e.entity, 'publish-all-feedback');
-					const submissionListHref = this._getRelativeUriProperty(e.entity, empty);
 
+					const submissionListSubEntity = e.entity.getSubEntityByRel(Rels.Assessments.submissionApplication);
+					let submissionListHref = '';
+					if (submissionListSubEntity.properties && submissionListSubEntity.properties.path) {
+						submissionListHref = submissionListSubEntity.properties.path;
+					}
+
+					const evaluateAllSubEntity = e.entity.getSubEntityByRel(Rels.Assessments.assessAllApplcation);
```
maboelsoudd2l: `assessAllApplcation` should be `assessAllApplication`

maboelsoudd2l: also me and @mtjandra notices that `e.entity.getSubEntityByRel` fails silently (in other words when `e.entity.getSubEntityByRel` is given an undefined value, it still returns a sub-entity). is there a way we can change this?

vix633: Yeah I noticed that before as well. Probably something worth changing, but not in scope here.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
maboelsoudd2l: @mtjandra as of this implementation, it doesn't seem like `d2l-quick-eval-activity-card-action-button` receives a url or link property. One alternate implementation of this, is that instead of dispatching an event, and then handling it in `d2l-quick-eval-activities` we can simply implement it in `d2l-quick-eval-activity-card-action-button` by putting `window.location.href = href;`. I think that is a simpler implementation and allows us to handle bad-links & disabled-buttons elegantly (though not sure what the downsides are of that approach).

mtjandra: > we can simply implement it in d2l-quick-eval-activity-card-action-button by putting window.location.href = href;

Yes great idea I like this. If this is the behavior of the component no matter where it is used, I would prefer this as well

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
mtjandra: Yes I would say this change is out of scope, there may be other component who maybe relies on this behavior? Who knows.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
maboelsoudd2l: Spoke with @vix633 offline, i changed my mind, I disagree with my own suggestion. because action-buttons should be dumb.

PR #283, opened 2019-08-07
US109177 - disable publish all button if api does not allow it

AlexBedley:
Problem: the publish all button is always enabled, but only certain activity types and permission sets allow it
Solution: disable the publish all button if the api does not allow it


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -219,17 +219,18 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
-								<d2l-quick-eval-activity-card-action-button 
-									icon-name="view-submission-list" 
+								<d2l-quick-eval-activity-card-action-button
+									icon-name="view-submission-list"
 									text="[[localize('submissionList')]]"
 									on-click="_dispatchViewSubmissionListEvent"></d2l-quick-eval-activity-card-action-button>
-								<d2l-quick-eval-activity-card-action-button 
-									icon-name="publish-all" 
+								<d2l-quick-eval-activity-card-action-button
+									icon-name="publish-all"
```
AlexBedley: Atom really hates trailing spaces

PR #285, opened 2019-08-08
Add back to QE query string

vix633:


test/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -110,6 +110,7 @@ suite('d2l-quick-eval-activities', function() {
 		assert.notEqual(getComputedStyle(list).display, 'none');
 	});
 	test('data is imported correctly', (done) => {
+		//console.log(JSON.stringify(act._data));
```
mtjandra: nit: remove

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -141,6 +141,13 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 			}.bind(this));
 	},
 
+	_formBackToQuickEvalLink(url) {
```
mtjandra: nit: `_fromBackToQuickEvalLink`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -358,6 +358,13 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		return result;
 	}
 
+	_formBackToQuickEvalLink(url) {
+		if (url) {
+			return url + '&cft=qe';
```
mtjandra: Do we know if this is always an additional query string param. IE. there is already a `?something=value` to append to?

test/d2l-quick-eval/data/assignmentEvaluationStatus.json
```diff
@@ -39,7 +39,7 @@
                 "https://assessments.api.brightspace.com/rels/assess-all-application"
             ],
             "properties": {
-                "path": "/assignment/all"
+                "path": "/assignment/all&cft=qe"
```
mtjandra: Shouldn't this just be `?ctf=qe` if there is only one query string param?

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
vix633: to form, as in to make

components/d2l-quick-eval/d2l-quick-eval-submissions.js
vix633: Right. This actually doesn't work really well as the value wont be interpreted. ref: https://tools.ietf.org/html/rfc3986#section-3.4

test/d2l-quick-eval/data/assignmentEvaluationStatus.json
vix633: Yeah, fat fingered this

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -141,6 +141,21 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
+		let newUrl;
+
+		if (!url) {
+			return url;
+		}
+
+		if (url.indexOf('?') > -1) {
+			newUrl = url + '&cft=qe';
+		} else {
+			newUrl = url + '?&cft=qe';
```
mtjandra: I think this should just be `?cft=qe`

AlexBedley:
Since we no longer have to support IE11 we should not be manually parsing urls. We should just use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

vix633:
> Since we no longer have to support IE11 we should not be manually parsing urls. We should just use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

Yeah, so i spoke to Stacey about using this and she said we can use it after January when we drop IE11 support. However as you just informed me we dropped IE11 support for QE, i'll remove this grossness and use the search params builder

svanherk:
>>Since we no longer have to support IE11 we should not be manually parsing urls. We should just use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

>Yeah, so i spoke to Stacey about using this and she said we can use it after January when we drop IE11 support. However as you just informed me we dropped IE11 support for QE, i'll remove this grossness and use the search params builder

Hopefully this is no longer true, but back when we were dealing with issues with `polymer-siren-behaviors`, I was seeing `url.searchParams.entries()` (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js#L22) not working at all.  I believe this was in Chrome, and that's why we were getting the random extra `?n=e` query parameter.  Hopefully that's been fixed, just a heads up if it doesn't behave as you expect.

AlexBedley:
@svanherk I thought that was because of the polyfill. Does BSI still automagically polyfill `URL` and `URLSearchParams`?

svanherk:
> @svanherk I thought that was because of the polyfill. Does BSI still automagically polyfill URL and URLSearchParams?

So now I'm concerned I'm mis-remembering everything lol, but from what I remember... BSI has never polyfilled URL or URLSearchParams, and there was hesitation towards adding it to the bundle to be loaded on every page.  That's why https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-quick-eval/compatability/ie11shims.js exists.  We were the first people to ever trigger the code path in `polymer-siren-behaviors`, so no one else had seen that a polyfill was needed.  And I'm pretty sure this behaviour was in Chrome, where the polyfill wouldn't have been used

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -141,6 +141,21 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
 
+	_formBackToQuickEvalLink(url) {
+		let newUrl;
+
+		if (!url) {
+			return url;
+		}
+
+		if (url.indexOf('?') > -1) {
+			newUrl = url + '&cft=qe';
```
maboelsoudd2l: soooooo in conclusion, can we actually use URLSearchParams?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -358,6 +358,21 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		return result;
 	}
 
+	_formBackToQuickEvalLink(url) {
```
maboelsoudd2l: Can we find a way have this function defined only once instead of in 2 different files?

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
vix633: We're using something else but similar.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
vix633: I'm sure if we sit down, really put our heads together we can find a way. but for now I'm ok with this being in two places

vix633: Mohamed made an excellent point offline, that i can just pull this implementation entirely since this class uses siren helpers.

PR #287, opened 2019-08-08
US109172 - show toast after successful publish all

AlexBedley:
Problem: users desire confirmation after taking actions
Solution: display a toast message after a successful publish all action

This PR does not contain langterms. Also, it does not fully cover the requirements - the requirements are to put the activity name in the toast as well as the message. As I mentioned in standup this morning, I want to merge this thin slice, followed by the activity name in the toast, and then lastly with langterms. Doing it this way will allow us to test the lion's share of the functionality before adding the complexity of langterms/activity name.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -4,6 +4,7 @@ import {QuickEvalLogging} from './QuickEvalLogging.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
 import 'd2l-alert/d2l-alert.js';
+import 'd2l-alert/d2l-alert-toast.js';
 import 'd2l-common/components/d2l-hm-filter/d2l-hm-filter.js';
 import 'd2l-common/components/d2l-hm-search/d2l-hm-search.js';
 import 'd2l-polymer-behaviors/d2l-id.js';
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -317,6 +323,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 						.then(evalStatusEntity => {
 							const evaluationStatusHref = this.getEvaluationStatusHref(evalStatusEntity);
 							this._updateEvaluationStatus(evaluationStatusHref, evalStatusEntity);
+
+							// _displayPublishAllToast is not reset if the toast is not manually closed
+							this._displayPublishAllToast = false;
```
AlexBedley: The problem here is that `d2l-alert-toast` does not fire an event or provide ANY public API to notify consumers that the toast has hidden itself. So if you:

1. click publishAll, it is successful
2. toast appears. you did it.
3. toast disappears after 2.5 seconds
4. click publishAll on a different activity, it is successful
5. toast will never appear again

Since there is no proper way to tell if the toast has hidden itself, I figured this way (forcing the animation to occur by flip flopping) was the lesser crime.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -372,5 +382,9 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		}
 		return false;
 	}
+
+	_onPublishAllToastClosed() {
```
AlexBedley: This event is fired when you click the little X on the toast message to dismiss it. This is not the same as the toast closing itself after 2.5 seconds (unfortunately)

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: Personally I was able to get the toast multiple times. But regardless, even if this bug does happen occasionally I kinda think that this is a bug that we shouldn't launch QE with (since a user clicking publish all a couple of times is a common use-case). But if the fix is not trivial (or impossible), then let's ship this PR first and worry about it later.

AlexBedley:
> Personally I was able to get the toast multiple times.

What do you mean by this? You commented out this line and were able to let the toast expire and appear again?

PR #291, opened 2019-08-12
clears options and searches on hidden

maboelsoudd2l:
Should clear search and filter when switching toggles from submissions to activities and vice-versa.

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
@@ -57,6 +57,13 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		this._clearErrors();
 	},
 
+	clearSearchResults: function() {
+		this.searchLoading = true;
+		const search = this.shadowRoot.querySelector('d2l-hm-search');
+		search.clearSearch();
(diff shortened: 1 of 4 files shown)


AlexBedley:
Looks good! I'm super happy you managed to make this a simple change.

One thing I noticed: when you switch back it display the filtered stuff, then with no warning just removes the filter. We should display the spinner when you switch back to tell the user we are loading.

AlexBedley:
Another thing is we don't reset sort state. However I think this is less egregious than not resestting filter /search state and can be addressed in a future PR (if at all)

components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -57,6 +57,13 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		this._clearErrors();
 	},
 
+	clearSearchResults: function() {
+		this.searchLoading = true;
+		const search = this.shadowRoot.querySelector('d2l-hm-search');
+		search.clearSearch();
```
AlexBedley: I can't find the source of this function which is sketchy. Do you know where it is?

AlexBedley: Oh nvm I found it now

PR #293, opened 2019-08-12
add text overflow

vix633:
image

Looks something like this ^

### DIFF
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -28,7 +28,7 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 				}
 				.d2l-activity-name-text {
 					vertical-align: middle;
-					max-width: 24ch;
+					max-width: 24rem;
 					text-overflow: ellipsis;
 					white-space: nowrap;
 					overflow: hidden;
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -27,15 +27,20 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 					padding-left: 0.6rem;
 				}
 				.d2l-activity-name-text {
+					display: inline;
 					vertical-align: middle;
+				}
+				.d2l-activity-name-text-and-icon {
 					max-width: 24rem;
 					text-overflow: ellipsis;
 					white-space: nowrap;
 					overflow: hidden;
 				}
 			</style>
-			<d2l-icon class="d2l-activity-name-icon" icon="[[_activityIcon]]"></d2l-icon>
-			<div title="[[_activityName]]" class="d2l-activity-name-text">[[_activityName]]</div>
+			<div class="d2l-activity-name-text-and-icon">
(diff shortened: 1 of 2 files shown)


components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -43,6 +47,10 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 				.d2l-quick-eval-activities-course-name-heading {
 					margin-top: 1.8rem;
 					margin-bottom: 1.2rem;
+					max-width: 24rem;
```
svanherk: I don't think you need this repeated here - since the above is outside any media query, it will always be applied.  This media query is then only used for changes/overrides when the screen is larger

components/d2l-activity-name/d2l-activity-name.js
```diff
@@ -27,9 +27,11 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 					padding-left: 0.6rem;
 				}
 				.d2l-activity-name-text {
```
svanherk: Some of these styles already exist here: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-quick-eval/d2l-quick-eval-submissions-table.js#L93
I'd either:
1) Keep the `activity-name` dumb, and apply these styles wherever you use the activity name (in the submissions table and in the activity card)
2) Add these styles here, but remove the duplicated styles from the submissions table

@jeroach I feel like you talked to design about where these styles should live originally (i.e. if it made sense for an activity name to always truncate, or for the client to truncate it)

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -50,7 +54,7 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 			<dom-repeat items="[[courses]]" as="c">
 				<template>
 					<li>
-						<h2 class="d2l-quick-eval-activities-course-name-heading">[[c.name]]</h2>
+						<h2 title= class="[[c.name]]" d2l-quick-eval-activities-course-name-heading">[[c.name]]</h2>
```
svanherk: Things got mixed up here

components/d2l-activity-name/d2l-activity-name.js
maboelsoudd2l: @svanherk I feel like the only things that would be duplicated are `overflow: hidden; text-overflow: ellipsis; `. Other than that the styling tends to be different (like padding, vertical-align and max-width). 

@AlexBedley thoughts?

vix633: Guyyyys, maybe no need to have the full team review a css change for text overflow, on shipping week 🥇

AlexBedley: http://everyonegetinhere.com/

To me this is more of a design question but my suspicion is we don't always want to truncate activity names. So I think it makes sense for it to live in the consumers (`activities` and `submissions`)

jeroach: @svanherk that does sound familiar but I don't remember what the outcome was :(

I think Viktor is doing a review with design so that should sort it out

PR #294, opened 2019-08-13
US109319 - New discussion tooltip lang

mtjandra:
* Update tool tip for discussion

### DIFF
--- a/components/d2l-quick-eval/build/lang/ar.js
+++ b/components/d2l-quick-eval/build/lang/ar.js
@@ -53,7 +53,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'newSubmissions': 'New Submissions',
 			'newSubmissionDetails': '{newNum} new, {resub} resubmissions',
 			'newPosts': 'New Posts',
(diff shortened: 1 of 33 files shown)


components/d2l-quick-eval/build/lang/ar.js
```diff
@@ -53,7 +53,7 @@ const LangArImpl = (superClass) => class extends superClass {
 			'newSubmissions': 'New Submissions',
 			'newSubmissionDetails': '{newNum} new, {resub} resubmissions',
 			'newPosts': 'New Posts',
-			'newPostDetails': '{newNum} new, {resub} reposts',
+			'newPostDetails': '{numInteractions} threads or replies',
```
maboelsoudd2l: so instead of the if-statement in `d2l-quick-eval-activity-card-unread-submissions.js`, why don't we change this lang term to 'newPostDetails': '{newNum} threads or replies', and we wouldn't need the if-statement at all.

mtjandra: I think I need the if statement still. I'm thinking I would always want to add `unread + resubmitted` just incase something ever changes in the backend

PR #296, opened 2019-08-13
Display activity name in toast

AlexBedley:
This PR adds:

* The activity name is now present in the toast
* The activity name is truncated to 48 characters
* Multiple toasts now appear on top of one another instead of replacing the older one
* English langterms. Other languages to follow after initial review

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -407,9 +407,9 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const maxLength = 48;
 
 		if (toast.activityName.length > maxLength) {
-			return this.localize('publishAllToastMessageTruncated', 'truncatedActivityName', toast.activityName.substring(0, maxLength))
+			return this.localize('publishAllToastMessageTruncated', 'truncatedActivityName', toast.activityName.substring(0, maxLength));
 		}
-		return this.localize('publishAllToastMessage', 'activityName', toast.activityName)
+		return this.localize('publishAllToastMessage', 'activityName', toast.activityName);
 	}
 }
 window.customElements.define(D2LQuickEvalActivities.is, D2LQuickEvalActivities);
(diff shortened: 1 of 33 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -337,11 +341,11 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
 						.then(evalStatusEntity => {
 							const evaluationStatusHref = this.getEvaluationStatusHref(evalStatusEntity);
-							this._updateEvaluationStatus(evaluationStatusHref, evalStatusEntity);
+							const publishedActivity = this._updateEvaluationStatus(evaluationStatusHref, evalStatusEntity);
 
-							// _displayPublishAllToast is not reset if the toast is not manually closed
-							this._displayPublishAllToast = false;
-							this._displayPublishAllToast = true;
+							if (publishedActivity) {
+								this.push('_publishAllToasts', { activityName: publishedActivity.activityName });
```
mtjandra: Not really sure how to test this, for some reason publish all isn't working from demo page. But I am a bit confused on why we hold all the toast activity names in an array. Don't we only want to show one toast at a time, not multiple?

AlexBedley: Publish all won't work in the demo page unless you override the global dialog stuff. Personally, I use 
```
D2L.LP = {
				Web: {
					UI: {
						Html: {
							JavaScript:{

							}
						}
					}
				}
			};

			const actuallyPublishAll = true;

			D2L.LP.Web.UI.Html.JavaScript.Confirm = function() {
				return {
					AddListener: (fn) => {
						fn(actuallyPublishAll);
					}
				};
			};
```

AlexBedley: The reason I hold them all in an array is so I can get rid of some of the hacks of the previous implementation (for example, `d2l-alert-toast` doesn't inform you when the toast auto-hides after 2.5 seconds). Secondly, this displays a new toast on top of the old one (the animation of the second one overlaps the old one) Previously, it would just immediately disappear and re-appear if you rapidly publish'd all.

PR #299, opened 2019-08-14
Cbusman/componentize subtitle

cbusmand2l:
New component for a bullet-separated subtitle. 

image

### DIFF
--- a/components/d2l-subtitle/d2l-subtitle.js
+++ b/components/d2l-subtitle/d2l-subtitle.js
@@ -1,7 +1,6 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
-import {QuickEvalLocalize} from '../QuickEvalLocalize.js';
 
-class D2LQuickEvalActivityCardSubtitle extends QuickEvalLocalize(PolymerElement) {
+class D2LSubtitle extends PolymerElement {
 	static get template() {
 		const subtitleTemplate =  html`
 			<style>
@@ -48,4 +47,4 @@ class D2LQuickEvalActivityCardSubtitle extends QuickEvalLocalize(PolymerElement)
 	}
 }
 
-window.customElements.define('d2l-quick-eval-activity-card-subtitle', D2LQuickEvalActivityCardSubtitle);
+window.customElements.define('d2l-subtitle', D2LSubtitle);
--- a/components/d2l-subtitle/d2l-subtitle.js
+++ b/components/d2l-subtitle/d2l-subtitle.js
@@ -43,7 +43,7 @@ class D2LSubtitle extends PolymerElement {
 	}
 
 	_computeDisplayText(text) {
-		return text.filter(t => t);
+		return text.filter(t => t !== null && t !== undefined && t !== '');
 	}
 }
 
--- a/demo/d2l-subtitle/d2l-subtitle.html
+++ b/demo/d2l-subtitle/d2l-subtitle.html
@@ -4,7 +4,7 @@
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
 
-		<title>d2l-quick-eval-activity-card-subtitle demo</title>
+		<title>d2l-subtitle demo</title>
 		<script src="../../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 
 		<!-- For IE11 -->
@@ -16,8 +16,7 @@
 			import '@polymer/iron-demo-helpers/demo-pages-shared-styles';
 			import '@polymer/iron-demo-helpers/demo-snippet';
 			import 'd2l-typography/d2l-typography.js';
-			import './update-direction.js';
-			import '../../components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-subtitle.js';
+			import '../../components/d2l-subtitle/d2l-subtitle.js';
 
 			const $_documentContainer = document.createElement('template');
 			$_documentContainer.innerHTML = `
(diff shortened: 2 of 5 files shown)


components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
...
+			},
+			_displayText: {
+				type: Array,
+				computed: '_computeDisplayText(text)'
+			}
+		};
+	}
+
+	_computeDisplayText(text) {
+		return text.filter(t => t);
```
AlexBedley: nit: explicitly filter out the things we don't want (empty string, null, undefined)

demo/d2l-subtitle/d2l-subtitle.html
```diff
@@ -0,0 +1,85 @@
...
+		<div class="vertical-section-container fixedSize">
+			<style>
+				d2l-subtitle {
+					display: block;
+				}
+			</style>
+			<h3>d2l-subtitle demo</h3>
+			<demo-snippet>
+				<template strip-whitespace>
+					<d2l-subtitle class="one-item"></d2l-subtitle>
```
AlexBedley: you can embed JSON into attributes by single quoting it `<d2l-subtitle text='["hello", "world"]'></d2l-subtitle>`

cbusmand2l: Always lamented that there wasn't a way to do this for the demos! Apparently there is! Thanks!

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+
+class D2LSubtitle extends PolymerElement {
+	static get template() {
+		const subtitleTemplate =  html`
+			<style>
```
dbatiste: The typography here almost matches our [core typography for d2l-body-small](https://github.com/BrightspaceUI/core/blob/master/components/typography/typography.js#L44).  The key differences being:

* color for `d2l-body-small` is tungsten
* responsive break-point is 615px
* line-height with small view-port is 0.9rem

Is it possible to stay with the standard?

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
...
+			}
+		};
+	}
+
+	_computeDisplayText(text) {
+		return text.filter(t => t !== null && t !== undefined && t !== '');
+	}
+}
+
+window.customElements.define('d2l-subtitle', D2LSubtitle);
```
dbatiste: Note: `d2l-subtitle` is a very generic name.  While it is unlikely that `core` will define such an element with this name, it's still possible.  Components in this repo should be named to effectively avoid such collisions which would cause `customElements.define` to throw a JS error.  Eventually this sort of component may be placed in the "labs" GitHub org, in which case it would be called `d2l-labs-subtitle`.

m6jones: This is almost the exact item I want to make `d2l-labs-list-horizontal` I've been talking to Colin about it. 

This would be amazing if you could transfer this component to labs. Make sure responsive behaviour is good :)

components/d2l-subtitle/d2l-subtitle.js
m6jones: On the projects I've been working on we have been using `d2l-body-small` so we should standardize it.

components/d2l-subtitle/d2l-subtitle.js
m6jones: (Note, if you want any help with that let me know and I can help setup repo and so on.)

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
...
+				}
+				@media (min-width: 900px) {
+					span {
+						font-size: .7rem;
+						line-height: .7rem;
+					}
+				}
+			</style>
+			<template is="dom-repeat" items="[[_displayText]]">
+				<span>[[item]]</span>
```
m6jones: For labs I would want it to be subelements. 
```
<d2l-labs-list-horizontal>
    <div> Item one </div>
     ....
    <div> item n </div>
</d2l-labs-list-horizontal>
```
My use case tends to want to show dates and such. So we need to add dom instead of plain text.

dbatiste: Yeah, if this is moved to [labs](https://github.com/BrightspaceUILabs), we should consider  name and API.

components/d2l-subtitle/d2l-subtitle.js
dlockhart: IMO it'd be cool if such a component had a name that was a little less semantic than `subtitle` or `list-horizontal` and more about the design intent? I think designers refer to these as "activity metadata" (or "object metadata"), so then the component could be something `d2l-labs-activity-metadata` and `d2l-labs-activity-metadata-item`. Then we could render it like a horizontal list today, but maybe down the road it changes to something else but that's OK because we didn't bake semantics into the component name.

m6jones: I like it!

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
```
maboelsoudd2l: nit: reminder to eventually change the crlf to lf line-endings

m6jones: This nit seems like the editor configuration isn't working.

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -0,0 +1,50 @@
...
+					font-size: .6rem;
+				}
+				span::before {
+					content: "\u2022";
+					margin: .2rem;
+				}
+				:host span:first-of-type::before {
+					content: "";
+					margin: 0;
+				}
```
maboelsoudd2l: nit: this would have been simpler
```
span + span::before {
    content: "\u2022";
    margin: .2rem;
}
```

components/d2l-subtitle/d2l-subtitle.js
maboelsoudd2l: spoke offline with @cbusmand2l, she just got a new laptop 😄

m6jones: Story checks out. :)

PR #301, opened 2019-08-14
add lang term for quiz resubmissions

vix633:
image

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
@@ -87,6 +87,9 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 		if (activityType === 'discussion') {
 			return this.localize('newPostDetails', 'numInteractions', unread + resubmitted);
 		}
+		if (activityType === 'quiz') {
+			return this.localize('newAttemptsDetails', 'newNum', unread, 'reAttemptNum', resubmitted);
+		}
 		return this.localize(activityTypeLocalizeDetail[activityType], 'newNum', unread, 'resub', resubmitted);
 	}
 }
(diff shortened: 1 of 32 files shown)


components/d2l-quick-eval/lang/en.json
```diff
@@ -48,7 +48,7 @@
   "newPosts": "New Posts",
   "newPostDetails": "{numInteractions} threads or replies",
   "newAttempts": "New Attempts",
-  "newAttemptsDetails": "{newNum} new",
+  "newAttemptsDetails": "{newNum} new, {reAttemptNum} reattempts",
```
jeroach: You need the `lang` file for all the other languages too

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-unread-submissions.js
```diff
@@ -87,6 +87,9 @@ class D2LQuickEvalActivityCardUnreadSubmissions extends QuickEvalLocalize(Polyme
 		if (activityType === 'discussion') {
 			return this.localize('newPostDetails', 'numInteractions', unread + resubmitted);
 		}
+		if (activityType === 'quiz') {
+			return this.localize('newAttemptsDetails', 'newNum', unread, 'reAttemptNum', resubmitted);
```
maboelsoudd2l: nit: can we instead of adding another if statement, is there a way we can use this line? 
`return this.localize(activityTypeLocalizeDetail[activityType], 'newNum', unread, 'resub', resubmitted);` 
perhaps we can change the lang term `newAttemptsDetails` to incorporate the `resub` variable?

vix633: Since quizzes have "attempts" as the accepted name for submissions in this case, I'd like to keep that naming around.

 I'll be adding a lot more ifs here in the next day or two as well. So maybe i'll refactor logic at that point if I see a good way to do it; I won't be changing the naming as long as quizzes have attempts.

PR #308, opened 2019-08-15
move cat image down via padding

vix633:
as per design on https://rally1.rallydev.com/#/238990806412d/detail/userstory/329131006348?fdp=true


Surekh:
Viktor said merge it.

PR #310, opened 2019-08-16
Support pluralization cases for tool tip

mtjandra:
Try playing with it using tool https://format-message.github.io/icu-message-format-for-translators/editor.html

### DIFF
--- a/components/d2l-quick-eval/build/lang/ar.js
+++ b/components/d2l-quick-eval/build/lang/ar.js
@@ -56,11 +56,11 @@ const LangArImpl = (superClass) => class extends superClass {
 			'toggleIndicatorLabel': 'Perform Actions on {target}',
 			'tryAgain': 'المحاولة مرة أخرى',
 			'newSubmissions': 'New Submissions',
-			'newSubmissionDetails': '{newNum} new, {resub} resubmissions',
+			'newSubmissionDetails': '{newNum, plural, =0 {{resub, plural, =1 {1 resubmission} other {{resub} resubmissions}}} other {{resub, plural, =0 {{newNum} new} =1{{newNum} new, 1 resubmission} other {{newNum} new, {resub} resubmissions}}}}',
 			'newPosts': 'New Posts',
-			'newPostDetails': '{numInteractions} threads or replies',
+			'newPostDetails': '{numInteractions, plural, =1 {1 thread or reply} other {{numInteractions} threads or replies}}',
 			'newAttempts': 'New Attempts',
-			'newAttemptsDetails': '{newNum} new, {reAttemptNum} reattempts',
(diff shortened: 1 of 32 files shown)


mtjandra:
@dlockhart I believe this is how you would expect pluralization to work when using ICU format, added you on in case you are interested.

components/d2l-quick-eval/build/lang/ar.js
```diff
@@ -56,11 +56,11 @@ const LangArImpl = (superClass) => class extends superClass {
 			'toggleIndicatorLabel': 'Perform Actions on {target}',
 			'tryAgain': 'المحاولة مرة أخرى',
 			'newSubmissions': 'New Submissions',
-			'newSubmissionDetails': '{newNum} new, {resub} resubmissions',
+			'newSubmissionDetails': '{newNum, plural, =0 {{resub, plural, =1 {1 resubmission} other {{resub} resubmissions}}} other {{resub, plural, =0 {{newNum} new} =1{{newNum} new, 1 resubmission} other {{newNum} new, {resub} resubmissions}}}}',
```
dlockhart: Took me a while to figure out what's going on here... haha it gets complicated when you want to omit the first part of the string like this when there are no results. I have a suspicion our translators are going to have trouble with this...

mtjandra: Yeah I'm almost certain it will come back incorrect. Would've also been great if I could format on multiple lines. But the json file doesn't allow that I don't think

dlockhart: I'm not sure what Jeff S thinks, but in a complex case like this we could possibly split this into 2 separate language terms and then bring them together using the comma in the UI code. Then we'd just drop the first one if there weren't any. That way the translators would translate them independently and not have to worry about nesting.

mtjandra: On one extreme, we don't need to use ICU formatting at all for pluralization, you could put all the logic in code, and have one langTerm for each case, decided what to do in the application. On the other extreme you can place all that logic in the language term. The last option is having some logic in the langTerm and some in code.

My personal opinion is that translators will eventually need to figure out how to deal with nested formats. If unavoidable, we could build tooling so that devs don't need to worry about this, and terms can transformed in a way that is digestible for translators. But I'd say this is a problem to be solved between the translation vendor and their translators.

dlockhart: Yeah I agree. I like having most of the logic in the translations, since for a lot of languages -- especially with pluralization cases like this -- they only need to handle a couple of the options. In English, we often deal with zero, one and more-than-one. Some other languages are much more complex.

mtjandra:
> I'm down to take the risk

When talking to @Surekh she seemed to be fine to try it out as well. Will merge now so it will get picked up by Serge

components/d2l-quick-eval/build/lang/ar.js
mtjandra: Haha yes. To take it further, translators should ideally be able to add their own cases (for 2, 3 etc.) that make sense in their language. Not even sure we even support that today

PR #315, opened 2019-08-22
Dont request activities data when activities view is disabled

AlexBedley:
This PR contains 2 changes:

1. when `activitiesViewEnabled` is set to `false`, we set `d2l-quick-eval-activities` `href` to empty string. [This causes the entitybehavior to not request the href](https://github.com/Brightspace/polymer-siren-behaviors/blob/9bc765ab8f130a980fd7d947f7925c2eea5bd2f4/store/entity-behavior.js#L70) since it will 404 anyways. This is from https://d2l.slack.com/archives/CBTK3JB6D/p1566491095322300
2. Cleanup of some old code that is no longer necessary (the LMS sends `submissionHref` now instead of `href`)


AlexBedley:
FYI: @amauryperez

PR #316, opened 2019-08-22
added white box-shadow to action-card items

maboelsoudd2l:
there's a small miniscule dot under "evaluate all", so i found a way to hide it using a white `box-shadow`
Screen Shot 2019-08-22 at 3 08 24 PM
Screen Shot 2019-08-22 at 3 09 15 PM

### DIFF
--- a/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
+++ b/components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
@@ -12,6 +12,9 @@ class D2LQuickEvalActivityCardItems extends mixinBehaviors([D2L.PolymerBehaviors
 					display: flex;
 					align-items: stretch;
 					justify-content: space-between;
+					-webkit-box-shadow: 0px 7px 0px 0px rgb(255, 255, 255);
+					-moz-box-shadow: 0px 7px 0px 0px rgba(255,255,255);
+					box-shadow: 0px 7px 0px 0px rgb(255, 255, 255);
 				}
 				.d2l-quick-eval-activity-card-items-container ::slotted(*) {
 					width: 7.5rem;


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
```diff
@@ -12,6 +12,9 @@ class D2LQuickEvalActivityCardItems extends mixinBehaviors([D2L.PolymerBehaviors
 					display: flex;
 					align-items: stretch;
 					justify-content: space-between;
+					-webkit-box-shadow: 0px 7px 0px 0px rgb(255, 255, 255);
```
dbatiste: Are you sure the vendor prefixes are needed?  Normally I think only the `box-shadow` is needed nowadays.

dbatiste: Curious what that dot is coming from...?

maboelsoudd2l: from the bottom of the "p" in "completed"
Screen Shot 2019-08-22 at 3 22 26 PM

maboelsoudd2l: I have no idea, do we need vendor prefixes?

AlexBedley: http://shouldiprefix.com/#box-shadow

doesn't look like it but a quick test can validate this easily

maboelsoudd2l:
I was discussing this offline with Carol and our concern is that hiding the underlying content might possibly affect screenreaders

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-items.js
m6jones: Design for the meter measured out the space for meter between the top and a flat word. So instead of 60px it is 62px. I talked with Bob about it and he said it was fine and expected.

dbatiste:
Yeah I think we should address this in the meter component if possible.  Otherwise this is bound to come up for some other consumers of it.  In addition, this work-around assumes a white background, which I don't think is safe.

Re the vendor prefixes: I don't think they should be needed for `box-shadow`.  If you see them for box-shadow, it's probably older code, or someone has copied older code.

m6jones:
I don't think this is a meter problem. The component is 62px in height.
The design: 
image
Didn't take into account all the letters.

Another option is to hide the meter while the overlay is active.

maboelsoudd2l:
is going to be implemented in https://github.com/BrightspaceHypermediaComponents/activities/pull/345

PR #329, opened 2019-09-03
Setting aria-pressed to string.

cbusmand2l:
No visual change. Verified that NVDA announces the difference between pressed and not pressed.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
@@ -89,13 +89,13 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 			</style>
 			<label>[[localize('viewBy')]]</label>
 			<button
-				aria-pressed$="[[_isSelected(_viewTypes.submissions, currentSelected)]]"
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/d2l-quick-eval-view-toggle.js
```diff
@@ -146,6 +146,9 @@ class D2LQuickEvalViewToggle extends QuickEvalLocalize(PolymerElement) {
 	_isSelected(view) {
 		return this.currentSelected === view;
 	}
+	_isPressed(view) {
```
maboelsoudd2l: nit: name is misleading.

AlexBedley: nit: `_isSelected` doesn't really need to exist since it isn't called anywhere else. I would just merge the two functions and call it `_isSelected`

cbusmand2l: `_isSelected` is bound to the `selected$` attribute on the button, which needs to be a boolean, not a string.

AlexBedley: oh duh. I must have misread, sorry!

PR #336, opened 2019-09-06
Vy/default view telemetry

vix633:
This is intended to capture what view is used by default


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -104,6 +109,7 @@ class D2LQuickEval extends  QuickEvalLocalize(PolymerElement) {
 		} else {
 			this._showActivitiesView = true;
 		}
+		this.logViewQuickEvalEvent(e.detail.view);
```
AlexBedley: So this will fire on the first load, but also every time the toggle is clicked. Is that what you want? The PR description specifically calls out that we want to measure the default view. If you want to measure just the default you can use an observer.

Additionally there are some nuances to this - technically this will fire with the currently displayed view which isn't necessarily the view that was passed in. For example, if you pass in `null` it will display `submissions`. Or if you pass in `activities` but the activities flag is disabled, it will display `submissions`. Not sure if you want to differentiate between these cases (it is possible)

vix633: Discussed offline

PR #342, opened 2019-09-10
fix ellipsis and icons being cut off on all browsers

vix633:
This should fix the weird icon clipping across browsers, it should also make the ellipsis work.

### DIFF
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -20,7 +20,8 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 				.d2l-activity-name-icon {
 					display: inline-block;
 					padding-right: 0.6rem;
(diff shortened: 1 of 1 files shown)


vix633:
> Ellipsis looks fixed on Chrome and Firefox on Windows. The activities don't load on Edge, but that's not specific to this PR.
> 
> What's the icon cut off issue? I didn't notice anything odd, but I'm not sure what I should be looking for.

The bottom of the domain specific icon would get cut off in activities view on some browsers.

AlexBedley:
> The activities don't load on Edge, but that's not specific to this PR.

Is this a known issue? We are supposed to work on Edge for this release afaik

jeroach:
> > The activities don't load on Edge, but that's not specific to this PR.
> 
> Is this a known issue? We are supposed to work on Edge for this release afaik

Switched back to `master` locally and Edge is still broken, so not specific to this PR.

Also QE works in Edge on Quad so I think we're okay

PR #343, opened 2019-09-10
[WIP] Perceptual diff tests

DiljotSG:


m6jones:
Woot!! first step!

maboelsoudd2l:
@DiljotSG @dlockhart @AlexBedley is there a way that we can integrate this into Travis CI? and if yes, can we control which PR's need a visual diff, and which PR don't? (so for example, if we are adding a feature, we expect the visual to look different, the visual diff will obviously fail so we don't want it to run)

dlockhart:
Totally -- have a look at how [core does it](https://github.com/BrightspaceUI/core/blob/master/.travis.yml#L11). It's storing the goldens in an S3 bucket, but I know @margaree (on vacation right now) is [working on getting them checked into source control](https://github.com/BrightspaceUI/core/pull/203).

package.json
```diff
@@ -18,6 +20,7 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@babel/polyfill": "^7.0.0",
+    "@brightspace-ui/visual-diff": "0.0.2",
```
dlockhart: You'll probably want a semver range like `^0.0.2` so you can automatically pick up patch releases.

dlockhart:
One note about goldens for the future: make sure the goldens that get generated that you're comparing against are generated in Travis, not on your local machine. There's too much variance between OS/browser with font kerning and anti-aliasing and there's just no way to get a consistent result from different machines.

AlexBedley:
@dlockhart how do we signal that we want to update the goldens in the S3 bucket?

dlockhart:
Currently we do it manually. We create a custom build in Travis that runs the special `npm run test:diff:golden` script, which will overwrite the goldens.

This isn't what we want long-term though. Ideally I think we'd have a GitHub Application you install with a little workflow where you could first look at the diffs, and then "approve" them, which would update your goldens automatically. But that's further down the road.

DiljotSG:
@dlockhart What is the procedure for updating the goldens with travis? I can see there is a script [here](https://github.com/BrightspaceUI/core/blob/a66b35bfe4c3153892b5f8c6c769604a8d6d68cc/cli/commit-goldens.js) that updates the goldens, however, I am not sure how that script gets run.

PR #344, opened 2019-09-11
add a nullcheck to address errors bubbling up to kibana

vix633:
This is to address 
```
TypeError: Cannot read property 'hasClass' of null
    at HTMLElement.<anonymous> (https://s.brightspace.com/lib/bsi/20.19.9-75/es6-bundled/web-components/bsi.js:1344:5787)
```

I checked the instance system logs and there is nothing logged during the window that this happened, I can only attribute it to a network blip or something innocuous.

### DIFF
--- a/components/d2l-activity-name/d2l-activity-name.js
+++ b/components/d2l-activity-name/d2l-activity-name.js
@@ -95,6 +95,9 @@ class D2LActivityName extends mixinBehaviors([D2L.PolymerBehaviors.Siren.EntityB
 	async _getActivityPromise(activityEntity) {
 		let rel;
 		let activityIcon;
+		if (!activityEntity) {
+			return Promise.resolve();
+		}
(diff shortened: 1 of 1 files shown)


vix633:
> We shouldn't silently fail here since this function is pretty important to QE (it would cause the row to render without a title). I'm ok receiving these errors in kibana - if a large percentage of users get this then we might have an issue.
> 
> I think alternative solutions could be: changing our alert threshold (one of these issues isn't worth our time, but 100 of them is probably something we're doing), moving logging up to line 77 of this file (inside the catch statement), or doing nothing.

I don't see this much differently that resolving the promise if there are no usages of the type we expect(few lines below, in the same method). Are you sure this will have the effect you're describing?

vix633:
Closing as this has an unwanted side-effect(maybe)

PR #348, opened 2019-09-12
Vy/view load performance telemetry

vix633:
Going to ponder on tests for a bit here as well. So far this seems to be logging everything I want it to. 

If you guys see a concrete, better way to approach this please let me know.

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
@@ -63,7 +63,7 @@ D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
 		window.performance.measure(viewName, startMark, endMark);
 		const eventBody = new window.d2lTelemetryBrowserClient.PerformanceEventBody()
 			.addUserTiming(window.performance.getEntriesByName(viewName))
-			.addCustom('ViewName', `${viewName}LoadTime` || 'unknown');
+			.addCustom('ViewName', `${viewName}LoadTime`);
 
 		this._logEvent(eventBody);
 	}
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
@@ -70,10 +70,7 @@ D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
 		window.performance.clearMarks(endMark);
 	},
 	_markExists(markName) {
-		return window.performance.getEntries({
-			name: markName,
-			entryType: 'mark'
-		}) ? true : false;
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -60,7 +60,10 @@ class D2LQuickEval extends
 			<d2l-quick-eval-activities href="[[_activitiesHref(activitiesViewEnabled)]]" token="[[token]]" logging-endpoint="[[loggingEndpoint]]" hidden$="[[!_showActivitiesView]]"></d2l-quick-eval-activities>
 		`;
 	}
-
+	connectedCallback() {
```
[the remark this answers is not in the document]
AlexBedley: They are both called "at the same time" when the element is attached to the document, except ready is only called the first time, and connectedCallback is called every time. https://polymer-library.polymer-project.org/3.0/docs/devguide/custom-elements#element-lifecycle You could do a quick console.log test to figure out which is called first.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -239,6 +240,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			this._updateSearchResultsCount(this._data);
 			this._clearErrors();
 			this._handleLoadSuccess();
+			this.perfMark('activitiesLoadEnd');
+			this.logPerformanceEvent('activities', 'qeViewLoadStart', 'activitiesLoadEnd');
```
AlexBedley: you forgot `logAndDestroyPerformanceEvent`

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -49,6 +49,31 @@ D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
...
+			return;
+		}
+		window.performance.measure(viewName, startMark, endMark);
+		const eventBody = new window.d2lTelemetryBrowserClient.PerformanceEventBody()
+			.addUserTiming(window.performance.getEntriesByName(viewName))
+			.addCustom('ViewName', `${viewName}LoadTime`);
+
+		this._logEvent(eventBody);
+		window.performance.clearMarks(startMark);
+		window.performance.clearMarks(endMark);
```
AlexBedley: might as well also `clearMeasures(viewName)` to provide that clean slate

components/d2l-quick-eval/behaviors/d2l-quick-eval-telemetry-behavior.js
```diff
@@ -49,6 +49,31 @@ D2L.PolymerBehaviors.QuickEval.TelemetryBehaviorImpl = {
...
+	},
+	_markExists(markName) {
+		return window.performance.getEntries({
```
AlexBedley: this will always return `true` because `[]` is truthy

PR #349, opened 2019-09-12
Screen readers should read tooltip text for draft icon.

cbusmand2l:
Should be no visual change. Screen readers should now read the tooltip when navigating to the draft icon.


AlexBedley:
per @maboelsoudd2l 's instruction, I'm testing this on VoiceOver

AlexBedley:
Hmm no matter what I try I cannot get VoiceOver to read the tooltip, it does not think the draft icon is focus/selectable. It's possible I just don't know how to use VoiceOver

AlexBedley:
Discussed offline with Carol: `aria-labelledby` is not working on VoiceOver, switching to `aria-label` as a workaround

PR #370, opened 2019-09-18
Skeleton components, tests, demo pages for assignment activity editor

mdgbayly:
Has a dependency on https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/63, so CI on this PR will fail until that is merged.

These are just placeholder components for us to start adding assignment editor functionality.

I've added some basic localization support following the patterns i've seen being used in the latest BrightspaceUI/core stuff but not hooked it up to the serge files yet until we have some really lang terms.

Also added starting demo pages and test files.

The components are structured as follows:
```
<d2l-activity-assignment-editor>
    <d2l-activity-editor>
        <d2l-activity-assignment-editor-detail>
```
`<d2l-activity-assignment-editor>` is the component that is written into the LMS page. It will be initialized with an href for the assignment activity usage API.

`<d2l-activity-editor>` can be used to give some consistent 'chrome' (e.g. the left/right panes etc) to all activity editors and could even potentially be used as a component on it's own to give basic activity editing capabilities if we find such a thing useful as we get deeper. 

`<d2l-activity-assignment-editor-detail>` is slotted into `<d2l-activity-editor>` and is the initial component that will manage the assignment specific details of adding activities. It's initialized with the href of the assignment folder API.

This is all fluid and may change as we start adding in specific properties and finding the best ways to combine editing activity properties and assignment properties!


mdgbayly:
@ryantmer Yes, that's correct regarding the rendering.

Here is a copy of a slack discussion we were having last week about this.
Vincent is working on the renderer. Initially, we'll just hardcode it to be the assignment component to get things up and running and then we'll add the plugin stuff later.

> We'll probably eventually want to make the LMS page dynamically choose the component based on the type of activity being edited.
the way that kind of thing is usually accomplished in the activities domain is by using a plugin which can be implemented by each activity type.
using a mechanism similar to this:
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox/Plugins/Activities/LMSAssignmentActivityUsageProvider.cs?r=12678700#105
Activitities domain code polls the provider passing in the ILMSActivity or ILMSActivityUsage until it finds a provider that knows how to handle the current activity type.
here is an example of polling providers
https://search.d2l.dev/xref/lms/le/activities/D2L.LE.Activities/Service/Default/LMSActivityProviderService.cs?r=6176bc5f#61
So I was thinking we might want to setup a new plugin that allows us to configure the web component based on the activity usage info.
initially it will just have the name of the tag which we'll use to write the web component tag in the page.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -0,0 +1,79 @@
...
+				};
+			}
+		}
+
+		return null;
+	}
+
+	constructor() {
+		super();
+		this._setEntityType(AssignmentActivityUsageEntity);
```
m6jones: Would it make sense to add the `_onAssignmentActivityUsageChange` to this call. So then we can avoid `set _entity(entity)` boilerplate?

mdgbayly: @m6jones  Sorry, didn't see this earlier - been fighting with wct tests again all day...

Don't totally follow? Do you mean adding some kind of callback to `_setEntityType`.

m6jones: Yeah, so setting the callback  in the `_setEntityType` method instead of calling it in the `set` method.

svanherk:
@mdgbayly Yeah that test failure you're seeing is because of this: https://github.com/Polymer/tools/pull/727
The fix is merged in but currently not released: https://github.com/Polymer/tools/blob/master/packages/build/CHANGELOG.md#change-log
Once this happens, we can re-enable Edge testing in the repos we've had to disable it in, including `core` and anything using `d2l-icons`.

PR #377, opened 2019-09-23
US109690 - display currently searched term on initial load

AlexBedley:
Problem: we want to be able to persist search state during evaluation
Problem: search state is persisted, but upon load the search state is not displayed to the user
Solution: use the new `initialValue` property of `d2l-hm-search` from https://github.com/BrightspaceHypermediaComponents/common/pull/32
 
FYI @jeroach since you found the bug in the first place

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
@@ -26,6 +26,10 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		searchLoading: {
 			type: Boolean,
 			value: false
+		},
+		searchTerm: {
+			type: String,
+			computed: '_computeSearchTerm(entity)'
 		}
 	},
 
@@ -43,6 +47,22 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 
 	_computeSearchAction: function(entity) {
 		return this._getAction(entity, 'search');
+	},
+
+	_computeSearchTerm: function(entity) {
+		if (!entity) {
+			return;
+		}
+		const searchTerm = entity
+			.getActionByName('search')
+			.getFieldByName('collectionSearch')
+			.value;
+
+		if (searchTerm) {
+			this.searchApplied = true;
+		}
+
+		return searchTerm;
 	},
 
 	onSearchResultsLoading: function() {
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -130,6 +130,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 					search-action="[[searchAction]]"
 					placeholder="[[localize('search')]]"
 					aria-label$="[[localize('search')]]"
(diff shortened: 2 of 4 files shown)


components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -45,6 +49,22 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		return this._getAction(entity, 'search');
 	},
 
+	_computeSearchTerm: function(entity) {
+		if (!entity) {
```
AlexBedley: `entity` is null on initial load (before any api requests happen). this just protects against an exception in that case

components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -45,6 +49,22 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
...
+			return;
+		}
+		const searchTerm = entity
+			.getActionByName('search')
+			.getFieldByName('collectionSearch')
+			.value;
+
+		if (searchTerm) {
+			this.searchApplied = true;
```
AlexBedley: this is important otherwise the search ecosystem behaves weirdly (ie search summary is not shown). It's a bit annoying that we have to duplicate the check here (since searching on initial load doesn't trigger the normal search workflow) but it wasn't egregious enough that I thought it was worth a large refactor.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -145,7 +146,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			</d2l-alert>
 			<d2l-quick-eval-search-results-summary-container
 				search-results-count="[[_searchResultsCount]]"
-				hidden$="[[!searchApplied]]"
+				hidden$="[[!_showSearchSummary]]"
```
AlexBedley: This now depends on the loading status, as well as `searchApplied`. During testing it would blip onto the screen and show 0 search results before blipping again to show the correct number of results. So I added this check to both a activities and submissions such that it only displays once loading is completely finished.

components/d2l-quick-eval/behaviors/d2l-hm-search-behavior.js
```diff
@@ -45,6 +49,22 @@ D2L.PolymerBehaviors.QuickEval.D2LHMSearchBehaviourImpl = {
 		return this._getAction(entity, 'search');
 	},
 
+	_computeSearchTerm: function(entity) {
+		if (!entity) {
+			return;
+		}
+		const searchTerm = entity
+			.getActionByName('search')
```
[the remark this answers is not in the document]
AlexBedley: If we get nulls in this chain then something went wrong. I'm ok not silently failing in that case. In contrast to the null `entity` check above - in that case I'm guarding against a specific edge case which I do want to silently ignore.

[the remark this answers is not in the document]
AlexBedley: Spoke offline, tried to weight the tradeoffs between adding extra code to handle errors and the odds of this actually happening and what could cause it. Decided to just leave it as-is

PR #352, opened 2019-09-13
Perceptual Diff Tests - Stage 3 - Travis CI Improvements / Visual Diff Bot

DiljotSG:

PR #346, opened 2019-09-11
Perceptual Diff Tests - Stage 2 - Travis CI

DiljotSG:
This PR adds the ability for Travis CI to run the visual difference tests on each pull request. The current state will be compared to the goldens stored in the S3 bucket. Goldens can be regenerated by triggering a custom travis build (See README.md).

PR #378, opened 2019-09-23
added translation "view in info on (activity name)" 

maboelsoudd2l:


cbusmand2l:
I assume we don't get `fi` translations?

maboelsoudd2l:
@cbusmand2l guess not...
also `fi.json` looks awfully like english

Surekh:
Hmm... looks like nothing has changed since last approval. So, looks good to me. Why is Visual Difference Tests taking this long - @DiljotSG? We should resolve that at the earliest to unblock this PR and then backport this PR change to 20.19.10 release.

DiljotSG:
Hi @Surekh, sorry about that but the visual difference tests aren't actually running and won't run until #346 and #352 are merged in. I was still testing the visual difference bot and it happened to mark Mohammed's repo, even though there were no tests running. It is safe to ignore that check and merge it anyway.

PR #379, opened 2019-09-23
Allow quick eval to dynamically respond to activitiesHref changes

AlexBedley:
Problem: Activities does not work when you dynamically set `activitiesHref` because the `_activitiesHref` function is only called when one if its parameters change
Problem: This was preventing auth-me (the dev tool I'm creating to automatically fetch keys) from working with activities
Solution: force polymer to recalculate `_activitiesHref` when `activitiesHref` changes


AlexBedley:
> On question though: it seems that we call `_activitiesHref(activitiesViewEnabled)` to enable lazy-loading for the activities view. what's stopping us from doing the same for submissions?

It's actually to support when the activities flag is disabled, but yes this is a mechanism we could use to enable lazy loading!

PR #380, opened 2019-09-23
US110193 display assignment name, non-ediable if no update-name actio…

mdulat:
…n present
Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/64

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -1,9 +1,9 @@
+import 'd2l-inputs/d2l-input-text.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
-import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
-import 'd2l-inputs/d2l-input-text.js';
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 
 class AssignmentEditorDetail extends EntityMixinLit(LocalizeMixin(LitElement)) {
 
@@ -79,8 +79,8 @@ class AssignmentEditorDetail extends EntityMixinLit(LocalizeMixin(LitElement)) {
 				<d2l-input-text
 					id="assignment-name"
 					value="${this._name}"
-					on-change="${this._saveName()}"
-					on-input="${this._saveNameOnInput()}"
+					@change="${this._saveName}"
+					@input="${this._saveNameOnInput}"
 					aria-label="${this.localize('name')}"
 					?disabled="${!super._entity.canEditName()}"
 					prevent-submit>
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -2,6 +2,7 @@ import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import 'd2l-inputs/d2l-input-text.js';
 
 class AssignmentEditorDetail extends EntityMixinLit(LocalizeMixin(LitElement)) {
 
@@ -15,6 +16,7 @@ class AssignmentEditorDetail extends EntityMixinLit(LocalizeMixin(LitElement)) {
 		return css`
 			:host {
 				display: inline-block;
+				padding: 20px;
 			}
 			:host([hidden]) {
 				display: none;
@@ -58,13 +60,36 @@ class AssignmentEditorDetail extends EntityMixinLit(LocalizeMixin(LitElement)) {
 	_onAssignmentChange(assignment) {
 		if (assignment) {
 			this._name = assignment.name();
+
+			if (assignment.canEditName()) {
+				this.shadowRoot.getElementById('assignment-name').removeAttribute('disabled', '');
+			} else {
+				this.shadowRoot.getElementById('assignment-name').setAttribute('disabled', '');
+			}
 		}
+	}
+
+	_saveName() {
+
+	}
+
+	_saveNameOnInput() {
+
 	}
 
 	render() {
 		return html`
-			<h2>Hello ${this._name}!</h2>
-			<div>Localization Example: ${this.localize('myLangTerm')}</div>
+			<div id="assignment-name-container">
+				<label for="assignment-name">${this.localize('name')}*</label>
+				<d2l-input-text
+					id="assignment-name"
+					value="${this._name}"
+					on-change="${this._saveName()}"
+					on-input="${this._saveNameOnInput()}"
+					aria-label="${this.localize('name')}"
+					prevent-submit>
+				</d2l-input-text>
+			</div>
 		`;
 	}
 }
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -1,71 +1,96 @@
...
+			this._onAssignmentChange(entity);
+			super._entity = entity;
+		}
+	}
+
+	_onAssignmentChange(assignment) {
+		if (assignment) {
+			this._name = assignment.name();
+
+			if (assignment.canEditName()) {
```
mdgbayly: This is an example of where I think the current integration mechanism with the entity store is awkward.
I think we should be able to set the 'disabled' attribute declaratively?
https://lit-element.polymer-project.org/guide/templates#bind-properties-to-child-elements
In fact, I'm not entirely sure why we need the _entity setter at all...
I just copied this from the example of the entity-lit-mixin usage  in the siren sdk.

Possibly we could just do something like 
```
get _canEditName() {
    return super._entity.canEditName();
}

<d2l-input-text ?disabled="${this._canEditName}"....>

```
But even that feels like a lot of boiler plate...

I'm not sure if you can reference functions in lit property bindings. But if we can maybe it would be better to just bind to `super._entity.canEditName()`.

ryantmer: I'm just kind of learning LitElement now (slash updating my WC knowledge in general), but it does seem like making DOM changes outside `render` is recommended against

> Also, avoid making DOM updates outside of `render`. Instead, express the element’s template as a function of its state, and capture its state in properties. - https://lit-element.polymer-project.org/guide/templates 

I'm fairly certain calling functions from within the DOM template was allowed in Polymer, but I haven't figured out whether that's still the case with LitElement or not, sadly.

mdgbayly: I too am just learning to crawl with lit...

I'm pretty sure you can call functions from within the render/template, just not sure about property bindings, or if you can, if it's deemed good practice. The guidance you quote above also says things like `Make sure your render function...only depends on your element's properties'.

But I think that is because the lit lifecycle hooks into the properties and initiates an update whenever the lit implemented property setters/getters detect a property change. If you create your own property setters/getters then I think you are responsible for initiating the update.

https://lit-element.polymer-project.org/guide/properties#accessors-custom

I think there are possibly some other bugettes in the current `entity-mixin-lit`. I'm not sure if it will currently re-render correctly when the entity changes outside of changes to the token/href, which is something we're definitely going to be contending with given we're an editor.

But as Maya mentioned, I think we should feel empowered to take the lit mixin and entity store integration forward, and if we can come up with better patterns for how to use it and make it easier to use - then yay.

dbatiste: This is true - modifying the DOM outside of `render` is not recommended.  In short, the UI = f(data) where `f` is the `render` method.  And there should be no side effects.  If this principle is followed, it should also be true that the inverse is true, data = f<sup>-1</sup>(UI).

It is definitely possible to call functions from within the render template.  Those functions would themselves returns template results. It is useful in a variety of cases (factoring rendering into mixins, reuse, conditional rendering, etc).

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -1,71 +1,97 @@
...
+				display: inline-block;
+				padding: 20px;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+		`];
+	}
+
+	static async getLocalizeResources(langs) {
```
dbatiste: Given that the resources are essentially static, we could avoid this duplicate logic by using the `LocalizeStaticMixin` with the `resources` static getter. ex. [dialogs](https://github.com/BrightspaceUI/core/blob/master/components/dialog/dialog.js#L60). Note: when there are missing translations, this mixin will automatically fallback.

If we want to use Serge (which this repo currently appears to not use) then we need to spend a bit more time flushing out how we can make that work with some fallback capability.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: I also started a conversation with Maya very recently about the `entity-mixin-lit` and the possibility of using the [run-async](https://github.com/BrightspaceUI/core/blob/master/directives/run-async.js) directive. I think it would clean-up some of the boiler plate getters/setters/requestUpdate and generally make it a lot cleaner. This is something she was very interested in as well.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -1,71 +1,97 @@
...
+			super._entity = entity;
+		}
+	}
+
+	_onAssignmentChange(assignment) {
+		if (assignment) {
+			this._name = assignment.name();
+
+			if (assignment.canEditName()) {
+				this.shadowRoot.getElementById('assignment-name').removeAttribute('disabled', '');
```
dbatiste: So yeah, we definitely don't want to be modifying the DOM here.  It should be done in the template as Martin suggested.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: @dbatiste Yeh, I've been looking into the `run-async` directive to get a better understanding of what it is doing and to try to gain some insight into how it could be used. We are especially interested, in using something like it to delay rendering of the UI (showing a skeleton instead) until we have sufficient data from all the different APIs we're calling. I had been considering an event based solution like the PendingContainer stuff that is used in the lit async demos, but then when you mentioned the `run-async` stuff that seemed like it might be a great off the shelf solution.

I plan on pinging you this morning about it. It looks cool but I'm not clear how such a technique could be used when we require the rendering of some of our components to update when the entities which they render are changed by other components. In these cases the entity is refreshed in the entity store, and notifications sent out, but the URL key being rendered by the component doesn't change. So I'm not clear yet how such a requirement could be used with the run-async approach, and whether we'd be better off just building some other type of shared integration that manages firing `pending-state` type events when entities are being loaded by the store.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Can this approach be used when the resources are being loaded from a separate file?
We still need to spend some more time digging into our localization options. This was initially just copied from the lit-element template repo, where it looked like the intention was to have a separate file per language rather than having all static resources?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: Just spit-balling here: is it possible to update the `href` when the entity changes in the store?  For example, by appending a timestamp, id, version, etc.?  That way, `render` would get called again, and the new entity (updated) fetched from the store when `run-async` is called again?  Or perhaps that's forcing the pattern that doesn't quite fit what we're trying to do?

Another approach might be to store the entity on the element (ex. as `this._entity`). Initially it would be `null`, and `run-async` would use the `href` to request it, and then assign it.  If it's not `null` then `render` would just use it.  If the component is listening with a callback for changes to the entity in the store, it would assign the new value for `this._entity` and then force a `requestUpdate`. With this approach, technically it would be possible to render the component without an `href`, and I'm not sure that's a valid scenario.  Or, `this._entity` gets out of sync with the `href`.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: Technically they can be imported and referenced in the static resources, however each import would be a separate request, and `LocalizeStaticMixin` relies on them to resolve missing translations, so this would not be advised.  @dlockhart and I had a lengthly conv about this for `core`, and @m6jones did some work previously to aggregate and convert Serge resources, along with some work-in-progress loader for them.  I think what we want for these components is up for conversation, but there seems to be desire to enable Serge integration.

PR #385, opened 2019-09-24
US105733 - remove search/filter criteria on initial load

AlexBedley:
Problem: There was a bug in QE with filtering and returing to QE (invisible filter) https://d2l.slack.com/archives/CBTK3JB6D/p1568743954028400
Solution: As a slight workaround, let's ship a long sought-after feature from the designers where returning to QE after evaluation removes your criteria if that criteria causes your list to be empty

One thing I wasn't a huge fan of was that the diff for both files are basically identical. I considered putting this code into a behaviour but it would need to inherit from search and filter and I didn't have the appetite for such a large refactor right now. Wondering what your thoughts on this are.


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -235,6 +239,13 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			return;
 		}
 		this._loading = true;
+
+		if (this._initialLoad && entity.hasClass('empty')) {
+			this._clearFilterAndSearch();
+			this._initialLoad = false;
+			return;
```
AlexBedley: The reason we `return` here is semantically, we don't want this to count as a load (outside of setting `_initialLoad`), so we don't want to run any of the cleanup stuff we do. This ensures that the loading spinner is displayed the whole time so no UI is briefly flashed on the screen or anything.

AlexBedley: aaaaand I just thought of a bug where if you truly have no items and no criteria it will be stuck loading. I will cover that case

AlexBedley: Ok the most recent commit should fix that case

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -235,6 +239,16 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			return;
 		}
 		this._loading = true;
+
+		if (this._initialLoad &&
```
maboelsoudd2l: why don't we use `ready()`? it runs once on initial load

AlexBedley: we don't have `entity` when `ready` is called

maboelsoudd2l:
i encountered the same idea of using a behaviour. So i thought of creating a "ClearBehaviour"
 Thing is that most likely since it clears both filters and search so I thought that the search-behaviour and filter-behaviour would inherit the "ClearBehaviour". but then we introduce a hidden dependency, the `hidden` property. since that is bad practice I thought "oh ill just implement it in Activities and submissions". The problem then becomes that "ClearBehaviour" would need filters to clear... thus another hidden dependency...

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: second question, what happened to "display currently searched term on initial load" PR that we merged this morning? Why are we clearing the search now on initial load?

AlexBedley: the 2 features live in harmony. We only clear the search if the list is empty

maboelsoudd2l:
I wish early on in the project we created a base class for Activities and Submissions pages. It would be the perfect place to put the clearing logic without fear of inconsistency, as well as a ton of other stuff. (maybe we'll revisit this conversation if someone figures out time-travel)

AlexBedley:
@CarolBusman I added some tests as per our offline conversation

PR #408, opened 2019-09-27
US110193 empty name validation

wongvincent:
- Prevents a save name request from occurring if the name field is empty
- Add aria-invalid red border on the input
- Show tooltip with "Name is required" error. Made it generic so that it can also handle a "save name" error. d2l-tooltip adds an aria-live on the input element, so we don't need to do anything to get it to announce the tooltip.
- Make name field full width

I took a look at the Rubrics error handling behavior, and considered copying it, but it seems overkill as d2l-tooltip does what we want without needing to implement extra functions. Such as getting the tooltip to add/remove "force-show", show/hide the bubble, fire iron-announce, etc.


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -90,8 +101,16 @@ class AssignmentEditorDetail extends SirenFetchMixinLit(EntityMixinLit(LocalizeM
 					@input="${this._saveNameOnInput}"
 					aria-label="${this.localize('name')}"
 					?disabled="${!super._entity.canEditName()}"
+					aria-invalid="${this._nameError ? 'true' : ''}"
 					prevent-submit>
 				</d2l-input-text>
+				<d2l-tooltip
```
mdgbayly: I think we had lots of perf issues with tooltips in rubrics, where including the tooltip even when there were no errors causes a big impact on page load time. Maybe check what we did with rubrics to prevent them from being added to the DOM until required.

Unless the issue has been fixed...

I do think we'll probably want to come up with some more generalized error handling code but we can probably refactor to that over time as we add more stuff.

wongvincent: > including the tooltip even when there were no errors causes a big impact on page load time.
Good point, I didn't consider that. And yes, it looks like you made them conditional to work around performance issues https://github.com/Brightspace/d2l-rubric/commit/2ee0dfa5bbcc8896424819808d77fa0240d3e439

I'll make it conditional. The tooltip doesn't automatically show on render, and is the reason why we have those `bubble.show()` calls in the rubric error handling behavior. I'll need to do something similar here.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -108,13 +108,15 @@ class AssignmentEditorDetail extends SirenFetchMixinLit(EntityMixinLit(LocalizeM
...
 					prevent-submit>
 				</d2l-input-text>
-				<d2l-tooltip
-					id="name-tooltip"
-					for="assignment-name"
-					position="bottom"
-					?showing="${this._nameError}">
-					${this._nameError}
-				</d2l-tooltip>
+				${this._nameError ? html`
```
mdgbayly: You can probably keep the main template cleaner by using a technique like this one shown here under `Conditionals with if statements`
https://lit-html.polymer-project.org/guide/writing-templates#conditional-templates

PR #413, opened 2019-10-01
US110688 - fix lazy loading regression

AlexBedley:
Problem: QE was behaving weirdly when you returned back to QE after doing evaluation and you had search/filter criteria. It would filter the list based on the criteria but not display that you had criteria active
Solution: After digging in a bit, I saw 2 problems:

1. after loading a page, we would set the href to `''` when we toggled and set it back to `href` when it was active again. However, this was causing the page to load again. The fix here is to keep `href` the same if `href` has already been loaded
2. There was a small period of time where the 2 variables that controlled which page was active did not match. Rather than handle this case specifically, I just unified the variables.

So now `_showActivitiesView` no longer exists, and the page is solely controlled by `toggleState`.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -7,6 +7,9 @@ import './d2l-quick-eval-activities.js';
 import './d2l-quick-eval-submissions.js';
 import './behaviors/d2l-quick-eval-telemetry-behavior.js';
 
+const submissions = 'submissions';
+const activities = 'activities';
+
 /**
  * @customElement
  * @polymer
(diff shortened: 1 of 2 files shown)


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -92,46 +91,51 @@ class D2LQuickEval extends
...
+	_toggleView(e) {
+		this.toggleState = e.detail.view;
 		this.logViewQuickEvalEvent(e.detail.view);
 	}
 
-	_computeLazySubmissionsHref(href, state) {
-		if (!state) {
-			return href;
+	_computeLazySubmissionsHref(submissionsHref, toggleState) {
+		if (toggleState === submissions || this._lazySubmissionsHref) {
```
AlexBedley: This is fix 1 `|| this._lazySubmissionsHref`

AlexBedley:
I will add some tests to Carol's suite to cover this specific scenario (loading with criteria and then switching views)

AlexBedley:
I tried adding tests for that specific case, but I forgot that you can't control the output based on query strings (for example, `data/unassessedActivities.json?collectionSearch=hello` is the same as `data/unassessedActivities.json`)

However, I ensured that the hrefs don't change after they've already been loaded. This means that quick-eval leaves the subpages alone when it comes to `href`, which is good enough for me.

PR #415, opened 2019-10-01
US110605 Implement getLocalizeResources

ryantmer:
This is a Serge-friendly, dynamically-loaded, fallback-enabled i10n solution that allows us to generate complete translation sets on load.

Advantages of this approach are that we don't need a build step and can instead work directly with the JSON files that Serge works with. When the element is loaded, it will load all the languages and locales given to `getLocalizeResources` and builds a fallbacked map of langterms and translations, with higher-preference translations being loaded last (and therefore overriding lower-preference translations). This approach has been tested to work with supported and unsupported localized and non-localized languages.

The primary disadvantage of this approach is that it is more work on page load - however, getting around this would require a build step, which is something we may do eventually. In addition, we're also sending more data than required, since the Serge files also contain the `context` clues used for translation purposes.


ryantmer:
The lint failure I'm not going to address yet - this implementation of `getLocalizeResources` should be put into a mixin or something so that we can use it across multiple components, not left where it is; I'll reorganize things once the approach has been vetted a bit.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -30,23 +32,55 @@ class AssignmentEditorDetail extends SirenFetchMixinLit(EntityMixinLit(LocalizeM
...
-			if (translations && translations.assignment) {
+			return langterms;
+		}
+
+		let langterms = await getLangterms({}, 'en');
+
+		let localeCode;
+		for (localeCode of langs) {
+			if (localeCode === 'en') {
 				return {
```
mdgbayly: Is this early return shortcutting potential `en` lang variants. Not sure if we ever have any but we seem to have 3 en options in the LMS (us, ca, gb).

ryantmer: `langs` in this case is an ordered list of locale codes, so if you're set to `en-gb`, then `langs = ["en-gb", "en"]`, meaning we won't shortcut out here until we're on the last option anyway. (Really this isn't super necessary, was just a shortcut for the always-least-priority-case-of-"en".)

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -30,23 +32,55 @@ class AssignmentEditorDetail extends SirenFetchMixinLit(EntityMixinLit(LocalizeM
...
-					language: lang,
-					resources: translations.assignment
+					language: localeCode,
+					resources: langterms
 				};
 			}
+
+			const languageCode = localeCode.split('-')[0];
+
+			if (languageCode === localeCode && supportedLanguages.indexOf(languageCode) > -1) {
```
mdgbayly: Doesn't the `LocalizeMixin` already do this language splitting/generation for us?
https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.js#L137

So I was assuming we could just attempt to load language files for each of the exact values in the `langs` property passed to the function, and do this concurrently e.g. using something like `Promise.all`.

Then build up the final language object by iteratively processing the set of language objects returned so that the overrides win. Localize static mixin does this by processing the langs in reverse order because I think that's how they are ordered for precedence.

I'm kind of confused by why `LocalizeMixin` always pushes `en` and `en-us` onto the list, as that would seem like it would lead to duplicate processing, which if we are making network requests would mean unnecessary requests. But I could be misunderstanding what it is doing there.

https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.js#L160

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -30,23 +32,55 @@ class AssignmentEditorDetail extends SirenFetchMixinLit(EntityMixinLit(LocalizeM
...
-			switch (lang) {
-				case 'en':
-					translations = await import('./lang/en.js');
-					break;
+		async function getLangterms(langterms, locale) {
+			const response = await fetch(`./lang/${locale}.json`);
+			const json = await response.json();
+
+			for (const langterm in json) {
+				langterms[langterm] = json[langterm].translation;
```
dbatiste: It looks like we'd be sending the "context" values down to the client as well ya?  Ideally we would not be sending that down to the client... it's really just for our translation workflow as far as I know.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: Well dang, so it does! I'll remove that from here in that case, which leaves exactly as you said - we still need to define and check `supportedLanguages` to avoid doing network calls for languages that we (the component) don't support. I didn't put any thought into where `langs` is coming from, but relying on it being a list of correctly-ordered values does make this logic a lot simpler - just need to do things in the reverse order as you said.

I am a little concerned about loading multiple language files concurrently on page load, but as discussed in standup, we can just do it that way and rework it if that proves to be a problem (although I suspect the answer there would be adding a build step, but we'll cross that bridge when/if we come to it).

`en` makes sense to me, but yeah `en-us` seems strange... Either way, with the addition of `supportedLanguages`, we won't even try to fetch anything that we know isn't supported, so not a biggie.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: But doesn't it still cause the language of the returned object to always be set to `en` even if the user is using some other language? Not sure what the language property on the returned object is used for. In `LocalizeStaticMixin` it looks like it's always set to the most specific language definition available, because they process 'en' first by processing the langs in reverse order.

ryantmer: Ah true, I suppose it could - pretty sure that based on your comment above, though, that this will be removed anyway. Update incoming, with much simpler logic!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: Yeah, that's how I understand this logic as well Martin.

In the case of `core`, we will have many components, each having their own `lang` folders.  I don't think we really want to be making multiple requests for the language files, per component.  That was the motivation behind the original build-script that aggregated them, and subsequently the motivation behind baking them into the core components.  Will HM components be similar in that regard... possibly several components, each having their own translation files?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: There wouldn't be a way to avoid that without the build step though, correct?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: Quick Eval is structured as a single lang file per language for the whole app, not per component - I suspect this is the same approach we'll be taking, as I imagine most/all of the components will be app-specific. Perhaps apps and reusable components should be treated differently due to this, where we'd want static resources for reusable components, but for whole-page experiences dynamically loading might make more sense?

mdgbayly: @dbatiste Regarding whether HM components will each have their own `lang` folders. It seems to be somewhat flexible currently. I think the main driver seems to be the extent to which components will be used together. e.g. `d2l-quick-eval` has a single `lang` folder for the many components that make up `d2l-quick-eval`. Whereas `d2l-activity-evaluation-icon` has it's own `lang` folder and only has a single lang term.

We were working on similar assumptions that we'd probably have a `lang` folder for all the generic `d2l-activity-editor` components and a `lang` folder for all the `d2l-activity-assignment-editor` components. So a similar approach to that being used by `quick-eval`.

So I was thinking from that point of view it's not too many requests to be making at load time, but I guess it will frequently be 4 if we always load `en`, `en-us`, `es`, `es-MX` e.g .for a Spanish Latin American user and they would likely be cached by the browser after the first ht.

But I could be convinced that this is a bad idea still....

And it may still be for the other reasons you give which I had overlooked - like having all the translator context downloaded too.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Yep, that is a good point!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Oops - sorry realized @ryantmer had already replied with similar comments - should have reloaded the page.

mdgbayly: @ryantmer Yeh, the question of whether reusable vs app components should be treated differently is an interesting question. I wonder if that distinction will always be clear.  And ideally we're just building components that can be used anywhere even if we don't expect them to be up front. But maybe that is too idealistic...

And if you have lots of reusable components each with their own small set of static resources for every language we support, does there come a point where it still makes sense to load only the current users lang terms even if that means an extra 30-60-90++ HTTP calls? (given they will be cached after first load...) ... maybe not... what to do...

ryantmer: Yeah, totally valid point... I'm hesitant to make final calls about that before we have a better idea of how ACE develops, really - I do agree that ultimately we want to aim for standalone components, but I think generally a good portion of them will end up being single-use, just since they're too specific to our use case. I suppose we have a bit of a unique opportunity here, though, in that ACE will eventually support multiple types of activities, which will probably result in a lot _more_ reuse than I'd expect. At that point, is it "components used once, we load dynamically, components used in multiple places have static langterms"? Sounds reasonable to me at this point, but overall still not totally sure!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-localize-mixin.js
```diff
@@ -0,0 +1,39 @@
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
+import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
+
+const SUPPORTED_LANGUAGES = ['en', 'fr'];
+const cache = {};
+const baseUrl = import.meta.url;
+
+export const ACELocalizeMixin = superclass => class extends LocalizeMixin(superclass) {
+	static async getLocalizeResources(langs) {
```
dbatiste: Seems like this could just be an ES6 function import ([like Maya's was](https://github.com/BrightspaceUI/core/blob/eb50db43fdda3d6f67c03ad91fb2272d4bc6e14e/tools/localization-load.js)).

Normally I'd only make something a mixin if it needs to store/access element state, or participate in the element lifecycle.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -1,78 +1,58 @@
...
+		// assignmentActivityUsage.onAssignmentChange((assignment) => {
+		// 	this._name = assignment.name();
+		// });
+	}
+
+	render() {
+		return html`
+			<d2l-activity-editor>
+				<d2l-activity-assignment-editor-detail
+					.href="${this._assignmentHref}"
```
dbatiste: It is interesting that the `href` and `token` are being bound to properties instead of attributes.  Right or wrong, I haven't found myself binding directly to properties very often.

ryantmer: Admittedly I didn't actually change this part, just whitespace - but it does seem correct that these could probably just be `href/token` (works in the demo, at least!)

mdgbayly: i think i decided to bind to props instead of attributes because of some issues i was having where a null/undefined value for the prop from the ancestor component was being translated to the string value `"undefined"` rather than the actual JS undefined and that was causing issues in some of the `siren-sdk` stuff where it checked if `href` was defined before making API calls. So we were making API calls to things like `/api/someroute/undefined`.

Maybe there was a better way to fix that.

components/d2l-activity-editor/d2l-activity-assignment-editor/localization.js
```diff
@@ -0,0 +1,36 @@
...
+		const request = fetch(url).then(res => res.json());
+		cache[url] = request;
+		return request;
+	});
+
+	const responses = await Promise.all(requests);
+
+	const langterms = {};
+	responses.forEach(language => {
+		for (const langterm in language) {
```
mdgbayly: Maybe this is micro-optimizing, but if we were to extract the lang terms from the serge structure as part of the fetch handling, we avoid storing all the translator context, and we can probably then just use `Object.assign` rather than a loop to construct the language object with the overrides?

ryantmer: Good call! Mostly that second part I find appealing, as it's more explicitly an overriding action.

PR #417, opened 2019-10-02
Fix New Travis Config

DiljotSG:
Fix for the Travis Config causing the version number to be bumped twice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,7 @@ jobs:
         npm run test:diff || travis_terminate 1;
       fi
   - stage: update-version
-    script:
-    - |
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,8 +25,10 @@ jobs:
         echo "Running visual difference tests...";
         npm run test:diff || travis_terminate 1;
       fi
-after_success:
-- frauci-update-version && export TRAVIS_TAG=$(frauci-get-version)
+  - stage: update-version
+    script:
+    - |
```
AlexBedley: what's with the `|` ? I thought that was just formatting to make the `if` work above

DiljotSG: True, we definitely probably don't need that, I think it's if you want a multi-line script.

PR #418, opened 2019-10-02
basic activity admin list

jlai-d2l:


components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -1,8 +1,18 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import 'd2l-common/components/d2l-hm-search/d2l-hm-search.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
+import 'd2l-button/d2l-button.js';
 
 class AdminList extends LitElement {
 	static get properties() {
-		return {};
+		return {
+			header: {
```
AllanKerr: Can we change `header` to `title` since the header contains the title and _create new_ button?

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -13,12 +23,34 @@ class AdminList extends LitElement {
...
 		`;
 	}
 
 	render() {
 		return html`
-			<div>D2L Activity Admin List</div>
+		<div class="d2l-activity-admin-list-container">
+			<h1>${this.header}</h1>
+			<d2l-button primary>Create Learning Path</d2l-button>
+			<d2l-hm-search
```
AllanKerr: Search and filter are more-so stretch goals so we want to leave them off for now until we've got the rest of the list and admin page working as intended.

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -13,12 +23,34 @@ class AdminList extends LitElement {
 			:host([hidden]) {
 				display: none;
 			}
+			d2l-hm-search {
+				display: inline-block;
+				width: 250px;
+				flex: 1;
+			}
+			d2l-button {
+				float: right;
```
AllanKerr: Rather than using `float` I would recommend adding a header `div` with `display: flex` and `justify-content: space-between` that contains both the button and title.

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -1,8 +1,18 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import 'd2l-common/components/d2l-hm-search/d2l-hm-search.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
```
dbatiste: It looks like typography is not used in this component.  Can we remove this import?

Related, if we need shared styles from typography in a LitElement, we should import them from `@brightspace-ui/core`, include them in the static styles for the component, and use the class name in the template where needed.  The meter component has [an example](https://github.com/BrightspaceUI/core/blob/master/components/meter/meter-radial.js) of this, but our buttons do as well.

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -1,8 +1,18 @@
...
+import 'd2l-button/d2l-button.js';
```
dbatiste: We could import from core instead.
```
import '@brightspace-ui/core/components/button/button.js';
```

components/d2l-activity-admin-list/d2l-activity-admin-list.js
dbatiste: We should be careful using `title`, since it's a native attribute on HTML elements that displays a native tooltip when hovering anywhere over the element.  `title-text`, or something else may be better.

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -13,12 +23,23 @@ class AdminList extends LitElement {
...
+				align-items: baseline;
+			}
 		`;
 	}
 
 	render() {
 		return html`
-			<div>D2L Activity Admin List</div>
+		<div class="d2l-activity-admin-list-container">
+			<h1>${this.title}</h1>
```
dbatiste: This is using title, but the property above still says `header`.  Maybe only half the change was made?

It's technically possible to have more than one `h1` on the page, but not common.  Are we sure that's the heading level that we want?  Aside from the semantic level, we _probably_ want to import from typography (see previous comment) and add the appropriate heading level style class here

components/d2l-activity-admin-list/d2l-activity-admin-list.js
svanherk: The `d2l-hm-search` also needs to be updated to the latest version of the UI component, which has been on my list for forever but keeps getting de-prioritized.  Would probably be easier to do before you need it, when are you thinking?

svanherk: Whoops, nvm!  I was thinking `d2l-hm-filter`!  No upcoming update to `d2l-hm-search`, you can ignore this 😄

components/d2l-activity-admin-list/d2l-activity-admin-list.js
```diff
@@ -1,24 +1,44 @@
...
+				align-items: center;
+			}
+		` ];
 	}
 
 	render() {
 		return html`
-			<div>D2L Activity Admin List</div>
+		<div class="d2l-activity-admin-list-container">
+			<div class="d2l-heading-1">${this.titleText}</div>
```
AllanKerr: We'll still want this on an `h` tag. From the designs I believe an `h1` should be appropriate.

PR #419, opened 2019-10-02
US110194 Add autosave component

ryantmer:
When the assignment name is updated, this will show "Saving..." as the network request occurs. When the network request successfully completes, it shows "✔️ Saved" after a few moments (the delay is set by the component itself, and is not configurable). If the network request fails, the component hides itself (behaviour that is also defined by the component itself, and is not configurable).

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -1,4 +1,5 @@
 import 'd2l-inputs/d2l-input-text.js';
+import 'd2l-save-status/d2l-save-status.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
@@ -59,9 +60,18 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 
 	_saveName(value) {
 		if (super._entity.canEditName()) {
+			const saveStatus = this.shadowRoot.querySelector('#save-status');
+			saveStatus.start();
+
 			const action = super._entity.getSaveNameAction();
 			const fields = [{ 'name': 'name', 'value': value }];
-			this._performSirenAction(action, fields);
+			this._performSirenAction(action, fields)
+				.then(() => {
+					saveStatus.end();
+				})
+				.catch(() => {
+					saveStatus.error();
+				});
 		}
 	}
 
@@ -99,9 +109,16 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 		}
 	}
 
+	_getSaveStatus() {
+		return html`
+			<d2l-save-status id="save-status"></d2l-save-status>
+		`;
+	}
+
 	render() {
 		return html`
 			<div id="assignment-name-container">
+				${this._getSaveStatus()}
 				<label class="d2l-label-text" for="assignment-name">${this.localize('name')}*</label>
 				<d2l-input-text
 					id="assignment-name"
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -1,5 +1,4 @@
 import 'd2l-inputs/d2l-input-text.js';
-import 'd2l-save-status/d2l-save-status.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
@@ -39,6 +38,17 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 	constructor() {
 		super();
 		this._setEntityType(AssignmentEntity);
+
+		const redispatch = e => {
+			this.dispatchEvent(new CustomEvent('d2l-activity-editor-autosave', {
+				detail: { message: e.type },
+				bubbles: true,
+				composed: true
+			}));
+		};
+		this.addEventListener('d2l-siren-entity-save-start', redispatch);
+		this.addEventListener('d2l-siren-entity-save-end', redispatch);
+		this.addEventListener('d2l-siren-entity-save-error', redispatch);
 	}
 
 	set _entity(entity) {
@@ -60,18 +70,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 
 	_saveName(value) {
 		if (super._entity.canEditName()) {
-			const saveStatus = this.shadowRoot.querySelector('#save-status');
-			saveStatus.start();
-
 			const action = super._entity.getSaveNameAction();
 			const fields = [{ 'name': 'name', 'value': value }];
-			this._performSirenAction(action, fields)
-				.then(() => {
-					saveStatus.end();
-				})
-				.catch(() => {
-					saveStatus.error();
-				});
+			this._performSirenAction(action, fields);
 		}
 	}
 
@@ -109,16 +110,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 		}
 	}
 
-	_getSaveStatus() {
-		return html`
-			<d2l-save-status id="save-status"></d2l-save-status>
-		`;
-	}
-
 	render() {
 		return html`
 			<div id="assignment-name-container">
-				${this._getSaveStatus()}
 				<label class="d2l-label-text" for="assignment-name">${this.localize('name')}*</label>
 				<d2l-input-text
 					id="assignment-name"
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -38,17 +38,6 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 	constructor() {
 		super();
 		this._setEntityType(AssignmentEntity);
-
-		const redispatch = e => {
-			this.dispatchEvent(new CustomEvent('d2l-activity-editor-autosave', {
-				detail: { message: e.type },
-				bubbles: true,
-				composed: true
-			}));
-		};
-		this.addEventListener('d2l-siren-entity-save-start', redispatch);
-		this.addEventListener('d2l-siren-entity-save-end', redispatch);
-		this.addEventListener('d2l-siren-entity-save-error', redispatch);
 	}
 
 	set _entity(entity) {
--- a/components/d2l-activity-editor/d2l-activity-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-editor.js
@@ -24,35 +24,25 @@ class ActivityEditor extends LitElement {
 		`;
 	}
 
-	constructor() {
-		super();
-		this.addEventListener('d2l-activity-editor-autosave', e => {
-			const saveStatus = this.shadowRoot.querySelector('#save-status');
-			switch(e.detail.message) {
-				case 'd2l-siren-entity-save-start':
-					saveStatus.start();
-					break;
-				case 'd2l-siren-entity-save-end':
-					saveStatus.end();
-					break;
-				case 'd2l-siren-entity-save-error':
-					saveStatus.error();
-					break;
-			}
-		});
-	}
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
 
-	_getSaveStatus() {
-		return html`
-			<d2l-save-status
-				id="save-status">
-			</d2l-save-status>
-		`;
+		this.addEventListener('d2l-siren-entity-save-start', () => {
+			this.shadowRoot.querySelector('#save-status').start();
+		});
+		this.addEventListener('d2l-siren-entity-save-end', () => {
+			this.shadowRoot.querySelector('#save-status').end();
+		});
+		this.addEventListener('d2l-siren-entity-save-error', () => {
+			this.shadowRoot.querySelector('#save-status').error();
+		});
 	}
 
 	render() {
 		return html`
-			${this._getSaveStatus()}
+			<d2l-save-status
+				id="save-status">
+			</d2l-save-status>
 			<div ?hidden="${!this.loading}">Loading ...</div>
 			<div ?hidden="${this.loading}">
 				<slot name="editor"></slot>
--- a/components/d2l-activity-editor/d2l-activity-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-editor.js
@@ -1,3 +1,4 @@
+import 'd2l-save-status/d2l-save-status.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 
 class ActivityEditor extends LitElement {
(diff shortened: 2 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -74,9 +75,18 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
...
+
 			const action = super._entity.getSaveNameAction();
 			const fields = [{ 'name': 'name', 'value': value }];
-			this._performSirenAction(action, fields);
+			this._performSirenAction(action, fields)
+				.then(() => {
+					saveStatus.end();
+				})
+				.catch(() => {
+					saveStatus.error();
```
ryantmer: We should probably do something nicer here, although unclear what - the `d2l-save-status` component doesn't appear to support showing an error or anything, so we probably would need to surface the error somewhere else.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -74,9 +75,18 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 
 	_saveName(value) {
 		if (super._entity.canEditName()) {
+			const saveStatus = this.shadowRoot.querySelector('#save-status');
+			saveStatus.start();
+
 			const action = super._entity.getSaveNameAction();
 			const fields = [{ 'name': 'name', 'value': value }];
-			this._performSirenAction(action, fields);
+			this._performSirenAction(action, fields)
```
mdgbayly: We might want to hook into this in a higher level (ancestor) component so that we handle all actions.
Possibly in `d2l-activity-editor` if we assume that all editing components will be slotted into that component at some point.

I think there is still some debate as to where exactly the `Saving...` label will finally show up. Some designs have it on a floating footer which might pose us some more challenges. But for now I think it would be fine to add to the `d2l-activity-editor` component.

The `performSirenAction` stuff generates events that we can listen for to drive the save status component.

This is how we did it in the rubric editor

https://github.com/Brightspace/d2l-rubric/blob/6e08bc9d4fb98fc3a48b99f9741aed1b396d9d0c/d2l-rubric.js#L329

And yes, currently I think the assumption is that errors will be handled elsewhere. Maybe when we first wrote the `save-status` component there was a potential for it showing save errors but maybe we ditched that design.

ryantmer: The `d2l-siren-entity-save-*` events don't currently bubble, so unfortunately they can't escape the `d2l-activity-assignment-editor` currently. My initial intuition is to make them bubble, but I suspect this use case is the exception rather than the rule? If so, then I'll just re-dispatch them from `d2l-activity-assignment-editor` to `d2l-activity-editor`.

(Mostly just posting this to explain the thought process behind the upcoming commit 😅)

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -54,6 +53,17 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
...
 		this._setEntityType(AssignmentEntity);
+
+		const redispatch = e => {
+			this.dispatchEvent(new CustomEvent('d2l-activity-editor-autosave', {
+				detail: { message: e.type },
+				bubbles: true,
+				composed: true
+			}));
+		};
+		this.addEventListener('d2l-siren-entity-save-start', redispatch);
```
dbatiste: Just curious if these events can fire before this component is connected?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -124,16 +125,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 		}
 	}
 
-	_getSaveStatus() {
-		return html`
-			<d2l-save-status id="save-status"></d2l-save-status>
-		`;
-	}
-
 	render() {
```
dbatiste: I'm not against having it outside, but could the template for `_getNameTooltip` be include inside?  

As we continue building out components using LitElement, we will undoubtedly see lots of factoring of template creation.  It may be good for readability sake to adopt a convention for naming template helpers (`_getNameTooltip`).  

My inclination is that they should be named like `_renderNameTooltip` or as in Lit-Element documentation, `nameTooltipTemplate`. The `_getNameTooltip` leads me to think a fetch will be performed.  Aside, `d2l-icon`'s uses the "get", but it actually does do an async request for the template. Anyway, as mentioned, such a convention may make code easier to understand, even if it's not followed 100% of the time.

ryantmer: No longer modified in this PR, but the same comment applies to the `_getSaveStatus` method I added - I'll merge it back in, since it's a pretty simple template anyway (and I agree, `get` implies a fetch).

components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -1,31 +1,63 @@
...
+	_getSaveStatus() {
+		return html`
+			<d2l-save-status
+				id="save-status">
+			</d2l-save-status>
+		`;
+	}
+
+	render() {
+		return html`
```
dbatiste: I think it should be possible to yield cleaner DOM in this case.  This is a really nice aspect of Lit-Element, in contrast to Polymer's `dom-if` that would creates lots of DOM that wasn't really wanted.  

```
return html`
   ${this._getSaveStatus()}
   ${this.loading ? html`<div>Loading...</div>` : html`<div><slot name="editor"></slot></div>`}
`;
```
It could be factored out if readability became a concern, but unless we really need the hidden DOM, I'd be inclined to avoid it.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: I don't _think_ they could, since they're triggered by the PATCH operation to update the name, but I'm not sure - is the concern that somebody would `document.createElement('d2l-activity-assignment-editor-detail')`, then do some more JS fiddling to modify the name that way, all without actually sticking the element in the DOM?

components/d2l-activity-editor/d2l-activity-editor.js
ryantmer: Much nicer! I hadn't actually modified that part of the DOM here (line endings causing noise here), but I'll change that over.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: The Lit-Element guide has a great short [section](https://lit-element.polymer-project.org/guide/events) on this.  Basically we usually defer event wire-up as late as possible as a perf optimization. When wiring up events to elements in the template, the `@someEvent` directive is always best, since it leaves it to Lit to wire-up at the appropriate time, and it also follows the principle of not modifying the DOM outside of `render`. Anyway, I recommend taking a quick peek.

ryantmer: Admittedly I tried `@d2l-siren-entity-save-start` in both places, but neither was working 😕 Is the `@event` one preferred only because it leaves the decision up to Lit? The options are listed in the order of `@event, constructor, firstUpdated, connectedCallback`, which I thought meant `@event` would actually be adding it as early as possible, and `connectedCallback` is the ideal/latest-possible - but it sounds like `@event` is kind of a special case outside the ordering of `constructor/firstUpdated/connectedCallback`?

dbatiste: The `@event` will happen after `connectCallback`.  In fact, if one tries to query for a template's element in `connectedCallback`, it won't find it initially because `render` has not been called yet.  The `host` is "connected", but the internal local DOM has not be rendered yet.  In addition, when using the `@event` directive, you don't need to worry about removing the event handlers.

But `@event` doesn't help us when we need to wire-up to `this` (the host element).  In such case, it would be preferred to defer the work as late as possible, so not the `constructor`.  It should be `firstUpdated` (after the first time `render` operation) or `connectedCallback`.  If using `connectedCallback`, it should be removed in `disconnectedCallback` to ensure we don't get duplicate handlers if the element is disconnected and reconnected.  Same is true if we need to listen for events outside of the element, for instance the window or body.

(their doc lists the `constructor` 2nd, but they also indicate to defer, and that wiring-up in the constructor is an option if necessary).

ryantmer: Ah, okay - the docs say to wire up events "as late as possible", but I guess to figure out what the ordering is, the Lifecycle page is really where you need to do it. Thanks for the clarification! I'll give the `@event` approach another shot today, since maybe I was just not seeing something obvious yesterday, but if not then I'll move it to `connectedCallback`.

dbatiste: To summarize that, I think if it's for an element in the local template, wire-up should be done with `@event`, always.  Otherwise, the preference is `firstUpdated`, followed by `connectedCallback`, and then `constructor`.

ryantmer: Oh whoops, okay - `firstUpdated` in that case! The order they give in the docs is really no order whatsoever, it turns out 😬

dbatiste: I'm not sure if `SirenFetchMixinLit` or `EntityMixinLit` have something that would influence this.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: > My initial intuition is to make them bubble, but I suspect this use case is the exception rather than the rule? 

@ryantmer Actually bubbling is the norm rather than the exception :-), so yes, I think we should be making them bubble and we should not need to be redispatching them here.

I think this must have broken when the Siren action implementation was converted from being a Polymer behaviour to being a Lit mixin. The behaviour used `fire` and the Lit Mixin is is using 
`dispatchEvent` but it isn't setting the events to bubble which it should be.

https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/mixin/siren-fetch-mixin-lit.js#L84

So that would be the correct place to fix this issue.

ryantmer: Hooray! I'll fix them there in that case - actually kind of a big relief, because I didn't like the redispatching stuff I added haha. I'll get a PR up there first, then will update things here.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Yeh I think firstUpdated would make sense here given that we don't really have an element to which we can attach the @event handler to, as these events could come from any child element.

mdgbayly: Well, actually, as mentioned in my other comment, we don't need to be adding a listener here, as the events should be bubbling, but i think using `firstUpdated` would make sense in the `d2l-activity-editor`?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: I think it will also by default be `composed: false`.

components/d2l-activity-editor/d2l-activity-editor.js
mdgbayly: I'm not clear if using the approach suggested by @dbatiste will actually work here.
Basically the way the loading indicator works is that the components on the negative side of the loading conditional need to be being rendered whilst we are showing a loading skeleton. This is because it is the rendering of those hidden components that generate the events that the top level component will ultimately use to flip the loading property on the `d2l-activity-editor` component.

Possibly we could clean it up a bit by using a simple conditional on the bit of DOM that shows the loading skeleton/message so that it is removed from the DOM once we are no longer loading, but I think we may have to keep the `hidden` attribute that is guarding the display of the slotted content, unless you guys can think of a better way of achieving this.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Yeh, as Dave mentions we will need to set `composed` to true. I don't recall what the default value is for `bubbles` and don't seem to be able to find it in MDN, but we'll need it to do both i guess.

ryantmer: Both `bubbles` and `composed` appear to default to false, at least according to https://lit-element.polymer-project.org/guide/events#custom-events. So, both it shall be! (see https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/69)

components/d2l-activity-editor/d2l-activity-editor.js
ryantmer: I've tried the conditional given above and it does work - but now you've raised a good point and I'm not sure why, as it seems like the `<slot>` wouldn't get rendered until `loading == true`, but `loading == true` doesn't happen until the contents of the (unrendered?) `<slot>` say it should 🤔

In that case, I won't actually change this here (since I didn't touch it anyway), until we can figure that out.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: @dbatiste When you say
> but could the template for `_getNameTooltip` be include inside

... do you mean rather than factoring into a helper function/property getter?

The reason I suggested to Vincent that we put that error tooltip in a helper template outside the main template was because I think we are going to end up with lots of 'error' tooltips as we add more stuff to our assignment editor, and to me they seem to clutter up the main point of the template which is the edit controls rather than error tooltips.

The convention I've been adopting in some other components I'm working on is to use a getter property for the template if it needs access to `this`. It just looks slightly cleaner to me than having a function call when you are composing templates.

But yeh, I think adopting a convention like `nameTooltipTemplate` makes sense. 

Are you guys using methods or getters when you compose templates?

components/d2l-activity-editor/d2l-activity-editor.js
mdgbayly: It's possible it's because I think in the initial implementation I have not forced `loading` (or `_hasPendingChildren` in the `PendingContainerMixin` which is what drives this), to `true` initially. So maybe it is rendering the slotted content initially to get the ball rolling before it actually flips the `loading` flag to true, when the first child indicates it is pending.

But I had been considering that I might need to provide a way to allow a container to initiate things in the `loading` state in case we see a flash of content then loading. But I wasn't really seeing that with this initial setup so I decided to just leave it until we have a bit more experience with using it.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: Yeah that's what I was thinking, and yep, fair point to pull out to make things cleaner with the anticipation of more to come.

So far I think we've been using methods for those helper templates - I probably followed suit with an example in their documentation somewhere, but they do use getters in their example [here](https://lit-element.polymer-project.org/guide/templates#compose-a-template-from-other-templates).  That being said, sometimes the helper might need an arg passed to it, in which case the method would be needed. In some cases it may be possible to avoid passing the arg by moving some of the logic.

PR #420, opened 2019-10-03
US109858 - Switch to New Confirmation Dialog Provided by Core

DiljotSG:


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -364,33 +371,23 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
-							const evaluationStatusHref = this.getEvaluationStatusHref(evalStatusEntity);
-							const publishedActivity = this._updateEvaluationStatus(evaluationStatusHref, evalStatusEntity);
-
-							if (publishedActivity) {
-								this.push('_publishAllToasts', { activityName: publishedActivity.activityName });
-							}
-						});
-				}
+		this._publishAllDialogMessage = evt.detail.confirmMessage;
+		const dialog = this.shadowRoot.querySelector('d2l-dialog-confirm');
```
AlexBedley: In general, we want to avoid querying our own `shadowRoot`. I think this is more a matter of style/design consistency rather than actual performance.

We can accomplish that here by binding `dialog.opened` in the html template to a property (maybe `_dialogOpen`) and just setting `this._dialogOpen` here instead.

We can also attach the event listener in the html by binding this listener to a `on-d2l-dialog-confirm` attribute on the dialog

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -171,6 +171,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				on-d2l-quick-eval-activity-view-evaluate-all="_navigateEvaluateAll"
 				>
 			</d2l-quick-eval-activities-list>
+			<d2l-dialog-confirm title-text="[[localize('confirmation')]]" text="[[_publishAllDialogMessage]]">
+			<d2l-button slot="footer" primary dialog-action="yes">[[localize('yes')]]</d2l-button>
+			<d2l-button slot="footer" dialog-action="no">[[localize('no')]]</d2l-button>
```
AlexBedley: nit: these 2 d2l-buttons should be indented by 1

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -364,33 +371,23 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
-							if (publishedActivity) {
-								this.push('_publishAllToasts', { activityName: publishedActivity.activityName });
-							}
-						});
-				}
+		this._publishAllDialogMessage = evt.detail.confirmMessage;
+		const dialog = this.shadowRoot.querySelector('d2l-dialog-confirm');
+		dialog.opened = true;
+
+		dialog.addEventListener('d2l-dialog-close', (result) => {
```
dbatiste: It is great to be using the `d2l-dialog-close` event, since it will ensure you are aware of the dialog closing no matter way (ex. if the user presses `escape` key).  Typically, we should be careful to avoid duplicate handlers for event(s) - i.e. if the same dialog is opened a 2nd time.  Another good option is to use the `open` method that will return you a promise - it will take care of listening for this event, etc.  Either approach is fine, but if you wire up to the event yourself, it's something to be careful of. 

```
const result = await dialog.open();
...
```
or 
```
dialog.open().then((result) => {
   ...
});
```

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -171,6 +171,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				on-d2l-quick-eval-activity-view-evaluate-all="_navigateEvaluateAll"
 				>
 			</d2l-quick-eval-activities-list>
+			<d2l-dialog-confirm title-text="[[localize('confirmation')]]" text="[[_publishAllDialogMessage]]">
+			<d2l-button slot="footer" primary dialog-action="yes">[[localize('yes')]]</d2l-button>
```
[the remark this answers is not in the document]
DiljotSG: That would be great IMO!

components/d2l-quick-eval/d2l-quick-eval-activities.js
DiljotSG: This would be nice, but as @AlexBedley mentioned we want to avoid grabbing elements from ShadowRoot, and that would be the only way for me to currently call the open method on the dialog.

dbatiste: My personal view is that it's fine to access elements in the component local DOM scope.  What is typically not good is to access DOM outside the components scope of ownership.

DiljotSG: @dbatiste Does the `open.then(...` get called after the dialog is closed?

dbatiste: Correct.  The `open` method will return a promise for the action result when the dialog is closed, regardless of whether it was closed as a result of clicking on workflow button (marked up with `dialog-action`), the user presses `escape`, or in the case of the generic dialog, if they click the `[x]` button in the top right corner.

If the user presses `escape` or the `[x]` button, the action will be `abort`.  If the user clicks a button marked up with `dialog-action`, the action will be the value of that attribute.

DiljotSG: @dbatiste got it thank you!

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -379,33 +387,21 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
-							const evaluationStatusHref = this.getEvaluationStatusHref(evalStatusEntity);
-							const publishedActivity = this._updateEvaluationStatus(evaluationStatusHref, evalStatusEntity);
-
-							if (publishedActivity) {
-								this.push('_publishAllToasts', { activityName: publishedActivity.activityName });
-							}
-						});
-				}
+		this._publishAllDialogMessage = evt.detail.confirmMessage;
+		const dialog = this.shadowRoot.querySelector('d2l-dialog-confirm');
```
maboelsoudd2l: Nit: I'm not the biggest fan of using `this.shadowRoot.querySelector`, because I usually don't like defining components and implicitly selecting and manipulating them somewhere else. This would be very hard for a newcomer to follow this code if they see it for the first time. This is similar the JQuery paradigm where you would define a dom-element and then somewhere else start using it by doing `$(".my-element").dostuff()`. This then cause code to be tightly coupled and hard to refactor. I thought that this is the kind of paradigm that we should be trying to move away from. I would think that "functional programming" is what we should aspire to follow, where we would use attributes and properties that explicitly defined.

I spoke offline with @AlexBedley and @DiljotSG and see that its a compromise. Besides this is a stylistic decision and doesn't affect performance. I just wanted to post a comment somewhere where I can hopefully have start a discussion of how we handle dynamic behaviours. I'm interested to hear from people from Gaudi on this @dbatiste @svanherk @dlockhart

svanherk: I personally don't have any problem with this, as it's only grabbing/controlling a component it knows about and is responsible for.  If you're digging into a `shadowRoot` that isn't "yours" (for example, `this.shadowRoot.querySelector('d2l-dialog-confirm').shadowRoot.querySelector('d2l-button-icon');`), then that's not good, since this component should not know that the dialog has that button.  I don't find this hard to follow, especially since it's grabbed and then immediately used.  Maybe I'm misunderstanding your concern?

maboelsoudd2l: My contention was that this approach was favoured over using properties of `d2l-dialog` in a declarative way (like this `<d2l-dialog-confirm opened="[[_isOpened]]"></d2l-dialog-confirm>`). I wanted to ask why this approach of `dialog.opened()` is favoured instead? My question is not specific to this dialog (since it's low impact to our project) but in general.

AlexBedley: From a nimbus point of view I'd just like us to be consistent. I think we should strive to pass things in through attributes, but when that pattern falls apart we can do what makes sense.

The underlying issue here isn't `dialog` IMO, it's because this `activities` controller doesn't know how to publish all (the event emitter, or in this case the activity card, tells the consumer how to do it), so when `activities` receives a publish all event it would have to store a bunch of temporary state in order to pass things to `dialog` via attributes.

dbatiste: I think there's nothing wrong with querying for elements in local DOM (`this.shadowRoot.querySelector`) - some of our core components may do this, for example to make a measurement that is required for some behavior.

It is a fair point that it is not preferred to **mutate** the local DOM from a variety of places.  While this is not uncommon in many Polymer components, it's not great.  For a LitElement, all DOM state should be reflected in the `render` method. To follow this practice for LitElement, we indeed would prefer the `opened` attribute **in the HTML template**.  It would then be recommended to listen for the `d2l-dialog-close` event, especially if the state of the component needs to be updated when the dialog closes, since the dialog may be closed in a couple of ways out of the app's control, and this event handler is ideally wired up in `render` also. 

To expand on my original comment, if the `opened` attribute is being assigned in JavaScript (which is what the code was doing), the app should be listening for the `d2l-dialog-close` event and care should be taken to avoid duplicate handler wire-up, etc.  To avoid the app from having to do this themselves, it would be preferred to use the `open` method as it simplifies their code, and avoids possible mistakes handling the event.  For background, the original motivation for the `open` method was for use in some React/Redux app where confirmation was being chained in JS.

This is what I would imagine for a LitElement component using `d2l-dialog`:
```
render() {
   return html` 
      ...
      <d2l-dialog opened="${this._someProp}" @d2l-dialog-close="${this._handleDialogClose}">
         ...
      </d2l-dialog>
      ...
   `;
}

_handleDialogClose(e) {
   if (e.detail.action === '...') {
      ...
   }
}
```
It may be possible to follow a similar approach for a Polymer component, though I am not sure for this particular component.

PR #421, opened 2019-10-03
Revert "US110605 Implement getLocalizeResources"

ryantmer:
Reverts BrightspaceHypermediaComponents/activities#415

Seems there is something that ends up removing the JSON files in BSI - locally, this works fine, but it got released in [this PR](https://github.com/Brightspace/brightspace-integration/pull/1645) (my bad for not realizing how the new BSI releases work, previously things weren't so automatic), and now ACE 403s and breaks when trying to fetch `en.json`. Reverting this (and will create a new BSI release) while I figure this out...

Very likely going to revert this revert (along with a fix, whatever it is) shortly, but figured better to un-break things quickly.


ryantmer:
All good actually - since we're dynamically importing the JSON files via `fetch` rather than using `import`, the analyzer doesn't know we need them - a simple fix of using `extraDependencies` in polymer.json, turns out! I'll close this and create a new PR to fix that.

PR #424, opened 2019-10-04
basic collection edit

jlai-d2l:


AllanKerr:
It would be a good idea to add the max-width, body coloring with a placeholder div, and and header border so we can get a better idea of what the entire page will look like. Its your call whether you want to add that now or in a future PR.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -1,24 +1,70 @@
...
 
 	render() {
 		return html`
-			<div>D2L Activity Collection Editor</div>
+			<div class="d2l-heading-4">Edit Learning Path</div>
+			<div class="d2l-activity-collection-header">
+				<h1 class="d2l-heading-1">Employee Onboarding</h1>
+				<div class="toggle-container">
+					<d2l-switch aria-label="Visibility Toggle" label-right @click="${this._updateVisibility}"></d2l-switch>
+					<div class="d2l-label-text"><d2l-icon icon=${this._visibility.icon}></d2l-icon> ${this._visibility.text}</div>
```
dbatiste: Presumably, this will eventually get localized.  In such case, this would become:
```
<div class="d2l-label-text">
   <d2l-icon icon=${this._visibility.icon}></d2l-icon> 
   ${this.localize(this._visibility.term)}
</div>
```
Food for thought though, the object outside the component could be avoided and property simplified to a boolean with `render` like:
```
render() {
   const icon = (this._visible ? 'tier1:visibility-show' : 'tier1:visibility-hide');
   const term = (this._visible ? 'visible' : 'hidden');
   return html`
      ...
      <div class="d2l-label-text">
         <d2l-icon icon=${icon}></d2l-icon>
         ${this.localize(term)}
      </div>
      ...
   `;
}
```

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -1,24 +1,70 @@
...
+		this._visibility = _hidden;
 	}
 
 	render() {
 		return html`
-			<div>D2L Activity Collection Editor</div>
+			<div class="d2l-heading-4">Edit Learning Path</div>
+			<div class="d2l-activity-collection-header">
+				<h1 class="d2l-heading-1">Employee Onboarding</h1>
+				<div class="toggle-container">
```
dbatiste: Be careful of "unqualified" class names such as this because for IE11 and Edge, this DOM will not be scoped since they don't support shadow-DOM.  If something else on the page defines CSS with this class, this element could be targeted by that CSS resulting in some mysterious unexpectedly styles being applied.  This has come up in the wild due to some 3rd party CSS that was being included on the page. To reduce the change of that happening, our best practice (until IE11 is gone, and Edge support shadow-DOM) is to qualify our class names in components with the component name.

PR #428, opened 2019-10-04
Adding visual diff tests for action buttons.

cbusmand2l:


test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card-action-button.visual-diff.js
```diff
@@ -0,0 +1,50 @@
...
+	[
+		'text-and-icon',
+		'button-disabled',
+		'icon-only',
+		'text-only',
+		'tall-height',
+		'small-screen-width'
+	].forEach((name) => {
+		describe(name, () => {
+			[ 'normal', 'hover', 'focus' ].forEach((state) => {
```
AlexBedley: nice

AlexBedley: I wonder if we could do this for rtl... it would mean 2 times as many tests but we'd have better rtl coverage... probably worth discussing

test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card-action-button.visual-diff.js
```diff
@@ -0,0 +1,50 @@
...
+						case 'hover':
+							await page.hover(`.${name}`);
+							break;
+						case 'focus':
+							await page.focus(`.${name}`);
+							break;
+					}
+					if (name.indexOf('width') !== -1) {
+						if (name.indexOf('small') !== -1) {
+							await page.setViewport({width: 899, height: 800, deviceScaleFactor: 2});
```
maboelsoudd2l: tbh if I were you I would have just created a new test below instead of if-statements that queries the string

cbusmand2l: Future proofing - future tests that want to check against the smaller screen width can just be named appropriately to get the smaller viewport.

test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card-action-button.visual-diff.js
dbatiste: For what it's worth, I've only repeated them (in core) if there's a chance RTL will affect those other states... just in the interest in keeping our test runtimes down.

test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card-action-button.visual-diff.js
maboelsoudd2l: I replicated some of your logic in https://github.com/BrightspaceHypermediaComponents/activities/pull/429/files#diff-bea8985f5fb8c85acfab2712f6be249bR42
It would be cool if we can create our own "directives" or basically keywords that informs the test suite to test something in a different viewport or on hovered state (e.g. something like "my-test-width-500px-hovered")

AlexBedley: last friday when we were writing these I was thinking it'd be cool to have like a polymer component to do this.

```html
<test-matrix hover rtl large-screen>
  <thing-under-test></thing-under-test>
</test-matrix>
```

And it would test the component with each of those things toggled on and off.

dbatiste: yeah i like the array approach to reduce repetition in simple cases, but as soon as we add a bunch of conditions it starts to defeat the purpose and my personal preference is to have simpler code that's easier to read with a little repetition if need-be. it's kinda of a fuzzy thing for me, but any more than one condition and i tend to start leaning away.

maboelsoudd2l: and also it would be cool if the test suit would just photograph every `<test-matrix>` element. Because currently you have to constantly sync up the `<div id="test">` elements with the array. And if you forget to do so, the element would not be photographed, and it might be hard to notice these things in code reviews.

DiljotSG:
@CarolBusman You'll need to regenerate the Goldens but the build should pass after that and then can be merged.

PR #429, opened 2019-10-04
added visual diff test coverage for activity card

maboelsoudd2l:
I know that visual diff is blocked but I added this regardless and hoping that by Monday this we'll have a fix for this.


DiljotSG:
@maboelsoudd2l It looks good, but you'll need to merge master in to get the latest version of the visual-diff package. I will approve it after that.

PR #431, opened 2019-10-04
US110605 Followup: Serge run_command

ryantmer:
[Based on the Serge docs](https://serge.io/docs/plugins/callback/run_command/), there's a `run_command` plugin with which you can run arbitrary commands. This adds a command to build our language files for us, so that (in theory) Serge will return to us both the translated JSON files, as well as built JavaScript files, which we can then import. Importing these by name allows the BSI analyzer to actually include the files correctly - we load only the files we need initially, and lazy-load the others if they're needed (e.g. if the page language changes).

I say "in theory" above as I've not been able to actually test this on the Serge service itself - the command works as expected locally, but I'm still not 100% sure about how the Serge integration actually works - still tracking down some of those details. For one, I don't know if the files that `run_command` generates will actually get committed back to the repo by Serge or...? But yeah, still looking.

### DIFF
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -97,7 +97,7 @@
     "callback_plugins": [{
       "plugin": "run_command",
       "data": {
-        "command": "echo \"export const translations = JSON.parse(\\`$(cat %OUTFILE%)\\`);\" > %OUTPATH%/build/%LANG%.js"
+        "command": "echo \"export const translations = $(cat %OUTFILE%);\" > %OUTPATH%/build/%LANG%.js"
       }
     }]
   }
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -97,17 +97,7 @@
     "callback_plugins": [{
       "plugin": "run_command",
       "data": {
-        "command": "export LANG_PATH=./components/d2l-activity-editor/d2l-activity-assignment-editor/lang"
-      }
-    },{
-      "plugin": "run_command",
-      "data": {
-        "command": "rm -rf $LANG_PATH/build && mkdir -p $LANG_PATH/build"
-      }
-    }, {
-      "plugin": "run_command",
-      "data": {
-        "command": "for language in $(find $LANG_PATH -name \"*.json\" -type f -printf \"%f\\n\" | sed 's/\\.json$//'); do echo \"export const translations = JSON.parse(\\`$(cat $LANG_PATH/$language.json)\\`);\" > $LANG_PATH/build/$language.js; done"
+        "command": "echo \"export const translations = JSON.parse(\\`$(cat %OUTFILE%)\\`);\" > %OUTPATH%/build/%LANG%.js"
       }
     }]
   }
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -93,6 +93,22 @@
       "tr-tr tr",
       "zh-cn zh",
       "zh-tw zh-tw"
-    ]
+    ],
+    "callback_plugins": [{
+      "plugin": "run_command",
+      "data": {
+        "command": "export LANG_PATH=./components/d2l-activity-editor/d2l-activity-assignment-editor/lang"
+      }
+    },{
+      "plugin": "run_command",
+      "data": {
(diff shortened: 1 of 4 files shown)


activities.serge.json
```diff
@@ -56,5 +56,30 @@
       "zh-cn zh",
       "zh-tw zh-tw"
     ]
+  },{
+    "name": "activityAssignmentEditor",
+    "source_dir": "components/d2l-activity-editor/d2l-activity-assignment-editor/lang",
+    "output_file_path": "components/d2l-activity-editor/d2l-activity-assignment-editor/lang/%LANG%.json",
+    "output_lang_rewrite": [
```
dbatiste: Any chance you know anything more about this?

activities.serge.json
```diff
@@ -56,5 +56,30 @@
...
+      "pt-br pt",
+      "sv-se sv",
+      "tr-tr tr",
+      "zh-cn zh",
+      "zh-tw zh-tw"
+    ],
+    "callback_plugins": [{
+      "plugin": "run_command",
+      "data": {
+        "command": "echo \"export const translations = JSON.parse(\\`$(cat %OUTFILE%)\\`);\" > %OUTPATH%/build/%LANG%.js"
```
dbatiste: Do you think it would be possible to only output the translations, so that clients aren't downloading the `context`.  Really we just want something like: 
```
{
   'en': { close: 'Close this dialog' },
   ...
}
   ```
Also, what do you think about doing fallback resolution here, as opposed to runtime? This would save processing client-side, as well as avoid potential multiple requests per component.

File: components/d2l-activity-editor/d2l-activity-assignment-editor/lang/build/en.js [hunk 3]
```diff
@@ -0,0 +1,10 @@
+export const translations = JSON.parse(`{
```
dbatiste: It looks like this doesn't need to be a string, in which case the `JSON.parse` is not really needed.  Ya?

activities.serge.json
ryantmer: Yes actually! Finally found it, from https://serge.io/docs/configuration-files/reference/:

> [OPTIONAL] Map telling to use different language name in language macros
   (see the list of available macros in `ts_file_path` parameter description)
   and when constructing output path (see `output_file_path` parameter).
   This allows to tweak language/locale names for different projects.

As far as I understand that, it just means that the translation will end up being `ar-sa` (I'm guessing since the person/service doing the translation is Saudi?), but we want to use that for _all_ `ar` locales.

activities.serge.json
ryantmer: We could, and I think it's worth it, but I'd rather do it by calling an external script rather than a big one-liner that we're just echoing into a file. Once I manage to actually get access to a Serge service (or run one locally?) then I can properly test this, at which point yeah, we could probably do the fallbacks at this stage!

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/build/en.js
ryantmer: Ya indeed! I was thinking the spacing might cause issues, but that makes no sense whatsoever...

ryantmer:
Build failure is due to the removal of the `JSON.parse` in the built files - going to ignore that for now, as I would like to externalize this into a build script rather than a one-liner, which will also allow for properly formatting the built file so that eslint doesn't yell about it.

maboelsoudd2l:
so given that the serge service PR is merged in this PR https://git.dev.d2l/projects/RE/repos/serge-localize/pull-requests/110/overview, does this work? how do we verify?

ryantmer:
Working on that as we speak - this isn't going to work anyway since Serge runs on Windows, but the command here is bash.

That said, this doesn't block #458 - that can be merged as a way to generate the files manually, whereas this would generate them automatically.

ryantmer:
Closing this in favour of #468, which uses the built-in `parse_js` parser in Serge to work exclusive with JS files. So long as Serge doesn't do anything weird with JS files (aside from translate the terms), that approach allows us to use built-in tooling directly with ES6 modules.

PR #434, opened 2019-10-08
US105733 - add method to detect if search or filter is applied immediately

AlexBedley:
Previous PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/385

Problem: In test, sometimes the filters would not be loaded before `_loadData` ran. This meant that `filtersApplied` was `undefined`. Thus in this case, the criteria would not be reset on an empty response.
Solution: Instead of waiting for `/filters` to load, just detect if the query string param is present in the `href`.

Obviously this solution breaks our design because now the UI knows too much about the filters implementation. But I can't think of any other way of doing it (short of the HM route returning this information - I think we should eventually do this). An advantage of this solution is that if we do go forward with lazily loading the filters then the features that rely on `filtersApplied` will still work.

What are your thoughts?

- [x] add tests for `_hasNonEmptyQueryParam`

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -248,11 +248,15 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		}
 		this._loading = true;
 
+		if (this._initialLoad) {
+			this._setApplied();
+		}
+
 		if (this._initialLoad &&
 			entity.hasClass('empty') &&
 			(this.searchApplied || this.filterApplied)
 		) {
-			this._clearFilterAndSearch();
+			await this._clearFilterAndSearch();
 			this._initialLoad = false;
 			return;
 		}
@@ -323,6 +327,15 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		}
 		if (this.searchApplied) {
 			await this.clearSearchResults();
+		}
+	}
+
+	_setApplied() {
+		if (this._hasNonEmptyQueryParam(this.href, 'filter')) {
+			this.filterApplied = true;
+		}
+		if (this._hasNonEmptyQueryParam(this.href, 'collectionSearch')) {
+			this.searchApplied = true;
 		}
 	}
 
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -249,7 +249,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		this._loading = true;
 
 		if (this._initialLoad) {
-			this._setApplied();
+			this.filterAppliedShortcut();
+			this.searchAppliedShortcut();
 		}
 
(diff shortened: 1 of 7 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -248,11 +248,15 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		}
 		this._loading = true;
 
+		if (this._initialLoad) {
+			this._setApplied();
```
cbusmand2l: Maybe have a `_setFilterApplied` and `setSearchApplied` in the respective behaviours, calling them both in this `if` block? That would keep the bulk of the new logic segregated, no?

AlexBedley: The reason I didn't do it this way is because I felt that sniffing the href for query params seemed hacky and I wanted to avoid polluting the behaviours with this QE-specific hack. Also since the search/filter behaviours don't have access to `d2l-siren-helper-behaviour`, I'd have to introduce a dependency or duplicate the `_hasNonEmptyQueryParam` code.

But it does keep the filter stuff in the filter behaviour.... I think if we planned on making this a "feature" of the filter behaviour then it might be ok to have this duplication.

cbusmand2l: Ah - I figured we could just keep the hacky part in the siren behaviour and use the filter/search `set` just be a wrapper for `_hasNonEmptyQueryParam` from the filter/search behaviour (since they're bundled together and we already do it with `_getHref`).

AlexBedley: oh wait you're right, they do have access to `d2l-siren-helper-behaviour`.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -390,6 +394,15 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		}
 	}
 
+	_setApplied() {
+		if (this._hasNonEmptyQueryParam(this.href, 'filter')) {
```
maboelsoudd2l: Nit: in the spirit of the "UI knowing too much about the implementation of filter/search" we can also take this to the next level. I think the controller shouldn't know about `'collectionSearch'` and `'filter'`. i think we can create an abstraction in `d2l-siren-helper-behavior.js` is filter applied
```javascript
_isfilterApplied: function (href) {
    return this._hasNonEmptyQueryParam(href, 'filter');
},
_isSearchApplied: function (href) {
    return this._hasNonEmptyQueryParam(href, 'collectionSearch');
}
```

AlexBedley:
Alright I updated the PR based on both of your comments.

I gave the functions weird names (`Shortcut`) because I wanted to make it obvious that this is a hack and they should not be used unless absolutely necessary

AlexBedley:
> why is travis-CI failing tho?

ugh I forgot a semi-colon. Will update

PR #436, opened 2019-10-10
US110195 Add d2l-html-editor

ryantmer:
This adds a `d2l-ace-html-editor` component, which wraps `d2l-html-editor` and allows for editing the HTML instructions for an assignment. Editing in this field is functional - it fetches the editor config for the assignment, displays the current instructions, autosaves on change, and uses the instructions edit Siren action to POST the new value. Formatting and ISF also appear to work as expected.

Still doing some final checking and validation of things locally, but this is ready to go barring any final changes from that testing.


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-ace-html-editor.js
```diff
@@ -0,0 +1,184 @@
...
+import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
+
+let htmlEditorId = 0;
+function getNextEditorId() {
+	return ++htmlEditorId;
+}
+
+class AceHtmlEditor extends EntityMixinLit(LitElement) {
```
mdgbayly: So far we've been using `d2l-assignment-*` for assignment components and `d2l-activity-*` for generic activity components?

I see this as a generic activity component, so would rather we stuck with the prefix `activitity` rather than introducing a new `Ace` prefix which is kind of tied to our project?

mdgbayly: And following on from that, yes, I think we should be making this non-assignment specific and it should live a level up in the component hierarchy.

ryantmer: Ah, yeah fair point! I think I just forgot to change the name on this one, I'll update it.

ryantmer: For the non-assignment-specific part, we would need to somehow convey the type of entity that the component is working with, or pass in the richtext config. I'm not a fan of the latter, since it means the parent component needs to be richtext-config-aware (which is super leaky), but I also don't see a nice way to pass in a class in LitElement 🤔

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-ace-html-editor.js
```diff
@@ -0,0 +1,185 @@
...
+				border-width: 1px;
+				box-shadow: var(--d2l-input-boxshadow);
+				padding: var(--d2l-input-padding);
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this._setEntityType(AssignmentEntity);
```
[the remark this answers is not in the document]
mdgbayly: Yeh i think something like that would be preferable. The goal here was to try and make an html editor wrapper that could work with any activity.

The way we handled this in rubrics was to push the richtext-editor config into the store explicitly which was I acknowledge kind of cludgy.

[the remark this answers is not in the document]
mdgbayly: but yeh, how to do that in a cleaner way!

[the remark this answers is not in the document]
ryantmer: Ah okay, then the retrieval happens [here]()? I was wondering about how that worked - would be nice for the component to be able to retrieve the config itself, since it's not relevant to the parent component. On that note, though, perhaps this can be structured such that instead of an `AssignmentEntity`, it's a `RichTextConfigEntity` (or, some better name) so that it's activity-type-agnostic - which I think makes sense, since it's kind of nonsensical for the HTML editor to care about the type of activity of the richtext that it's editing!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -106,15 +130,29 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
...
 					prevent-submit>
 				</d2l-input-text>
 				${this._getNameTooltip()}
 			</div>
+
+			<div id="assignment-instructions-container">
+				<label class="d2l-label-text" for="assignment-instructions">${this.localize('instructions')}</label>
+				<d2l-ace-html-editor
+					id="assignment-instructions"
+					.href="${this.href}"
```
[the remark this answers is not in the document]
mdgbayly: I do see the html editor wrapper as a purely UI component kind of like `d2l-text-area` so I don't think it specifically needs to be entity aware. Until we come up with a better place for the rich text editor config to live e.g. maybe we do decide to make it an org level addressable entity, couldn't we just pass the config in as a prop to the `d2l-ace-html-edtor`?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-ace-html-editor.js
```diff
@@ -0,0 +1,184 @@
+import 'd2l-html-editor/d2l-html-editor.js';
+import 'd2l-html-editor/d2l-html-editor-client.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
+
+let htmlEditorId = 0;
```
dbatiste: Note: we have a helper that can be used for unique ids.

https://github.com/BrightspaceUI/core/blob/master/helpers/uniqueId.js

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-ace-html-editor.js
```diff
@@ -0,0 +1,184 @@
...
+		this._setEntityType(AssignmentEntity);
+		this._htmlEditorUniqueId = `htmleditor-${getNextEditorId()}`;
+	}
+
+	set value(newValue) {
+		const oldValue = this.value;
+
+		const editorContainer = this.shadowRoot.querySelector('d2l-html-editor > .d2l-html-editor-container');
+		if (editorContainer) {
+			editorContainer.innerHTML = newValue;
```
dbatiste: Note: this seems to go against the idea that the DOM should not be modified outside of `render`.

If we indeed need this here, should we bail if `editorContainer` does not exist yet? Or does it matter?

ryantmer: Yeah, I'll give this another look this morning, but the explanation here is that templating the value into the inside of the `d2l-html-editor-container` doesn't work, as it ends up being rendered as a string, rather than as HTML. There may be a way to escape it somehow, though, which perhaps new-week-Ryan can figure out!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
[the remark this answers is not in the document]
ryantmer: Mentioned in my last commit, but for this I'll have to loop back around to this after some further LMS work - in particular, we are right now always rendering the `d2l-activity-html-editor`, but there's that user config that may mean we should render a (not-yet-created) `d2l-activity-plaintext-editor` instead. Once the LMS side is figured out, I'll add the `plaintext-editor` to the UI as well, and probably wrap them both in a `d2l-activity-text-editor` or something.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-ace-html-editor.js
ryantmer: Looked into this, but sadly no luck - tried using `this.value` directly in the template, or calling a helper render function that uses `` return html`${this.value}` ``, but same result as both end up just using it as text content, rather than rendering the HTML.

ryantmer: Looked into this, but sadly no luck - tried using `this.value` directly in the template, or calling a helper render function that uses `` return html`${this.value}` ``, but same result as both end up just using it as text content, rather than rendering the HTML.

mdgbayly: This is partly historical and partly due to the nature of the `d2l-html-editor`. It is not a lit-element for one thing.

In other scenarios e.g. free range apps, a `content` prop is used to initialize the content of the `d2l-html-editor`. However, at some point, someone modified the implementation of that to integrate filtering the initial HTML by calling the LMS API (via ifrau) to apply HTML filtering. I don't recall why it was done like that vs just ensuring filtering is applied up front. Could be historical - could be because that is how the legacy HTML editor does it. I don't recall.

But we found that approach was killing performance for rubrics, where we have many HTML editors on a page, and the initial contents are all pre-filtered in the API that is used to load the rubric.

So instead we switched to just setting the innerHtml of the div that is used to render the tinymce editable area.

There might be better ways to do that. 
e..g lit-html does have an unsafeHtml directive?

https://lit-html.polymer-project.org/guide/template-reference#unsafehtml

But I think the other complication here is we only want to call this when the entity value associated with this component changes. Once we start editing it's effectively a uncontrolled component as tinymce is managing all the state.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
```diff
@@ -0,0 +1,181 @@
...
+		super();
+		this._htmlEditorUniqueId = `htmleditor-${getUniqueId()}`;
+	}
+
+	set value(newValue) {
+		const oldValue = this.value;
+
+		const editorContainer = this.shadowRoot.querySelector('d2l-html-editor > .d2l-html-editor-container');
+		if (editorContainer) {
+			editorContainer.innerHTML = newValue;
```
mdgbayly: I'm still a bit unclear whether setting this here every time value changes will do funny things to the HTML editor internals

e.g. does this called again after you enter some content in the editor, we save the changed content, and then we re-render the UI with the updated entity?

I'm vaguely recall tinymce is adding stuff to the HTML to track things like selection and caret position and wonder if that will get lost if we update the innerHMTL again.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
```diff
@@ -0,0 +1,181 @@
...
+		const editor = this.shadowRoot.querySelector('d2l-html-editor');
+		if (editor) {
+			editor.d2lPluginSettings = editorConfig.properties || {};
+		}
+
+		this.requestUpdate('richtextEditorConfig', oldValue);
+	}
+
+	_resolveUrl() {
+		return resolveUrl('../../../');
```
mdgbayly: Is this used somewhere?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
```diff
@@ -0,0 +1,181 @@
...
+				max-rows="1000"
+				fullpage-enabled="0"
+				toolbar="bold italic bullist d2l_isf"
+				plugins="lists paste d2l_isf">
+
+				<div id="toolbar-shortcut-${this._htmlEditorUniqueId}" hidden=""></div>
+				<div
+					class="d2l-html-editor-container"
+					id="${this._htmlEditorUniqueId}"
+					value="${this.value}"
```
mdgbayly: I doubt this value attribute is doing anything?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
ryantmer: Yeah, it defines the `app-root` in the DOM. I'll admit I don't really know what this is, only that it works with it, and doesn't work without it 😅

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
ryantmer: I totally added a `_valueSet` to avoid re-setting this every time it changes, but forgot to push it! Now this code only ever gets executed the first time the value is set - after that, it's all up to the editor.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-html-editor.js
mdgbayly: oh - right - my searching must have been wonky..

yes - it's for some more tinymce ugliness where we have to pass absolute paths to various files required by tinymce and some of the plugins (css/lang files/skins etc)

ryantmer: Ah okay, that explains it. I looked into it briefly but decided it was "TinyMCE magic" and left it at that 😆

PR #442, opened 2019-10-17
Delaying filter loading until after the activities are loaded.

cbusmand2l:


cbusmand2l:
@AlexBedley Ah! I forgot that the skeleton state was based on whether the filters were loading or not. On that note, would it also make sense to remove it from the search summary state as well?

maboelsoudd2l:
To add to what @AlexBedley said i captured this screenshot (through the magic of browser network throttling), which shows the filters loaded and the entity not-loaded. but that could also be due to the skeleton as you guys said.

image

maboelsoudd2l:
The entangling of `filtersLoading` with the loading skeleton was something I added a while ago (probably because I thought that `filtersLoading` was related to filtering the data).

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -532,12 +532,12 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		return _loadingMore;
 	}
 
-	_computeShowLoadingSkeleton(_loading, filtersLoading, searchLoading) {
-		return _loading || filtersLoading || searchLoading;
+	_computeShowLoadingSkeleton(_loading, searchLoading) {
+		return _loading || searchLoading;
```
maboelsoudd2l: If we're removing `filtersLoading`, by that logic should we also be removing `searchLoading` as well?

cbusmand2l: I don't know that the search has near as much overhead as the filters. We may want to just double check whether it's something we want to do, but either way I don't see it as a reason to hold up this PR.

maboelsoudd2l: fair enough

AlexBedley: Search doesn't really have loading associated with it, it was just to make the UI a bit snappier so it immediately displays the skeleton when you press enter

PR #443, opened 2019-10-17
US110196 Fixup: HTML editor change event

ryantmer:
This adds a more predictable `change` event to the `d2l-activity-html-editor`, which is internally hooked up to both the `input` and `change` event from `d2l-html-editor`. This ensures the outer component generates an event for all changes within the HTML editor, whereas the previous implementation wouldn't pick up e.g. inserting an image or changing the text format until the input text was also changed. This means that the parent component (`d2l-activity-assignment-editor-detail` in this case) only has to listen for this `change` event, and can use that to drive auto-saving the instructions value.

This also updates the size of the editor to be a minimum of 3 rows, which matches the design.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -101,8 +101,8 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 		}
 	}
 
-	_saveInstructionsOnInput(e) {
-		const instructions = e.target.getContent();
+	_saveInstructionsOnChange(e) {
+		const instructions = e.detail.content;
 
 		this._debounceJobs.instructions = Debouncer.debounce(
 			this._debounceJobs.instructions,
@@ -148,8 +148,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 					id="assignment-instructions"
 					value="${this._instructions}"
 					.richtextEditorConfig="${this._richtextEditorConfig}"
-					@change="${this._saveOnChange('instructions')}"
-					@input="${this._saveInstructionsOnInput}"
+					@change="${this._saveInstructionsOnChange}"
 					aria-label="${this.localize('instructions')}"
 					?disabled="${super._entity && !super._entity.canEditInstructions()}">
 				</d2l-activity-html-editor>
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -148,7 +148,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 					id="assignment-instructions"
 					value="${this._instructions}"
 					.richtextEditorConfig="${this._richtextEditorConfig}"
-					@change="${this._saveInstructionsOnChange}"
+					@d2l-activity-html-editor-change="${this._saveInstructionsOnChange}"
 					aria-label="${this.localize('instructions')}"
 					?disabled="${super._entity && !super._entity.canEditInstructions()}">
 				</d2l-activity-html-editor>
--- a/components/d2l-activity-editor/d2l-activity-html-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-editor.js
@@ -151,11 +151,9 @@ class ActivityHtmlEditor extends LitElement {
 		return `${import.meta.url}/../../../`;
 	}
 
-	_onContentChange(e) {
-		e.stopPropagation();
-
+	_onContentChange() {
 		const content = this.shadowRoot.querySelector('d2l-html-editor').getContent();
-		this.dispatchEvent(new CustomEvent('change', {
+		this.dispatchEvent(new CustomEvent('d2l-activity-html-editor-change', {
 			bubbles: true,
 			composed: true,
 			detail: {
--- a/components/d2l-activity-editor/d2l-activity-html-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-editor.js
@@ -151,8 +151,17 @@ class ActivityHtmlEditor extends LitElement {
 		return `${import.meta.url}/../../../`;
 	}
 
-	getContent() {
-		return this.shadowRoot.querySelector('d2l-html-editor').getContent();
+	_onContentChange(e) {
+		e.stopPropagation();
+
+		const content = this.shadowRoot.querySelector('d2l-html-editor').getContent();
+		this.dispatchEvent(new CustomEvent('change', {
+			bubbles: true,
+			composed: true,
+			detail: {
+				content: content
+			}
+		}));
 	}
 
 	render() {
@@ -161,8 +170,10 @@ class ActivityHtmlEditor extends LitElement {
 				id="assignment-instructions"
 				editor-id="${this._htmlEditorUniqueId}"
 				app-root="${this._resolveUrl()}"
+				@change="${this._onContentChange}"
+				@input="${this._onContentChange}"
 				inline="1"
-				min-rows="1"
+				min-rows="3"
 				max-rows="1000"
 				fullpage-enabled="0"
 				toolbar="bold italic bullist d2l_isf"


mdulat:
The `input` event in the name field is so that changes are saved periodically even when focus doesn't leave (like if user is typing a lot or slowly).

ryantmer:
Hmm, I still see the `change` event happening in those cases. There's the `change` event that is part of the HTMLElement spec (which [fires on focus change](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event)), but the `change` event from `d2l-text-input` is [a `CustomEvent`](https://github.com/BrightspaceUI/inputs/blob/master/d2l-input-text.js#L120) that is basically a re-broadcast of that event, but also seems to fire anytime I change the text in that field... I'll remove that change from this PR though, since it's unrelated at best, and breaks things (in a way I can't figure out) at worst! 😄

components/d2l-activity-editor/d2l-activity-html-editor.js
```diff
@@ -161,8 +170,10 @@ class ActivityHtmlEditor extends LitElement {
 				id="assignment-instructions"
 				editor-id="${this._htmlEditorUniqueId}"
 				app-root="${this._resolveUrl()}"
+				@change="${this._onContentChange}"
+				@input="${this._onContentChange}"
```
wongvincent: Does the html-editor fire an "input" event?

mdgbayly: I didn't think the tinymce HTML editor fired input events? Are you seeing this input handler being called.
But yeh this is a confusing area. In previous integrations where we have wanted to provide a consistent event interface for both text and html inputs we've adapted the events in the wrapper component that conditionally rendered a text input or an html editor based on the user's editing preferences.

e.g. here.
https://github.com/Brightspace/d2l-rubric/blob/master/editor/d2l-rubric-text-editor.js#L84

But it might make sense to do that adaption directly in the d2l-html-editor wrapper as you are doing here.

To be honest, i'm not entirely sure what this is doing as, like I say, I didn't think the HTML editor fired input events, but maybe I'm wrong about that.

https://github.com/Brightspace/d2l-rubric/blob/master/editor/d2l-rubric-text-editor.js#L93

mdgbayly: Basically tinymce, fires 'change' events on every redo/undo level change, which means it fires events much more frequently than a regular <input> change event.

mdgbayly: ditto - lol

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -148,8 +148,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 					id="assignment-instructions"
 					value="${this._instructions}"
 					.richtextEditorConfig="${this._richtextEditorConfig}"
-					@change="${this._saveOnChange('instructions')}"
-					@input="${this._saveInstructionsOnInput}"
+					@change="${this._saveInstructionsOnChange}"
```
wongvincent: Do we want to immediately trigger the save on focus out (i.e. flush the debounce)? If we want to match what the name input is doing, then the `d2l-activity-html-editor` should fire an event that results in calling `@event-1234="${this._saveOnChange('instructions')}"`

components/d2l-activity-editor/d2l-activity-html-editor.js
mdgbayly: Also maybe instead of all the shenanigans with having to interpret whether the change event is a html editor or regular browser change/input event we should have just modified the HTML editor to make it more consistent.
For normal events you have to get the value by calling `event.target.value`. Maybe we could have implemented a 'value' property getter on the `d2l-html-editor`, which just returns `this.getContent()`

ryantmer: It does indeed fire an input event, although I'm honestly not sure where it's coming from. The `change` event is the one that acts like the HTMLElement `change` event, which doesn't always fire when the text changes - it does fire consistently when I type the first time, change text formatting, insert an image, or the field loses focus, but it doesn't fire as I type (except the first time). The input event, on the other hand, fires on every keypress, but it _doesn't_ fire on formatting change or when I insert an image.

The end result of this two-events-into-one is that I get a `change` event from `d2l-activity-html-editor` for all of the following:
- keypress
- focus leaves
- insert image
- change formatting
- paste text

whereas using just `change` or `input` gives me events only on a subset of these 🤔

ryantmer: See the reply below :D

mdgbayly: interesting - what is the target of the input event?

wongvincent: ... but not as frequently as an input event. the changes that cause a redo/undo levels are also hard to pin down.

I think the `change` event from `d2l-html-editor` could be treated as `input` and cause a debounced save. While a `focus-out` event could be treated as a `change` event that causes an immediate save (i think it either fires focus out or change on a focus out.)

ryantmer: For the target of the event, the `change` event is coming from the `d2l-html-editor id=assignment-instructions` , whereas the `input` event is coming from the `div class=d2l-html-editor-container`. I think this makes intuitive sense, since the div is being effectively upgraded to an input element by TinyMCE, if I understand correctly, and I suppose as part of that process it starts to look and act like an input element, which includes dispatching `input` events?

Yeah, I found that `change` is slightly unpredictable in how it works, hence the inclusion of `input` handling here as well. I didn't actually think of the situation of treating `focus-out` differently and immediately saving - aside from the debounce delay, is there a benefit to that? (I'm just thinking from a naive-saving perspective, if you type some stuff then change focus, wouldn't that cause a debounced autosave as well as a syncronous immediate save?)

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: I can try it out! Mostly I'd be curious as to what the flush does to an in-progress save - would the network call get cancelled, or does it just wait until the in-progress one is finished, then do an immediate save after? (If so, I'd imagine the majority of cases we'd be doing a network request to save the value to a value that isn't different than the current value, no?)

components/d2l-activity-editor/d2l-activity-html-editor.js
mdgbayly: oh - yeh i don't think an input event from a content editable is really a std browser thing. And it probably wouldn't allow you to call `target.value` like a regular input event. But using here as another way to get the html editor content probably doesn't hurt.

wongvincent: It won't cause an immediate save. It would call the debounced autosave.

With the `name` input. An `input` event causes a debounced autosave. A `change` event (i.e. focus out) causes the immediate save.

So in this `html-editor` case, I think `change` and `inputs` could all be handled as a debounced auto save. If we wanted to do an immediate save, the `html-editor` fires a `blur` event that we could listen for. But it's not really important and I don't know if the order of the `blur` and `change` event firing back to back could cause a double save.

ryantmer: There's the [input](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) event that is standard to input/select/textarea, so I was thinking TinyMCE maybe makes the div you give it into a fake-textarea or something, including dispatching that event, although I really don't know.

All that said, upon some further testing, the `change` event I think is _supposed_ to be enough - when I do things like paste or cut text, change formatting, insert an image, or `focusout` it fires, and it seems like it _eventually_ also fires for when I modify the text content, but that last one is super weird. I'll see if maybe there's some sort of "undo stack timer", where a new undo level is added every `n` milliseconds automatically or something, which would probably be sufficient to trigger the autosave and remove the need to listen for `input` at all!

mdgbayly: I think it's just that when we use tinymce in inline mode, it makes the div that we tell it to act as the target a contenteditable https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable.

But it looks like the guys who did the fixes to the rubrics edits that we were discussing the other day were making use of this to prevent autosaves from potentially overwriting the what the user is currently typing. Although, having said that, I think because we never update tinymce, that shouldn't be a problem anyway. Why does this have to be so complicated....

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
wongvincent: pretty sure it doesn't wait for anything. Theoretically, if save 2 begins after save 1, it could complete before save 1. I think...

components/d2l-activity-editor/d2l-activity-html-editor.js
ryantmer: Ohh, gotcha okay. I misunderstood how TinyMCE does it's upgrade magic!

I will admit that this change doesn't specifically address the multiple-edits-at-once problem, but I did trigger multiple edits at once using the network throttler in Chrome and everything did work as expected - that said, I'd imagine that wouldn't be the case if the network responses came back in a different order.

Also above I said it seems to auto-`change` every once in a while, that doesn't actually appear to be the case - I was thinking it was something like `vim`'s undo stack, where each inserting is one thing on the undo stack. It looks like for TinyMCE, the only way to predictably add new text onto the undo stack is by pressing Enter, or Home/End/PageUp/PageDown, which is kind of weird - no auto-stack-entry after 100 characters, or after 500ms or anything like that (at least, not by default) 😢

ryantmer: Aha - so it looks like it _could_ automatically generate a `change` event every `n` milliseconds or every `x` keypresses, but we'd have to do it explicitly (within `d2l-html-editor` itself), using [`UndoManager.add`](https://www.tiny.cloud/docs-3x/api/class_tinymce.UndoManager.html/#add). Not sure if that would be worth pursuing, though, since it would affect how it works elsewhere.

components/d2l-activity-editor/d2l-activity-html-editor.js
```diff
@@ -151,8 +151,17 @@ class ActivityHtmlEditor extends LitElement {
 		return `${import.meta.url}/../../../`;
 	}
 
-	getContent() {
-		return this.shadowRoot.querySelector('d2l-html-editor').getContent();
+	_onContentChange(e) {
+		e.stopPropagation();
```
dbatiste: Is the `stopPropagation` needed?

components/d2l-activity-editor/d2l-activity-html-editor.js
dbatiste: It seems like it would be value-add for our wrapper component once we figure out what events are needed to have some auto-save.

In the LMS HtmlEditor I think we wire-up to some type of node-change event.  We tried to minimize how often, and the work done in those handlers for the obvious possible performance implications.  That event may be useful. The user may move focus out to the toolbar to perform an action, and focus may or may not return back to the content. I think there are a few scenarios that we'd want auto-save to be triggered.

ryantmer: Yeah - it would be nice to use the undo layers thing to only emit an event "once in a while", which could _potentially_ avoid the need for a debounce outside the component (although still probably a good idea to have one). I'm thinking the ideal circumstance would be the existing TinyMCE `change` + an automatic change event every `x` milliseconds or `y` keypresses. This would probably still generate a fairly large number of `change` events, granted, hence I think having the debounce outside the component would still be wise.

components/d2l-activity-editor/d2l-activity-html-editor.js
ryantmer: I kind of wanted the `d2l-activity-html-editor` component to be almost invisible, hence going with the same-named `change` event, but I'm realizing this morning that it's pretty inconsistent with how we do things elsewhere. I'll change it to `d2l-activity-html-editor-change` or something!

PR #441, opened 2019-10-17
Updating admin-list and collection-editor to use the new d2l-list.

jlai-d2l:
image
image

PR #444, opened 2019-10-18
Fix siren-sdk version

ryantmer:
Due to magic I don't entirely understand, BSI requires all dependencies to only specify the major version. This change was introduced in #441, but breaks the BSI build when `activities` is updated there.

It seems like this was put in place to ensure getting at least a specific version of `siren-sdk` that presumably has some new functionality, but BSI will always get the newest version anyway (until a breaking/major version change).

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,6 @@
     "fastdom": "^1.0.8",
     "merge-stream": "^1.0.1",
     "require-dir": "^1.2.0",
-    "siren-sdk": "BrightspaceHypermediaComponents/siren-sdk#semver:^1.4.2"
+    "siren-sdk": "BrightspaceHypermediaComponents/siren-sdk#semver:^1"
   }
 }


dbatiste:
I think this is fine.  When BSI is updated it should be pulling in the new (latest version).

jlai-d2l:
@ryantmer I thought so too. But for some reason, the build would fail as it _didn’t_ pick up the latest version. Hoping this doesn’t happen here.

svanherk:
@jlai-d2l It might have been because of a Travis cache.  If you see it again, I'd try clearing these if you have permission: https://travis-ci.com/BrightspaceHypermediaComponents/activities/caches
Or, better yet, you can turn off caching completely by adding this to the Travis file: https://github.com/BrightspaceUI/navigation/blob/master/.travis.yml#L1-L2

PR #446, opened 2019-10-21
US110582 Add due date editor

ryantmer:
This `d2l-datetime-picker` control will show the current due date for the activity, and can be used to set a new due date for the activity. There are a few caveats to this currently that are worth noting:

- The `d2l-datetime-picker` component in use here is obsolete, and is going to be replaced in the near future, but there is not currently a timeline for this change. Once a new component exists, we should switch to use it.
- There is a change in the LMS to pull in `moment.js` to allow this component to work, as it doesn't do it itself. When we switch to a new picker component, this change can be removed, as it's two network calls to large-ish libraries that we won't need to make anymore.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -1,4 +1,5 @@
 import 'd2l-inputs/d2l-input-text.js';
+import '../d2l-activity-due-date-editor.js';
 import '../d2l-activity-html-editor.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
@@ -18,7 +19,8 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 			_name: { type: String },
 			_nameError: { type: String },
 			_instructions: { type: String },
-			_richtextEditorConfig: { type: Object }
+			_richtextEditorConfig: { type: Object },
+			_activityUsageHref: { type: String }
 		};
 	}
 
@@ -58,6 +60,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 			this._name = assignment.name();
 			this._instructions = assignment.instructionsEditorHtml();
 			this._richtextEditorConfig = assignment.getRichTextEditorConfig();
+			this._activityUsageHref = assignment.activityUsageHref();
 		}
 	}
 
@@ -68,7 +71,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 	_saveName(value) {
 		if (super._entity.canEditName()) {
 			const action = super._entity.getSaveNameAction();
-			const fields = [{ 'name': 'name', 'value': value }];
+			const fields = [{ name: 'name', value: value }];
 			this._performSirenAction(action, fields);
 		}
 	}
@@ -140,6 +143,14 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 					prevent-submit>
 				</d2l-input-text>
 				${this._getNameTooltip()}
+			</div>
+
+			<div id="duedate-container">
+				<label class="d2l-label-text" for="due-date">${this.localize('dueDate')}</label>
+				<d2l-activity-due-date-editor
+					.href="${this._activityUsageHref}"
+					.token="${this.token}">
+				</d2l-activity-due-date-editor>
 			</div>
 
 			<div id="assignment-instructions-container">
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -146,21 +146,22 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 			</div>
 
 			<div id="duedate-container">
-				<label class="d2l-label-text" for="due-date">${this.localize('dueDate')}</label>
+				<label class="d2l-label-text">${this.localize('dueDate')}</label>
 				<d2l-activity-due-date-editor
+					dateLabel="${this.localize('dueDate')}"
+					timeLabel="${this.localize('dueTime')}"
 					.href="${this._activityUsageHref}"
 					.token="${this.token}">
 				</d2l-activity-due-date-editor>
 			</div>
 
 			<div id="assignment-instructions-container">
-				<label class="d2l-label-text" for="assignment-instructions">${this.localize('instructions')}</label>
+				<label class="d2l-label-text">${this.localize('instructions')}</label>
 				<d2l-activity-html-editor
-					id="assignment-instructions"
 					value="${this._instructions}"
 					.richtextEditorConfig="${this._richtextEditorConfig}"
 					@d2l-activity-html-editor-change="${this._saveInstructionsOnChange}"
-					aria-label="${this.localize('instructions')}"
+					ariaLabel="${this.localize('instructions')}"
 					?disabled="${super._entity && !super._entity.canEditInstructions()}">
 				</d2l-activity-html-editor>
 			</div>
--- a/components/d2l-activity-editor/d2l-activity-due-date-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-due-date-editor.js
@@ -2,13 +2,14 @@ import 'd2l-datetime-picker/d2l-datetime-picker';
 import { css, html, LitElement } from 'lit-element/lit-element';
 import { ActivityUsageEntity } from 'siren-sdk/src/activities/ActivityUsageEntity';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
-import { PendingContainerMixin } from 'siren-sdk/src/mixin/pending-container-mixin';
 import { SirenFetchMixinLit } from 'siren-sdk/src/mixin/siren-fetch-mixin-lit';
 
-class ActivityDueDateEditor extends PendingContainerMixin(SirenFetchMixinLit(EntityMixinLit(LitElement))) {
+class ActivityDueDateEditor extends SirenFetchMixinLit(EntityMixinLit(LitElement)) {
 
 	static get properties() {
 		return {
+			dateLabel: { type: String },
+			timeLabel: { type: String },
 			_date: { type: String },
 			_overrides: { type: Object }
 		};
@@ -73,6 +74,8 @@ class ActivityDueDateEditor extends PendingContainerMixin(SirenFetchMixinLit(Ent
 					hide-label
 					name="date"
 					id="date"
+					date-label="${this.dateLabel}"
+					time-label="${this.timeLabel}"
 					datetime="${this._date}"
 					overrides="${this._overrides}"
 					@d2l-datetime-picker-datetime-changed="${this._onDatetimePickerDatetimeChanged}"
--- a/components/d2l-activity-editor/d2l-activity-due-date-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-due-date-editor.js
@@ -1,0 +1,86 @@
+import 'd2l-datetime-picker/d2l-datetime-picker';
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { ActivityUsageEntity } from 'siren-sdk/src/activities/ActivityUsageEntity';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { PendingContainerMixin } from 'siren-sdk/src/mixin/pending-container-mixin';
+import { SirenFetchMixinLit } from 'siren-sdk/src/mixin/siren-fetch-mixin-lit';
+
+class ActivityDueDateEditor extends PendingContainerMixin(SirenFetchMixinLit(EntityMixinLit(LitElement))) {
+
+	static get properties() {
+		return {
+			_date: { type: String },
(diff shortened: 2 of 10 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -142,6 +145,14 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 				${this._getNameTooltip()}
 			</div>
 
+			<div id="duedate-container">
+				<label class="d2l-label-text" for="due-date">${this.localize('dueDate')}</label>
+				<d2l-activity-due-date-editor
+					.href="${this._activityUsageHref}"
+					.token="${this.token}">
```
ryantmer: @dbatiste A while ago you mentioned not having to bind directly to properties, but this doesn't work as just attributes. Any ideas as to why that might be? This especially confuses me here, since these are both just strings so there shouldn't be any weirdness in using them as attributes.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -142,6 +145,14 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 				${this._getNameTooltip()}
 			</div>
 
+			<div id="duedate-container">
+				<label class="d2l-label-text" for="due-date">${this.localize('dueDate')}</label>
```
mdulat: What is this label supposed to be attached to, i.e. with id `due-date`? I don't think it will work, I think you will have to use the label property of the `d2l-datetime-picker` so that screen readers will associate the label with the input component.

ryantmer: Ahh yep, the `d2l-activity-due-date-editor` previously had an ID on it. I'll fix it - I'm not sure if the HTML editor and due date editor components will have their own langterms - they don't right now, but probably will. The other option is that the entire `d2l-activity-editor` component has one set of langterms, I suppose.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -142,6 +145,14 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
...
+				<d2l-activity-due-date-editor
+					.href="${this._activityUsageHref}"
+					.token="${this.token}">
+				</d2l-activity-due-date-editor>
+			</div>
+
 			<div id="assignment-instructions-container">
 				<label class="d2l-label-text" for="assignment-instructions">${this.localize('instructions')}</label>
```
mdulat: I didn't notice this before, but I think this label attached to `assignment-instructions` using the `for` attribute won't work either, because labels can only go with specific input components. In particular <button>, <input>, <meter>, <output>, <progress>, <select>, or <textarea> element, according to https://www.w3schools.com/tags/tag_label.asp

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: token is usually a function rather than a string? 
Not sure about the href - but I've had issues with hrefs where the component first gets rendered when the href is not yet defined (e.g. this._activityUsageHref) and so it's get serialized as the string "undefined" and that causes any code in the component that tests for truthiness on the href to fail and you get requests being made to `d2l/api/something/undefined`

ryantmer: Oh wow, yeah token is definitely a function and I knew that... not sure what I was on about above! So that explains it, I'll try not using the property setter for the `href` though, since that should always be a string.

As for the undefined case, `_activityUsageHref` is only set when `d2l-activity.assignment-editor-detail`'s assignment is "a thing", so it is indeed always initially rendered with a falsy `_activityUsageHref`. This is handled gracefully within `EntityMixinLit` though, as it only fetches the href if it's truthy, so no failed network requests.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: Huh, did not know that! I can fix that one here as well - I'll check it with NVDA, which I admittedly did not do before.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdulat: I think it's more than the id existing though, it has to be associated with a specific input element. See my comment about the instructions label.

ryantmer: There are `dateLabel` and `timeLabel` properties on the datetime-picker component too - I'll see if I can just use those instead. They may not be fully accessible, but d2l-datetime-picker is not long for this world anyway.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -142,14 +145,23 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SirenFetchMixinLit(Entit
 				${this._getNameTooltip()}
 			</div>
 
+			<div id="duedate-container">
+				<label class="d2l-label-text">${this.localize('dueDate')}</label>
+				<d2l-activity-due-date-editor
+					dateLabel="${this.localize('dueDate')}"
```
mdgbayly: I think I would expect these localizations to be part of the `d2l-activity-due-date-editor` for consistency?

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -0,0 +1,90 @@
+import 'd2l-datetime-picker/d2l-datetime-picker';
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { ActivityUsageEntity } from 'siren-sdk/src/activities/ActivityUsageEntity';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { PendingContainerMixin } from 'siren-sdk/src/mixin/pending-container-mixin';
+import { SirenFetchMixinLit } from 'siren-sdk/src/mixin/siren-fetch-mixin-lit';
+
+class ActivityDueDateEditor extends PendingContainerMixin(SirenFetchMixinLit(EntityMixinLit(LitElement))) {
```
mdgbayly: I don't think these leafier components need to be pending containers?
The Pending container is on the top level entity to allow it to avoid rendering until all it's children are ready?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: oh wait - i just realized you are passing in the label maybe for some kind of a11y purpose the main label is on the outside...

mdgbayly: i guess that's an interesting question.... should the labels be part of the component or the client.

ryantmer: Yeah, I can go either way on this. In particular, in #440, I added the `assignmentActivityEditor` as a Serge project, but I'm wondering if it should instead just the `activityEditor`, and we'd have a single lang file per language, containing all langterms for the whole project? The downside to this, of course, is that when someday there is a version of this editor for quizzes and surveys, they'll be pulling in a bunch of unused langterms. But, splitting them up means that these generic, non-domain-specific components _also_ need a set of langterms. I kind of avoided the question here, but it's one that we'll have to answer sooner rather than later, I'd imagine.

(Perhaps just having a generic `activityEditor` Serge config as well as the domain-specific `assignmentActivityEditor` config would be most obvious.)

components/d2l-activity-editor/d2l-activity-due-date-editor.js
ryantmer: Oh okay - I was under the impression that the parent container only knows that its children are ready if they also extend `PendingContainerMixin`, but looking at the code it looks like they just need to extend `EntityMixinLit`, and the `PendingContainerMixin` watches for events from that?

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -0,0 +1,90 @@
...
+		this._overrides = document.documentElement.intlOverrides || '{}';
+	}
+
+	set _entity(entity) {
+		if (!this._entityHasChanged(entity)) {
+			return;
+		}
+
+		if (entity) {
+			this._date = entity.dueDate();
```
mdgbayly: dates are such a PITA....

I wonder if we would be better to pass date objects from the siren-sdk?
But that might mean resorting to using the std JS Date.parse on our ISO string from the API.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: On the note of whether these should be part of the component or the client, I suppose in this case it's called a `due-date-editor` (I tried to make it date-type-agnostic but it was suuuuper gross), so I suppose it could also render its own label? The datetime-picker is kind of like that, I'm suppressing its built-in label via the `hide-label` attribute though.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
mdgbayly: correct - yes

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: yeh - i don't think i have a strong opinion on whether the label should be part of the editor.
Having the capability to at least have it outside the editor gives a lot more leeway for integrating the due date component into an existing layout which won't necessarily be consistent across all activity types.
But yeh - it's kind of gross to have the lang term duplicated, and ideally that would be consistent across all activities.

But I also like the ability to have separate lang term files for activity specific stuff and activity general stuff.

Could we have the specific activities like assignments and quizzes each load their own lang terms but also the activity shared lang terms?

components/d2l-activity-editor/d2l-activity-due-date-editor.js
ryantmer: They very much are!

I did consider just providing the entity, but I figured that some of the utility of the `siren-sdk` is to hide a lot of that Siren-specific logic out of the realm of the consumer. I think if we did need the whole due date entity for some reason, we could just add a `dueDateEntity` getter in the `siren-sdk`.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -0,0 +1,90 @@
...
+
+		super._entity = entity;
+	}
+
+	_onDatetimePickerDatetimeCleared() {
+		this._updateDueDate('');
+	}
+
+	_onDatetimePickerDatetimeChanged(e) {
+		if (e.detail.isSame(this._date)) {
```
mdgbayly: I think my other comment relating to whether this._date should be a date or a string was also related to this. but i guess moment is maybe handling that for us? and allowing a comparison between a moment object and a date string.

I was just confused about whether we ever set `this._date` to the object we get back from the date picker. but in looking closer i guess we never do currently.

So after updating the date , we are expecting the whole component to be re-rendered after the action to save the date completes, so that `this._date` gets set to the updated date, such that if they open the date picker again, they see the last date they picked.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: I think so! I can get a PR up to add a Serge config for just `activityEditor`, which would contain all the generic langterms. Then someday when we add an activity editor for surveys or checklists, we'll add a new Serge config for `surveyActivityEditor/checklistActivityEditor` or what-have-you.

There will likely be _some_ overlap, but I don't see that as much of a concern. (e.g. it's entirely possible that somewhere in the assignment-specific stuff, we'll need to write out "due date" as well as in the generic `due-date-editor` component.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
ryantmer: Yeah, this is `moment.js` and is just a safety check more than anything, so we don't do the network request when the value hasn't actually changed. Honestly, this should really be handled within the component itself, but it isn't - if you select a date, unfocus the control, then select the same date again, it generates a `d2l-datetime-picker-datetime-changed` event - but the datetime hasn't actually changed.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
mdgbayly: it wasn't so much the whole entity i was referring to - just whether the siren-sdk `entity.dueDate()` helper should return a 'string` (as it's currently doing) or a `Date`.

ryantmer: Ah! Fair - I don't really have a strong opinion, as they're easy enough to convert. In this case we need the string, so I just went with that 😄

ryantmer:
Sounds good - I'll make a note to refactor things slightly to allow for the generic `activityEditor` langterms in Serge. Better to do it now when it's simpler!

PR #448, opened 2019-10-23
Add Serge config for generic components

ryantmer:
In #440, a Serge config was added for the `assignmentActivityEditor`, but there will be several components within `d2l-activity-editor` that are not specific to assignments. To address this (based largely off [this thread](https://github.com/BrightspaceHypermediaComponents/activities/pull/446#pullrequestreview-306128414)), this adds a new config for the generic `activityEditor` langterms. There will likely be overlap in these terms, but this is the cleanest way to keep the domain-specific and domain-generic terms from cluttering each other up. This also means that eventually, when there are other specialized activity editors (e.g. `d2l-activity-survey-editor` or something), we'll have to add new Serge configs for them as well.

In order to support this, and use the same `getLocalizationResources` in all places, a minor refactor was required such that the method now takes in `import.meta.url` from any implementing class.

### DIFF
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -57,6 +57,25 @@
       "zh-tw zh-tw"
     ]
   },{
+    "name": "activityEditor",
+    "source_dir": "components/d2l-activity-editor/lang",
+    "output_file_path": "components/d2l-activity-editor/lang/%LANG%.json",
+    "output_lang_rewrite": [
+      "ar-sa ar",
+      "de-de de",
+      "es-mx es",
+      "fi-FI fi",
+      "fr-ca fr",
+      "ja-jp ja",
+      "ko-kr ko",
+      "nl-nl nl",
+      "pt-br pt",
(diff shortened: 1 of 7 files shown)


ryantmer:
Sadly I don't think there's a built-in way to do that, but one option might be to pursue some sort of "one translation set, multiple langterm sets" approach, i.e. there's the giant (hopefully-namespaced) JSON file of all things that need to be translated, then each sub-component would generate its `lang` files based on the translations that come back from that. I don't really see a good way to structure that, sadly, but it was a theory (and I don't _think_ would require changes to `LocalizeMixin`? but would definitely require changes here...)

But ultimately, I think we might not even have to cross that bridge when we come to it, in particular due to the `context` values. Something as generic as "Due date" is probably going to be localized the same in different contexts, but other phrases may be different depending on that. Ultimately, I don't think double-translating a few things is much of a cost anyway? But [citation needed] on that since I really don't know 😕

PR #450, opened 2019-10-25
Use new siren-sdk functionality

ryantmer:
Previously, for saving name/instructions/due date, we were getting the Siren action from the `siren-sdk`, building the field values, and calling `SirenFetchMixinLit._performSirenAction`. However, the purpose of the `siren-sdk` is to abstract out Siren almost entirely, so that there is an easy to use interface on a given entity type that can be used by consumers - e.g. just call `setName(newName)`. There are two main benefits to this:

1. The UI code becomes cleaner - ideally, to the point where the UI code knows nothing about Siren at all, and only knows how to speak in the language of `siren-sdk`.
2. Consistency - first, because ideally all clients only speak `siren-sdk` and therefore changes to the actual details of the flow need to only be changed in one place. In addition, this reduces the possibility of different UIs doing slightly different things to accomplish the same task.

One caveat of this change is that the `siren-sdk` doesn't have knowledge of the events that drive a `d2l-save-status` element, as this is a UI concern. In order to overcome this, the `SaveStatusMixin` was also added here.

See also https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/78

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -9,10 +9,10 @@ import { ErrorHandlingMixin } from '../error-handling-mixin.js';
 import { getLocalizeResources } from './localization.js';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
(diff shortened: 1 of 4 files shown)


ryantmer:
@mdulat / @mdgbayly I ended up changing a name, so will requite re-approval 😬 Sorry, I forgot that this repo auto-dismisses review on changes! (As it probably should but I was trying to sneak in the name change.)

ryantmer:
...and then I merged a different change that conflicted with this so I need another re-approval 🤦‍♂

PR #451, opened 2019-10-25
[experimental] add a helpful script to add lang terms

maboelsoudd2l:
Sometimes, i hate having to add lang terms manually.
So instead I created this little script that might speed up the process.
Any feedback is welcome, it's by no means polished.


components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,57 @@
...
+const [command, langKey, langPhrase] = params;
+
+switch (command) {
+	case "add":
+		appendLangTerm(langKey, langPhrase);
+		break;
+	case "remove":
+		removeLangTerm(langKey, langPhrase);
+		break;
+	case "change": // add and change do the same thing, but easier in terms of usability
```
AlexBedley: you could combine this with the `add` case like:

```js
case "add":
case "change":
    appendLangTerm(...);
```

components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,57 @@
...
+		appendLangTerm(langKey, langPhrase);
+		break;
+	default:
+		throw "invalid command"
```
AlexBedley: nit: we should throw an `Error` object, not a string (probably doesn't matter here but a good habit to get into)

components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function removeJsonLangTerm(langKey, filePath) {
+	const parsedJSON = JSON.parse(fs.readFileSync(filePath));
+	delete parsedJSON[langKey]
+	fs.writeFileSync(filePath, JSON.stringify(parsedJSON, null, 3).replace(/": "/g, '" : "')+"\n");
+}
+
+const params = process.argv.slice(2);
+if (params.length < 2){
```
AlexBedley: shouldn't this be `< 3`?

cbusmand2l:
Could we get something added to the readme? Even if just to let people know it's an option, but an example usage would also be useful.

components/d2l-quick-eval/lang/insert-lang-term.js
maboelsoudd2l: the "remove" takes one less argument. This was my way of saying "if you have 2 arguments, you could be doing something wrong, but if you have 1 argument you are DEFINITELY doing something wrong"

AlexBedley: Ah ok I was confused since we call `removeLangTerm` with 2 arguments. It might be clearer if we parse `params` fully after parsing `command`

maboelsoudd2l:
I'll revisit this in a few days once we have figured out what to do with the lang-term debacle.

m6jones:
This is awesome. I would like to see context added.

maboelsoudd2l:
@m6jones we don't have time in our sprint to add context. maybe a subsequent PR?

components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function CLIRemoveLangTerm(langKey) {
+	if (!langKey) throw 'add the lang term that you want to remove';
+	traverseJSONDir(LANG_PATH, removeJsonLangTerm.bind(null, langKey));
+}
+
+function CLISyncJsonFiles() {
+	const enJsonFile = path.resolve(LANG_PATH, 'en.json');
+	const jaJsonFile = path.resolve(LANG_PATH, 'ja.json');
```
jeroach: Why are we checking against `ja` only? Shouldn't we check all?

components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function removeJsonLangTerm(langKey, filePath) {
+	const parsedJSON = JSON.parse(fs.readFileSync(filePath));
+	delete parsedJSON[langKey];
+	fs.writeFileSync(filePath, JSON.stringify(parsedJSON, null, 3).replace(/": "/g, '" : "') + '\n');
+}
+
+function CLIAppendLangTerm(langKey, langPhrase) {
+	if (!langPhrase) throw 'add the lang term that you want to add/modify and the content of the lang term';
```
jeroach: nit: IMO would be more readable if this was on multiple lines w/ brackets

Same goes for the function below

components/d2l-quick-eval/lang/insert-lang-term.js
maboelsoudd2l: I'm honestly conflicted on this. My intent is to check the english file for any new lang terms and any removed lang terms. That way, in order to add new lang terms, just add to the `en.json` file, and in order to remove lang terms just remove them from `en.json`.
 The way i thought of doing it is to compare the `en.json` with any other lang file. So I just assumed that all lang files _other_ than `en.json` would be identical.

I honestly don't have high expectations for robustness, since this is an internal tool (which will likely get phased out soon anyways by the serge integration once i get it to work).

jeroach: If this is indeed a temporary measure, then I suppose that assumption is okay.

components/d2l-quick-eval/lang/insert-lang-term.js
```diff
@@ -0,0 +1,83 @@
+/* eslint-env node, es6 */
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const LANG_PATH = path.resolve('components/d2l-quick-eval/lang');
```
dmackenz: You should consider using path.join for OS compatibility.

Also, you should use __dirname because this assumes which folder you are executing the script from.

maboelsoudd2l: sank you

dmackenz: Another thing that I think would be really nice is if this was built as an npm module instead of inside of the codebase. That way, I can just call it from the command line and don't need to worry about which branch I'm on (since this isn't merged) or what folder I'm executing it from.

maboelsoudd2l:
closed in favor of this https://github.com/BrightspaceHypermediaComponents/activities/pull/612

PR #452, opened 2019-10-28
US110198 Add Submission Type select list

ryantmer:
This adds a styled `select` element that presents the options for `submissionType` on the assignment. There are two caveats to this change:

1. The `completionType` dropdown is not included, which means the On Paper and Observed in Person submission types can't be used correctly (they will both just use "Automatically on due date" as the completion type for now)
2. The styles for the `select` have just been copy+pasted into a `select-styles` file, but are planned to be separated out somewhere common.

Both of these should be rectified soon - the first by another US, and the second once some further discussions occur with Gaudi et al.

This also includes some minor changes to not access `super._entity` directly for getters, which is somewhat bad form it turns out - instead, it's better to use local values that are set in the `_onAssignmentChange` handler.

Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/80

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -26,7 +26,8 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 			_richtextEditorConfig: { type: Object },
 			_canEditInstructions: { type: Boolean },
 			_activityUsageHref: { type: String },
-			_submissionTypes: { type: Array }
+			_submissionTypes: { type: Array },
+			_canEditSubmissionType: { type: Boolean }
 		};
 	}
 
@@ -40,6 +41,11 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 			}
 			:host > div {
 				padding-bottom: 20px;
+			}
+
+			select {
+				width: 250px;
+				display: block;
 			}
 		`];
 	}
@@ -75,6 +81,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		this._canEditInstructions = assignment.canEditInstructions();
 		this._activityUsageHref = assignment.activityUsageHref();
 		this._submissionTypes = assignment.submissionTypeOptions();
+		this._canEditSubmissionType = assignment.canEditSubmissionType();
 	}
 
 	_saveOnChange(jobName) {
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -154,9 +185,16 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 					.richtextEditorConfig="${this._richtextEditorConfig}"
 					@d2l-activity-html-editor-change="${this._saveInstructionsOnChange}"
 					ariaLabel="${this.localize('instructions')}"
-					?disabled="${super._entity && !super._entity.canEditInstructions()}">
+					?disabled="${!this._canEditInstructions}">
 				</d2l-activity-html-editor>
 			</div>
+
+			<div id="assignment-submission-type-container">
+				<label class="d2l-label-text">${this.localize('submissionType')}</label>
```
mdulat: `select` is an input type which the label `for` attribute can be bound to, so this should have a `for` corresponding to an id on the `select`.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -20,14 +21,17 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		return {
 			_name: { type: String },
 			_nameError: { type: String },
+			_canEditName: { type: Boolean },
 			_instructions: { type: String },
 			_richtextEditorConfig: { type: Object },
-			_activityUsageHref: { type: String }
+			_canEditInstructions: { type: Boolean },
+			_activityUsageHref: { type: String },
+			_submissionTypes: { type: Array }
```
mdulat: Is it possible for a user to not have permissions to edit a submission type?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: Someday I'll get this right 😅

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: Good question, and I don't actually know the answer - but I'd argue knowing the answer is unimportant, since this should have a check for it anyway! I'll add that.

PR #453, opened 2019-10-29
US111526 Add completion type selector

ryantmer:
Most of the logic for this selector is done in `siren-sdk`, so all it needs to do is call `setCompletionType`. This change also removes the debounce autosave from these dropdowns, as it causes odd UI behaviour and lags the showing/hiding of the completion type selector.

Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/84

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -27,7 +27,10 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 			_canEditInstructions: { type: Boolean },
 			_activityUsageHref: { type: String },
 			_submissionTypes: { type: Array },
-			_canEditSubmissionType: { type: Boolean }
+			_canEditSubmissionType: { type: Boolean },
+			_completionTypes: { type: Array },
+			_canEditCompletionType: { type: Boolean },
+			_showCompletionType: { type: Boolean }
 		};
 	}
 
@@ -44,7 +47,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 			}
 
 			select {
-				width: 250px;
+				width: 300px;
 				display: block;
 			}
 		`];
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -82,6 +86,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		this._activityUsageHref = assignment.activityUsageHref();
 		this._submissionTypes = assignment.submissionTypeOptions();
 		this._canEditSubmissionType = assignment.canEditSubmissionType();
+		this._completionTypes = assignment.completionTypeOptions();
+		this._canEditCompletionType = assignment.canEditCompletionType();
+		this._showCompletionType = assignment.submissionType().value === 2 || assignment.submissionType().value === 3;
```
mdulat: Can this be computed based on whether the current `submissionType` has any associated `completionTypeOptions` instead of hard-coding the specific submission types?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -131,13 +134,14 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 	}
 
 	_saveSubmissionTypeOnChange() {
-		const submissionType = this.shadowRoot.querySelector('select').value;
+		const submissionType = this.shadowRoot.querySelector('select#assignment-submission-type').value;
+		this._showCompletionType = submissionType === 2 || submissionType === 3;
```
mdulat: Wouldn't this get set after saving the submission type? Is it done here because of lag when saving?

ryantmer: Yeah, this is just for UI responsiveness 😢 There isn't a huge delay when running locally, but it could be a few hundred ms on a normal instance.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ryantmer: It occurs to me that this is actually much easier to compute actually - `assignment.completionTypeList().length > 0` would do the same thing, I think!

PR #456, opened 2019-10-30
Using lazy option loading, returning skeleton and search summary to be dependent on filter loading state.

cbusmand2l:


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -124,7 +124,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			<div class="d2l-quick-eval-activity-list-modifiers">
 				<d2l-hm-filter
 					href="[[filterHref]]"
-					token="[[token]]">
+					token="[[token]]"
+					lazy-load-options>
```
cbusmand2l: Adding this attribute (and its equivalent in submissions) is the added lazy behaviour. The rest is reversing the original lazy loading implementation as there would have been issues with the UI sometimes displaying inaccurate (late) information.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -388,8 +388,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		return _data.length && !_showLoadingSkeleton;
 	}
 
-	_computeShowLoadingSkeleton(_loading, searchLoading) {
-		return _loading || searchLoading;
+	_computeShowLoadingSkeleton(_loading, filtersLoading, searchLoading) {
+		return _loading || filtersLoading || searchLoading;
```
maboelsoudd2l: wasn't this the cause of an issue we had in https://github.com/BrightspaceHypermediaComponents/activities/pull/442?

AlexBedley: yes but it's no longer an issue due to the changes that Carol made in hm-filter and we need to revert the change in #442

maboelsoudd2l: cool just being thorough

PR #458, opened 2019-10-31
Refactor polymer localization

maboelsoudd2l:
The current method we store lang files in js is by storing them as a mixin, which makes certain assumptions about frameworks and structure of the repo.

This becomes problematic at times like this where we might want to consider upgrading.

I also realized that the localization process is shared across `d2l-quick-eval`, `d2l-evaluation-icon` and `d2l-activity-list-item`. Hence updating one means all the others at once.

My long term goal is to be able to easily adapt these existing lang term files to be used used in a Lit-component as well as a Polymer-component. But in order to do this I need to remove the `dedupingMixin` middleware.

This PR will be affecting a number of different projects, so I have added a number of people who are working on these repos (to the best of my knowledge, feel free to add yourself as a reviewer).


components/d2l-activity-editor/build/lang/en.js
```diff
@@ -0,0 +1,11 @@
+export const LangEn = {
```
maboelsoudd2l: @ryantmer not sure what to do about this generated file

components/d2l-activity-editor/d2l-activity-assignment-editor/build/lang/en.js
```diff
@@ -0,0 +1,23 @@
+export const LangEn = {
```
maboelsoudd2l: @ryantmer this one too

components/d2l-quick-eval/build/lang/ar.js
```diff
@@ -1,72 +1,60 @@
...
-	}
+export const LangAr = {
+	'activities': 'النشاطات',
+	'activityName': 'اسم النشاط',
+	'assignment': 'الفرض',
+	'caughtUp': 'لقد انتهيت من مهامك!',
+	'checkBackOften': 'تحقق باستمرار من احتمال توفر عمليات إرسال جديدة.',
+	'clearSearch': 'مسح البحث',
+	'close': 'إغلاق',
+	'completed': 'تم الإكمال',
```
maboelsoudd2l: I had to re-indent this because eslint was sad 😢

templates/lang-mixin.ejs
```diff
@@ -1,14 +1,2 @@
...
-/* @polymerMixin */
-const Lang<%- data.properLang %>Impl = (superClass) => class extends superClass {
-	constructor() {
-		super();
-		this.<%- data.lang %> = <%- data.resources %>;
-	}
-};
-
-export const Lang<%- data.properLang %> = dedupingMixin(Lang<%- data.properLang %>Impl);
+export const Lang<%- data.properLang %> = <%- data.resources %>;
```
maboelsoudd2l: This template affects all projects in this repo. We can discuss if we don't want it to apply to every project.

components/d2l-quick-eval/QuickEvalLocalize.js
```diff
@@ -18,6 +18,15 @@ import {LangTr} from './build/lang/tr.js';
 import {LangZhtw} from './build/lang/zh-tw.js';
 import {LangZh} from './build/lang/zh.js';
 
+const LangImpl = (prefix, langObj, superClass) => class extends superClass {
```
maboelsoudd2l: this `LangImpl` tries to do exactly what exists in the js lang-term files.

components/d2l-quick-eval/QuickEvalLocalize.js
```diff
@@ -39,9 +48,10 @@ const QuickEvalLocalizeImpl = (superClass) => {
 		LangZh
 	];
 	let mixinLang = mixinBehaviors([D2L.PolymerBehaviors.LocalizeBehavior], superClass);
-	for (const langMixin of langMixins) {
-		mixinLang = langMixin(mixinLang);
-	}
+	LANGUAGES.forEach((langPrefix, index)=> {
+		mixinLang = dedupingMixin(LangImpl.bind(null, langPrefix, langMixins[index])).call(null, mixinLang);
```
maboelsoudd2l: this does the same exact process that existed before, but more compressed
(tested on the demo pages in multiple languages)

components/d2l-activity-editor/build/lang/en.js
ryantmer: I think they're fine to leave in - there are separate Serge configs for each project in this repo, but it makes some sense to me to have a unified langtermification model for all of them. `d2l-activity-editor` is new, so doesn't even have its first set of translations yet (see #457), but if this method works for other projects then it should work for `d2l-activity-editor` too (the only difference being that `d2l-activity-editor` is entirely Lit)

maboelsoudd2l: my thought process is that Lit or Polymer, new or old, I know that we need a js file. so I figured that you won't mind that I generated some js files for you. I would like to know  whether this format works for you (or maybe you don't like it :(((((    ).

components/d2l-activity-evaluation-icon/ActivityEvaluationIconBaseLocalize.js
```diff
@@ -16,6 +16,15 @@ import {LangTr} from './build/lang/tr.js';
...
 import {LangZh} from './build/lang/zh.js';
 
+const LangImpl = (prefix, langObj, superClass) => class extends superClass {
+	constructor() {
+		super();
+		this[prefix] = langObj;
+	}
+};
+
+const LANGUAGES = ['ar', 'de', 'en', 'es', 'fi', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh' ];
```
AlexBedley: is `da` missing from this list?

maboelsoudd2l: yeah, they don't use it, but we do. I don't know why

components/d2l-activity-editor/build/lang/en.js
ryantmer: Haha, no it's good! It's pretty much exactly what would end up getting generated by #431 anyway, which is what I plan on working on during the upcoming Inspiration sprint. (In effect, this means that for this component specifically, running the gulpfile after a Serge translation PR _should_ have no effect - since the results from Serge would exactly match this file! 😄)

components/d2l-activity-evaluation-icon/ActivityEvaluationIconBaseLocalize.js
```diff
@@ -16,6 +16,15 @@ import {LangTr} from './build/lang/tr.js';
 import {LangZhtw} from './build/lang/zh-tw.js';
```
maboelsoudd2l: @AlexBedley notice that `da` wasn't imported

components/d2l-quick-eval/QuickEvalLocalize.js
```diff
@@ -18,6 +18,15 @@ import {LangTr} from './build/lang/tr.js';
 import {LangZhtw} from './build/lang/zh-tw.js';
```
maboelsoudd2l: @AlexBedley but here `da` is imported

components/d2l-activity-evaluation-icon/ActivityEvaluationIconBaseLocalize.js
ryantmer: https://translate.google.com/#view=home&op=translate&sl=ru&tl=en&text=da

components/d2l-activity-editor/build/lang/en.js
dbatiste: @ryantmer : with the Serge integration, would the `context` still be present?  Wouldn't the output from Serge just be the key and translation?  ex. `dueDate: 'Due Date'`?

ryantmer: You make a good point! Strangely we got back translated `context` strings this time, which isn't supposed to happen - so these files would actually look slightly different due to using `context. I think the options in that case are therefore

1. Stop using `context`
2. Use `context` everywhere
3. Ignore the gulpfile output for `d2l-activity-editor` and just use the Serge stuff

It would be neato to use `context` everywhere, since I believe @m6jones said it gets us closer to having same-day translations, but that would also be a larger change - so for now, I'm good with doing 3 and just having Serge replace these files upon translation completion.

maboelsoudd2l: isn't context the extra info about the lang term? why don't we just ignore the `context`?

ryantmer: It is yeah - it should only be present in the JSON file, not the JS output. However, until I figure out what happened with our Serge translations (where the `context` strings themselves got translated in some languages), I'd say no issue here - Serge should take the JSON file (with `context`) and produce a JSON file _without_ context, but it didn't.

In other words, these changes are fine from my perspective; more changes may be required, but until I figure out what happened with the most recent translation, and until I finish the Serge changes, 🤷‍♂

dbatiste: Yeah the `context` is just for the translators, so they can provide a better translation.  It's not a super big deal, but we don't really want to be sending it down to the browser.  I'd actually expect the `translation` key to be removed as well.  

So if we start with this JSON:
```
'dueDate': {
   'translation': 'Due Date',
   'context': 'ARIA label for the due date field when creating/editing an activity'
},
```
I'd expect this to be in the JS:
```
'dueDate': 'Due Date'
```

ryantmer: Yeah - it seems to me like the source JSON file would be the first format you've used, then the JSON/JS we get back from Serge would be the second format - but in #457, the JSON returned from Serge is the first format. Still working to track down why that is, as it seems to be the right format.

maboelsoudd2l: Personally i think the inclusion of `context` would not be directly relevant to this PR, since the context issue would be present regardless of my change.
That being said if we are trying to get rid of context we should modify the gulpfile.
Specifically i see that the JSON gets stringified like so:
```
JSON.stringify(localeResources[lang], null, '\t\t\t').replace(/'/g, '\\\'').replace(/"/g, '\'').replace(/\n\}/g, '\n\t\t}'),
```
i would add a pre-processing step that removes the context out of every lang term:
```
delete localeResources["en"]["dueDate"]["context"];
```
but in some kind of for loop.

ryantmer: Yep - there are (going to be) two ways to get the built lang files for `d2l-activity-editor`: either via this gulpfile, or from Serge. Both should give the same results, but right now the gulpfile isn't stripping the `context`, which Serge would. That said, I'll be working on this stuff during the Inspiration sprint after Infusion, so will look into it then. For now, I'm basically saying do whatever you want to the files in `d2l-activity-editor/lang/build`, since we're not actually referencing them in the code anyway :)

maboelsoudd2l:
@AlexBedley spoke offline and I followed your suggestion, created diff tests for quick eval and found & addressed these issues https://github.com/BrightspaceHypermediaComponents/activities/pull/458/commits/f32764fe78b4dbb10cd34b74721e5c156908d046 and https://github.com/BrightspaceHypermediaComponents/activities/pull/462

anything else?

components/d2l-activity-editor/build/lang/en.js
ryantmer: To close the loop here, mystery solved about the `context` being translated - I didn't include `"parser_plugin": { "plugin": "parse_d2l_fra" }` in the config for my project - adding it in #464 so that it works correctly in the next translation run.

As for this thread in general, these files LGTM - whether it's being generated by Serge or by a build step, these are the intended output files (or, will be, once the above PR is merged and the next translation run comes back).

maboelsoudd2l:
changes have been merged in https://github.com/BrightspaceHypermediaComponents/activities/pull/470

PR #459, opened 2019-10-31
US111620 Add plaintext instructions editor

ryantmer:
This passes down a user config setting to determine whether to use the HTML editor or a plaintext editor for the instructions. This could be represented in an API response somehow, but after some thought and trying to flesh that out, this was a much simpler solution. It's unfortunate that the interim components needs to pass down an attribute that they don't really care about (`htmlEditorEnabled`), but it's not all that much more code or complexity, and the presence of other text editing fields in the future would be easy to support.

The new `d2l-activity-text-editor` acts more-or-less the same as the HTML editor did, but now will display a `d2l-input-textarea` when the user doesn't want to use richtext editing.

Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/86

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -1,6 +1,6 @@
 import 'd2l-inputs/d2l-input-text.js';
 import '../d2l-activity-due-date-editor.js';
-import '../d2l-activity-html-editor.js';
+import '../d2l-activity-text-editor.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
@@ -17,11 +17,12 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 
 	static get properties() {
 		return {
+			htmlEditorEnabled: { type: Boolean },
 			_name: { type: String },
 			_nameError: { type: String },
 			_canEditName: { type: Boolean },
 			_instructions: { type: String },
-			_richtextEditorConfig: { type: Object },
+			_instructionsRichTextEditorConfig: { type: Object },
 			_canEditInstructions: { type: Boolean },
 			_activityUsageHref: { type: String },
 			_submissionTypes: { type: Array },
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -81,7 +82,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		this._name = assignment.name();
 		this._canEditName = assignment.canEditName();
 		this._instructions = assignment.instructionsEditorHtml();
-		this._richtextEditorConfig = assignment.getRichTextEditorConfig();
+		this._instructionsRichTextEditorConfig = assignment.instructionsRichTextEditorConfig();
```
ryantmer: This comes from https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/86 - I just realized that the previous name was pretty misleading, as this config comes from a single richtext field.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -13,6 +13,7 @@ class AssignmentEditor extends PendingContainerMixin(EntityMixinLit(LocalizeMixi
 
 	static get properties() {
 		return {
+			htmlEditorEnabled: { type: Boolean },
```
ryantmer: @mdgbayly We talked about this a while back, and how passing down parameters is a bit of a bummer, but I looked deeper into the idea of adding a settings HM domain and it just looks like it would be a case of https://xkcd.com/927/ - there are dozens of API routes that return settings in the LMS (even a few in HM already), so while I still think the idea is sound, it would probably require changing at least the existing HM settings routes over to use it, otherwise we're adding more standard 😢 (In addition, this isn't prescriptively only about assignment instructions at least - this applies to all richtext fields in any activity editor, present or future.)

components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -0,0 +1,64 @@
...
+		}));
+	}
+
+	render() {
+		if (this.htmlEditorEnabled) {
+			return html`
+				<d2l-activity-html-editor
+					ariaLabel="${this.ariaLabel}"
+					value="${this.value}"
+					?htmlEditorEnabled="${this.htmlEditorEnabled}"
```
mdulat: Is `htmlEditorEnabled` needed here?

components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -0,0 +1,64 @@
...
+					?disabled="${this.disabled}"
+					@d2l-activity-html-editor-change="${this._onRichtextChange}"
+					.richtextEditorConfig="${this.richtextEditorConfig}">
```
mdulat: Is it intentional that `.richtextEditorConfig` has a `.` in front? What does it mean?

components/d2l-activity-editor/d2l-activity-text-editor.js
ryantmer: Hah, sure isn't - I'll remove it.

components/d2l-activity-editor/d2l-activity-text-editor.js
ryantmer: It's intentional - it's the LitElement syntax for [property binding](https://lit-element.polymer-project.org/guide/templates#bind-properties-to-templated-elements). I think the alternative would be to explicitly `JSON.stringify` the config, so this ends up being nicer. (Without the `.`, this just doesn't work at all.)

PR #461, opened 2019-11-01
Adding "more" action button with dropdown.

cbusmand2l:
Left side shows default, hover, and focused states, right side shows the expanded dropdown (screenshots taken from the demo; black outline is not part of the button but only to show how the button is aligned within its parent).
image
Added icon override to the base action button (`more` only has a t1 version).


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -0,0 +1,42 @@
...
+			}
+			d2l-dropdown {
+				height: 100%;
+				width: 100%;
+			}
+		`;
+	}
+	render() {
+		return html`
+			<d2l-dropdown text="More Actions">
```
maboelsoudd2l: perhaps we can add a story to add lang terms to this button and the ellipsis button once https://github.com/BrightspaceHypermediaComponents/activities/pull/458 is merged

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
```diff
@@ -55,6 +56,14 @@ <h3>d2l-quick-eval-activity-card demo</h3>
 				<template strip-whitespace>
 					<div class="button-wrapper-big"><d2l-quick-eval-activity-card-action-button text="button text" icon-name="publish-all"></d2l-quick-eval-activity-card-action-button></div>
 					<div class="button-wrapper-big"><d2l-quick-eval-activity-card-action-button text="button text" icon-name="publish-all" disabled></d2l-quick-eval-activity-card-action-button></div>
+					<div class="button-wrapper-big">
+						<d2l-quick-eval-activity-card-action-button-more>
+							<d2l-menu-item text="First thing"></d2l-menu-item>
```
maboelsoudd2l: nit: there is an implicit assumption to use `d2l-menu-item` here. I wouldn't have known unless I read the docs at https://github.com/BrightspaceUI/dropdown.

Not sure how to mitigate this though.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -140,6 +140,10 @@ class D2LQuickEvalActivityCardActionButton extends mixinBehaviors(
 			iconName: {
 				type: String
 			},
+			// Icon override, in case you don't want the default behaviour (t2 on small widths, t3 on large)
+			iconFullName: {
+				type: String
```
maboelsoudd2l: maybe we should be adding the `more` icon in t2 and t3? i remember looking for it for my spike and didn't find it except in t1.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -0,0 +1,42 @@
...
+				<d2l-quick-eval-activity-card-action-button 
+					icon-full-name="tier1:more"
+					text="More Actions"
+					class="d2l-dropdown-opener"></d2l-quick-eval-activity-card-action-button>
+				<d2l-dropdown-menu>
+					<d2l-menu label="More Actions">
```
maboelsoudd2l: When creating the ellipsis menu, I remember using this label, but I felt that it didn't really do anything. You can't see it anywhere, and it might show up in the tooltips but I think Dan didn't put that in the specs.

I'm interested to know what you think.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
maboelsoudd2l: I'm assuming we cant use `d2l-dropdown-more`?

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
```diff
@@ -18,6 +18,7 @@
 			import 'd2l-typography/d2l-typography.js';
```
maboelsoudd2l: I'm assuming we'll be adding visual diff coverage?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
dbatiste: This label is used for a11y: [it is rendered as an `arial-label`](https://github.com/BrightspaceUI/menu/blob/master/d2l-menu.js#L216) on the menu. It is required for the root menu (like this one)... whereas for nested menus, the label is automatically set based on the menu-item parent/opener of the nested menu.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
dbatiste: Ditto : the purpose of `d2l-dropdown-more` was to provide consistency.  If we need to diverge, I'd be asking if this is really a snowflake, or if something new needs to be added to the core component.

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
dbatiste: Hopefully this works out ok.  The prebuilt-components are slightly different, in that they require the consumer to provide the `d2l-menu`, which results in the menu and menu-items in the same DOM scope. They may also do it that way because nested menus are possible... and something like the following would be a little odd:
```
d2l-some-dropdown-menu
      -- shadow-dom
         d2l-menu
   d2l-menu-item
      d2l-menu
         d2l-menu-item
```

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
cbusmand2l: One problem with using `d2l-dropdown-more` is that the icon won't be aligned with the ones in the other action buttons.
The top of the hover area (which includes padding around the icon) is aligned to the top of the parent
image
vs the top of the icon being aligned.
image
We can nudge it around with absolute positioning, but I feel like that's just asking for trouble.

Also as far as I can tell, `d2l-dropdown-more` only shows hover effects when you're directly over the icon, and only allows clicking directly on the icon, so we'd need special css and click handlers to allow hovering/clicking on the entire region (as with the other action buttons).

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
cbusmand2l: Yes - since this is a net-new component and my first foray into Lit, I wanted to hash out any issues with the actual component first, then I'm going to add in visual-diff.

demo/d2l-quick-eval/d2l-quick-eval-activity-card-action-button.html
cbusmand2l: We can add a comment to the component that specifies to use the menu item? If we only use this component in quick eval then I don't think that's too big a deal.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
dlockhart: Those seem like pretty serious bugs that we should fix dropdown. I'm not seeing any of that behaviour though in other places that use `d2l-dropdown-more` (in the demos for the component or in the Brightspace navbar)... so I'm curious what's different about your case.

maboelsoudd2l:
visual diff-bot is sad, perhaps we should regenerate the goldens?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
dlockhart: After chatting with @cbusmand2l and seeing the design she's trying to implement, I think using a custom opener instead of `d2l-dropdown-more` (which is just a subtle button) is the right approach. If this pattern of using a subtle button with text under it starts to show up elsewhere, we should centralize it.

cbusmand2l:
@AlexBedley No idea what the deal is there, since I didn't touch them in the PR (in fact, I don't think they were even in this branch originally).

Now that I've done the merge from master, I can actually see the files now and I'm hoping I can just fix them manually.

maboelsoudd2l:
The PR should show the changes compared to master, if these lang terms are in master, then they shouldn't show up here, right?

AlexBedley:
@CarolBusman @maboelsoudd2l yeah looks some weird git thing is happening here, lemme know if you need helping sorting it out

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -140,6 +140,10 @@ class D2LQuickEvalActivityCardActionButton extends mixinBehaviors(
 			iconName: {
 				type: String
 			},
+			// Icon override, in case you don't want the default behaviour (t2 on small widths, t3 on large)
+			iconFullName: {
```
AlexBedley: nit: I would have named this `iconNameOverride` and gotten rid of the comment

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -0,0 +1,42 @@
...
+			}
+		`;
+	}
+	render() {
+		return html`
+			<d2l-dropdown text="More Actions">
+				<d2l-quick-eval-activity-card-action-button 
+					icon-full-name="tier1:more"
+					text="More Actions"
+					class="d2l-dropdown-opener"></d2l-quick-eval-activity-card-action-button>
```
AlexBedley: nit: Is this class `d2l-dropdown-opener` used?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -0,0 +1,42 @@
...
+				width: 100%;
+			}
+		`;
+	}
+	render() {
+		return html`
+			<d2l-dropdown text="More Actions">
+				<d2l-quick-eval-activity-card-action-button 
+					icon-full-name="tier1:more"
+					text="More Actions"
```
AlexBedley: for my understanding: will this eventually be replaced by a langterm?

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
cbusmand2l: Since we're using a "custom" button for the dropdown, this is how we mark which element opens the dropdown.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
cbusmand2l: Yes. When I add the button to the activity card, I'll put in langterms. They weren't added here as they weren't yet ready for lit components.

PR #462, opened 2019-11-01
fix overtranslation of korean langterms

maboelsoudd2l:
It seems that the translators did not translate it properly, we will have to ask them to retranslate them. At this point I'm just doing damage control so that the master branch wouldn't crash when the language is switched to korean. 

I also did a rough check in all the other languages to see if this problem exists elsewhere but didn't find anything else.


Surekh:
Thanks, @maboelsoudd2l. We had shared specific instructions for translators to support translating this specific term (ICU format), it looks like it was not followed for Korean. Let us merge this fix and Tina will be following up with the translation team to get a new Korean translation for this term.

maboelsoudd2l:
lol Alex `other` is the only thing it shouldn't translate 😆

maboelsoudd2l:
image
So this is a screenshot of a typical ICU message
The "one", "other" are part of the syntax, they shouldn't be translated, just like the variables we use

AlexBedley:
Ah ok I get it. thanks!

PR #468, opened 2019-11-12
Use Serge's parse_js parser

ryantmer:
In the ongoing saga of a better integration between Serge and our components, this converts FACE to use [the `parse_js` parser](https://serge.io/docs/plugins/parser/parse_js/) in Serge. This addresses most of the requirements we had for improved i10n:

- Works with Serge
- Language files include context - for the JS parser, this is provided in the form of a comment after the langterm
- No build step - Serge takes the ES6 modules as inputs, and produces ES6 modules as output
- Only change one file - we'll only ever touch `en.js`
- No gulp!

This approach partially addresses the following:

- Automatically sort the langterms - not explicitly, but Serge returns them sorted it seems
- Easy to develop and test without requiring Serge - assuming it's existing langterms, yes

This explicitly _doesn't_ address the following:

- ~Language files per component - as far as I can tell, this is not actually possibly in conjuction with `import`, see below~ Does now!
- Github action to check langterms - not explicitly forbidden by this approach, just not addressed by it either.

~Worth noting that this sort of undoes #448, and flattens FACE back down to one set of langterm files - this is because if `import` is handed anything other than an absolute path, bundling doesn't work correctly. Not sure what the best solution is here, but generally we have only one lang folder per component anyway, so maybe not a big deal.~ No longer the case, landed on a solution that still has multiple `lang` folders.


mdgbayly:
I'm not super keen on having to merge the different activities lang terms into a single file.
Is this just a limitation of the polymer build/bundler? After all the point of dynamic imports is that they are supposed to be dynamic. (https://v8.dev/features/dynamic-import).

Are there other approaches to identifying the lang files to the polymer build e.g. as fragments, that would still allow them to be resolved at runtime?

ryantmer:
Hmm, fair point - I'll look into that. Other projects generally seem to only have one `lang` folder, which made me think that perhaps the structure of `d2l-activity-editor` is strange - i.e. if we're defining different `lang` folders, perhaps we should be splitting things up even more than that. However, that might be overly-split-up-happy, and I totally forgot that `sources` and fragments and such are things (although they're Polymer things, which FACE isn't, _but_ it's still being build by polymer build, so I guess we can leverage that!)

ryantmer:
Sadly, it appears none of the options work if defined in FACE's `polymer.json`. This makes some sense, since it's BSI that's building everything (I figured it would obey each sub-component's `polymer.json`, but perhaps not?), but it means we'll either need to leave the [`extraDependencies` entry](https://github.com/Brightspace/brightspace-integration/blob/39f2e6f71c34c5a6e2e47bbcd0426826623a3fc8/polymer.json#L29) in BSI, or `localization.js` will become very application-specific. (For other projects like quick eval with a single `lang` folder for the whole project, this isn't an issue as it can use constants to refer to its lang files.)

components/d2l-activity-editor/lang/fr.js
```diff
@@ -0,0 +1,9 @@
+export default {
+	assignmentCompletionType: 'Le Marked as completed', // Label for the completion type field when creating/editing an assignment
```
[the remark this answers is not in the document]
ryantmer: It is not, oops 😆 This was for testing, forgot to remove them. I'll do so with an upcoming change to address Martin's concerns above.

components/d2l-activity-editor/localization.js
```diff
@@ -1,39 +1,27 @@
...
-		if (cache[url]) {
-			return cache[url];
+export async function getLocalizeResources(langs) {
+	const imports = [];
+	let supportedLanguage;
+	for (const language of langs.reverse()) {
+		switch (language) {
+			case 'en':
+				supportedLanguage = 'en';
+				imports.push(import('./lang/en.js'));
```
maboelsoudd2l: Why dynamically import languages when it looks like they will be bundled in BSI? I read this documentation https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.md

and they recommend `LocalizeStaticMixin` for static lang files.

maboelsoudd2l: if we're gonna do dynamic imports, then might as well stick with JSONs

ryantmer: Not sure I follow your question; to clarify, there are bundled and bundled files in BSI, and the approach here (which I'm about to modify slightly) will result in lang files being included by BSI, but not bundled into the base package. This is what we want, as we don't want to be sending along (potentially large) langterm files for languages we're not supporting. On the other hand, JSON files are not included _at all_ by BSI, unless explicitly specified - which leads to: the benefit of JS over JSON in this case is that we don't need to tell BSI to [explicitly include those files](https://github.com/Brightspace/brightspace-integration/blob/39f2e6f71c34c5a6e2e47bbcd0426826623a3fc8/polymer.json#L29), and can instead rely on `polymer build`'s knowledge of `import` to include the JS files for us.

On the other hand, using `LocalizeStaticMixin` means the langterms are bundled into the base package for all languages - this is a good approach for small components since it's simpler, but for more complex stuff like FACE it'll just mean we have a bundled package that includes all languages (i.e. no lazy-loading).

maboelsoudd2l: What's the difference between it being included by BSI, vs bundled in the base package? I imagine that BSI generates one massive file that has EVERYTHING. What difference does it make whether to dynamically import or not? If BSI loads one file with all the js, then there is no _real_ lazy-loading

ryantmer: It creates a number of files - if you check in the `build/es6-bundled` folder of your local BSI you can see the result. For FACE, it creates a single `d2l-activity-editor.js` file, but that doesn't actually include any langterms; instead, the code above pulls them in when they're needed based on the document language.

That said, it is worth noting that the `d2l-activity-editor.js` file that is created is like 250KB, so it's not like we're saving the world by excluding a few KB of language files - but it's an easy enough thing to not include languages we don't need that it's worth doing to keep that size a little smaller, at least.

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/fr.js
```diff
@@ -0,0 +1,8 @@
+export default {
+	completionType: 'Le Marked as completed', // Label for the completion type field when creating/editing an assignment
```
dbatiste: Are we expecting the translated files to contain the hints?

components/d2l-activity-editor/localization.js
```diff
@@ -1,39 +1,49 @@
...
+		}
+	}
 	return {
-		language: supportedLanguages[supportedLanguages.length - 1],
+		language: supportedLanguage,
 		resources: langterms
 	};
 }
+
+export async function getLocalizeResourcesForAssignmentEditor(langs) {
```
dbatiste: It seems like there's some duplication of logic arising here.  Do you think these component specific methods are needed, or could be avoided?

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/fr.js
ryantmer: Not actually sure - we'll have to wait for a translation run to come back. I suspect they would, as I'm pretty sure all that gets replaced on translation runs is the actual translated value - the rest of the file remains untouched.

components/d2l-activity-editor/localization.js
ryantmer: These are needed to avoid [explicitly including the lang files](https://github.com/Brightspace/brightspace-integration/blob/39f2e6f71c34c5a6e2e47bbcd0426826623a3fc8/polymer.json#L29) in BSI. This particular case arises since we have multiple lang folders for one project, which makes sense for FACE but I don't think is done elsewhere. However, this approach also means we can't really move this `localization` functionality into a generic module either, which is kinda crap.

In other words

1. Having BSI not be aware of its dependcies' folder structures
2. Working with JS `import` rather than `fetch`
3. Making `localization.js` a generic module

Pick two. Right now, this approach solves 1 and 2, but I think @mdgbayly has convinced me that 1 isn't so bad, so I'll change this so that it fulfills 2 and 3, since having this able to be a generic module (possibly somewhere in `core`) would be a great benefit, I think (once it's been proven and tested a bit more, including this repo going through another translation run).

ryantmer: Argh, except, the linter fails with `Cannot analyze dynamic imports with non-literal arguments` if we solve 2 and 3 instead of 1 and 2. I really want to figure out a way to solve them all, so I'll keep trying combinations of things.

mdgbayly: > Cannot analyze dynamic imports with non-literal arguments

Which linter is that? The polymer linter or eslint?  If it's the polymer linter then it again sounds like a polymer build limitation...which we shouldn't be so concerned with.

ryantmer: Yeah, the polymer linter - so perhaps just ignoring it and moving on is indeed the right answer! 😄

ryantmer:
Finally ready for review again - sorry for all the back-and-forth, this is what happens when I don't just accept that the world isn't perfect!

Short summary, based on all the different approaches I tried, we can pick two:

1. Having BSI not be aware of its dependencies' folder structures
2. Working with JS import rather than fetch
3. Making localization.js a generic module

This change picks 2 and 3 as being the most important, since 1 is kind of just a quirk of BSI.

Once this is merged and has gone through a translation run to validate that it's working, we can consider moving `localization.js` into somewhere more common (perhaps [somewhere in core](https://github.com/BrightspaceUI/core/tree/master/helpers)), but I don't want to do that until it's been proven here first.

activities.serge.json
```diff
@@ -59,10 +59,11 @@
   },{
     "name": "activityEditor",
     "parser_plugin": {
-      "plugin": "parse_d2l_fra"
+      "plugin": "parse_js"
     },
+    "source_match": "en\\.js$",
     "source_dir": "components/d2l-activity-editor/lang",
```
maboelsoudd2l: Doesn't this directory contain JSON files? If we don't need JSON files, then wouldn't this line need to be removed?

ryantmer: This PR also removes the JSON files, leaving just the new JS ones (for English and French - I've decided I'm leaving the incorrectly-translated French one to see how that gets handled).

ryantmer:
I've been able to test the config here on a local Serge installation (after converting it to Serge's actual config format from JSON), as well as checking things locally with a full BSI build and all that with no surprises - but we'll see what the real-world translation process shows in about a week I suppose!

wongvincent:
Thanks for all the work you've done on this @ryantmer!

PR #469, opened 2019-11-13
better messaging at getActivityName

vix633:


components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -206,7 +206,11 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 	},
 
 	_getActivityName: function(entity) {
-		return this._followHref(this._getActivityNameHref(entity))
+		const activityNameHref = this._getActivityNameHref(entity);
+		if (activityNameHref === '') {
```
AlexBedley: while we are specifically interested in empty string (due to the behaviour of `_getActivityNameHref`) I think we also want to reject on null or undefined, maybe doing `if (!activityNameHref)` is better

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -206,7 +206,11 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
...
+			return Promise.reject('Activity name href was empty');
```
AlexBedley: if we stringify `entity` and put it in the log message it might help the investigation when one crops up in the slack channel

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
vix633: Yeah i was thinking that too

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
vix633: Sure thing, might be a big entity, any PII to consider when doing this?

AlexBedley: good question, not sure about the PII. and yeah maybe a subset of the entity - whatever we need to identify it (self link? not sure)

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -206,7 +206,11 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 	},
 
 	_getActivityName: function(entity) {
-		return this._followHref(this._getActivityNameHref(entity))
+		const activityNameHref = this._getActivityNameHref(entity);
+		if (!activityNameHref) {
+			return Promise.reject(`Activity name href was empty. Content was "${JSON.stringify(entity.Rels.self)}"`);
```
AlexBedley: I don't think `entity.Rels.self` will work. Maybe like `entity ? entity.getLinkByRel('self') : entity`

PR #471, opened 2019-11-13
Added visual diff for quick-eval component

maboelsoudd2l:
This was sitting on my local machine, so i thought we might want this in our master branch. Though it might have consequences.

Arguments for:
- It helps catch regression tests for the overall quick-eval product
- We talked about creating visual diffs for the quick-eval component
- It catches issues relating to RTL and different languages.

Arguments against:
- you need to set the hrefs and tokens from quad sites in order for this test to be useful (SIDE NOTE: how come our dummy data doesn't work anymore?)
- This test will fail every-time we introduce a new feature. Regenerating the goldens might lead us to miss a legitimate regression error.
- It adds extra time


AlexBedley:
I'm ok keeping these tests in the repo for local validation, but I think they should be ignored in CI

test/d2l-quick-eval/perceptual/d2l-quick-eval.visual-diff.html
```diff
@@ -0,0 +1,34 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+		<title>d2l-quick-eval visual diff tests</title>
+		<script type="module" src="../../../components/d2l-quick-eval/d2l-quick-eval.js"></script>
+		<style>
```
dbatiste: If you find yourself repeating these, you may prefer to move these styles into a common file.  We did that in [brightspace-ui/core/test](https://github.com/BrightspaceUI/core/blob/master/test/styles.css).

maboelsoudd2l:
i moved the test files into this new directory, though not too sure about the naming. Also the commands don't work, it gives me errors for not finding the selectors.

PR #472, opened 2019-11-14
created a localization mixin helper for Lit components in QE

maboelsoudd2l:
Here is a screenshot of a lit components working in different lang terms, tested under different lang terms.
image

Testing for this helper would be through the quick diff test I wrote in https://github.com/BrightspaceHypermediaComponents/activities/pull/471.

Thanks to @dbatiste and @ryantmer for their input.

The design of this component is derived straight from the documentation https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.md

It's not the most elegant design. I had an idea of changing `QuickEvalLocalize` to give you localization based on the component being polymer or lit. That idea was discarded because it was too risky.

### DIFF
--- a/components/d2l-quick-eval/LitQuickEvalLocalize.js
+++ b/components/d2l-quick-eval/LitQuickEvalLocalize.js
@@ -1,0 +1,69 @@
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export const LitQuickEvalLocalize = superclass => class extends LocalizeMixin(superclass) {
+	static async getLocalizeResources(langs) {
+		let translations;
+		for (const lang of langs) {
+			switch (lang) {
+				case 'ar':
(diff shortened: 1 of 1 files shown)


maboelsoudd2l:
@ryantmer would you like to share a common implementation of localization? I feel like its very similar to what you built in FACE. (The only thing that might be problematic is that we cant share lang files, don't know how we can reconcile this)

ryantmer:
The short answer is "yes", but the longer answer is "soon" 😄 The `localization.js` I've got in #468 I think can be shared around more broadly, but until it's actually gone through a Serge translation run, might be best to keep it contained for now. It's also currently only supporting "language files are in `./lang`" and "lang files use `export default`", neither of which is the case for QE right now (easily solved, but worth noting as a current difference).

maboelsoudd2l:
will do

PR #473, opened 2019-11-14
added experimental serge parsing

maboelsoudd2l:
I saw @ryantmer do a thing and its pretty cool.

So i wanted to pilot an experimentation new serge integration that pulls us only js files. I shamelessly stole this code from https://github.com/BrightspaceHypermediaComponents/activities/pull/468.

@ryantmer is there a way to test this? Do I have to submit a lang term, then wait and see? Any modifications do i need to get this running?

### DIFF
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -57,6 +57,29 @@
       "zh-tw zh-tw"
     ]
   },{
+    "name": "quickEval_EXPERIMENTAL",
+		"parser_plugin": {
+      "plugin": "parse_js"
+    },
+    "source_match": "en\\.js$",
+    "source_dir": "components/d2l-quick-eval/lang",
+    "output_file_path": "components/d2l-quick-eval/lang-experimental/%LANG%.js",
+    "output_lang_rewrite": [
+      "ar-sa ar",
+      "de-de de",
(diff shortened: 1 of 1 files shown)


activities.serge.json
```diff
@@ -56,6 +56,29 @@
...
       "zh-tw zh-tw"
     ]
+  },{
+    "name": "quickEval_EXPERIMENTAL",
+		"parser_plugin": {
+      "plugin": "parse_js"
+    },
+    "source_match": "en\\.js$",
+    "source_dir": "components/d2l-quick-eval/lang",
+    "output_file_path": "components/d2l-quick-eval/lang-experimental/%LANG%.js",
```
maboelsoudd2l: I changed the output file so that it doesn't overwrite the existing lang terms

activities.serge.json
```diff
@@ -56,6 +56,29 @@
       "zh-cn zh",
       "zh-tw zh-tw"
     ]
+  },{
+    "name": "quickEval_EXPERIMENTAL",
```
maboelsoudd2l: creating a new integration means that serge will send us 2 identical lang terms, on in json form like before and a new one in js form

PR #475, opened 2019-11-15
Adding the visibility switch editor

kieranderson:
Adding the ability to use the switch control to edit the draft status of an activity. Created in a separate component that also handles the read only view when you can't edit the draft status.

I know that the lang area has had lots of changes recently so if any of that needs changed let me know!


components/d2l-activity-editor/lang/en.json
```diff
@@ -7,8 +7,16 @@
     "translation": "Due Time",
     "context": "ARIA label for the due time field when creating/editing an activity"
   },
+  "hidden": {
```
ryantmer: This file doesn't actually exist anymore (hence the conflict) - you'll just need to add these to `en.js` instead, with the context as a comment

components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -0,0 +1,92 @@
...
+
+	static async getLocalizeResources(langs) {
+		return getLocalizeResources(langs, baseUrl);
+	}
+
+	constructor() {
+		super();
+		this._setEntityType(ActivityUsageEntity);
+		this._isDraft = false;
+		this._overrides = document.documentElement.dataset.intlOverrides || '{}';
```
ryantmer: `_overrides` doesn't appear to be used? For context, `document.documentElement.dataset.intlOverrides` is the date/time overrides and such from the `html` tag, which `d2l-due-date-editor` needs in order to format the due date correctly, but it's not needed everywhere

components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -0,0 +1,92 @@
...
+			this._onActivityUsageChange(entity);
+		}
+
+		super._entity = entity;
+	}
+
+	_onActivityUsageChange(activityUsage) {
+		if (activityUsage) {
+			this._isDraft = activityUsage.isDraft();
+			this._switchText = (this._isDraft ? this.localize('hidden') : this.localize('visible'));
```
ryantmer: I don't believe you actually need `this._switchText` or `this._icon` at all if you don't want them, as you can do this logic within `render` if you want. That said, also kind of convenient to have them extracted - mostly just pointing out that we can do simple stuff in the HTML template in `render` itself, when desired 😄

ppaskaris-d2l: I was going to comment on this, but I thought that was maybe a React-ism. I recall from Polymer that we were strongly encouraged to make use of `computed` properties, and figured this was the equivalent.

:+1:

ryantmer: Yeah, I don't believe there's much of a difference either way here - nice to have `this._switchText` around for debugging, but `?:` can be used in the templates if desired. Whatever works and feels right, I figure!

ppaskaris-d2l: In Reactsville, USA, I strongly prefer this pattern:

```
render() {

    // easy breezy debugging
    const someValue = this.props.flag ? 'a' : 'b';

    return <div>{someValue}</div>;
}
```

ryantmer: Actually yeah, even nicer - removes `_switchText` and `_icon` from `this`, which is 👌

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -170,6 +171,12 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 
 	render() {
 		return html`
+			<div id="assignment-visibility-container">
+				<d2l-activity-visibility-editor
+					.href="${this._activityUsageHref}"
+					.token="${this.token}">
+				</d2l-activity-due-date-editor>
```
mdulat: typo? (due-date-editor)

components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -0,0 +1,92 @@
...
+
+	static get styles() {
+		return css`
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			d2l-switch .d2l-label-text {
```
mdulat: Is it possible to use label styles here instead like: https://github.com/BrightspaceHypermediaComponents/activities/pull/380/commits/795f3012656b272978302a8266b48b05bd4f0b3d

kieranderson: For the font-weight I think so, but I need to set the colour specifically, because the default is green!

I think the default in the switch control should be switched to not green, then I think we can make use of the label styles.

PR #476, opened 2019-11-15
Fix lang file format

ryantmer:
Sadly, support for single quotes and non-quoted keys was only added to Serge's `parse_js` parser in evernote/serge@8d42a5d#diff-ab91cf35eb431bb9c7e44e833b0a724f (in August), and hasn't yet been released - despite the documentation saying it's supported. So, we need to use the double-quotes-everywhere syntax for now.


maboelsoudd2l:
I never knew you could force merge a PR without an approval. 

Imagine all the possibilities...

ryantmer:
If you're a repo admin you can 😜 (I use this power very selectively, usually only in cases like this where it's largely unimportant)

PR #480, opened 2019-11-22
Dismiss dialog work

maboelsoudd2l:
Sorry for the lang term spam. this PR includes the stuff from https://github.com/BrightspaceHypermediaComponents/activities/pull/479 because it hasn't been merged yet.

I added an `dismissed-activities-list` component just for demo purposes. I'm planning to do more work on this `dismissed-activities-list` component but I just want something that isn't blank for us to see. I will replace this activity card, with a dismissed activity card that is yet to be designed.
image

PR #481, opened 2019-11-22
Cbusman/add more button to card

cbusmand2l:
image


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -1,16 +1,16 @@
...
 import './d2l-quick-eval-activity-card-action-button.js';
 
-class D2LQuickEvalActivityCardActionButtonMore extends LocalizeMixin(LitElement) {
+class D2LQuickEvalActivityCardActionButtonMore extends LitQuickEvalLocalize(LitElement) {
 
 	static get styles() {
 		return css`
 			:host([hidden]) {
-				display: none;
+				display: none !important;
```
cbusmand2l: The `display: flex` from the activity card was overriding the styling from `hidden`. If there is a way to have this be priority without `!important`, that would be awesome.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -422,6 +438,36 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 		);
 	}
 
+	_dispatchDismissUntilEvent() {
```
cbusmand2l: This (and `_dispatchEditActivityEvent`) are just boilerplate for now - will be implemented in a separate PR.

components/d2l-quick-eval/build/lang/en.js
```diff
@@ -10,8 +10,10 @@ export const LangEn = {
 	'confirmation': 'Confirmation',
 	'courseName': 'Course',
 	'discussion': 'Discussion',
+	'dismissUntil': 'Dismiss until...',
```
cbusmand2l: Only added the en localization for now, to keep this PR lighter.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -413,6 +420,14 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		});
 	}
 
+	_dismissUntil() {
```
cbusmand2l: This (and `_editActivity`) are just stubs for now, since the events that they're handling are not yet implemented.

maboelsoudd2l:
the design doc says "when there's nothing to publish, the publish all button should be disabled". I went in quad and published all for "First Topic with Grade Item" so I can test this behaviour. but then looking at the card the publish all button isn't disabled.
image

design doc says "until publish all is possible for quizzes that item wont be on the list". I tested that in quad, the publish all is disabled. But here it isn't.
image

I think there is a common thread in both of these, maybe the logic for disabling the publish all button needs to be moved to this new dropdown.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
maboelsoudd2l: i looked at `d2l-quick-eval-activity-card-action-button.js` and they do this too so I'm not too worried.

Also this seems like a good use of `!important` I would want my component to be hidden *always* whenever i give it the `hidden` tag, no exceptions.

maboelsoudd2l:
might be a good time to change `demo/d2l-quick-eval/d2l-quick-eval-activity-card.html` and `test/d2l-quick-eval/d2l-quick-eval-activity-card.html` to include a version of the card the has dissmiss-enabled.

cbusmand2l:
@maboelsoudd2l We're not testing anything in the unit tests that have to do with the buttons, so I don't think we need any new tests in `test/d2l-quick-eval/d2l-quick-eval-activity-card.html`.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -263,12 +266,21 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
...
 									on-click="_dispatchPublishAllEvent"
 									disabled$="[[_disablePublishAllButton(publishAll)]]"
 									tab-index-number="[[_computeTabstop(_indicatorPressed, _desktopView)]]"
 									aria-hidden$="[[_computeOppositeIndicatorPressed(_indicatorPressed, _desktopView)]]"></d2l-quick-eval-activity-card-action-button>
+								<d2l-quick-eval-activity-card-action-button-more
+									hidden$="[[!dismissEnabled]]"
+									tab-index-number="[[_computeTabstop(_indicatorPressed, _desktopView)]]"
+									aria-hidden$="[[_computeOppositeIndicatorPressed(_indicatorPressed, _desktopView)]]">
+										<d2l-menu-item text="[[localize('publishAll')]]" on-click="_dispatchPublishAllEvent" disabled$="[[_disablePublishAllButton(publishAll)]]"></d2l-menu-item>
+										<d2l-menu-item text="[[localize('dismissUntil')]]" on-click="_dispatchDismissUntilEvent"></d2l-menu-item>
```
maboelsoudd2l: So I used a bunch of `on-click` listeners in the ellipsis button. But then I realized on Friday, that keyboard users can't really do much with on-click listeners. I found that `d2l-menu-item-select` seems to work better for this. Thoughts?

maboelsoudd2l:
oh sorry @cbusmand2l, LOL I meant `d2l-quick-eval-activity-card.visual-diff.html`. Oopsies 😆

maboelsoudd2l:
This PR, I assume,  should fail the visual diff tests cause we gotta regenerate the goldens with dropdown now. Also you can open the dropdown in the visual test like what I did in my PR https://github.com/BrightspaceHypermediaComponents/activities/pull/480, 
```javascript
await page.evaluate(() => {
	document.querySelector('#dialog-opened d2l-quick-eval-ellipsis-menu').shadowRoot.querySelector('d2l-quick-eval-ellipsis-dialog').opened = true;
});
```

components/d2l-quick-eval/build/lang/en.js
AlexBedley: love it

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
AlexBedley: yeah I'm ok with `!important` here... `:host([hidden])` will likely not be the most specific selector, but we definitely want to hide stuff if it's present

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button-more.js
```diff
@@ -24,19 +24,28 @@ class D2LQuickEvalActivityCardActionButtonMore extends LocalizeMixin(LitElement)
...
 					</d2l-menu>
 				</d2l-dropdown-menu>
 			</d2l-dropdown>
 		`;
 	}
+	static get properties() {
+		return {
+			tabIndexNumber: {
+				type: Number,
+				value: 0
```
AlexBedley: does this work in Lit? seems like the Lit way is to use the constructor

https://lit-element.polymer-project.org/guide/properties#initialize

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: I'm not a huge fan of committing unfinished work, since I feel like PRs should be the "good copy" and we wouldn't want to ship the code like this, but I can see that it's convenient, especially if we're handing it off to someone else

maboelsoudd2l:
@AlexBedley I want to populate the `_dismissUntil()` function with the stuff in #480, but i can't until this PR gets merged to master.

AlexBedley:
> @AlexBedley I want to populate the `_dismissUntil()` function with the stuff in #480, but i can't until this PR gets merged to master.

Right but my point is that `_dismissUntil()` should be added in the PR that uses it. Even if #480 was in I think that should be it's own PR

PR #482, opened 2019-11-22
Changed single quotes to double quotes

maboelsoudd2l:
One of the requirements of Serge.io is to have double quotes instead of single quotes in the lang file. This should hypothetically be the last change that we need in order for serge to work with js files.


AlexBedley:
it's pretty scary that it matters (since it implies they are parsing the file not as a js file), but whatever works!

components/d2l-quick-eval/build/lang/ar.js
```diff
@@ -1,60 +1,61 @@
+/* eslint quotes: 0 */
```
AlexBedley: I forget if these are generated, but if they are, do we need to include this comment in the generator code?

maboelsoudd2l: This was totally added by had, to every file.

ryantmer:
@AlexBedley [Your worst fears, confirmed!](https://github.com/evernote/serge/blob/master/lib/Serge/Engine/Plugin/parse_js.pm#L35) 😄

maboelsoudd2l:
oh yeah @AlexBedley forgot to inform you that the gulp generator is broken now. 😢 (and I don't know how to fix it). That was kinda the reason why I didn't add anything to the generator.

components/d2l-quick-eval/build/lang/ar.js
AlexBedley: Right. So if a computer (such as serge) edits this file, will you have to manually add the comment in every single time?

AlexBedley:
> @AlexBedley [Your worst fears, confirmed!](https://github.com/evernote/serge/blob/master/lib/Serge/Engine/Plugin/parse_js.pm#L35) 😄

components/d2l-quick-eval/build/lang/ar.js
maboelsoudd2l: Good question. Serge still remains a mystery to me as to how it decides to change files. @ryantmer do you know the answer to this?

ryantmer: Should be okay - based on the line I linked below, Serge works line-by-line and ignores every line it doesn't recognize. But, will know for sure soon, once the first set of FACE translations come back (which are using this method and which got sent out for the currently-going translation run).

maboelsoudd2l:
this is useless now since Alex added the build lang terms to be ignored by eslint in https://github.com/BrightspaceHypermediaComponents/activities/pull/573

PR #483, opened 2019-11-22
Integrated Edit-In-Place into title and description of activity collection editor

BhamelinD2L:
### Changes
- Learning path Title and Description are now editable. Users may select and edit the text of the title and description of the activity collection editor in-place by clicking them.
- Hovering over either item will change their text colour to blue and the mouse to a pointer to denote the functionality.
- Long, one word titles shouldn't overflow. Long titles won't compress the `visible` icon and label over top each other.
- The header's margins have been compressed to more correctly match the design sample.

image
image
image
image

### DIFF
--- a/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
+++ b/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
@@ -7,11 +7,13 @@ import { DescribableEntityMixin } from 'siren-sdk/src/entityAddons/describable-e
 import { SimpleEntity } from 'siren-sdk/src/es6/SimpleEntity.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/button/button.js';
+import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/list/list.js';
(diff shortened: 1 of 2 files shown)


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -82,6 +84,16 @@ class CollectionEditor extends EntityMixinLit(LitElement) {
 				justify-content: space-between;
 				align-items: center;
 			}
+			.d2l-activity-collection-title-header {
+				min-height:52px;
+				margin-top:6px;
+				margin-bottom:6px;
+				margin-right: 6px;
+				overflow:hidden;
+			}
```
BhamelinD2L: Edit-in-place needs a minimum buffer size to prevent shifting the page content when entering edit mode. This occurs naturally with normal text, but the header seems to cause a small shift without additional space.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -82,6 +84,16 @@ class CollectionEditor extends EntityMixinLit(LitElement) {
...
+			.d2l-activity-visbility-label {
+				white-space: nowrap;
```
BhamelinD2L: This is needed to prevent the eyeball icon and `visible` text from being pushed atop each other by long titles.

PR #485, opened 2019-11-26
Maboelsoudd2l/dismiss dialog

maboelsoudd2l:
Creating a new PR cause the old one was weird.


maboelsoudd2l:
???
tests seem to fail due to visual diff but yet can't regenerate the goldens
😠

PR #489, opened 2019-11-27
Adding dismissed activities list.

cbusmand2l:
image


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -275,7 +275,7 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 									aria-hidden$="[[_computeOppositeIndicatorPressed(_indicatorPressed, _desktopView)]]"></d2l-quick-eval-activity-card-action-button>
 								<d2l-quick-eval-activity-card-action-button-more
 									hidden$="[[!dismissEnabled]]"
-									tab-index-number="[[_computeTabstop(_indicatorPressed, _desktopView)]]"
+									tabindexnumber="[[_computeTabstop(_indicatorPressed, _desktopView)]]"
```
cbusmand2l: lit doesn't use hyphens when translating attributes <-> properties.

components/d2l-quick-eval/build/lang/en.js
```diff
@@ -13,6 +13,7 @@ export const LangEn = {
 	'discussion': 'Discussion',
 	'dismissedActivities': 'Dismissed activities',
 	'dismissedActivitiesList': 'Dismissed Activities List',
+	'dismissedOn': 'Dismissed on {date}',
```
cbusmand2l: Only adding en to keep PR clean.

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -1,36 +1,34 @@
-import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
 
-class D2LSubtitle extends PolymerElement {
-	static get template() {
-		const subtitleTemplate =  html`
-			<style>
+class D2LSubtitle extends LitElement {
```
cbusmand2l: The old component was having trouble with the `dom-repeat` not rendering in visdiff tests, so rewrote this as a lit component that seems to have fixed things.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
```diff
@@ -1,58 +1,49 @@
...
-					newSubmissions="5"
-					resubmitted="6"
-					due-date="2019-10-10"
-					activity-type="quiz"></d2l-quick-eval-activity-card>
-			</d2l-input-checkbox>
+			<d2l-list separators>${this.dismissedActivities && this.dismissedActivities.length ? this.dismissedActivities.map(act => html`
+				<d2l-list-item selectable>
+					<d2l-icon slot="illustration" icon="${this._computeIcon(act.type)}" aria-label="${this.localize(act.type)}"></d2l-icon>
+					<d2l-list-item-content>
+						<div>${act.name}</div>
```
maboelsoudd2l: why are we wrapping it in divs? same goes for `<div slot="secondary">` below

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -1,36 +1,34 @@
...
-			</template>
+			}
 		`;
-		subtitleTemplate.setAttribute('strip-whitespace', 'strip-whitespace');
-		return subtitleTemplate;
+	}
+	render() {
+		return html`${this._displayText && this._displayText.length ?
+			this._displayText.map(dt => html`<span>${dt}</span>`) :
+			''}`;
```
maboelsoudd2l: just double checking here: If `this._displayText` is undefined then that means its gonna render null?

package.json
```diff
@@ -62,7 +62,7 @@
   "main": "d2l-activities.js",
   "dependencies": {
     "@brightspace-ui-labs/edit-in-place": "^1.0.7",
-    "@brightspace-ui/core": "^1.14.3",
+    "@brightspace-ui/core": "^1.17.0",
```
maboelsoudd2l: Core might be causing issues here https://d2l.slack.com/archives/CBTK3JB6D/p1574959686178700
Don't know if this new version will fix the problem or make it worse.

Im updating core too to version 1.17.0 so we're on the the same boat.

test/d2l-quick-eval/perceptual/d2l-quick-eval-dismissed-activities-list.visual-diff.html
```diff
@@ -0,0 +1,25 @@
...
+				margin: 15px;
+			}
+		</style>
+		<script type="module">
+			import '@brightspace-ui/core/components/typography/typography.js';
+		</script>
+	</head>
+	<body class="d2l-typography">
+		<div id="default">
+			<d2l-quick-eval-dismissed-activities-list dismissedactivities='[{"name":"Activity A","course":"Course A","type":"quiz","dismissedDate":"2019-05-05"},{"name":"Activity B","course":"Course A","type":"discussion","dismissedDate":"2019-06-06"},{"name":"Activity C","course":"Course B","type":"quiz","dismissedDate":"2019-07-07"},{"name":"Activity D","course":"Course B","type":"assignment","dismissedDate":"2019-08-08"}]'></d2l-quick-eval-dismissed-activities-list>
```
maboelsoudd2l: We are making assumptions about the data model before it is finalized. @AlexBedley is this similar what we expect to get from the HM api?

cbusmand2l: This is the data we ultimately need for this component, and it's a dumb component. Whatever format it comes through in the api, something (likely `d2l-quick-eval-activities`) is going to have to parse it out from the entity.

components/d2l-subtitle/d2l-subtitle.js
cbusmand2l: Shouldn't render null, should just render nothing: no element tags.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
cbusmand2l: Partly because of the example from https://github.com/BrightspaceUI/core/tree/master/components/list, and partly leftover from previous iterations of the component that didn't make the final cut. I can probably cut them out - I'll take a look.

maboelsoudd2l: fair, sometimes you need them, but sometimes they just bloat the dom. It's one of @dlockhart 's recommendations.
also if necessary, `act.name` seems like text, it could be wrapped in a `p` tag if we need to `display: block`.

components/d2l-subtitle/d2l-subtitle.js
maboelsoudd2l: cool awesome!

test/d2l-quick-eval/perceptual/d2l-quick-eval-dismissed-activities-list.visual-diff.html
maboelsoudd2l: agreed 👍

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
dlockhart: Paragraph tags will likely come with margins that you don't want. If it can just go right in the content -- `<d2l-list-item-content>${act.name}</d2l-list-item-content>` that's ideal!

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
```diff
@@ -1,58 +1,47 @@
...
-					assigned="20"
-					completed="2"
-					published="3"
-					evaluated="4"
-					newSubmissions="5"
-					resubmitted="6"
-					due-date="2019-10-10"
-					activity-type="quiz"></d2l-quick-eval-activity-card>
-			</d2l-input-checkbox>
+			<d2l-list separators>${this.dismissedActivities && this.dismissedActivities.length ? this.dismissedActivities.map(act => html`
```
dbatiste: Note: `separators` is not a boolean.  The valid values are `all`, `between`, and `none`... `all` being the default.

components/d2l-subtitle/d2l-subtitle.js
maboelsoudd2l: nice!

maboelsoudd2l:
alignment?
image

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -39,12 +37,19 @@ class D2LSubtitle extends PolymerElement {
...
-				computed: '_computeDisplayText(text)'
+				type: Array
 			}
 		};
 	}
 
+	updated(changedProperties) {
+		changedProperties.forEach((_, propName) => {
+			if (propName === 'text') {
+				this._displayText = this._computeDisplayText(this.text);
```
AlexBedley: wow this seems so annoying to do in Lit that it seems to me that this is no longer the correct way to do this stuff. I wonder if they want us to do it in the `render` method or something. In this particular case it would work well since you are doing some validation `render` anyways

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
```diff
@@ -1,58 +1,47 @@
...
+	_computeIcon(type) {
+		switch (type) {
+			case 'quiz':
+				return 'tier2:quizzing';
+			case 'assignment':
+				return 'tier2:assignments';
+			case 'discussion':
+				return 'tier2:discussions';
+		}
+		return '';
```
AlexBedley: if we got to here, I think we would consider this a serious error (quick eval should not be returning activities we don't support). I would prefer to throw here, or log something at the very least

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -1,36 +1,34 @@
...
-			<template is="dom-repeat" items="[[_displayText]]">
-				<span>[[item]]</span>
-			</template>
+			}
 		`;
-		subtitleTemplate.setAttribute('strip-whitespace', 'strip-whitespace');
-		return subtitleTemplate;
+	}
+	render() {
+		return html`${this._displayText && this._displayText.length ?
```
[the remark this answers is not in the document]
maboelsoudd2l: according to the docs https://lit-html.polymer-project.org/guide/writing-templates
This is one of the 2 ways to do conditional rendering

components/d2l-subtitle/d2l-subtitle.js
maboelsoudd2l: I agree, it seems like you are trying to use the `updated` function as an observer. I think this is the polymer way of doing things. This is no longer needed in Lit, because Lit is designed to observe all changes and calls the render function whenever there is a change.

Hence,  in the render function you can do something like this:
```javascript
render() {
const computedText = this._computeDisplayText(this.text);
return html`${computedText && computedText.length ?
// stuff...
```
Also Lit listens to changes to all the properties, even if these properties are changed by a parent (which in this case, `this.text` is an example of this). What this means is that you don't need to call `updated(changedProperties)` in order to capture any changes, theyre already captured in the render function! I even made a demo to demonstrate this: https://stackblitz.com/edit/qgkcgk?file=my-element.js
Reading the docs, https://lit-element.polymer-project.org/guide/lifecycle#updated, it seems like it is used for post-processing changes (the docs for it aren't the best).

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
maboelsoudd2l: @vix633 FYI

components/d2l-subtitle/d2l-subtitle.js
cbusmand2l: I was trying to get the automatic updating to work, but I was having trouble with it for some reason, which is why I ended up using `updated`. Probably just didn't have the syntax exactly right - I'll look into this.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
cbusmand2l: This was more just in case we don't get anything for the type, rather than that we might get an unsupported activity - just so that we're not passing a `null` or `undefined` to the icon.

cbusmand2l: This was more just in case we don't get anything for the type, rather than that we might get an unsupported activity - just so that we're not passing a `null` or `undefined` to the icon.

vix633: Lets get away from the empty string returns. It deals us a disservice when we see errors in prod. Let's throw an unsupported type exception, something verbose that tells us it was not one of the types known to quick eval

components/d2l-subtitle/d2l-subtitle.js
```diff
@@ -26,6 +26,8 @@ class D2LSubtitle extends LitElement {
 		`;
 	}
 	render() {
+		this._displayText = this._computeDisplayText(this.text);
```
maboelsoudd2l: Super sorry to be annoying, but we shouldn't be assigning properties inside our `render()` function. 

Assigning properties causes lit-element to automatically re-render and call the render function again.

Consequently, assigning properties in the render function is not a good idea (side-note: in React, it would typically cause an infinite recursion and break the app. In lit-element, they actually prevented that, but it seems to be a bug rather than a feature as discussed here: https://github.com/Polymer/lit-element/issues/800)

Point is, According to Lars den Bakker, `render()` is based on immutable data. You shouldn't be modifying things in it. 
Instead, I recommend using creating a `const` variable for this 
```javascript
render() {
const computedText = this._computeDisplayText(this.text);
return html`${computedText && computedText.length ?
// stuff...
```
We should only use properties when we need to. Only use properties if we need our data to remain persistent outside our scope (and even then we can update our properties in callbacks and lifecycle changes, instead of the render).

I know that this is a significant departure from our polymer code, where we used to use properties for everything, but this is why I provide these long-drawn out explanations. I want us to shift our polymer habits, and it'll take some re-adjusting at first.

FINAL NOTE: One BIG exception to this rule is the callbacks inside the `render()` function, such as `@click="${()=> this.myprop = "new value"}"`. In this case, we **should** be assigning properties there. Because that is the appropriate situation where we want our component to update based on a "click" or other user input.

dlockhart: 100% agree. Yep the Polymer -> Lit transition will definitely take some getting used to! I've done several migrations and am still learning new things.

components/d2l-subtitle/d2l-subtitle.js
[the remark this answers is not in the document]
AlexBedley: right and I think the other way is more readable

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
AlexBedley: But then we're just moving the error further down right? If the true error is "we couldn't get the activity type" but the symptom is "the icon doesn't render cuz it gets an empty string icon" it will take forever to debug this in prod.

If we can't get the activity type, that's an exception. If the activity type is unsupported, that's an exception. This way the error is raised where the error happens.

components/d2l-subtitle/d2l-subtitle.js
AlexBedley: Thanks for the writeup Mohamed!

components/d2l-subtitle/d2l-subtitle.js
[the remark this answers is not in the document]
maboelsoudd2l: @AlexBedley so according to the docs https://lit-html.polymer-project.org/guide/writing-templates#conditional-templates, would you prefer that we write conditionals like this?
```javascript
const displayText = this._computeDisplayText(this.text);

if (!displayText || !displayText.length) {
  return html``; // or just return null
}
return html`${displayText.map(dt => html`<span>${dt}</span>`)}`;
```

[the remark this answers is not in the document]
AlexBedley: sure something like that. might just be personal preference, but I find that very easy to read. at a glance you can tell what's going on.

PR #490, opened 2019-11-27
Added action dialog

maboelsoudd2l:
I'm learning from my mistakes now by creating smaller changes in my PR's.

Issues:
1. it looks ugly (things are unnecessarily clipped out, photo below) 😢 
2. also I don't know if it's just me, but isn't the main text a bit too big? I just used `<p>` tags, any thoughts?
image


components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+
+class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)) {
+
+	static get styles() {
+		return radioStyles;
+	}
+
+	render() {
+		return html`
+			<d2l-dialog title-text="Dismiss Activity">
```
cbusmand2l: The localization for all this text is going in a subsequent PR?

maboelsoudd2l: yeah, not going to repeat the mistake of adding a million files to PR 😭

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+				<p>Dismissing an activity hides it from Quick Eval, but won't affect the activity</p>
+				<p>Dismiss until...</p>
+				<label class="d2l-input-radio-label">
+					<input type="radio" name="myGroup">
```
maboelsoudd2l: what name should we give "myGroup" 😝

cbusmand2l: Do screen readers/etc. "see" radio group names?

maboelsoudd2l: Just tried it, it says `dismiss until, radio button, 1 of 3`. I think these group names are made so that all the options are working together. (so if each one has a different group name, it clicking one radio doesn't affect the other). 

Due to possible future naming conflicts, i recommend we either give it a complicated name or generated ID.

Thoughts?

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -421,7 +423,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 	}
 
 	_dismissUntil() {
-		//console.log('dismiss until');
+		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dialog');
+		actionDialog.open();
```
maboelsoudd2l: i don't like doing this here because I won't be able to test this in the demo page.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	open() {
+		const actionDialog = this.shadowRoot.querySelector('d2l-dialog');
+		actionDialog.open();
+	}
+
+}
+
+window.customElements.define('d2l-quick-eval-action-dialog', D2LQuickEvalActionDialog);
```
AlexBedley: I don't like that component name doesn't match the filename. can we make it consistent?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: The backend calls these `DismissType`s but I think uniqueness is more important here. I'd be ok with some kind of `d2l-id` thing

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+				<label class="d2l-input-radio-label">
+					<input type="radio" name="myGroup">
+					A specific date
+				</label>
+				<label class="d2l-input-radio-label">
+					<input type="radio" name="myGroup">
+					Forever
+				</label>
+				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
+				<d2l-button slot="footer" dialog-action>${this.localize('cancel')}</d2l-button>
```
AlexBedley: ah I didn't know you could inject multiple components into a single slot!

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+import { LitQuickEvalLocalize } from '../LitQuickEvalLocalize.js';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { radioStyles } from '@brightspace-ui/core/components/inputs/input-radio-styles.js';
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import '@brightspace-ui/core/components/button/button.js';
+
+class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)) {
+
+	static get styles() {
+		return radioStyles;
```
AlexBedley: this is pretty cool

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: If you want more control over this, you can use a property and pass it to `d2l-dialog` via the `opened` attribute

https://github.com/BrightspaceUI/core/blob/60be5b8893d0cce0d4e1831c26c95996562c272f/components/dialog/dialog-mixin.js#L25

Then in the demo you can short circuit the event by just setting a pass-thru property

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: I love it

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: what's a pass-thru property?
In order for me to get this dialog to show in the activity card demo page, I have to do things in demo/d2l-quick-eval/d2l-quick-eval-activity-card.html

(in retrospect, those "things" didn't even work 😢 )

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: Wasn't that totally a polymer thing too?

AlexBedley: probably haha

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: what's even more cool, is that you can chain multiple styles like so 
```javascript
static get styles() {
		return [ labelStyles, bodyStandardStyles, radioStyles, css`
			.radio-container {
				margin-top: 0.5rem;
			}
			.datepicker-container {
				margin-bottom: 0.9rem;
			}
			` ];
	}
```
i prefer this way more to `<style include="d2l-typography">`

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: a "pass-through property" in this case is a property a parent takes but does nothing with, except to pass it to the child component

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,50 @@
...
+import 'd2l-polymer-behaviors/d2l-id.js';
+
+class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)) {
+
+	static get styles() {
+		return radioStyles;
+	}
+
+	render() {
+		const groupID = this._genGroupID();
```
AlexBedley: is this expensive? we could probably cache this across renders. Could just be a constant in this file really

dbatiste: To answer...

> is this expensive?

It is not expensive.  😄

maboelsoudd2l: If I keep causing re-renders, then it'll be expensive. But the good news is that there's no moving components here (***YET***) that causes it to ever re-render so its not too bad.
(I just wanna get this PR out the door, and i can move in the constructor).

PR #491, opened 2019-11-27
US111949 Add attachments editor

ryantmer:
This adds the first phase of components that provide the ability to add/remove attachments associated with an assignment. As of this change, the `d2l-activity-attachments-editor` is quite simple and bare, and intentionally designed such that it could be used with the concept of activity attachments (as well the existing assignment attachments).

What this does do:

- Display an attachment-picker component in FACE
- Show action buttons with tooltips based on what types of (link) attachments are available to be added

What this doesn't do:

- Do stuff when you click on a button - coming soon (next user story), still figuring that out
- Do anything related to file attachments, as these are being handled separately (but should be compatible with these changes)

This was added as a new folder/Serge config since it's _probably_ generic enough that it could be moved up in the repo, outside of `d2l-activity-editor`, but I didn't want to go that far just yet. In particular, if we do decide to make file attachments an activities-wide concept, then moving this up might make sense. (Link attachments are still only supported in the assignments domain, but file attachments are supported elsewhere.)

Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/96, will fail until that's merged.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
@@ -1,0 +1,48 @@
+import './d2l-activity-attachments-picker';
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { AttachmentCollectionEntity } from 'siren-sdk/src/activities/AttachmentCollectionEntity';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+
+class ActivityAttachmentsEditor extends EntityMixinLit(LitElement) {
+	static get properties() {
+		return {
+			_canAddAttachments: { type: Boolean }
+		};
+	}
+
+	static get styles() {
+		return css`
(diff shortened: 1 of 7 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
```diff
@@ -0,0 +1,48 @@
...
+
+		if (entity) {
+			this._canAddAttachments = entity.canAddAttachments();
+		}
+
+		super._entity = entity;
+	}
+
+	render() {
+		return html`
```
mdulat: Will this also eventually show the current attachments, with option to remove?

ryantmer: Yeah - I'm thinking it'll be a sibling element to `d2l-activity-attachment-picker`, some sort of `d2l-activity-attachment-list` or something, and each `d2l-activity-attachment` item in the list would have its own button for removal.

PR #494, opened 2019-11-28
Kanderson/us109692 edit start end dates

kieranderson:
This is adding the availability dates to the editor, start and end date.


components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -0,0 +1,121 @@
...
+
+	render() {
+		return html`
+		<label class="d2l-label-text" ?hidden=${!this._canEditStartDate}>${this.localize('startDate')}</label>
+		<div id="startdate-container" ?hidden=${!this._canEditStartDate}>
+			<d2l-datetime-picker
+				hide-label
+				name="startDate"
+				id="startDate"
+				date-label="${this.localize('startDate')}"
```
mdulat: Trying to remember how this works for accessibility. Is this `date-label` read out by the screen reader even though we're using `hide-label`?

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -0,0 +1,121 @@
...
+			_canEditEndDate: {type: Boolean},
+		};
+	}
+
+	static get styles() {
+		return [labelStyles, css`
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
```
ppaskaris-d2l: needed? isn't this done by the browser / polyfilled?

m6jones: This is required because 
```
:host {
	display: block;
}
```
Will override the `hidden` attribute to hide. So we need to define `:host([hidden])` for expected behaviour. 

I don't think there is a polyfill that fixes this issue.

ppaskaris-d2l: that's unfortunate. also highlighted by the dinosaur people

>**Note**: Changing the value of the CSS display property on an element with the hidden attribute overrides the behavior. For instance, elements styled display: flex will be displayed despite the hidden attribute's presence.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
kieranderson: Yes, the date-label is read out.

PR #497, opened 2019-11-29
Maboelsoudd2l/specific date

maboelsoudd2l:
Added a datetime component in the dialog
This is what the dialog looks like
image
And when specific date is selected:
image
and when the date component is selected:
image


components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+			</d2l-dialog>
+		`;
+	}
+
+	open() {
+		this._getDialog().open();
+	}
+
+	constructor() {
+		super();
```
jeroach: Maybe this is a Lit thing, but doesn't the default constructor already call `super()` by default

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+						<div class="datepicker-container">
+							<d2l-datetime-picker placeholder="11|12|2019" hide-label></d2l-datetime-picker>
+						</div>
+						`: null }
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "forever")}" type="radio" name="myGroup">
+						Forever
+					</label>
+				</div>
+				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
```
jeroach: Why does the primary action button say "Restore"? Shouldn't it be "Dismiss"?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+						<input @change="${this._updateProp.bind(this, "submission")}" type="radio" name="myGroup">
+						Next submission
+					</label>
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "date")}" type="radio" name="myGroup">
+						A specific date
+					</label>
+					${this.selectedRadio === "date"? html`
+						<div class="datepicker-container">
+							<d2l-datetime-picker placeholder="11|12|2019" hide-label></d2l-datetime-picker>
```
jeroach: Might want to check with Dan as to what we want this placeholder to be.

e,g, MM|DD|YYYY might be preferable

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+		vaadin-date-picker-overlay {
+			z-index: 1000;
+		}
+	`;
+	var head = document.head || document.getElementsByTagName('head')[0];
+	var style = document.createElement('style');
+	head.appendChild(style);
+	style.type = 'text/css';
+	style.appendChild(document.createTextNode(styling));
+})()
```
maboelsoudd2l: This function is very hacky. Its designed to hack the datetime components into showing it on top of the dialog, which turns out to be extremely messy.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+			` ];
+	}
+
+	render() {
+		return html`
+			<d2l-dialog title-text="Dismiss Activity">
+				<div class="d2l-body-standard">Dismissing an activity hides it from Quick Eval, but won't affect the activity</div>
+				<br/>
+				<p class="d2l-label-text">Dismiss until...</p>
+				<div class="radio-container">
```
maboelsoudd2l: I added these containers whenever I wanna space things out. I usually add some kind of spacer, but @AlexBedley doesn't like those.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+		super();
+	}
+
+	_getDialog() {
+		return this.shadowRoot.querySelector('d2l-dialog');
+	}
+
+	_updateProp(propName) {
+		this.selectedRadio = propName;
+		this._getDialog().resize();
```
maboelsoudd2l: the dialog content grows and shrinks due to the disappearing datetime component.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: lol oops 😄

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: yeah I guess i forgot to clean it up.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+
+	render() {
+		return html`
+			<d2l-dialog title-text="Dismiss Activity">
+				<div class="d2l-body-standard">Dismissing an activity hides it from Quick Eval, but won't affect the activity</div>
+				<br/>
+				<p class="d2l-label-text">Dismiss until...</p>
+				<div class="radio-container">
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "submission")}" type="radio" name="myGroup">
```
AlexBedley: Dismiss until next submission got bumped until next release.

You can use the hm api to detect what's supported and render the buttons. Or you could just render `forever` and `until` and make a PR when the backend supports `dismiss until next submission`.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+import { radioStyles } from '@brightspace-ui/core/components/inputs/input-radio-styles.js';
+import { labelStyles, bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import '@brightspace-ui/core/components/button/button.js';
+import 'd2l-datetime-picker/d2l-datetime-picker.js';
+
+// shim for vaadin calendar to overlay on top of the dialog
+(()=>{
+	if (window.D2L && window.D2L.DialogMixin && window.D2L.DialogMixin.preferNative) {
+		window.D2L.DialogMixin.preferNative = false;
```
AlexBedley: what does this do?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	var styling = `
```
AlexBedley: `var`?!

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+		vaadin-date-picker-overlay {
+			z-index: 1000;
+		}
+	`;
+	var head = document.head || document.getElementsByTagName('head')[0];
+	var style = document.createElement('style');
+	head.appendChild(style);
+	style.type = 'text/css';
+	style.appendChild(document.createTextNode(styling));
```
AlexBedley: this seems a bit heavy-handed to set a single css property. It also feels like this might belong in the calendar repo, and not here. Can you talk a bit about what you tried and what led you to here?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+						<input @change="${this._updateProp.bind(this, "date")}" type="radio" name="myGroup">
+						A specific date
+					</label>
+					${this.selectedRadio === "date"? html`
+						<div class="datepicker-container">
+							<d2l-datetime-picker placeholder="11|12|2019" hide-label></d2l-datetime-picker>
+						</div>
+						`: null }
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "forever")}" type="radio" name="myGroup">
```
AlexBedley: I dont think these `bind`s are necessary... event handlers get called with the target already (usually `event.target`).  You can put an attribute on the `input` and fetch it in the event handler to determine which one was pressed.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+				<div class="radio-container">
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "submission")}" type="radio" name="myGroup">
+						Next submission
+					</label>
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, "date")}" type="radio" name="myGroup">
+						A specific date
+					</label>
+					${this.selectedRadio === "date"? html`
```
AlexBedley: date should probably be a constant

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+
+	constructor() {
+		super();
+	}
+
+	_getDialog() {
+		return this.shadowRoot.querySelector('d2l-dialog');
+	}
+
+	_updateProp(propName) {
```
AlexBedley: I think this will also have to emit an event eventually, but we can cross that bridge later

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
dbatiste: The `d2l-dialog` component will internally render the native `dialog` element when possible (Chrome only), otherwise it will render it's own custom "dialog". The native dialog is preferred for a number of reasons (one of which is that the browser ensures the dialog is on the "top layer"), however `d2l-dialog` was built with this `preferNative` flag so that we could turn it off if necessary, falling back to our custom dialog solution in all browsers regardless of their native support.

From what I understand, Mohamed encountered an issue with the Vaadin date picker when using the native dialog.  In short, their date picker (which has dialog-like behavior, which typically renders elements at the `document.body` level) was being displayed behind the dialog instead of on top, due to the way it was implemented.

We _really_ don't want to use this flag, and it was never really intended to be used in this way. So, the earlier we get our own date picker implemented the better.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
dbatiste: Hmmm. Contemplating whether the dialog should try to automatically detect these changes.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+					</label>
+				</div>
+				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
+				<d2l-button slot="footer" dialog-action>${this.localize('cancel')}</d2l-button>
+			</d2l-dialog>
+		`;
+	}
+
+	open() {
+		this._getDialog().open();
```
dbatiste: Nit: if exposing an open method to consumers, you may also want to return the promise.  If this is only used internally and you know the consumer doesn't need it, then no biggie.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: do you prefer const?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: Yup pretty much, I have more details below.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: So as @dbatiste pointed out, I came across a big problem while implementing the vaadin datepicker on `d2l-dialog`. Namely, the datepicker appears _behind_ the dialog instead of in-front of it. Then I told myself "no-worries, I'll just increase my z-index until its in-front of the dialog". Didn't work. Because the d2l-dialog uses this chrome-native thing called "overlay", which makes it on top of ABSOLUTELY EVERYTHING! So using a z-index of one million did nothing.

So the first thing I had to do is disable that native overlay thing, which @dbatiste was explaining with the flag. Now with that flag disabled, the dialog now uses z-index, and all I have to do is find the right z-index to get the datepicker on top. Great!

Except that the datepicker is not managed by D2L, it's managed by Vaadin (https://github.com/vaadin/vaadin-date-picker). I don't know how much time it will take for us to get a PR in their repo (and probably you won't be able to approve it 😄 ) so I assumed that is off the table. Furthermore, from my experimentation I was not able to modify one webcomponent from inside a totally different webcomponent. 

One could then argue that we don't need to "modify" the webcomponent, we just need to add a style to the class `vaadin-date-picker-overlay`. But i won't do this in the demo page because it obviously won't work in quad LOL. So I need to find a way to modify any page that this component is inserted in, so that this datepicker would work. I figured that the `body` will always contain everything, so can set css there as global styling. besides, when we transition to the new datepicker, deleting this file, will delete this added css in the body.

Like I said this is very hacky, and if anyone found a more elegant solution, feel free to show me a demo of it.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: This is a good segue for us to bring up a discussion about how HM will play a role in this project (cause as you can tell, me and @cbusmand2l are kind of assuming stuff here). 

I'll speak to you off-PR about this. (also I vote that we let this PR through so that I can create a subsequent PR about this)

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: I'll make all of them constant (like an Enum)

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: Don't you think this is more elegant than adding attributes that a reader won't be able to easily correlate it's intended use. I'm pretty easygoing about this so up to you.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: is the d2l-dialog method async as well?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: I'm surprised this is not automatic (like sized using css)

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: i thought you don't like adding stubs 😆

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: I find the `bind`s confusing, even thought I know what `bind` does. Just tracking down what `this` is in the `render` method, and in the `_updateProp` method took a minute or so.

but yeah it's a matter of style really, I could see being confused by the other approach as well, i'm fine with whatever

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: missing semicolon

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: yes please

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: Thanks for the explanation, sounds like a wild ride. The approach makes sense now.

Re: my comment about where the fix belongs: I didn't mean the vaadin repo, I meant d2l-datetime-picker. Presumably when the real d2l datepicker is out, the changes will be in d2l-datetime-picker, which is also when we'd want to remove this "shim" code. but it sounds like you've put a lot more thought into this than I have so I'll let you decide

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: I would definitely remove `until next submission` in this PR, but happy to talk about the interaction with HM in a seperate medium

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
AlexBedley: yes

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: @dbatiste isn't the d2l-datetime picker archived? that means I can't request more PR's? Also im scared that other people will start getting more things than they bargained for. For instance, I know @BraydonKains is also using this same component for the same reason we are, would he be okay with us inadvertently modifying his CSS?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
maboelsoudd2l: Dan says its not a dealbreaker, I'll speak to @dbatiste over how to implement this, and if I do, i'll put this in a separate PR.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
BraydonKains: Surprised lint didn't see this as an issue. The default Brightspace lit config has this as a rule. Potentially worth adding it if you're using your own eslintrc rather than the standard.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
BraydonKains: Similar to my previous lint comment, surprised this wasn't caught.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
BraydonKains: I can adapt if something needs changing urgently on your end. However I'd still suggest opening a discussion in the #wc-date-time-inputs channel in slack just to make sure if you need to make a big change, to know who needs to be on those PRs.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+
+	getPlaceHOlderFormat() {
+		const overrides = this.tryParseHtmlElemAttr('data-intl-overrides', {});
+		if (overrides && overrides.date && overrides.date.formats && overrides.date.formats.dateFormats && overrides.date.formats.dateFormats.short) {
+			return overrides.date.formats.dateFormats.short;
+		}
+		return 'MM|DD|YYYY';
+	}
+
+	tryParseHtmlElemAttr(attrName, defaultValue) {
```
maboelsoudd2l: This is not the best idea that I copy-pasted this from https://github.com/BrightspaceUI/core/blob/master/helpers/localization.js

@dbatiste any advice? Should I try to export that method from there?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -92,6 +85,37 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
+			return html`
+				<div class="datepicker-container">
+					<d2l-datetime-picker placeholder="${this.getPlaceHOlderFormat()}" hide-label></d2l-datetime-picker>
+				</div>
+			`;
+		}
+		return null;
+	}
+
+	getPlaceHOlderFormat() {
```
AlexBedley: typo on a public function

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
-				<label class="d2l-input-radio-label">
-					<input type="radio" name="${groupID}">
-					A specific date
-				</label>
-				<label class="d2l-input-radio-label">
-					<input type="radio" name="${groupID}">
-					Forever
-				</label>
-				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
+				<div class="d2l-body-standard">Dismissing an activity hides it from Quick Eval, but won't affect the activity</div>
```
jeroach: Does this text need to be localized?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+				<br/>
+				<p class="d2l-label-text">Dismiss until...</p>
```
jeroach: Same here - do we need to localize?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+				margin-bottom: 0.9rem;
+			}
+			` ];
 	}
 
 	render() {
 		const groupID = this._genGroupID();
 
 		return html`
 			<d2l-dialog title-text="Dismiss Activity">
```
jeroach: Localization required?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -92,6 +85,37 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		this._getDialog().resize();
 	}
 
+	renderDatePicker(selectedRadio) {
+		if (selectedRadio === DISMISS_UNTIL_OPTIONS.date) {
+			return html`
+				<div class="datepicker-container">
+					<d2l-datetime-picker placeholder="${this.getPlaceHOlderFormat()}" hide-label></d2l-datetime-picker>
```
AlexBedley: hmm how does the datetime-picker placeholder work? I know we should be respecting `data-intl-overrides` when _displaying_ dates but the API has a specific datetime format it expects. Just making sure this format only affects how dates are displayed.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
-					Forever
-				</label>
-				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
+				<div class="d2l-body-standard">Dismissing an activity hides it from Quick Eval, but won't affect the activity</div>
+				<br/>
+				<p class="d2l-label-text">Dismiss until...</p>
+				<div class="radio-container">
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, DISMISS_UNTIL_OPTIONS.date)}" type="radio" name="${groupID}">
+						A specific date
```
jeroach: Localization?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+					</label>
+					${this.renderDatePicker(this.selectedRadio)}
+					<label class="d2l-input-radio-label">
+						<input @change="${this._updateProp.bind(this, DISMISS_UNTIL_OPTIONS.forever)}" type="radio" name="${groupID}">
+						Forever
```
jeroach: Localization?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+					</label>
+				</div>
+				<d2l-button slot="footer" primary dialog-action="done">Dismiss Activity</d2l-button>
```
jeroach: Localization - can use the same term as the dialog title on line 57

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
dbatiste: I don't think consumers should have to do this, or manually access `data-intl-overrides`.  Perhaps we should add a method or getter to the the shared localization helper to return the format(s).  @dlockhart is more familiar with that code though - he may have a different opinion.

AlexBedley: this code belongs in a mixin. putting it in a mixin will also allow others to take advantage. tbh I'm surprised a mixin doesn't already exist for this.

AlexBedley:
@jeroach localization is coming in the next PR

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,45 +1,122 @@
...
+			return html`
+				<div class="datepicker-container">
+					<d2l-datetime-picker placeholder="${this.getPlaceHOlderFormat()}" hide-label></d2l-datetime-picker>
+				</div>
+			`;
+		}
+		return null;
+	}
+
+	getPlaceHOlderFormat() {
```
jeroach: typo in method name `H0lder`

jeroach:
> @jeroach localization is coming in the next PR

Thanks - seemed a bit odd since some terms are translated, and others are not.

Still, I'm cool to wait until the next PR. Will mark comments as resolved to reduce noise.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
dlockhart: I agree that I think this belongs in our date-picker -- I can un-archive it temporarily that's not a big deal. If you don't want to impact others, you could add an attribute that when set, adds this crazy stuff. That way it's contained in the place that's actually using vaadin, and when you switch to the new date picker the code dies (for you anyway). Plus anyone else that runs into this will already have a fix.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
dlockhart: Yeah this can't know about all this `data-` business, and the overrides aren't even guaranteed to be there as they're a particular user's overrides that might not be present on every page.

How important is this placeholder format thing, and why aren't other places that use the date-picker setting it? Should it be something that's properly set from the date-picker itself instead of consumers needing to always set it?

AlexBedley: to clarify: it might not belong specifically in a mixin lol, what I really wanted to say is this should be somewhere shared, behind an interface that is specific to fetching the user's date format (not parsing html, which is an implementation detail).

maboelsoudd2l: I spoke offline with @AlexBedley and found that, if there will be some common datetime format stuff going on. It might be better to put it in some common mixin (that is if @dlockhart and @dbatiste agree to having this functionality in the first place).

Point is, Alex and I feel that date-formatting is something that we might need again. I'm thinking of removing this code in this PR, hardcoding the format to `'MM|DD|YYYY'`. Then in a separate PR, I'll modify https://github.com/BrightspaceUI/core/blob/master/helpers/localization.js to expose `tryParseHtmlElemAttr`, then call it in our 2 localization behaviours (`LitQuickEvalLocalize` and `QuickEvalLocalize`)

dlockhart: Totally, it should 100%  be coming from the user settings, as should the actual parsing/formatting algorithm it uses -- but that's the job of the date picker, not every page that uses the date picker. The new date picker will absolutely handle this automatically.

Until the new date picker is ready, I'm just wondering if we could ignore this issue by setting the placeholder to something like "Enter a date..."?

PR #498, opened 2019-11-29
US108929: Integrate the draft / publish toggle

AllanKerr:
# Changes
This pull request replaces the visibility toggle with the `d2l-activity-visibility-editor` component to allow the toggle to properly trigger the action to update the visibility status when changed.

### DIFF
--- a/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
+++ b/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
@@ -10,15 +10,14 @@ import '@brightspace-ui/core/components/button/button.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/list/list.js';
 import '@brightspace-ui/core/components/list/list-item.js';
-import '@d2l/switch/d2l-switch.js';
 import 'd2l-organizations/components/d2l-organization-image/d2l-organization-image.js';
 import '@brightspace-ui-labs/edit-in-place/d2l-labs-edit-in-place.js';
+import '../d2l-activity-editor/d2l-activity-visibility-editor.js';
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -58,11 +58,12 @@ class ActivityVisibilityEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMi
 	}
 
 	_updateVisibility() {
-		this.wrapSaveAction(super._entity.setDraftStatus(!this._isDraft));
+		const oldStatus = this._isDraft;
+		this._isDraft = !this._isDraft;
+		this.wrapSaveAction(super._entity.setDraftStatus(this._isDraft)).catch(() => this._isDraft = oldStatus);
```
AllanKerr: @kieranderson I noticed that the label text wasn't getting updated when the draft / published status changed. Are you good with this change going in?

AllanKerr: For anyone else reading this, the label not getting updated was a result of the entity getting returned having a `self` link that didn't match the `activity-usage` `self` link. This meant that the entity store never triggered the `_onActivityUsageChange` function. Because the component wasn't updating it manifested as a bug in the component.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -13,12 +13,12 @@ import '@brightspace-ui/core/components/list/list-item.js';
 import '@d2l/switch/d2l-switch.js';
```
jlai-d2l: If we're no longer using `d2l-switch`, is this import safe to remove?

PR #503, opened 2019-12-04
US111919 Add turnitin dialog launcher

ppaskaris-d2l:
[US111919](https://rally1.rallydev.com/#/detail/userstory/350454972296?fdp=true)

This PR adds the turnitin editor, which calls into the LMS to do the thing via a legacy dialog to the existing control. It doesn't currently display anything about the turnitin state, nor does it trigger a refresh of the assignment entity, both of which will happen in a follow-up when the designs catch up.

I also added an `if (!entity) { ... }` guard because I learned that it sure can be null or whatever, here and in the release conditions editor.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-assignment-turnitin-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-assignment-turnitin-editor.js
@@ -1,0 +1,117 @@
+import '@brightspace-ui/core/components/button/button.js';
+import { bodySmallStyles, heading4Styles } from '@brightspace-ui/core/components/typography/styles.js';
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-assignment-turnitin-editor.js
```diff
@@ -0,0 +1,117 @@
...
+			{
+				Text: this.localize('btnCancel'),
+				ResponseType: 2, // D2L.Dialog.ResponseType.Negative
+				IsPrimary: false,
+				IsEnabled: true
+			}
+		];
+
+		// Launch into our friend, the LMS, to do the thing.
+		D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
```
kieranderson: :+1:

JordanCorner:
Once we merge this one, can we bump BSI? We have a few changes queued up now with TII dialog, RC dialog, and start/end dates.

ppaskaris-d2l:
BSI was just bumped today actually!

PR #504, opened 2019-12-04
[US105733] Fix the "You are all caught up" bug

DiljotSG:
Having an empty filter applied will clear filters, and return you to a quick eval list without any filters.

Known issue: the "You are all caught up" cat appears on every initial load for a split second.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -400,16 +400,14 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 	async _handleFilterLoad() {
 		// When the filter finishes loading, we check to see if the filter is empty
 		// If the filter is empty, we remove all filters
+		this._loading = true;
 		if(this._initialLoad) {
-			this._loading = true;
-			await this._handleEmptyFilter().then (() => {
-				this._loading = false;
-			});
+			await this._handleEmptyFilter();
 		}
+		this._loading = false;
 	}
 
 	async _handleEmptyFilter() {
-		console.log(this.entity);
 		// Remove filter if we have no elements within the current filter.
 		if(this.entity && this.entity.entities && this.entity.getSubEntities().length === 0) {
 			await this._clearFilterAndSearch();
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -283,7 +283,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			await this._clearFilterAndSearch();
 			this._initialLoad = false;
 			return;
-		}	
+		}
 
 		try {
 			if (entity.entities) {
@@ -401,7 +401,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		// When the filter finishes loading, we check to see if the filter is empty
 		// If the filter is empty, we remove all filters
 		this._loading = true;
-		if(this._initialLoad) {
+		if (this._initialLoad) {
 			await this._handleEmptyFilter();
 		}
 		this._loading = false;
@@ -409,7 +409,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 
 	async _handleEmptyFilter() {
 		// Remove filter if we have no elements within the current filter.
-		if(this.entity && this.entity.entities && this.entity.getSubEntities().length === 0) {
+		if (this.entity && this.entity.entities && this.entity.getSubEntities().length === 0) {
 			await this._clearFilterAndSearch();
 		}
 	}
--- a/components/d2l-quick-eval/d2l-quick-eval-submissions.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-submissions.js
@@ -98,6 +98,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 					token="[[token]]"
 					category-whitelist="[[_filterIds]]"
 					result-size="[[_numberOfActivitiesToShow]]"
+					on-d2l-hm-filter-filters-loaded="_handleFilterLoad"
 					lazy-load-options>
 				</d2l-hm-filter>
 				<d2l-hm-search
@@ -265,23 +266,14 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 	}
 
 	async _loadData(entity) {
+		this._loading = true;
 		if (!entity) {
 			return Promise.resolve();
 		}
-		this._loading = true;
 
 		if (this._initialLoad) {
 			this.filterAppliedShortcut();
 			this.searchAppliedShortcut();
-		}
-
-		if (this._initialLoad &&
-			entity.hasClass('empty') &&
-			(this.searchApplied || this.filterApplied)
-		) {
-			await this._clearFilterAndSearch();
-			this._initialLoad = false;
-			return;
 		}
 
 		try {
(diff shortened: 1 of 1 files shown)


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,25 +266,16 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 	}
 
 	async _loadData(entity) {
+		this._loading = true;
 		if (!entity) {
 			return Promise.resolve();
 		}
-		this._loading = true;
```
maboelsoudd2l: ?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -396,6 +388,21 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
+		if(this._initialLoad) {
+			this._loading = true;
+			await this._filterLoadHandler().then ( () => {
+				this._loading = false;
+			} )
+		}
+	}
+
+	async _filterLoadHandler() {
+		if(this.entity && this.entity.getSubEntities().length == 0) {
```
maboelsoudd2l: `this.entity && this.entity.getSubEntities() && this.entity.getSubEntities().length === 0`
What happens if we got no subentities? does it display an empty list or null, if the latter then we'll get a bunch of junk errors in our logs.
This might not be necessary but you should check. But better safe than sorry.
(also nit: triple equals?)

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -265,25 +266,16 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
 
-		if (this._initialLoad &&
-			entity.hasClass('empty') &&
-			(this.searchApplied || this.filterApplied)
-		) {
-			await this._clearFilterAndSearch();
-			this._initialLoad = false;
-			return;
-		}
-
```
maboelsoudd2l: I was never a big fan of this code, but can you explain your thinking as to why you are removing this?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -98,6 +98,7 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 					token="[[token]]"
 					category-whitelist="[[_filterIds]]"
 					result-size="[[_numberOfActivitiesToShow]]"
+					on-d2l-hm-filter-filters-loaded="_handleFilterLoad"
```
maboelsoudd2l: tbh not a big fan of this. I feel like this should somehow belong in the behaviour, but I'm not sure how.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
DiljotSG: I moved it up, but this change didn't make a difference with trying to fix the issue I am currently having. I will try to include a gif of the current issue.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
DiljotSG: It never ran anyways, as the two Booleans it checks are still false as the filters haven't finished loading, etc.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
maboelsoudd2l: so why did you move it up?

DiljotSG: I can revert this change lol

maboelsoudd2l:
I'm interested to hear @AlexBedley 's opinion on my comments.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
cbusmand2l: Please do. If we end up not having an entity, then the component will stay in `_loading` state because there's nowhere it's being set to false.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
DiljotSG: Nevermind, I was mistaken.

AlexBedley:
I might be biased for having attempted this twice already, but it's a pretty complicated issue. Can you do a short writeup on why the current implementation is broken, and how your fix addresses it? It might also help us help you with the cat flashing problem.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: it depends if everyone that does filtering will have this problem, or just us. I think this is a "just us" problem so it probably belongs in our code. However if we end up duplicating this exact fix over in `activities` that could be an argument to put it in the behaviour.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: it runs. the problem is that it runs too soon. `_clearFilterAndSearch` get called before the "clear all" action is loaded, so it ends up throwing.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: This seems right. we only want to reset the filters if the filters cause the list to be empty (which is this conditional). Or maybe I'm not understanding your concern Mohamed

AlexBedley:
Some test cases to consider:

No filters applied, elements in list -> table
No filters applied, no elements in list -> cat
Filters applied, no elements in list, after filter is removed there are elements -> table without filters applied
Filters applied, no elements in list, after filter is removed still no elements -> cat
Filters applied, elements in list -> table with filters applied

repeat for search
check activities
check interaction with paging

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -396,6 +388,21 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		}
 	}
 
+	async _handleFilterLoad() {
+		if(this._initialLoad) {
```
AlexBedley: what happens if `_loadData` finishes (and thus sets `_initialLoad` to `false`) before the filters have loaded?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
AlexBedley: Ah I get it now. Actually this is incorrect usage of getSubEntities - we're missing the `entityRel` parameter. I would just do `this.entity.entities`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
DiljotSG: Good question - not sure... 🤷🏽‍♂️

I will be testing this more today.

DiljotSG:
So the problem was that when an instructor finished marking all of the items in the current filter in QuickEval, upon returning to QuickEval from the immersive marking screen, the filter would still be applied and the QuickEval page would not show any remaining items with the filter removed. 

This change attempts to fix that, and it seems to work, aside from the visual glitches. I added an "On filter loaded" event handler and when the filters are loaded, I check if the filter is empty. If it is empty, I clear all the filters/searches. I know Alex had tried this earlier in the `_loadData` function, however, at that point, the filters may or may not have loaded yet. This fixes that, since the code only runs when the filters are loaded. However, it is susceptible to that visual glitch (where the cat appears for a split second). I assume this is because the page has loaded (with the filter), so there are no items, then the cat appears. Then I set the `_loading` boolean to true, as the filters are loaded now, and then I check for it being empty, and then I clear the filter. So we need to fix this somehow.

DiljotSG:
image
I am getting different behaviour when revisiting this PR. This is what I see when I finish evaluating all the items in the current filter. This behaviour is different than what was seen in the video above, and there have not been any code changes since I last visited this PR (all I've done is a `full_all` back to my prior state).

PR #506, opened 2019-12-05
Passing dismissHref around, adding missed aria-disabled.

cbusmand2l:


components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
```diff
@@ -81,6 +81,7 @@ class D2LQuickEvalActivitiesList extends QuickEvalLocalize(PolymerElement) {
 										activity-name-href="[[a.activityNameHref]]"
 										activity-name="[[a.activityName]]"
 										dismiss-enabled="[[dismissEnabled]]"
+										dismiss-href="[[a.dismissHref]]"
```
maboelsoudd2l: what is this `dismiss-href` for?

cbusmand2l: This is the href that we'll use to get the dismiss actions for this activity. I'm passing around the href instead of the actions because we probably don't want to make another call for each activity until we actually need the actions - may as well grab them when the user indicates that they actually want to dismiss the activity.

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -352,6 +356,10 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 				type: Boolean,
 				value: false
 			},
+			dismissHref: {
+				type: String,
+				value: ''
```
AlexBedley: nit: i'm really not a fan of these safe default values

components/d2l-quick-eval/activities-list/d2l-quick-eval-activities-list.js
maboelsoudd2l: ok i understand now

maboelsoudd2l: ok i understand now

PR #508, opened 2019-12-05
Add Dismiss Forever

AlexBedley:
After this PR, you should be able to dismiss activities forever. Next up is `dismiss until`

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
@@ -266,6 +266,13 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 		}
 
 		return false;
+	},
+
+	_dismissActivity: function(dismissHref, actionName) {
+		return this._followHref(dismissHref).then(x => {
+			const action = this._getAction(x.entity, actionName);
+			return this.performSirenAction(action);
+		});
 	}
 };
 
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
@@ -268,9 +268,13 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 		return false;
 	},
 
-	_dismissActivity: function(dismissHref, actionName) {
+	_dismissActivity: function(dismissHref, actionName, dismissUntil) {
 		return this._followHref(dismissHref).then(dismissEntity => {
 			const action = this._getAction(dismissEntity.entity, actionName);
+			if (dismissUntil) {
+				const field = action.getField('dismissUntil');
+				field.value = dismissUntil;
+			}
 			return this.performSirenAction(action);
 		});
 	}
--- a/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
@@ -269,8 +269,8 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 	},
 
 	_dismissActivity: function(dismissHref, actionName) {
-		return this._followHref(dismissHref).then(x => {
-			const action = this._getAction(x.entity, actionName);
+		return this._followHref(dismissHref).then(dismissEntity => {
+			const action = this._getAction(dismissEntity.entity, actionName);
 			return this.performSirenAction(action);
 		});
 	}
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -19,7 +19,7 @@ import './d2l-quick-eval-activities-skeleton.js';
 import './behaviors/d2l-quick-eval-telemetry-behavior.js';
 import '@brightspace-ui/core/components/dialog/dialog-confirm.js';
 import './dismiss/d2l-quick-eval-action-dismiss-dialog.js';
-
+import { DISMISS_TYPES } from './dismiss/dismiss-types.js';
 /**
  * @customElement
  * @polymer
@@ -426,7 +426,14 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 
 	_dismissUntil() {
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
-		actionDialog.open();
+
+		actionDialog.open().then(action => {
+			if (action === DISMISS_TYPES.forever) {
+				console.log('dismiss forever');
+			} else if (action === DISMISS_TYPES.date) {
+				console.log('dismiss until');
+			}
+		});
 	}
 
 	_editActivity() {
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -427,9 +427,19 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 	_dismissUntil(evt) {
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
 
-		return actionDialog.open().then(action => {
-			if (action === DISMISS_TYPES.forever) {
-				return this._dismissActivity(evt.detail.dismissHref, action).then(() => {
+		return actionDialog.open().then((action) => {
+			const {selectedRadio, date} = JSON.parse(action);
+			if (selectedRadio === DISMISS_TYPES.forever) {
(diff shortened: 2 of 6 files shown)


components/d2l-quick-eval/dismiss/dismiss-types.js
```diff
@@ -0,0 +1,4 @@
+export const DISMISS_TYPES = {
+	forever: 'dismiss-forever',
+	date: 'dismiss-until'
```
AlexBedley: the values here must match the action names returned by the API

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -26,11 +27,6 @@ import 'd2l-polymer-behaviors/d2l-id.js';
 	style.appendChild(document.createTextNode(styling));
 })();
 
-const DISMISS_UNTIL_OPTIONS = {
```
AlexBedley: I renamed it to `DISMISS_TYPES` to match the backend api

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -60,16 +56,16 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
 					</label>
 					${this.renderDatePicker(this.selectedRadio)}
 					<label class="d2l-input-radio-label">
-						<input @change="${this._updateProp.bind(this, DISMISS_UNTIL_OPTIONS.forever)}" type="radio" name="${groupID}">
+						<input @change="${this._updateProp.bind(this, DISMISS_TYPES.forever)}" type="radio" name="${groupID}">
 						Forever
 					</label>
 				</div>
-				<d2l-button slot="footer" primary dialog-action="done">Dismiss Activity</d2l-button>
+				<d2l-button slot="footer" primary dialog-action="${this.selectedRadio}">Dismiss Activity</d2l-button>
```
AlexBedley: `selectedRadio` contains the dismiss type, which is the action name, which we use as the dialog result

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -424,9 +424,18 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+	_dismissUntil(evt) {
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
-		actionDialog.open();
+
+		return actionDialog.open().then(action => {
+			if (action === DISMISS_TYPES.forever) {
+				return this._dismissActivity(evt.detail.dismissHref, action).then(() => {
+					const selfHref = this._getSelfLink(this.entity);
+					// bypass cache
+					window.D2L.Siren.EntityStore.fetch(selfHref, this.token, true);
```
AlexBedley: this is a huge hack to bypass the entity cache. The entity behaviour has a way to bypass the cache, but it's global, and we only want to evict this particular member, so this seemed like the cleanest way to do it, but not the greatest, since it relies on some globals and private functions.

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -424,9 +424,18 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
 
-	_dismissUntil() {
+	_dismissUntil(evt) {
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
-		actionDialog.open();
+
+		return actionDialog.open().then(action => {
+			if (action === DISMISS_TYPES.forever) {
+				return this._dismissActivity(evt.detail.dismissHref, action).then(() => {
+					const selfHref = this._getSelfLink(this.entity);
```
AlexBedley: the reason I do this instead of `this.href` is because we never update `this.href` (which was probably the original intent, but we strayed from that design path long long ago). So `this.href` contains the original href QE was loaded with, not the current one, potentially with filters and search applied.

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -266,6 +266,13 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 		}
 
 		return false;
+	},
+
+	_dismissActivity: function(dismissHref, actionName) {
```
cbusmand2l: Dismiss until date (I assume) takes in the date as a param. We are going to update function to accept another parameter when we do the next step?

AlexBedley: yes, I need to find a way to pass that information out of the dialog. I haven't quite figured that out yet. Worst case we can serialize it inside the actionName - which is terrible, but it would work.

AlexBedley: yes, I need to find a way to pass that information out of the dialog. I haven't quite figured that out yet. Worst case we can serialize it inside the actionName - which is terrible, but it would work.

components/d2l-quick-eval/dismiss/dismiss-types.js
maboelsoudd2l: I think that eventually these should be returned by the HM. Something to keep in mind.

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -266,6 +266,17 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 		}
 
 		return false;
+	},
+
+	_dismissActivity: function(dismissHref, actionName, dismissUntil) {
+		return this._followHref(dismissHref).then(dismissEntity => {
+			const action = this._getAction(dismissEntity.entity, actionName);
+			if (dismissUntil) {
```
maboelsoudd2l: As a team, what are our best-practices for input validation? Or is input validation something that is elegantly handled in the Hypermedia layer?

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -103,15 +109,16 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		return D2L.Id.getUniqueId();
 	}
 
+	_onDatetimePickerDatetimeCleared() {
+		this.date = undefined;
```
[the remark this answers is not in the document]
AlexBedley: hmm yes we probably should disable the dismiss activity button if they've selected dismiss until but not put a date. I vote to defer this work however

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -91,8 +94,11 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 			return html`
 				<div class="datepicker-container">
 					<d2l-datetime-picker
-					@d2l-datetime-picker-datetime-changed="${this._onDatetimePickerDatetimeChanged}"
-					placeholder="MM|DD|YYYY" hide-label></d2l-datetime-picker>
+						@d2l-datetime-picker-datetime-changed="${this._onDatetimePickerDatetimeChanged}"
+						@d2l-datetime-picker-datetime-cleared="${this._onDatetimePickerDatetimeCleared}"
```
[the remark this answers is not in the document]
AlexBedley: what's the goal of that change? The datepicker already display `this._date`, since you selected it via the datepicker.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -103,15 +109,16 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		return D2L.Id.getUniqueId();
 	}
 
+	_onDatetimePickerDatetimeCleared() {
+		this.date = undefined;
+	}
+
 	_onDatetimePickerDatetimeChanged(e) {
-		if (e.detail.isSame(this._date)) {
```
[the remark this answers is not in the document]
AlexBedley: Honestly I found it more confusing than helpful - I think the one case this improves perf does not outweigh the 99% of other cases where it makes perf worse. The other repo had this code to avoid a webrequest, but we do not make a web request on change here.

components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
m6jones: The servers should be validating input but it makes for a better user experience if we do it in the UI too. I don't think the front-end hypermedia layer does this for us. Our input components probably do though.

AlexBedley: we do not handle errors here very well. I will address this in the next PR, and I'll take a look at the input validation as well.

PR #510, opened 2019-12-05
US112016 Add attachment list

ryantmer:
This adds a list view of (link) attachments in the editor, as well as support for removing them.

Thanks to the entity store and a small change in the LMS that adds `Link` headers to the add/delete attachment routes, this approach automatically re-renders the list when an attachment is added or removed without any additional effort. It's likely that once autosave is figured out a bit more, we'll have to add better interim states, but that remains to be seen.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
@@ -33,10 +33,7 @@ class ActivityAttachment extends EntityMixinLit(LitElement) {
 			this._attachment = {
 				id: entity.href(),
 				url: entity.href(),
-				name: entity.name(),
-				isDeleted: false,
-				isInFocus: false,
-				isNew: false
+				name: entity.name()
 			};
 			this._editing = entity.canDeleteAttachment();
 		}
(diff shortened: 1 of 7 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
```diff
@@ -0,0 +1,64 @@
...
+		if (!this._entityHasChanged(entity)) {
+			return;
+		}
+
+		if (entity) {
+			this._attachment = {
+				id: entity.href(),
+				url: entity.href(),
+				name: entity.name(),
+				isDeleted: false,
```
mdgbayly: i think some of these properties like isDeleted, isInFocus, isNew are actually properties of the d2l-labs-attachment component rather than the attachment property passed to the component, but as they are not really being used, probably no harm.

ryantmer: You are totally correct! I had them in both places at one point due to extreme confusion, but then forgot to remove them from here. I'll update that, then merge this - the unfurling and trusted sites change will come as its own PR, to keep things a bit easier to digest.

ryantmer:
Apologies for the noise! Travis wasn't getting the latest version of `siren-sdk`, but honestly not sure how things were working previously - we were referencing `^1` which provides no real guarantees. Honestly not sure how I didn't hit this problem previously, as I've been adding stuff for weeks to `siren-sdk`... (clearing Travis' caches didn't help, strangely)

Either way, fixed now, and I guess a note that `^1` isn't a very good idea! (There are a bunch of dependencies like that in this repo, but I'll only change things as they come up.)

PR #511, opened 2019-12-06
Add key to list item to keep track of selected activities.

cbusmand2l:


cbusmand2l:
The key gets passed back in the `d2l-list-item-selected` event from the list item when the checkbox state changes, will be handling this event in the controller:
```
_handleListItemSelected(e) {
	if (this._data)
	{
		const act = this._data[e.detail.key];
		if (act) {
			act.selected = e.detail.selected;
		}
	}
}
```

PR #512, opened 2019-12-06
Jroach/us112192

jeroach:
[US112192](https://rally1.rallydev.com/#/detail/userstory/353901460400?fdp=true): [Dismiss][UI] Hide dismiss filter from filter component

Thoughts on visual diff tests? Is there a way for me to have the activities list open with the filters open? I think that would be the only way it would work

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -126,6 +126,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				<d2l-hm-filter
 					href="[[filterHref]]"
 					token="[[token]]"
+					category-whitelist="[[_filterIds]]"
 					lazy-load-options>
 				</d2l-hm-filter>
 				<d2l-hm-search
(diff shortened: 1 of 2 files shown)


cbusmand2l:
I would think that the visdiff tests for filter whitelists would belong in the filter component itself, with unit tests here to be sure we're assigning the whitelist properly.

If we do want to do the visdiff here though, we can manually open the dropdown in the test with `page.evaluate`.

jeroach:
> with unit tests here to be sure we're assigning the whitelist properly.

I think I've covered this - please let me know if not.

> I would think that the visdiff tests for filter whitelists would belong in the filter component itself

Thoughts @AlexBedley ?

cbusmand2l:
> > with unit tests here to be sure we're assigning the whitelist properly.
> 
> I think I've covered this - please let me know if not.

Yes! Sorry, I meant that the unit tests that you've already got in this PR are what we want.

AlexBedley:
I agree with Carol, we should assume the whitelist works correctly, it's up to the filter component to test that.

We just need to validate that we are sending the correct whitelist to the filter component. which that test is basically doing. I'm ok with the current test coverage.

PR #514, opened 2019-12-09
Maboelsoudd2l/add action toast

maboelsoudd2l:
pretty simple

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -183,8 +183,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				<d2l-button slot="footer" dialog-action="no">[[localize('no')]]</d2l-button>
 			</d2l-dialog-confirm>
 			<d2l-quick-eval-action-dismiss-dialog></d2l-quick-eval-action-dismiss-dialog>
-			<d2l-alert-toast id="toast-dismiss-success" type="success">Activity Dismissed</d2l-alert-toast>
-			<d2l-alert-toast id="toast-dismiss-critical" type="critical">Activity could not be dismissed. Please try again.</d2l-alert-toast>
+			<d2l-alert-toast id="toast-dismiss-success" type="success">[[localize('activityDismissed')]]</d2l-alert-toast>
+			<d2l-alert-toast id="toast-dismiss-critical" type="critical">[[localize('failedToDismissActivity')]]</d2l-alert-toast>
 			<dom-repeat items="[[_publishAllToasts]]" as="toast">
 				<template>
 					<d2l-alert-toast type="default" open>[[_publishAllToastMessage(toast)]]</d2l-alert-toast>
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -450,6 +452,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				const selfHref = this._getSelfLink(this.entity);
 				// bypass cache and reload
 				window.D2L.Siren.EntityStore.fetch(selfHref, this.token, true);
+				this.shadowRoot.querySelector('#toast-dismiss-success').open = true;
```
AlexBedley: If I dismiss activity A, then dismiss activity B, will I see this toast after dismissing activity B?

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -450,6 +452,10 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+			}).catch((error) => {
+				this.shadowRoot.querySelector('#toast-dismiss-critical').open = true;
+				throw new Error(`could not dismiss activity, error: ${error}`);
```
AlexBedley: won't this cause the app to display the global error message and freeze? I think in this case we'd like the non-dismiss stuff to keep working, but happy to defer to Dan

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -183,6 +183,8 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				<d2l-button slot="footer" dialog-action="no">[[localize('no')]]</d2l-button>
 			</d2l-dialog-confirm>
 			<d2l-quick-eval-action-dismiss-dialog></d2l-quick-eval-action-dismiss-dialog>
+			<d2l-alert-toast id="toast-dismiss-success" type="success">Activity Dismissed</d2l-alert-toast>
+			<d2l-alert-toast id="toast-dismiss-critical" type="critical">Activity could not be dismissed. Please try again.</d2l-alert-toast>
```
AlexBedley: I think at this point we should be adding langterms

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: yeah it auto-closes, so you have to keep doing this everytime

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: what would you like instead?

AlexBedley: chatted offline, gonna log using the logging service

components/d2l-quick-eval/d2l-quick-eval-activities.js
maboelsoudd2l: cool

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: I could not reproduce this. have you tried it?

PR #515, opened 2019-12-09
US112016 Unfurling support

ryantmer:
This adds support for attachments being able to unfurl themselves into rich previews via the AF unfurling service. This is done using the newish Requester/Provider approach, which allows the top-level `d2l-activity-editor` component to provide the configuration necesssary for the nested `d2l-labs-attachment` component via the `d2l-request-provider` event.

The behaviour that is currently being encompassed in `d2l-activity-editor`'s constructor (via `addEventListener`) will soon be provided via `BrightspaceUI/core`, but until that is merged and released, this is a simplified version of it that unblocks this change. This also relies on the LMS setting the `unfurlEndpoint` attribute on the `d2l-assignment-activity-editor` that it renders, which has now been merged.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-editor.js
@@ -6,6 +6,10 @@ class ActivityEditor extends LitElement {
 	static get properties() {
 		return {
 			loading: { type: Boolean },
+			/**
+			 * API endpoint for attachment unfurling service
+			 */
+			unfurlEndpoint: { type: String }
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -36,6 +40,13 @@ class ActivityEditor extends LitElement {
 		this.addEventListener('d2l-siren-entity-save-error', () => {
 			this.shadowRoot.querySelector('#save-status').error();
 		});
+
+		this.addEventListener('d2l-request-provider', e => {
```
ryantmer: This will be simplified/centralized/improved by https://github.com/BrightspaceUI/core/pull/301, but figured this works until that is merged (which should be very soon, but figured no sense in blocking this change).

PR #516, opened 2019-12-09
Dmackenzie/US112183 localize dismiss

dmackenz:
Changed the dismiss UI to use localization and included the localization terms in the js lang files.


cbusmand2l:
Why are we only updating the English `.json` files if we're doing all the `.js` files?

dmackenz:
> Why are we only updating the English `.json` file if we're doing all the `.js` files?

dmackenz:
@CarolBusman Alex said this:

> the translators look at the json files to do translations. luckily we only have to edit the english one and they figure it out and translate it.
> the code looks at the js files. so they all need to be there otherwise if you set the language to like spanish and es.js is missing the langterm the app will crash

PR #518, opened 2019-12-10
US112016 Add trustedSitesEndpoint

ryantmer:
This is a method of determining whether or not a given link attachment is a trusted site or not. Sadly, the actual Trusted Sites tool doesn't have its own API, but we can instead engage the HTML Editor API to determine if a link is trusted. The approach is a bit hackish, but it more-or-less mimics the behaviour provided by the `filterHtml` ifrau service.


ryantmer:
Merged to validate this, as I don't have a local AF to test this on my local instance :( I'll validate that things work as expected, and will create a follow-up PR to add the alert that was mentioned this morning that will tell the user they've added an untrusted link.

PR #519, opened 2019-12-10
necessary code to get the dialog to work

maboelsoudd2l:
at it's current state the dialog wasn't working. So i made it work...


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -60,14 +60,14 @@ class D2LQuickEval extends
 					<h1 class="d2l-quick-eval-header-with-toggle" hidden$="[[!activitiesViewEnabled]]">[[headerText]]</h1>
 					<h1 class="d2l-quick-eval-header" hidden$="[[activitiesViewEnabled]]">[[headerText]]</h1>
 				</template>
-				<d2l-quick-eval-ellipsis-menu hidden$="[[!dismissEnabled]]"></d2l-quick-eval-ellipsis-menu>
+				<d2l-quick-eval-ellipsis-menu href="[[_lazyActivitiesHref]]" token="[[token]]" hidden$="[[!dismissEnabled]]"></d2l-quick-eval-ellipsis-menu>
```
cbusmand2l: I don't like checking in the wrong href just because we're not getting the right one generated yet. This should be bound to `dismissedActivitiesHref`.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -31,10 +30,15 @@ class D2LQuickEvalEllipsisDialog extends LitQuickEvalLocalize(LitElement) {
...
 	}
 
-	_onClose() {
-		this.dispatchEvent(new Event('on-close'));
+	_getDialog() {
+		return this.shadowRoot.querySelector('d2l-dialog');
 	}
 
+	updated() {
+		if (this._getDialog() && this.dismissedActivities && this.dismissedActivities.length) {
```
cbusmand2l: Is the dialog object going to change? Should we be storing it in a const somewhere?

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: sorry when I said "use the activities href" I mean add a property called `dismissHref` and pass the activities href into that so we're ready for the backend changes

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -31,10 +30,15 @@ class D2LQuickEvalEllipsisDialog extends LitQuickEvalLocalize(LitElement) {
...
+			setTimeout(()=>this._getDialog().resize(), 0);
```
AlexBedley: Can you explain why this is in a `setTimeout`?

components/d2l-quick-eval/d2l-quick-eval.js
AlexBedley: Ah there already is a property for this in the code. Let's just use that please

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -41,7 +42,11 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 			_isError: {
 				type: Boolean,
 				value: false,
-			}
+			},
+			opened: {
```
AlexBedley: why is this property in the controller? it looks like the dialog deals with it's own opened state? or did you mean to move all the code up to this level?

components/d2l-quick-eval/d2l-quick-eval.js
maboelsoudd2l: 👍

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: It's there to hook up the dialog to the ellipsis button. The dialog won't open when you click the dropdown unless we hook it up.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
maboelsoudd2l: the dialog has a resize function that doesn't work as expected unless it waits a certain amount of time until it recalculates the dialog size again. @dbatiste might be able to help with this.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
maboelsoudd2l: There are certain moments where it is not defined yet. It kept giving me errors when I define it once and setting it to a constant. It became to the point of bloat. I just figured we don't need more errors.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
dbatiste: Yeah the dialog doesn't know that UI changes have been made and it needs to wait before making new measurements.  I'd be curious if a `requestAnimationFrame` call would have been sufficient here.  We could consider applying this inside the `resize` method.

maboelsoudd2l: what's the difference operationally between a `requestAnimationFrame` and a `settimeout(,0)`?

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
cbusmand2l: Can we at least only call it once in this method? Put it into a `const` before the `if`?

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
AlexBedley: ohh this is just a pass-thru. ok.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
maboelsoudd2l: coming right up.

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
dbatiste: `requestAnimationFrame` tells the browser to perform some task, such as a UI update, via the provided callback before the next repaint (i.e. the next frame, after pending changes have been applied). Despite it's name, it is not just for performing animations over many frames.

`setTimeout` simply executes the specified task after the specified delay.  A delay of 0 is regarded ideally as immediately, but it may not be immediate.  The delay may be longer due to browser throttling `setTimeout` callbacks, or if the browser is busy with other tasks. The browser keeps a queue of these tasks, and it executes them as it deems possible. It could be executed the next frame, or it could be 10 frames later.

Before `requestAnimationFrame` was supported, developers frequently used `setTimeout(,0)`, to "push" some task to the next frame, for instance to wait for some pending changes.  It was heavy handled, kinda hacky, and not always perfect, but it was what people had to do.  That all being said, `setTimeout` is still sometimes useful, and sometimes simply waiting a single frame is not enough... but in such case, the consumer might be better to wait for some event, rather than waiting some arbitrary amount of time.

So, if we know that a single frame would be sufficient in this case, it may be ok to do that within the dialog's `resize`.

AlexBedley:
I approved it, but we still need to fix the build. The vdiff stuff needs to be updated I think

test/d2l-quick-eval/perceptual/d2l-quick-eval-ellipsis-menu.visual-diff.js
```diff
@@ -37,7 +37,9 @@ describe('d2l-quick-eval-ellipsis-menu', function() {
 		await page.click('#dialog-opened d2l-quick-eval-ellipsis-menu');
 		await page.waitFor(200);
 		await page.evaluate(() => {
-			document.querySelector('#dialog-opened d2l-quick-eval-ellipsis-menu').shadowRoot.querySelector('d2l-quick-eval-ellipsis-dialog').opened = true;
+			document.querySelector('#dialog-opened d2l-quick-eval-ellipsis-menu')
+				.shadowRoot.querySelector('d2l-quick-eval-dismissed-activities')
```
cbusmand2l: Technically since we have `opened` passing from the controller down to the dialog, we could set `opened=true` at this level instead of digging down into the `shadowRoot` again, but since this is a test (and a visdiff at that), as long as we end up in the expected visual state for testing, I don't see that there's any advantage either way.

maboelsoudd2l: I noticed that too. At this point, i think this test had outlived its usefulness. The list thing doesn't work anymore without an href and a token.
I think we should keep these visual tests limited to just the dropdown. Then we create new test for the list and card. I think that's a better return on investment.

PR #520, opened 2019-12-10
Maboelsoudd2l/submit restore

maboelsoudd2l:
Working code for restoring the dismissed activities.


components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -117,6 +137,22 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 		}
 	}
 
+	_submitData(e) {
+		if (e.detail.action === 'done') {
+			const selectedData = this._data.filter(d=>d.selected);
```
cbusmand2l: We already have `_getSelectedActivities()` written for this functionality.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -20,8 +20,12 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 	static get template() {
 		const quickEvalActivitiesTemplate = html`
 			<d2l-quick-eval-ellipsis-dialog
+				opened="[[opened]]"
 				dismissed-activities="[[_data]]"
+				on-d2l-dialog-close="_submitData"
 				on-d2l-quick-eval-dismissed-activity-selected="_handleListItemSelected"></d2l-quick-eval-ellipsis-dialog>
+			<d2l-alert-toast id="toast-dismiss-list-success" type="success">[[successMessage]]</d2l-alert-toast>
```
cbusmand2l: Do we still need to worry about unique IDs in components? We should probably make these classes instead.

maboelsoudd2l: what do you recommend? I honestly had no idea what to do with this to make it unique

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: cool whip

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
cbusmand2l: I would make it a class, and also add a prefix that ties it to this component; something like `d2l-quick-eval-dismissed-list-success`, etc.

AlexBedley: If you absolutely need to use an id, use `d2l-id`, of which you are familiar with (and many examples exist in this codebase)

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -47,6 +50,14 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 				type: Boolean,
 				value: false,
 			},
+			count: {
```
AlexBedley: This is confusing, because I'd assume this is the count of `_data`. Maybe `selectedCount`?

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -122,6 +137,22 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 		}
 	}
 
+	_submitData(e) {
+		if (e.detail.action === 'done') {
+			const selectedData = this._getSelectedActivities();
+			this.count = selectedData.length;
+			const result = Promise.all(selectedData.map((act)=> {
```
AlexBedley: hmmm this will spawn `n` web requests all at once. Might be premature optimization, but can we chain these promises together instead?

Also if a single one fails I think we'd still want to dismiss the other ones... I think the current code does that but Promise.all will return immediately on the first failure so it will be hard to determine what happened (since the in-flight promises could also fail).

I'd suggest wrapping the promises to handle the possible failure, then we can count the failures in the resulting promise.

AlexBedley: From a backend perspective, we might consider some kind of bulk API call if this gets too ridiculous (if users start doing 100s at a time for example)

maboelsoudd2l: In the activities controller `_parseActivities`, we use `Promise.all` and catch inside the async function. is that you want? Also, how do you expect to count the failures in the resulting promise?
Since this might not be clear I might speak to you about this offline.

AlexBedley:
- [ ] Change target branch back to master

AlexBedley:
- [ ] refresh dismissed and activities list after restoring

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -128,6 +160,28 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
+			}));
+			result.then((results)=> {
+				this.failedCount = results.filter(a=>a instanceof Error).length;
+				if (this.failedCount > 0) {
+					this._logError(e, {developerMessage: 'Error dismissing activities'});
+					this.shadowRoot.querySelector('.d2l-quick-eval-dismissed-list-critical').open = true;
+				} else {
+					this.shadowRoot.querySelector('.d2l-quick-eval-dismissed-list-success').open = true;
+				}
+			});
```
AlexBedley: we need to refresh both lists here. we can defer to the next PR if you'd like to get this in

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -149,13 +165,20 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 			const selectedData = this._getSelectedActivities();
 			this.selectedCount = selectedData.length;
 			const result = Promise.all(selectedData.map((act)=> {
-				return this.performSirenAction(act.unDismiss);
+				try {
+					return this.performSirenAction(act.unDismiss);
+				} catch (e) {
+					return new Error();
```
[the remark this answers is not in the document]
maboelsoudd2l: The content of this error would be junk, I just want it fail. I already do logging a few lines down, and spoke with Dan about the proper UI for handling failure.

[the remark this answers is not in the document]
AlexBedley: The contents of the error you made would be junk. I was referring to `e` which should contain enough information to identify the error. The log statement below does not contain enough information to fix the problem.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -128,6 +160,28 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
+				}
+			}));
+			result.then((results)=> {
+				this.failedCount = results.filter(a=>a instanceof Error).length;
+				if (this.failedCount > 0) {
+					this._logError(e, {developerMessage: 'Error dismissing activities'});
```
AlexBedley: `e` here is just a button pressed event, which isn't really related to the failure

AlexBedley: - [ ] log enough information so we can fix the issue

maboelsoudd2l: Alex I created a small story for this since this guy is merged.
https://rally1.rallydev.com/#/238990806412d/detail/userstory/355116681708

what info do you consider to be an "acceptable log message"? is it the number of failed activities? activity names? the error message from `performSirenAction`?

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: resolving this

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
AlexBedley: Vaguely: enough information to fix the issue
Specifically: I think the self hrefs should be enough. If we're worried about massive log messages, we could limit it to the first 5 self hrefs and the total count

PR #521, opened 2019-12-10
US105618 - Focus on the first name in the submissions table when navigation forward and backward to quick eval page.

dmackenz:
…gation forward and backward to quick eval page.


components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -320,6 +322,15 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 		}
 	}
 
+	_handleNameFocusOnPageForwardBack() {
+		if (!this.showLoadingSkeleton && this._data.length > 0 && window.performance && window.performance.navigation.type === window.performance.navigation.TYPE_BACK_FORWARD) {
```
AlexBedley: ahh unfortunately the user's wont be using the back button. The workflow is thus:

1. open QE
2. click a link in QE which takes you to evaluation
3. evaluate, click a link that takes you back to QE
4. QE loads and needs to focus the element you clicked (or the one after it)

dmackenz: Ohh I see, should I still leave this just incase they do actually use the back button?

maboelsoudd2l: I agree with @AlexBedley, I'm not sure of the full context of this, but I guess the ideal would be both the browser back button and the QE back button could focus the first element.

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -301,7 +302,8 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 	}
 	static get observers() {
 		return [
-			'_handleNameSwap(_headerColumns.0.headers.*)'
+			'_handleNameSwap(_headerColumns.0.headers.*)',
```
maboelsoudd2l: niccceeeeeeee 👍 👍 👍

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -320,6 +322,15 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 		}
 	}
 
+	_handleNameFocusOnPageForwardBack() {
+		if (!this.showLoadingSkeleton && this._data.length > 0 && window.performance && window.performance.navigation.type === window.performance.navigation.TYPE_BACK_FORWARD) {
+			const firstUserLink = this.shadowRoot.querySelector('.d2l-user-name-link');
```
maboelsoudd2l: not the biggest fan of this for 2 reasons:
1. how do we know this is the first link?
2. why not `.querySelector('d2l-link');`?
3. @dbatiste any conventions for class names? cause `d2l-user-name-link` might be too generic to me.

dmackenz: 1. querySelector always returns the first instance. querySelectorAll returns an array of them.
2. I figured it is safer to use the classname instead of the tag name because if the page changes in the future, there may be the same tag name added above the table, which would break 'd2l-link'.

dbatiste: In practice, we use the convention `tag-name-*`, so in this case `d2l-quick-eval-submissions-table-*`, but I know that tends to get pretty verbose. The main reason for this is, that shadyCSS does not prevent styles from the outer page from targeting elements in this components HTML. It only scopes _this_ element's styles from affecting HTML outside this component when shadow-DOM is not supported (Edge).

While the scope of this query is confined to this component's DOM, the possibility of the HTML being modified in the future is a valid concern with the use of `d2l-link` in this case.  Safe today, maybe not in the future if someone adds a link above without realizing the effect it has.  For smaller components or for components with good test coverage to catch this scenario, I wouldn't worry as much, but it's perfectly valid to make this code more resilient to such changes imo.  Just be careful of using class names that exist in CSS in the outer page, at least until Edge supports shadow-DOM.  😄

dbatiste: Could also mitigate that risk by querying for `d2l-td > d2l-link` or something along those lines.  Probably multiple way to go about it if you really want to avoid the extra class.

PR #522, opened 2019-12-10
Adding empty and loading list states.

cbusmand2l:
Verified button states with Daniel:
`Restore` (disabled) and `Cancel` buttons while loading
- if we end up with an empty list, switch to just `Close`
- if we end up with items in the list, keep the buttons as they are 
- when any items in the list are selected, enable the `Restore` button
- if the `Restore` button is enabled and we deselect the last item in the list (no more items are selected), `Restore` will become disabled.

Loading:
image
Data (none selected):
image
Data (>0 selected):
image
No data:
image


components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -112,11 +118,15 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
 			const act = this._data[e.detail.key];
 			if (act) {
-				act.selected = e.detail.selected;
+				this.set(`_data.${e.detail.key}.selected`, e.detail.selected);
 			}
 		}
 	}
 
+	_computeRestoreDisabled(loading, data) {
+		return loading || !(data.base && data.base.some(d => d.selected));
```
AlexBedley: what's `base`? I've never used an observer on `something.*` before

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -112,11 +118,15 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 		{
 			const act = this._data[e.detail.key];
 			if (act) {
-				act.selected = e.detail.selected;
+				this.set(`_data.${e.detail.key}.selected`, e.detail.selected);
```
AlexBedley: this is so the change is observable?

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
cbusmand2l: When you watch for a property on an object or an item in an array, the observer sends an object that wraps: the base object/array in the `base` property (in this case, `_data`), and a `value` property that's the new value of whatever changed (in this case, `selected`).

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
cbusmand2l: Yup! That way the `_data.*` in the `_computeRestoreDisabled` observer can be triggered on this change.

maboelsoudd2l: I like this 👍

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: `loading || !data.base || data.base.some(d => d.selected)`

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -1,40 +1,71 @@
...
 	render() {
 		return html`
 			<d2l-dialog
 				title-text="${this.localize('dismissedActivitiesList')}"
 				.opened="${this.opened}"
 				@d2l-dialog-close="${this._onClose.bind(this)}">
-				<d2l-quick-eval-dismissed-activities-list .dismissedActivities=${this.dismissedActivities}></d2l-quick-eval-dismissed-activities-list>
-				<d2l-button slot="footer" primary dialog-action="done">${this.localize('restore')}</d2l-button>
-				<d2l-button slot="footer" dialog-action>${this.localize('cancel')}</d2l-button>
+				${this._renderList()}
```
maboelsoudd2l: beautiful

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -1,40 +1,71 @@
...
 
+	_renderList() {
+		if (!this.loading) {
+			if (this.dismissedActivities && this.dismissedActivities.length) {
+				return html`<d2l-quick-eval-dismissed-activities-list .dismissedActivities=${this.dismissedActivities}></d2l-quick-eval-dismissed-activities-list>`;
+			} else {
+				return html`${this.localize('noDismissedActivities')}`;
+			}
+		}
+		return html`<d2l-loading-spinner size="100"></d2l-loading-spinner>`;
```
maboelsoudd2l: what's size 100?

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -1,40 +1,71 @@
...
+			dismissedActivities: { type: Array },
+			loading: { type: Boolean },
+			restoreDisabled: {type: Boolean}
 		};
 	}
 
+	static get styles() {
+		return css`
+			d2l-loading-spinner {
+				width: 100%;
```
maboelsoudd2l: you would think this would be so by default

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
AlexBedley: @cbusmand2l I know this is merged but I am also curious

cbusmand2l: The default size for the loading spinner is 50. It looked too small. So I made it bigger. 100 is arbitrary (it looked good to me) and I don't think it looks too big on a smaller screen (tested it in the mobile emulator in chrome not on an actual mobile device though).

PR #523, opened 2019-12-10
Fix alignment of tooltip

ryantmer:
The "Add link to existing activity" tooltip was being cut off, due to the boundary not being set. This applies the boundary to each tooltip - not strictly necessary, but this means that the order doesn't really matter. (If the styling were applied using CSS, we could `first-of-type`, but that doesn't allow for the computation that `d2l-tooltip` does with `boundary`.)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -13,8 +13,7 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 		return {
 			_canAddLink: { type: Boolean },
 			_canAddGoogleDriveLink: { type: Boolean },
-			_canAddOneDriveLink: { type: Boolean },
-			_tooltipBoundary: { type: Object }
+			_canAddOneDriveLink: { type: Boolean }
 		};
 	}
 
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -13,7 +13,8 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 		return {
 			_canAddLink: { type: Boolean },
 			_canAddGoogleDriveLink: { type: Boolean },
-			_canAddOneDriveLink: { type: Boolean }
+			_canAddOneDriveLink: { type: Boolean },
+			_tooltipBoundary: { type: Object }
```
mdgbayly: Does this need to be a property if we never update it?

ryantmer: It doesn't, mostly I was just going with "things used in `render` are properties", but I suppose that isn't actually necessary in cases like this. I'll move it to just exist within `render`.

mdgbayly: I don't feel strongly either way, but when I see a property it makes me think this is a reactive property.

I've used class properties before for stuff which is more constant if it feels cleaner than having to define a local variable in the render method. Also then if you want to break up your render method into functions it can make things slightly cleaner to keep it as a class property without being a lit property.

ryantmer:
Realized that I made this PR too soon - new one coming shortly with more small fixes!

PR #527, opened 2019-12-11
[US112301] Clear the selected options on opening the dialog

DiljotSG:
[Video coming soon]

### DIFF
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
@@ -76,7 +76,19 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		`;
 	}
 
+	_clearSelectedOptions() {
+		this.selectedRadio = undefined;
+		this._date = undefined;
+
+		// Get the Div holding the radio buttons and clear them all
+		let elements = this.shadowRoot.querySelector('.radio-container').children;
+		for(let element of elements) {
+			element.children[0].checked = false;
+		}
+	}
+
 	open() {
+		this._clearSelectedOptions();
 		return this._getDialog().open();
 	}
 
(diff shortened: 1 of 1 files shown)


DiljotSG:
zoom_1
Video of the changes

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -76,7 +76,19 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		`;
 	}
 
+	_clearSelectedOptions() {
+		this.selectedRadio = undefined;
+		this._date = undefined;
+
+		// Get the Div holding the radio buttons and clear them all
+		let elements = this.shadowRoot.querySelector('.radio-container').children;
```
AlexBedley: Can we make this a bit safer? Just add a class to the `input` s and select that instead of navigating the DOM

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -76,7 +76,19 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
+		for(let element of elements) {
+			element.children[0].checked = false;
```
maboelsoudd2l: instead of doing query selector stuff, I think there is a "Lit" way of doing things: for the 2 radio inputs up in the render function, we can bind them to the this.selectedRadio so that when we do `this.selectedRadio = undefined;` it will trigger a re-render and show the dialog with nothing being selected. I think this might also work. @DiljotSG what do you think?

DiljotSG: Sounds way better to me

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
DiljotSG: I think this way is better because it works regardless of how many inputs there are. IF we add more this still works since I iterate through `.radio-container`'s children.

DiljotSG: If I were to add a class to the `input`s it would have to be added to any new inputs.

PR #528, opened 2019-12-11
US112183 - Fxing dismissUntil, editActivity and moreActions lang terms

dmackenz:


components/d2l-quick-eval/lang/en.json
```diff
@@ -34,6 +34,7 @@
    "loadMore" : "Load more",
    "loading" : "Loading",
    "masterTeacher" : "Teacher",
+   "moreActions": "More Actions",
```
AlexBedley: Was this one being used but just missing? Or you've added it but it's not used yet?

AlexBedley: Ah I see, someone added it to the js file and not the json file. We should really have automation that prevents that :(

dmackenz: I feel like we should only need to add it to en.js, then it will sync to all js and json files.

PR #531, opened 2019-12-11
Abedley/us112313/no past dates

AlexBedley:
This PR disallows you from dismissing until a date in the past (but not a time in the past). The next PR will address the "time in the past" problem.

### DIFF
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
@@ -89,6 +89,12 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		this._getDialog().resize();
 	}
 
+	_getIso8601Date(date) {
+		const pad = x => x < 10 ? '0' + x : x;
+
+		return `${date.getFullYear()}-${pad(date.getMonth() + 1)}-${pad(date.getDate())}`;
+	}
+
 	renderDatePicker(selectedRadio) {
 		const now = new Date();
 		const minDate = this._getIso8601Date(now);
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
@@ -100,6 +100,7 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 	}
 
 	renderDatePicker(selectedRadio) {
+		/* global moment:false */
 		const now = moment();
 		const minDate = this._getIso8601Date(now);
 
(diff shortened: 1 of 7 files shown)


components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -89,14 +89,24 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		this._getDialog().resize();
 	}
 
+	_getIso8601Date(date) {
+		const pad = x => x < 10 ? '0' + x : x;
+
+		return `${date.getFullYear()}-${pad(date.getMonth() + 1)}-${pad(date.getDate())}`;
```
AlexBedley: This only works if the user's browser settings are the same as their LMS settings. Investigating...

maboelsoudd2l: since we have momentjs might as well use it:`moment().toISOString();` will get you the same result.
also according @dlockhart, the `date-timezone` tag in html has timezone data that could be parsed by momentjs 
like this `moment().tz("America/Toronto").utcOffset()` (i got that timezone from the html in my quad)

AlexBedley: yeah I went that route

demo/d2l-quick-eval/d2l-quick-eval.html
```diff
@@ -1,6 +1,7 @@
 <!doctype html>
 <html lang="en"
 	data-lang-default="en-us"
+	data-timezone="{&quot;name&quot;:&quot;Canada - Toronto&quot;,&quot;identifier&quot;:&quot;America/Toronto&quot;}"
```
maboelsoudd2l: ❤️

test/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -0,0 +1,28 @@
+suite('_getIso8601Date', () => {
+
+	let component;
+
+	setup(function() {
+		component = fixture('basic');
+	});
+
+	const testCases = [
```
maboelsoudd2l: ❤️

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -89,14 +89,28 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		this._getDialog().resize();
 	}
 
+	_getIso8601Date(date) {
+		const timezone = this.getTimezone().identifier;
+
+		if (timezone) {
+			return date.tz(timezone).format('YYYY-MM-DD');
```
maboelsoudd2l: what about the date format in `data-intl-overrides`?

AlexBedley: YYYY-MM-DD is what the datetime-picker expects for the `min` property, we don't display this to the user

maboelsoudd2l: 👍

PR #533, opened 2019-12-12
Add refresh capabilities to all 3 controllers

AlexBedley:
My approach here is to use events. We can fire global events when refreshes are needed and the controllers can listen for them. I was going to write this as a behaviour, but since that's polymer 1 stuff and we're on Lit now I decided rather than entangle myself in a framework I'd just get it done.

While testing this I noticed a weird bug where the UI does not match what the API is returning. Peggy reported a similar defect in slack. I'm wondering if we're not making observable changes to `this.entity`.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -449,9 +449,9 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			}
 
 			return dismissAction.then(() => {
-				const selfHref = this._getSelfLink(this.entity);
-				// bypass cache and reload
-				window.D2L.Siren.EntityStore.fetch(selfHref, this.token, true);
+				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh-activities'));
+				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh-submissions'));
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -121,7 +121,8 @@ class D2LQuickEval extends
 				value: submissions
 			},
 			dismissedActivitiesHref: {
-				type: String
+				type: String,
+				value: ''
```
AlexBedley: without this, the href in the dismiss controller gets set to `'undefined'` (literally the string, not `undefined`). I tried it with `null` and it sets it to the string `'null'`. I'm wondering if this is because we pass the Lit/Polymer barrier twice

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -529,5 +529,27 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		window.removeEventListener('d2l-quick-eval-refresh-activities', this._boundRefresh);
+	}
+
+	refresh() {
+		const selfHref = this._getSelfLink(this.entity);
+		if (selfHref) {
```
AlexBedley: these if-blocks protect against the lazy loading case (we don't have to reload if we haven't loaded yet)

PR #534, opened 2019-12-12
Cbusman/lazy load dismiss href

cbusmand2l:


components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-menu.js
```diff
@@ -42,7 +41,7 @@ class D2LQuickEvalEllipsisMenu extends LitQuickEvalLocalize(LitElement) {
 			</d2l-dropdown-menu>
 		</d2l-dropdown-more>
 		<d2l-quick-eval-dismissed-activities
-			href="${this.href}"
+			href="${this._computeLazyHref()}"
```
AlexBedley: I'm still confused to how this works. Why would it recompute this value if it has no dependencies? Polymer is a mystery to me sometimes haha

maboelsoudd2l: because this `render()` function is designed to run every single time. This is not an observer like we were used to in polymer 3.

PR #535, opened 2019-12-12
US112339 - Remove Edit Activity from activity card temporarily

dmackenz:


AlexBedley:
Ah you'll also need to update the goldens so the visual diff tests will pass (it's expecting 3 items in the dropdown)

PR #536, opened 2019-12-12
US112016 Add untrusted URL alert

ryantmer:
This uses `d2l-labs-attachment-list` to show a `d2l-alert` informing the user that they have attached an untrusted link when they do so. This also adds an `<li>` around `d2l-activity-attachment` as is being done in `assignments-ui`, as these are list elements, semantically speaking. The list is always in `editing` mode here, as FACE is an editor. This also required adding a sort of pass-through parameter, `creating`, to `d2l-activity-attachment` - this is because the `d2l-labs-attachment-list` looks for this parameter on the event target when deciding whether or not to show the untrusted URL alert.

A caveat to this change is that right now, the "Attachments" header shows above the `d2l-labs-attachment-list` when there are attachments, which we don't want to always show.

An additional caveat is that I was unable to test this locally due to there not being an AF service on the proddev domain - however, I mocked out that response locally and was able to see the intended alert show when I added a non-trusted URL.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
@@ -6,6 +6,7 @@ import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 class ActivityAttachment extends EntityMixinLit(LitElement) {
 	static get properties() {
 		return {
+			creating: { type: Boolean },
 			_attachment: { type: Object },
 			_editing: { type: Boolean }
 		};
@@ -36,6 +37,7 @@ class ActivityAttachment extends EntityMixinLit(LitElement) {
 				name: entity.name()
 			};
 			this._editing = entity.canDeleteAttachment();
+			this.creating = entity.canDeleteAttachment();
 		}
 
 		super._entity = entity;
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
@@ -37,7 +37,6 @@ class ActivityAttachment extends EntityMixinLit(LitElement) {
 				name: entity.name()
 			};
 			this._editing = entity.canDeleteAttachment();
-			this.creating = entity.canDeleteAttachment();
 		}
 
 		super._entity = entity;
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
@@ -37,6 +37,7 @@ class ActivityAttachment extends EntityMixinLit(LitElement) {
 				name: entity.name()
 			};
 			this._editing = entity.canDeleteAttachment();
+			this.creating = entity.canDeleteAttachment();
 		}
 
 		super._entity = entity;
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
@@ -7,8 +7,7 @@ import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 class ActivityAttachmentsEditor extends EntityMixinLit(LitElement) {
 	static get properties() {
 		return {
-			_canAddAttachments: { type: Boolean },
-			_creating: { type: Boolean }
+			_canAddAttachments: { type: Boolean }
 		};
 	}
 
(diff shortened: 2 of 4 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
```diff
@@ -6,6 +6,7 @@ import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 class ActivityAttachment extends EntityMixinLit(LitElement) {
 	static get properties() {
 		return {
+			creating: { type: Boolean },
```
ryantmer: This approach is a solid 3.6 to me - not great, not terrible. This is required [by `d2l-labs-attachment-list`](https://github.com/Brightspace/attachment/blob/master/components/attachment-list.js#L58) in the event handler for `d2l-attachment-untrusted` - the different being that `e.target` is now a `d2l-activity-attachment` rather than just a `d2l-labs-attachment`, which has (and uses) this parameter. I gave it some thought, but couldn't come up with a nicer way of doing this, but am open to ideas!

mdulat: Does this mean that the untrusted URL alert will show when an activity is edited after the fact, not just when the attachment is initially added?

ryantmer: That... is a good question. Right now in AF that isn't the case, and this is using the same components, so I think it should act the same, but I'll check.

ryantmer: I see what you mean - it does indeed always show. I'm not actually sure how to reconcile that - from the entity's perspective (and the perspective of the component), we don't know if an attachment is new, or if it was already in the collection. I'll give it some thought, since having it show all the time is annoying!

mdgbayly: yeh - this is a tough one -we had the ability in AF to determine if we have a new unsaved attachment and we only wanted to show the warning the first time a user added the attachment, but not on subsequent edits. Would have to think about it to come up with something workable.

mdgbayly: it looks like you are also going to get an `Attachments` title by using the d2l-labs-attachment-list component. We always use a title in other scenarios if there are attachments. We might have to add a new property to allow it to be disabled.

ryantmer: I'm thinking maybe I can set `creating` on the attachment component via the list, and it gets set the first time an attachment is added - I'll give it a shot, not sure if it will work, but it gives us a way to sort of have state, in a mostly-stateless place.

ryantmer: Yeah, the title is always there - amusingly, there's a (private) method that determines whether or not to show the title, but right now it's only concerned with whether or not there are attachments to display :)

mdgbayly: Yeh - the showTitle is currently just a Lit computed property effectively.

ryantmer: I've added what I would describe as "a solution" to this in 5617783.

I don't particularly like it, since it's passing state up and down, and doesn't entirely work since if you dismiss the alert, it comes back when you add another attachment (since `creating` is still true). That said, the fully-stateless alternative does indeed appear to be "always see the alert so long as there is an untrusted URL attached".

mdgbayly: You mean the alert comes back if you add a new attachment that is trusted? If so, then yeh, that doesn't seem great. Maybe this should be just another feature we ignore until we have save/cancel. I was looking at an alternative very hacky approach which I used to accomplish something similar in the rubric editor. 

It basically involved setting a flag on the collection the first time it was rendered, and using that flag in combination with a flag on the the collection item the first time it was rendered to determine whether an item was newly added.

e.g. if the collection previously existed and the item is new, then the item must have been just added. 

It used observables on properties checking if the property had a previous value to make this distinction.

But it was also kind of weird - so if we think we're going to do save/cancel in which case we'll have more control over this stuff, maybe we shouldn't bother.

ryantmer: Yeah, adding any attachment causes the whole collection to re-render, with the alert (if any in the collection are untrusted, that is). So yeah, not great, and I think you're probably right about potential save/cancel implications, and it will open the door to a more state-aware situation, where we should be able to handle this more appropriately. I'll remove that commit!

PR #539, opened 2019-12-12
pass property as property instead of attribute

AlexBedley:
Context is here: https://d2l.slack.com/archives/C0PHG3QB0/p1576179194361700

Basically, we were passing token as a string instead of an object. And token is an object in the LMS which is why it worked locally (where we use string tokens) but not in QUAD.


maboelsoudd2l:
what does this object token look like?

AlexBedley:
> what does this object token look like?

it's in the slack thread, but literally 

```js
function tokenGetter(){return D2L.LP.Web.Authentication.OAuth2.GetToken(scope)}
```

maboelsoudd2l:
OMG HAHAHAHA this is the funniest thing ive ever seen 🤣 😆 
image

PR #542, opened 2019-12-13
minor fixes

maboelsoudd2l:
Wanted to address small things that annoyed me. Just a mini-dragon-slayer.

Specifically:
When we press cancel in the dialog it logs errors.
I didn't like the way we reset the dialog.

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -434,6 +434,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
 
 		return actionDialog.open().then(action => {
+			if (!action) return;
 
 			const { selectedRadio, date } = JSON.parse(action);
 			let dismissAction;
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
@@ -57,12 +57,20 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 				<p class="d2l-label-text">${this.localize('dismissUntil')}</p>
 				<div class="radio-container">
 					<label class="d2l-input-radio-label">
-						<input @change="${this._updateProp.bind(this, DISMISS_TYPES.date)}" type="radio" name="${groupID}">
+						<input
+						type="radio"
+						name="${groupID}"
+						.checked="${this.selectedRadio === DISMISS_TYPES.date}"
+						@change="${this._updateProp.bind(this, DISMISS_TYPES.date)}">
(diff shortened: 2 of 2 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -434,6 +434,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 		const actionDialog = this.shadowRoot.querySelector('d2l-quick-eval-action-dismiss-dialog');
 
 		return actionDialog.open().then(action => {
+			if (!action) return;
```
maboelsoudd2l: if there is no action, don't try to do stuff, we are probably cancelling the dialog.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -57,12 +57,20 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 				<p class="d2l-label-text">${this.localize('dismissUntil')}</p>
 				<div class="radio-container">
 					<label class="d2l-input-radio-label">
-						<input @change="${this._updateProp.bind(this, DISMISS_TYPES.date)}" type="radio" name="${groupID}">
+						<input
+						type="radio"
+						name="${groupID}"
+						.checked="${this.selectedRadio === DISMISS_TYPES.date}"
```
maboelsoudd2l: way better than querying the children of a radio selector

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -77,14 +85,8 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 	}
 
 	_clearSelectedOptions() {
-		this.selectedRadio = undefined;
-		this._date = undefined;
-
-		// Get the Div holding the radio buttons and clear them all
-		const elements = this.shadowRoot.querySelector('.radio-container').children;
```
maboelsoudd2l: Never been a big fan of these selectors for a couple of reasons, specially class names for a few reasons:
1. they create an implicit dependency with the stuff in the render function.
2. they create an implicit dependency for stuff in the styles.

AlexBedley:
I like these changes but I'm hesitant to introduce more risk at this point. I think we should hold off until after the release

maboelsoudd2l:
@AlexBedley the `if (!action) return;` is necessary in my opinion because otherwise kibana will complain about it every time someone presses cancel.

components/d2l-quick-eval/d2l-quick-eval-activities.js
AlexBedley: can we be certain they are cancelling the dialog?

maboelsoudd2l: There are multiple buttons that close the dialog, we want any button that isn't submit, to cancel the behaviour.

PR #543, opened 2019-12-13
Fix null reference when dismiss flag is off.

cbusmand2l:


components/d2l-quick-eval/behaviors/d2l-quick-eval-siren-helper-behavior.js
```diff
@@ -84,7 +84,10 @@ D2L.PolymerBehaviors.QuickEval.D2LQuickEvalSirenHelperBehaviorImpl = {
 
 	getDismissHref: function(entity) {
 		//TODO: change once it's in its final form in HM constants.
-		return entity.getLinkByRel('https://api.brightspace.com/rels/dismiss').href;
+		const link = entity.getLinkByRel('https://api.brightspace.com/rels/dismiss');
```
AlexBedley: yeah this happens because the link does not appear when the flag is off

PR #544, opened 2019-12-13
Hide hidden/visible toggle

m6jones:
No courses no switch

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-visibility-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-visibility-editor.js
@@ -12,6 +12,7 @@ class ActivityVisibilityEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMi
 
 	static get properties() {
 		return {
+			disabled: { type: Boolean },
 			_isDraft: { type: Boolean },
 			_canEditDraft: {type: Boolean }
 		};
@@ -67,7 +68,7 @@ class ActivityVisibilityEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMi
 		const icon = (this._isDraft ? 'tier1:visibility-hide' : 'tier1:visibility-show');
 
 		return html`
-			<div ?hidden=${!this._canEditDraft}>
+			<div ?hidden=${!this._canEditDraft || this.disabled}>
 				<d2l-switch
 					aria-label="${switchVisibilityText}"
 					label-right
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -1,90 +1,91 @@
...
-			<div d2l-label-text ?hidden=${this._canEditDraft}>
-				<d2l-icon icon=${icon}></d2l-icon>
-				${switchVisibilityText}
-			</div>
-		`;
-	}
-
-}
-customElements.define('d2l-activity-visibility-editor', ActivityVisibilityEditor);
+import '@d2l/switch/d2l-switch.js';
```
m6jones: Line endings back to LF

components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -1,90 +1,91 @@
...
+		this.wrapSaveAction(super._entity.setDraftStatus(!this._isDraft));
+	}
+
+	render() {
+
+		const switchVisibilityText = (this._isDraft ? this.localize('hidden') : this.localize('visible'));
+		const icon = (this._isDraft ? 'tier1:visibility-hide' : 'tier1:visibility-show');
+
+		return html`
+			<div ?hidden=${!this._canEditDraft || this.disabled}>
```
AllanKerr: nit: Wrap the logic that checks `canEditDraft` and `disabled` in a function that can be re-used for both checks rather than duplicating it.

components/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -1,90 +1,91 @@
...
+import { getLocalizeResources } from './localization';
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { SaveStatusMixin } from './save-status-mixin';
+
+const baseUrl = import.meta.url;
+class ActivityVisibilityEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMixin(LitElement))) {
+
+	static get properties() {
+		return {
+			disabled: { type: Boolean },
```
AllanKerr: I assume lit guarantees boolean properties will be initialized to `false` or should we do this in the constructor?

m6jones: Yeah, it is falsy with out doing it so it is fine.

dbatiste: I think the lit-analyzer may complain in certain instances where the boolean is used but is undefined.  To avoid such error, initialization in the constructor fixes it.

components/d2l-activity-editor/d2l-activity-visibility-editor.js
m6jones: Actually, it probably a better idea to not hide and only render the proper html.

PR #547, opened 2019-12-14
Allow for better partial re-rendering

ryantmer:
With BrightspaceHypermediaComponents/siren-sdk#109, we can now more precisely control partial re-renders of the page as things load. This change makes it so that when the page is first loading, we show the "Loading" text in place of the full editor. Once the page has finished loading once (which we know by `d2l-pending-resolved` being fired), we won't ever show that text again. This allows for lower-level components on the page to re-fetch data without the whole page looking like it's reloading.

This approach is also more compatible with lower-level `PendingContainer` elements - previously, when we were fetching a new attachment, the whole page would go to the "Loading" state, which meant we wouldn't even see the nice attachment skeleton that `d2l-labs-attachment` has!

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -22,6 +22,7 @@ class AssignmentEditor extends PendingContainerMixin(EntityMixinLit(LitElement))
 			 */
 			trustedSitesEndpoint: { type: String },
 			_assignmentHref: { type: String },
+			_initialLoadComplete: { type: Boolean }
 		};
 	}
 
@@ -60,13 +61,20 @@ class AssignmentEditor extends PendingContainerMixin(EntityMixinLit(LitElement))
 		}
 	}
 
+	_onPendingResolved() {
+		// Once we've loaded the page once, this prevents us from ever showing
+		// the "Loading..." div again, even if page components are (re)loading
+		this._initialLoadComplete = true;
+	}
+
 	render() {
 		return html`
 			<d2l-activity-editor
-				?loading="${this._hasPendingChildren}"
+				?loading="${this._hasPendingChildren && !this._initialLoadComplete}"
 				unfurlEndpoint="${this.unfurlEndpoint}"
 				trustedSitesEndpoint="${this.trustedSitesEndpoint}"
-				@d2l-request-provider="${this._onRequestProvider}">
+				@d2l-request-provider="${this._onRequestProvider}"
+				@d2l-pending-resolved="${this._onPendingResolved}">
 
 				<d2l-activity-assignment-editor-detail
 					href="${this._assignmentHref}"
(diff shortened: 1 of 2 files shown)


ryantmer:
For some additional context, this fixes the issue you were seeing @mdulat where the whole page seems to reload (but without a network request) when you add an attachment. I'm not sure if this will fix the _double_ re-render you were seeing, but we'll see! (I can't replicate that one myself.)

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
```diff
@@ -23,6 +23,7 @@ class ActivityAttachment extends EntityMixinLit(LitElement) {
 	constructor() {
 		super();
 		this._setEntityType(AttachmentEntity);
+		this._attachment = {};
```
ryantmer: erm... I guess this was unrelated, but it prevents a momentary console error so I guess have a bonus tiny bugfix! 🤷‍♂

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-list.js
```diff
@@ -51,16 +52,19 @@ class ActivityAttachmentsList extends EntityMixinLit(LitElement) {
...
-				${repeat(this._attachmentUrls, href => href, href => html`
-					<li slot="attachment" class="panel">
-						<d2l-activity-attachment
-							href="${href}"
-							.token="${this.token}">
-						</d2l-activity-attachment>
-					</li>
-				`)}
-			</d2l-labs-attachment-list>
+			<div ?hidden="${!this._hasPendingChildren}">Loading ...</div>
```
mdulat: Should this be a lang term?

ryantmer: Fixed-by-removal! Turns out there was a better way here, so this PR is simpler now :)

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -3,7 +3,6 @@ import './d2l-activity-assignment-editor-detail.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { AssignmentActivityUsageEntity } from 'siren-sdk/src/activities/assignments/AssignmentActivityUsageEntity.js';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
-import { PendingContainerMixin } from 'siren-sdk/src/mixin/pending-container-mixin.js';
```
mdulat: Isn't this still being used?

ryantmer: You managed to comment on a change that was removed via a force-push... I think you might be in a time portal! (But you are correct, I forgot to remove this addition initially 😄)

PR #549, opened 2019-12-16
resized dialog on _onDatetimePickerDatetimeChanged

maboelsoudd2l:
I don't exactly love the way we are doing this. We are essentially calling the resize function at specific points where we know that the layout of this dialog changes.

This is due to the way that the dialog is designed. @dbatiste recommended that we do not overuse the resize function, but I don't see another way to fix this issue.

### DIFF
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
@@ -142,6 +142,7 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 
 	_onDatetimePickerDatetimeChanged(e) {
 		this._date = e.detail.toISOString();
+		this._getDialog().resize();
 	}
 
 	computeAction(selectedRadio, date) {


dbatiste:
Is this to address the issue with the datetime being hidden behind the footer, etc?  If so, I would like to find a way to get that floating content on the top layer above the footer.  In other cases, I think we should explicitly resize as dictated by design - for example, perhaps when transitioning between views in a multi-step task within the dialog. We can also consider automatically resizing when content within the dialog changes size - however it's probably worth considering different use-cases before adding this.

maboelsoudd2l:
@dbatiste nope, that is a different issue. 
my current issue at hand is where the datetime component grows in size due to the timezone info (refer to the pics below)
image
image
notice how the "forever" radio button dissapears?

maboelsoudd2l:
also the time dropdown is also important and I was planning to follow up with you on this. Of course I wouldn't fix the dropdown overlay issue like this, we would need a different approach for that problem.

dbatiste:
Ah I see. Yeah, the idea of resizing when the content inside changes was something in our backlog.  The date/time example wasn't one I was thinking of though - the one I was considering was expanding/collapsing of content in the dialog.  I wasn't sure if we would always want it to automatically resize, so the story got punted until a concrete use-case came up.

dbatiste:
It's not really any more of a perf concern than making other changes that would require layout and repaint.  That being said, I wouldn't put it a loop or call it recursively or anything like that.

maboelsoudd2l:
yeah basically this runs only on user input, so highly unlikely that it would cause perf issues. Also internally it seems that the resize function is just a bunch of math and changing dom properties, wouldn't call it anything serious.

PR #550, opened 2019-12-16
Adding visdiff tests

cbusmand2l:


test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.html
```diff
@@ -0,0 +1,32 @@
...
+			height: 500px;
+		}
+		</style>
+		<script type="module">
+			import '@brightspace-ui/core/components/typography/typography.js';
+		</script>
+	</head>
+	<body style="width: 884px;height: 884px;" class="d2l-typography">
+		<div id="default">
+			<d2l-quick-eval-action-dismiss-dialog></d2l-quick-eval-action-dismiss-dialog>
```
maboelsoudd2l: Would it make your life easier if you just add `<d2l-quick-eval-action-dismiss-dialog .selectedRadio="dismiss-until">` instead of doing all this clicking stuff?

Or does that not work?

maboelsoudd2l:
Sooo I'm not super opposed to the approach used in these diff tests, but based on our experience with testing dialogs, they were basically a nightmare. They kept failing when we didn't want them to, and the delays we had never worked out for us. Things were so bad we ended up ditching the test we wrote altogether (if you guys remember a few days ago).

Now not only are we adding adding those same tests we ditched, now there's three of them. Am I the only one who feels like we are shooting ourselves in the foot here?

I think there are two alternative approaches that we can take:
1. Find a way to divorce the content from the dialog and test the content separately (very annoying since we tightly-coupled the content to the dialog).
2. Just keep pestering @dbatiste until he can get d2l-dialog to have the option to load `.open` without the crashing.
3. keep playing with the delays till we get it right. @dbatiste mentioned that there might be a way we can add an async callback when the dialog fully opens.

AlexBedley:
looks like the `d2l-quick-eval-action-dismiss-dialog-specific-date.png` golden is cut off at the bottom

AlexBedley:
> Sooo I'm not super opposed to the approach used in these diff tests, but based on our experience with testing dialogs, they were basically a nightmare. They kept failing when we didn't want them to, and the delays we had never worked out for us. Things were so bad we ended up ditching the test we wrote altogether (if you guys remember a few days ago).
> 
> Now not only are we adding adding those same tests we ditched, now there's three of them. Am I the only one who feels like we are shooting ourselves in the foot here?
> 
> I think there are two alternative approaches that we can take:
> 
> 1. Find a way to divorce the content from the dialog and test the content separately (very annoying since we tightly-coupled the content to the dialog).
> 2. Just keep pestering @dbatiste until he can get d2l-dialog to have the option to load `.open` without the crashing.
> 3. keep playing with the delays till we get it right. @dbatiste mentioned that there might be a way we can add an async callback when the dialog fully opens.

It looks like internally `d2l-dialog` knows when it's opened. It listens for some kind of `transitionend` event and does stuff afterwards. Unfortunately it calls `stopPropagation` on the event so it won't bubble up. But maybe we can also hook into that `transitionend` thing, or expose something publically so we don't waste time during CI runs.

https://github.com/BrightspaceUI/core/blob/60be5b8893d0cce0d4e1831c26c95996562c272f/components/dialog/dialog-mixin.js#L247
https://github.com/BrightspaceUI/core/blob/60be5b8893d0cce0d4e1831c26c95996562c272f/components/dialog/dialog-mixin.js#L188

dbatiste:
> [Mohamed] Find a way to divorce the content from the dialog and test the content separately (very annoying since we tightly-coupled the content to the dialog).

I think we should remember that dialogs are already being tested in core, so I would only do visual-diff tests on a higher-level dialog component if the are "interesting" behaviors or visuals that it provides, beyond the content of the dialog.  As such, separating the content of the dialog may be advantageous to allow easier testing.

> [Mohamed] Just keep pestering @dbatiste until he can get d2l-dialog to have the option to load .open without the crashing.

I started to look into this but I thought this issue was resolved based on relies in GitHub.  I was also not able to reproduce, so I think the `.open` method should be working as expected.  Please let me know if this still needs looking into, but I will need something I can reproduce (note: currently I'm almost entirely occupied by reviews and looking into things, so I am not really making progress on fixing things... 😢  ).

> [Mohamed] keep playing with the delays till we get it right. @dbatiste mentioned that there might be a way we can add an async callback when the dialog fully opens.

> [Alex] It looks like internally d2l-dialog knows when it's opened. It listens for some kind of transitionend event and does stuff afterwards. Unfortunately it calls stopPropagation on the event so it won't bubble up. But maybe we can also hook into that transitionend thing, or expose something publically so we don't waste time during CI runs.

Yeah, animations are a real pain with visual-diffs. They are somewhat manageable in core where we are generally not dealing with nested components/shadow-roots. Dealing with nested animations in composite components was something we knew would be tricky, and we still don't have an amazing solution for them.  That's not to say we shouldn't use visual-diff, but we might be more limited in what we can reasonably test with the current version of visual-diff.

Wiring up to `transitionend` is not great for apps because the event is not composed, which means that you'd have to peek into the shadow-DOM of the dialog, which is definitely not what we want. That being said, the dialog does expose `d2l-dialog-open` and `d2l-dialog-close` events, and it is certainly possible to create async helper method for opening and closing the dialog that can be used in tests - this is what the dialog does for its visual-diff tests. However, unless the page is reloaded for each test, the dialog also needs to be closed/reset. In short, testing dialogs is non-trivial.

I would really like to provide a method to allow something like `await visualDiff.animations();`. I have an idea on how to do this, but unfortunately I haven't had time to work on it.

PR #552, opened 2019-12-16
resetting the dismissed activities list

maboelsoudd2l:
Basically this is code that resets the activities list whenever you close the dialog.

### DIFF
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
@@ -10,7 +10,7 @@ class D2LQuickEvalDismissedActivitiesList extends LitQuickEvalLocalize(LitElemen
 	render() {
 		return html`
 			<d2l-list separators="all">${this.dismissedActivities && this.dismissedActivities.length ? this.dismissedActivities.map((act, index) => html`
-				<d2l-list-item selectable key="${index}" @d2l-list-item-selected=${this._handleItemSelected}>
+				<d2l-list-item ?selected=${act.selected} selectable key="${index}" @d2l-list-item-selected=${this._handleItemSelected}>
 					<d2l-icon slot="illustration" icon="${this._computeIcon(act.type)}" aria-label="${this.localize(act.type)}"></d2l-icon>
 					<d2l-list-item-content>
 						${act.name}
@@ -23,7 +23,9 @@ class D2LQuickEvalDismissedActivitiesList extends LitQuickEvalLocalize(LitElemen
 
 	static get properties() {
 		return {
-			dismissedActivities: { type: Array }
+			dismissedActivities: { type: Array, hasChanged() {
+				return true;
+			}},
 		};
 	}
 
--- a/components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
+++ b/components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
@@ -10,7 +10,9 @@ class D2LQuickEvalEllipsisDialog extends LitQuickEvalLocalize(LitElement) {
 	static get properties() {
 		return {
 			opened: { type: Boolean },
-			dismissedActivities: { type: Array },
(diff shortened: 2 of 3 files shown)


components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
```diff
@@ -23,7 +23,9 @@ class D2LQuickEvalDismissedActivitiesList extends LitQuickEvalLocalize(LitElemen
 
 	static get properties() {
 		return {
-			dismissedActivities: { type: Array }
+			dismissedActivities: { type: Array, hasChanged() {
+				return true;
```
maboelsoudd2l: I have to add this code, because Lit-element has a hard time understanding something has changed when an array is involved.

This code basically says that whenever the parent re-renders, re-render this component *ALWAYS*, even if there are no changes. Now this may appear to be inefficient, but the alternative would be looking at every element in the array to see if it has changed. I just figured that this might be slower than a bunch of DOM re-rendering.

I might be wrong...

components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-dismiss-dialog.js
```diff
@@ -10,7 +10,9 @@ class D2LQuickEvalEllipsisDialog extends LitQuickEvalLocalize(LitElement) {
 	static get properties() {
 		return {
 			opened: { type: Boolean },
-			dismissedActivities: { type: Array },
+			dismissedActivities: { type: Array, hasChanged() {
+				return true;
```
maboelsoudd2l: same as what I explained in `d2l-quick-eval-dismissed-activities-list.js` applies here. The reason why I'm doing this twice is because the data travels 2 layers deep in order for the view to change.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities-list.js
AlexBedley: I think the correct way to do this in Lit is to treat properties as immutable, since it only checks `newVal !== oldVal`. This is a difference from polymer, which provides `this.set` which allows you to mutate the array in an observable way.

So to make the change observable in Lit you could do like:

```js
// make any change change
dismissedActivities[0].selected = false;
// use one of the many ways to shallow copy the array, in this case spread operator
dismissedActivities = [...dismissedActivities];
```

AlexBedley: In terms of what to do for this PR, I would prefer we try and do things "the lit way", but I'm not sure how Polymer will behave

AlexBedley: @maboelsoudd2l what do you think?

maboelsoudd2l: basically the solution to this is something along the lines of this:
```javascript
hasChanged(newList, oldList) {
					for (var i = 0; i < newList.length; i++) {
						if (newList[i].selected != oldList[i].selected) { // something changed
							return true;
						}
					}
					return false;
				}
```
What i did instead is `return true` instead of going through this whole array (at the cost of some performance).

maboelsoudd2l: UPDATE: people in the react world encountered this problem and have a solution for this called fast-deep-equal https://www.npmjs.com/package/fast-deep-equal

```javascript
import equal from 'fast-deep-equal';
// ...
hasChanged(newList, oldList) {
    return equal(newList, oldList);
}
```

AlexBedley: Right you already suggested that in your original comment. My comment was about making the code more idiomatic to Lit

AlexBedley: I'm ok with any of the suggested solutions, although I'd prefer not to involve another dependency. I think just duplicating the array in the controller is the most maintainable

PR #556, opened 2019-12-18
Abedley/us112373/refresh behavior

AlexBedley:
Problem: After modifying one of the 3 lists (activities, submissions, dismiss) the UI does not refresh in all cases (due to a bug)
Problem: The front-end entity store assumes things are cached forever

Solution: Instead of trying to be fancy by relying on the EntityBehaviour stuff (which we could not get to to work due to how we use it previously) just explicitly set `this.entity` to the result of `fetch`. Since we are calling `fetch` directly, we can send in the `bypassCache` flag to invalidate the cache.


components/d2l-quick-eval/behaviors/d2l-quick-eval-refresh-behavior.js
```diff
@@ -0,0 +1,38 @@
...
+window.D2L.PolymerBehaviors.QuickEval = window.D2L.PolymerBehaviors.QuickEval || {};
+
+/*
+* Behavior for refreshing the current href
+* @polymerBehavior
+*/
+D2L.PolymerBehaviors.QuickEval.D2LQuickEvalRefreshBehaviorImpl = {
+
+	created: function() {
+		this._boundRefresh = this.refresh.bind(this);
```
AlexBedley: we have to do this, otherwise the value of `this` in `refresh` will be `window` and not the controller

components/d2l-quick-eval/behaviors/d2l-quick-eval-refresh-behavior.js
```diff
@@ -0,0 +1,38 @@
...
+	attached: function() {
+		window.addEventListener('d2l-quick-eval-refresh', this._boundRefresh);
+	},
+
+	detached: function() {
+		window.removeEventListener('d2l-quick-eval-refresh', this._boundRefresh);
+	},
+
+	refresh: function() {
+		const selfHref = this._getSelfLink(this.entity);
```
AlexBedley: since we don't ever change `this.href` this is the best way to retrieve our current href. It does rely on the hm api being well-behaved (and supplying the true self href)

components/d2l-quick-eval/behaviors/d2l-quick-eval-refresh-behavior.js
```diff
@@ -0,0 +1,38 @@
...
+
+		if (selfHref) {
```
AlexBedley: this will be `false` in the case where we haven't loaded the controller yet (due to lazy loading)

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -449,9 +451,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			}
 
 			return dismissAction.then(() => {
-				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh-activities'));
-				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh-submissions'));
-				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh-dismissed'));
+				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh'));
```
AlexBedley: I consolidated the 3 events, since they all need to be refreshed when things change

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -548,28 +550,6 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
-
-	disconnectedCallback() {
-		super.disconnectedCallback();
-		window.removeEventListener('d2l-quick-eval-refresh-submissions', this._boundRefresh);
-	}
-
-	refresh() {
-		const selfHref = this._getSelfLink(this.entity);
-		if (selfHref) {
-			window.D2L.Siren.EntityStore.fetch(selfHref, this.token, true).then(x => this.entity = x.entity);
```
AlexBedley: Notice how submissions was different than the other 2. This is because we thought `activities `and `dismiss `"worked" but we were wrong, hence the bug.

AlexBedley:
I'm having trouble thinking of unit tests, will think about while this is being reviewed.

AlexBedley:
I found a way to write a unit test. Not sure how useful it ended up being, but it exists.

PR #558, opened 2019-12-19
US112361 Add file selector dialog for attachments

ryantmer:
Similar to change made in #500 for link attachments, this adds and connects a button to open a file picker dialog. Also similar to #500, this required getting deep into LMS dialog JavaScript, which is ugly and not very maintainable.

This approach assumes we are living on an LMS page that has this JS available - fine since that is currently the case, but not great. The code is also very finicky, with a bunch of mystery parameters and such. A better solution, which is currently in the works, is for us to have a `d2l-dialog` that supports a filepicker and use that - in the short term, it will likely still mean this code exists somewhere (within the `d2l-dialog`), but for our use case here, it will mean we're just interacting with a WC with a clean API.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -11,6 +11,7 @@ import { SaveStatusMixin } from '../save-status-mixin';
 class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeMixin(LitElement))) {
 	static get properties() {
 		return {
+			_canAddFile: { type: Boolean },
 			_canAddLink: { type: Boolean },
 			_canAddGoogleDriveLink: { type: Boolean },
 			_canAddOneDriveLink: { type: Boolean }
@@ -61,7 +62,8 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 		}
 
 		if (entity) {
-			this._canAddLink  = entity.canAddLinkAttachment();
+			this._canAddFile = entity.canAddFileAttachment();
+			this._canAddLink = entity.canAddLinkAttachment();
 			this._canAddGoogleDriveLink = entity.canAddGoogleDriveLinkAttachment();
 			this._canAddOneDriveLink = entity.canAddOneDriveLinkAttachment();
 		}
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -85,16 +90,82 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
...
 
 		event.AddListener(callback);
+	}
 
-		return orgUnitId;
+	_launchAddFileDialog() {
+		const opener = this.shadowRoot.querySelector('#add-file-button');
+
+		const params = new URLSearchParams();
+		params.set('ou', this._orgUnitId);
```
ryantmer: This is all entirely 100% gross. I'm now working on getting this over to `d2l-dialog`, so that at least it can be a facade for all of the grossness. Long-term, the plan is to somehow re-create the file selector as a webcomponent rather than relying on a bunch of LMS callbacks, but moving this over to `d2l-dialog` will make that transition smoother (hopefully invisible) for consumers like FACE.

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -108,8 +179,8 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 		// Required for the async handler below to work in Edge
 		const superEntity = super._entity;
 
-		const orgUnitId = this._openDialog(opener, settings, async event => {
-			const quicklinkUrl = `/d2l/api/lp/unstable/${orgUnitId}/quickLinks/${event.m_typeKey}/${event.m_id}`;
+		this._openDialog(opener, settings, async event => {
+			const quicklinkUrl = `/d2l/api/lp/unstable/${this._orgUnitId}/quickLinks/${event.m_typeKey}/${event.m_id}`;
 			const response = await fetch(quicklinkUrl);
 			const json = await response.json();
 			this.wrapSaveAction(superEntity.addLinkAttachment(event.m_title, json.QuickLink));
```
mdgbayly: I think we will still need to address this at some point? Currently it is attaching linsks with a hardcoded orgunitid? And we will need to come up with a way to use templated quicklinks and replace the orgunitid on render?

ryantmer: Right yes! Joseph asked about that yesterday - I'll attempt to fix both in one go, there is a different property that has the templated links, I'm just not sure what the implications of using it are yet. I'll look into that today!

PR #559, opened 2019-12-19
Cbusman/visdiff tests for ellipsis dismiss dialog

cbusmand2l:


cbusmand2l:
> Is there a way to move `waitForOpen` and `waitForClose` to a test helper (and pass the event name in) to reduce duplication?
Possibly? Unfortunately, this and the other dialog in QE both interact with `d2l-dialog` differently, so we would also need to pass in an "open" and "close" callback (the ellipsis dialog uses `open()` and doesn't itself pass through a close to the `d2l-dialog` so we have to drill to it directly, while this dialog uses an `opened` property that's bound directly to the underlying dialog).

jeroach:
> Possibly? Unfortunately, this and the other dialog in QE both interact with d2l-dialog differently, so we would also need to pass in an "open" and "close" callback (the ellipsis dialog uses open() and doesn't itself pass through a close to the d2l-dialog so we have to drill to it directly, while this dialog uses an opened property that's bound directly to the underlying dialog).

Yeah that sounds messy (aka more trouble than worth). This is good as is - it was just a thought :)

PR #562, opened 2019-12-20
Use QuickLinkTemplate to allow for copying

ryantmer:
If we use the QuickLink directly, copying the assignment to another course will end up causing issues as the copied assignment will have attachments in another course. Instead, we can use the QuickLinkTemplate so that when an assignment is copied to another course, the attachments will refer to the new org unit.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -112,7 +112,7 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 			const quicklinkUrl = `/d2l/api/lp/unstable/${orgUnitId}/quickLinks/${event.m_typeKey}/${event.m_id}`;
 			const response = await fetch(quicklinkUrl);
 			const json = await response.json();
-			this.wrapSaveAction(superEntity.addLinkAttachment(event.m_title, json.QuickLink));
+			this.wrapSaveAction(superEntity.addLinkAttachment(event.m_title, json.QuickLinkTemplate));
 		});
 	}
 


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -112,7 +112,7 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 			const quicklinkUrl = `/d2l/api/lp/unstable/${orgUnitId}/quickLinks/${event.m_typeKey}/${event.m_id}`;
 			const response = await fetch(quicklinkUrl);
 			const json = await response.json();
-			this.wrapSaveAction(superEntity.addLinkAttachment(event.m_title, json.QuickLink));
+			this.wrapSaveAction(superEntity.addLinkAttachment(event.m_title, json.QuickLinkTemplate));
```
ryantmer: Turns out, it is indeed this easy! Confirmed that if I copy this assignment to another course, any attachments are also updated to point to the new course.

mdgbayly: Yep, i think that's the correct solution for creating the attachments.

But what does this mean for rendering the attachments? e.g. now the link we include in the rendered `d2l-labs-attachment` will have an `ou={orgUnitId}` query parameter or something like that? So what happens when you click on that link? Does some other LMS magic automatically resolve it to the current org unit, or do we need to replace it prior to rendering somehow with the current org unit.

ryantmer: LMS magic is the answer, fortunately! The link attachment is stored with `{orgUnitId}`, but when it's fetched the orgUnitId is filled in. (Specifically, we end up using the `PreparedUrl` rather than the `Url`.)

ryantmer: Er, to be clear - that happens when we fetch the link attachment, so the UI doesn't ever see the `{orgUnitId}` at all.

mdgbayly: Interesting - do you know where that replacement happens? Just wondering on the implications for Save/Cancel.

dbatiste: The handler (in the QuickLinks code) for `/d2l/common/dialogs/quickLink/quickLink.d2l` resolves the link to point to the actual course component using the query string parameters (orgUnitId, QuickLink type, and resource code), and redirects the user to the actual content.

I think CopyCourseComponents has logic that will copy the resources, including the resource codes from the source to target (resource code stays the same, mapped ids are updated to point to new copied components).  Otherwise, if the the original source component is updated or deleted, the user would be redirected to the source component, or they would get a broken link. As such, it's very important that QuickLink URLS are handled in this way.

I also think CopyCourseComponents has some special code to replace orgUnitId in Content topics, since it looks like they are stored with the actual orgUnitId (since at the time it was developed, there was limited infrastructure for view-time replacements), and it may look for them using that URL (`quickLink.d2l`).

ryantmer: oops - we actually followed up on this offline a bit, but didn't close this loop. That's an excellent description though, thanks @dbatiste! Wasn't totally sure how it works, so it's good to have some insight beyond just "well, it seems to work via some ancient magic" 😄

dbatiste: Cool yeah.  My knowledge of this is several years old, and I am only really familiar with is as far as the UI-domain integration goes.  Glad you got it sorted... 😄

PR #566, opened 2019-12-23
Abedley/backport/refresh

AlexBedley:
I'm backporting the [refresh fix](https://github.com/BrightspaceHypermediaComponents/activities/pull/556) to 20.20.1 @dmackenz please read the other PR's annotated diff for an explanation on what the refresh bug is.

This PR was made with `git cherry-pick` so I'm reasonably certain the diffs are the same. We haven't made any major changes in 20.20.2 to these files, so there are no specific regressions I'm worried about, but that doesn't mean I couldn't have broken anything :P I am currently testing on a 20.20.1 LMS.


AlexBedley:
Ah I also needed Carol's change to delete the failing subtitle vdiff tests

PR #567, opened 2019-12-23
US112359 - Added always-show-time to dismiss specific date d2l-datetime-picker

dmackenz:
For this story, we need to make sure that a user is able to add a specific time for the "dismiss specific date" feature in the dismiss dialog. More info found here https://rally1.rallydev.com/#/42960320374d/detail/userstory/355285476896


components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -84,6 +84,13 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		`;
 	}
 
+	updated(changedProperties) {
```
AlexBedley: What's the difference between Mohamed's `resize()` call in `_updateProp` and your `resize()` call here? Do we need both of them?

dmackenz: I'm actually not sure why Mohamed's method was not working. It may be because it gets called before the rendering finishes? I tested it after removing mohamed's call to resize() and saw no difference.

dmackenz:
For this story, we need to make sure that a user is able to add a specific time for the "dismiss specific date" feature in the dismiss dialog. More info found here https://rally1.rallydev.com/#/42960320374d/detail/userstory/355285476896

PR #571, opened 2019-12-28
Us111489 add to grades

mdulat:


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	_toggleScoreState(isUngraded) {
+		const ungradedButton = this.shadowRoot.querySelector('#ungraded-button-container');
+		const scoreInfo = this.shadowRoot.querySelector('#score-info-container');
+
+		const toHide = isUngraded ? scoreInfo : ungradedButton;
+		const toShow = isUngraded ? ungradedButton : scoreInfo;
+
+		toHide.setAttribute('hidden', 'hidden');
```
mdgbayly: I think generally it is considered less optimal to make DOM changes outside of render.
https://lit-element.polymer-project.org/guide/templates#design-a-performant-template

Not sure if this really affects things much here, but I think it is less idiomatic lit-element to manipulate the DOM in ths way.

I'm also not sure on the benefits of using hidden attributes vs just using some conditional rendering logic. I think it is also considered more idiomatic to just use conditional rendering. I think the use of `hidden` attribute is a left over vestige from polymer `dom-if`s etc.

Switching to just use properties with conditional rendering might make it harder to apply the focus logic, but i think the more idiomatic way to do focus handling is to check for changed properties in the `updated` lifecycle callback and apply the focus change when the old/new property values indicate focus should change.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -0,0 +1,247 @@
...
+			this.setError(errorProperty, scoreErrorLangterm, tooltipId);
+		} else {
+			this.clearError(errorProperty);
+			this.wrapSaveAction(super._entity.setScoreOutOf(scoreOutOf, this._isUngraded() && !this._preventNewGrade));
+		}
+	}
+
+	_addToGrades() {
+		if (!this._scoreOutOfError) {
+			this._preventNewGrade = false;
```
mdgbayly: I'm having a bit of a hard time understanding the logic behind the `_preventNewGrade` property. It feels like it could be simplified to just use `_inGrades` and the `scoreOutOf` value, but maybe there is an edge case that I'm not following.

What is the use case where `entity.setScoreOutOf` might be called where a true value for `preventNewGrade` is required?

components/d2l-activity-editor/d2l-activity-score-editor.js
dbatiste: ➕ x 4.

dbatiste:
I notice the [sort member rules](https://github.com/Brightspace/eslint-config-brightspace/blob/master/sort-member-config.js) are disabled for this project - I guess there are some linting errors when they are enabled?  Ideally we should try to follow the order rules even if we have them disabled, since eventually we'd like them enabled.  Not a huge deal, but it definitely helps if you know where look for things in the code.  😄

ryantmer:
Just tried turning the rule back on - since this is a pretty big multi-project repo, the list of errors is quite long 😬

mdgbayly:
> I notice the [sort member rules](https://github.com/Brightspace/eslint-config-brightspace/blob/master/sort-member-config.js) are disabled for this project - 

Yeh, I think I had to disable it when I first started adding lit-element stuff to this repo, as before that the repo was using an older eslint config, and they were not conforming to many of the new rules.

I did add a folder specific .eslintrc to the `d2l-activity-editor` folder so that it could use the lit config. Possibly I could have added the sort class members rule back in? Don't recall if I tried that at the time or just had other things on my mind!

components/d2l-activity-editor/d2l-activity-score-editor.js
mdulat: Sorry should have explained this in the description. In the UI when you click the "ungraded" button you see this:
image

With auto-save, at this point based on what we see, we should create a new grade item with a score value of empty, which is not possible. Therefore we must display "in grades" even though the entity data does not agree with this because we didn't actually create a grade item yet. So `_preventNewGrade` is used to display "in grades" even though it is not in grades at this point.

Also, if the user initially selects "remove from grades" before entering a score value in the above screenshot, we don't want to create a grade item. Otherwise we DO create a grade item if it is currently ungraded. This property allows us to specify not to create a grade item, even though we are going from empty score to numerical value for score, which is normally the trigger to create a new grade item. Basically this is a hack that allows us to do things differently when the current entity state does not agree with what the user is seeing.

mdgbayly: Maybe it's just the naming of the variable which isn't really speaking to me, but it still feels like it might be superflous and we could achieve the desired functionality with just the existing `inGrades` property.

> With auto-save, at this point based on what we see, we should create a new grade item with a score value of empty, which is not possible. 

Couldn't we just set `inGrades` to true here?

> Also, if the user initially selects "remove from grades" before entering a score value in the above screenshot, we don't want to create a grade item. Otherwise we DO create a grade item if it is currently ungraded. 

Couldn't we just toggle the `inGrades` property when the user selects 'remove from grades' or 'add to grades' before they actually enter a score value? I think we need to check the score value anyway when they select 'add to grades'? What if they select remove from grades and then add to grades before they enter a score value?
Would it currently try to call `_entity.addToGrades` anyway because the `_scoreOutOfError` would still be false?

mdulat: > Couldn't we just set inGrades to true here?
Do we want to make inGrades inconsistent with the corresponding entity property? I thought this property is supposed to represent the current entity state?

> What if they select remove from grades and then add to grades before they enter a score value?
It does call addToGrades in this case, which throws an error, which the UI ignores. But it keeps showing "in grades" because of `_preventNewGrade`.

mdgbayly: > Do we want to make inGrades inconsistent with the corresponding entity property? I thought this property is supposed to represent the current entity state?

I don't think it has to necessarily reflect the entity state. As the entity state is still maintained by the entity object. So i think it's more what makes sense from the perspective of the UI component.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -0,0 +1,252 @@
...
+		}
+
+		super._entity = entity;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldValue, propName) => {
+			if (propName === '_showUngraded') {
```
mdgbayly: Looks good - i think prop is missing which is probably why the focus was breaking?

mdulat: Yup, fixed now.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -0,0 +1,252 @@
...
+		const scoreErrorLangterm = isScoreEmpty ? 'emptyScoreOutOfError' : 'invalidScoreOutOfError';
+		const errorProperty = isScoreEmpty ? '_emptyScoreOutOfError' : '_invalidScoreOutOfError';
+		const tooltipId = 'score-tooltip';
+
+		if ((this._inGrades && isScoreEmpty) || isScoreInvalid) {
+			this._scoreOutOf = scoreOutOf;
+			this.setError(errorProperty, scoreErrorLangterm, tooltipId);
+		} else {
+			this.clearError(errorProperty);
+			this.wrapSaveAction(super._entity.setScoreOutOf(scoreOutOf, this._shouldCreateNewGradeItem()));
```
mdgbayly: Not sure if this would make it any simpler, but my thought was that the action would just take the score value and a flag to indicate if the score is in grades. I guess the way the API was originally done was that the flag means explicitly "add to grades". It might have been simpler to just have the API take a flag which indicates whether the desire is for the score to be in grades or not, and let the backend worry about whether it needs to add or remove a grade association. e.g. the UI should be able to pass the inGrades flag as true even if there is already an association grade item and the API would just do the right thing.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldValue, propName) => {
+			if (propName === '_inGrades') {
+				const toFocus = this._isUngraded() ?
+					this.shadowRoot.querySelector('#ungraded') :
+					this.shadowRoot.querySelector('#score-out-of');
```
mdgbayly: Minor point - but does this cause focus to switch back to the score when you select remove from grades? I was thinking we might have a different prop which just controls whether we show the 'ungraded' button or the other controls. As it's slightly different to whether this is "in grades". But switching from ungraded to the other state would also set `inGrades` to true.

components/d2l-activity-editor/d2l-activity-score-editor.js
mdulat: Actually the API does just figure out whether a new grade item is required, so we can keep passing in `inGrades` even if the associated grade item exists. I needed a special check here before when I was using different params to keep track of the special cases, but now I don't need that any more, so you're right, this can be simplified. I'll fix that up.

PR #572, opened 2019-12-30
next submission option

maboelsoudd2l:
add the "next submission" option in the UI. Assuming that the backend is already done (seems to work just fine)
image


components/d2l-quick-eval/build/lang/en.js
```diff
@@ -46,6 +46,7 @@ export const LangEn = {
 	'newPosts': 'New Posts',
 	'newSubmissionDetails': '{newNum, plural, =0 {{resub, plural, =1 {1 resubmission} other {{resub} resubmissions}}} other {{resub, plural, =0 {{newNum} new} =1{{newNum} new, 1 resubmission} other {{newNum} new, {resub} resubmissions}}}}',
 	'newSubmissions': 'New Submissions',
+	'nextSubmission': 'Next Submission',
```
maboelsoudd2l: this should trigger serge. But it will silently fail in other languages. Are we okay with this?

AlexBedley: Since we will be testing other languages this iteration, I would prefer if master worked without any finagling. I think after you get a green review you should add it to all the language files.

maboelsoudd2l: the review is red because of the visual diff tests. If i get an Okay from you guys i'll update the goldens

AlexBedley: Carol is still on vacation, maybe ask Dylan? Or probably 1 approval is enough for this particular change

test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.js
```diff
@@ -56,7 +56,7 @@ describe('d2l-quick-eval-action-dismiss-dialog', function() {
 		await waitForOpen('#default d2l-quick-eval-action-dismiss-dialog');
 		await page.evaluate(() => {
 			const dialog = document.querySelector('#default d2l-quick-eval-action-dismiss-dialog');
-			const forever = dialog.shadowRoot.querySelectorAll('input')[1];
```
dmackenz: Should this Selector be changed? It seems very brittle to just be changing the array index every time the selector grabs the wrong HTMLElement.

test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.js
```diff
@@ -68,11 +68,23 @@ describe('d2l-quick-eval-action-dismiss-dialog', function() {
 		await waitForOpen('#default d2l-quick-eval-action-dismiss-dialog');
 		await page.evaluate(() => {
 			const dialog = document.querySelector('#default d2l-quick-eval-action-dismiss-dialog');
-			const specDate = dialog.shadowRoot.querySelectorAll('input')[0];
```
dmackenz: See other comment

components/d2l-quick-eval/build/lang/en.js
dmackenz: Shouldn't the lang term be added to all of the .js files, not just en.js? Also add the lang term to en.json?

dmackenz: Goldens look fine to me... As long as there aren't any issues with the dialog resizing and the overflow shadow.

maboelsoudd2l: > Shouldn't the lang term be added to all of the .js files, not just en.js? Also add the lang term to en.json?

Yeah alex seems to have said the same thing.





> Updating goldens seems fine to me... As long as there aren't any issues with the dialog resizing and the overflow shadow.

Seems fine to me https://s3.ca-central-1.amazonaws.com/visualdiff.gaudi.d2l/screenshots/BrightspaceHypermediaComponents/activities/action-dismiss-dialog/2019-12-30-18.39.28.688-6160715578f00236ff1201eedf5380530df08022-report.html

test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.js
maboelsoudd2l: Strongly agree, do you have any ideas for what we should change it to?

dmackenz: you could just give each of the inputs an id then select them by that id. `dialog.shadowRoot.querySelector('#<id>')`. Don't use `querySelectorAll`.

test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.js
```diff
@@ -68,11 +68,23 @@ describe('d2l-quick-eval-action-dismiss-dialog', function() {
...
 		const rect = await visualDiff.getRect(page, '#default');
 		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 		await waitForClose('#default d2l-quick-eval-action-dismiss-dialog');
 	});
+
+	it('next-submission', async function() {
+		await waitForOpen('#default d2l-quick-eval-action-dismiss-dialog');
+		await page.evaluate(() => {
+			const dialog = document.querySelector('#default d2l-quick-eval-action-dismiss-dialog');
+			const nextSub = dialog.shadowRoot.querySelectorAll('input')[0];
```
dmackenz: change this one too

PR #574, opened 2019-12-31
Localization and Loading Skeletons

m6jones:
Localization and loading skeletons for create learning path page (not including the Add Activity Dialog)

### DIFF
--- a/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
+++ b/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
@@ -35,8 +35,8 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		this._specialization = {};
 		this._showImages = false;
 		this._loadedImages = 0;
-		this._mainPageLoad = new Promise((resolve) => null);
-		this.ariaBusy = "true";
+		this._mainPageLoad = new Promise(() => null);
+		this.ariaBusy = 'true';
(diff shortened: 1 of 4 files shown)


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -294,58 +384,170 @@ class CollectionEditor extends EntityMixinLit(LitElement) {
...
+
+	_renderItemList() {
+		const items = repeat(this._items, (item) => item.self(), item => {
+			return html`
+				<d2l-list-item>
+					<div slot="illustration" class="d2l-activitiy-collection-list-item-illustration">
+						${this._renderCourseImageSkeleton()}
+						<d2l-organization-image class="d2l-activitiy-collection-organization-image" href=${item.self()} @d2l-organization-image-loaded="${this._onListImageLoaded}" ?hidden="${!this._showImages}"></d2l-organization-image>
+					</div>
+					<d2l-list-item-content>
```
BhamelinD2L: Is the course image skeleton always rendered? Does the real image obscure it, or is there a mechanism elsewhere that hides it at the time that the images load?

m6jones: The real image obscures it. It is the only way to prevent a flicker when the image loads. The list element shrinks in the frame when one hides and the other doesnt. This is the only solution i found.

PR #575, opened 2020-01-02
Fixed title going under toggle

m6jones:


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -224,7 +225,7 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 				position: relative;
 			}
 			.d2l-activity-collection-header-col1 {
-				max-width: 600px;
+				overflow: hidden;
```
BhamelinD2L: Justin pointed this out to me in a prior commit - having overflow hidden here crops the tab highlight box into being thinner than it should be. I think adjusting either the margin or padding will give it the space it needs to display correctly. I think this is also affecting the left side of the description.
example:
image

m6jones: We need to update those in the edit and place element. So doing that could fix the issue.

m6jones: But, lets re-visit when we have those fixed.

PR #577, opened 2020-01-02
US111218 - Toasts - Fix error in promise logic

AlexBedley:
US111218

Problem: Errors during restore were not triggering the error toast
Solution: Upon investigation, I noticed we were improperly catching exceptions (catch block instead of `Promise.catch`) I refactored the code to use `.catch`

Alex's notes: You can use the chrome dev tool feature called [Request blocking](https://developers.google.com/web/updates/2017/04/devtools-release-notes#block-requests) to test this locally, just block one or a few of the dismiss-info urls to trigger a failure.

I made a few formatting changes, as I suspect the formatting had a part to play in how complicated this block is. Best reviewed in side-by-side mode.


components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,21 +163,23 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 		}
 	}
 
-	_submitData(e) {
-		if (e.detail.action === 'done') {
+	_submitData(evt) {
```
AlexBedley: multiple variables named `e` meant that we were always logging the event, instead of the actual error

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,21 +163,23 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
+					return err;
+				});
+			})).then(results => {
+
+				const failures = results.filter(err => err instanceof Error);
+				this.failedCount = failures.length;
+
 				if (this.failedCount > 0) {
-					this._logError(e, {developerMessage: 'Error dismissing activities'});
+					this._logError(failures, { developerMessage: 'Error dismissing activities' });
```
AlexBedley: Logging the errors array now instead of the event (which didn't contain any useful information)

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,21 +163,23 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
-	_submitData(e) {
-		if (e.detail.action === 'done') {
+	_submitData(evt) {
+		if (evt.detail.action === 'done') {
 			const selectedData = this._getSelectedActivities();
 			this.selectedCount = selectedData.length;
-			const result = Promise.all(selectedData.map((act)=> {
-				try {
-					return this.performSirenAction(act.unDismiss);
-				} catch (e) {
```
AlexBedley: This was never triggered, since the Promise is asynchrounous

dmackenz: I feel like it may be better to refactor this block to use async/await instead of then/catch. I feel like it would read much better in async/await

AlexBedley: haha maybe. I guess I'm old school, I still prefer then/catch to async await, but only because I'm more comfortable with it. I'll give it a try and see what it looks like.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: haha 😆

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: I remember that we looked into this before. Async/await has no substitute for Promise.all. We have to use Promise.all. I think the most that we can do is something like:
```javascript
const result = await Promise.all(...)
```

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
maboelsoudd2l: what use is sending an array of failures? does it matter if one fail or 10?

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
dmackenz: Yeah, but then you wouldn't need all of the nested `.then`s and you wouldn't need to implement the promise chain with the awkward indentation. Also, you can use try/catch instead of `.catch` too.

dmackenz: While we're on the topic, another thing that I see a lot of in this codebase is using `Promise.resolve` and `Promise.reject`. Since async/await was introduced, I think you're supposed to use `return` instead of `Promise.resolve` in an async function and `throw` instead of `Promise.reject`. Seems like we're mixing different standards into the same functions haha.

AlexBedley: Yes there was (and is) some confusion around proper usage of `async`/`await`. Might be worth a short demo next Friday @dmackenz if you are interested!

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
AlexBedley: Having done some of these investigations after getting log messages, the more information the better IMO

AlexBedley:
Ok huge refactor last commit. `async` was much better, and it also made it much easier to restore the activities sequentially (rather than all at once), which is why there is no `Promise.all`. I think I prefer them one after another so as to not overload the client or server (in the case where you restore like 50), however that assumption may be invalid (I tested with up to 10).

AlexBedley:
I could easily refactor it back to use `Promise.all` with `async`, it just depends which we deem safer. This thread here seems to suggest chrome can handle 6 http requests at once: https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,30 +163,35 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
-					this._logError(e, {developerMessage: 'Error dismissing activities'});
-					this.shadowRoot.querySelector('.d2l-quick-eval-dismissed-list-critical').open = true;
-				} else {
-					this.shadowRoot.querySelector('.d2l-quick-eval-dismissed-list-success').open = true;
-				}
-			}).then(() => {
-				window.dispatchEvent(new CustomEvent('d2l-quick-eval-refresh'));
-			});
+	async _submitData(evt) {
+		if (evt.detail.action !== 'done') {
```
AlexBedley: I did this to avoid indenting the entire function

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,30 +163,35 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
+			} catch (err) {
+				failures.push(err);
+			}
 		}
+
+		this.failedCount = failures.length;
+
+		if (this.failedCount > 0) {
+			this._logError(failures, { developerMessage: 'Error dismissing activities' });
+			this.shadowRoot.querySelector('.d2l-quick-eval-dismissed-list-critical').open = true;
```
dmackenz: nit: this could bubble an error up the stack if the selector is not present or if there is a DOM sync error... https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

`querySelector(): HTMLElement | null`

dmackenz: If changed, just wrap the return value from querySelector in an if statement

dmackenz: One of the reasons I'm 100% for maybe using TypeScript in the future as this code wouldn't compile in ts.

AlexBedley: I'm torn a bit - while I agree this shouldn't crash the app, I also wouldn't want readers to assume we expect the toast DOM element to not be found, and I think it's extremely unlikely. So for those reasons, I'd like to keep it as-is. I wonder if we support the null propagation operator `?.`

maboelsoudd2l: my personal opinion on this is that it is assumed that a user dismisses activity at least a few minutes after the page loads. So in my opinion I personally am not super worried that the page would not have fully loaded yet, as the page takes only a few seconds to load. 

Again you can add this if statement as a best-practice, but i doubt that we would be preventing potential errors.

On the other hand @dmackenz in the PR you recently merged https://github.com/BrightspaceHypermediaComponents/activities/pull/581, timing was definitely important, because _loadData runs as soon as the page loads, and things like filters might not have loaded yet. So in that code I can see why would would have to check if things have loaded yet.

My personal rule of thumb is that, user input tends to happen pretty late.

dmackenz: All true, it's definitely not a big deal right now, but I was just thinking if the flow of the pages change in the future or the logic in which the components are rendered, this could cause a problem. I'm good for leaving it or changing it. I'm too jaded from getting compile errors from these types of things haha.

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,30 +163,35 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
 		}
 	}
 
-	_submitData(e) {
-		if (e.detail.action === 'done') {
-			const selectedData = this._getSelectedActivities();
-			this.selectedCount = selectedData.length;
-			const result = Promise.all(selectedData.map((act)=> {
```
maboelsoudd2l: so the reason why I wished we could use `Promise.all` as opposed to the for loop, is because currently the api calls do not get called consecutively. Here is a picture of this code in the network tab:
image
You can definitely see a waterfall that causes little delays that add up.

That being said, this photo is from quad today, which means that `Promise.all` might not be the right solution (maybe we're using promises wrong?). But point is, I think batching requests is something that we should try to figure out as it helps improve our performance.

AlexBedley: It's hard to tell from a screenshot, but maybe the overhead of making the request is longer than the request itself - `Promise.all`  does start all the promises "concurrently" - that is, without waiting for the previous one to resolve (or reject)

dmackenz: Is there API rate limiting or throttling on any of the endpoints that are being called?

dmackenz: Just to clarify
> Promise.all does start all the promises "concurrently"
`Promise.all` doesn't start the promises, it just waits for them to reject or resolve. They are started upon instantiation. `Promise.all` is just wrapping all of the promises into a single promise that resolves when all of the wrapped promises resolve.

AlexBedley: > Is there API rate limiting or throttling on any of the endpoints that are being called?

I don't think we do any rate limiting for our APIs

dmackenz: Synchronous calls are not ideal and buffering the requests would be better, but it's unclear how long you should make the request queues, especially considering the StackOverflow post on http connections is over 10 years old.

AlexBedley: > especially considering the StackOverflow post on http connections is over 10 years old.

there are more recent responses if you scroll down. In any case, I'm fine putting `Promise.all` back in

AlexBedley: done

components/d2l-quick-eval/dismiss/d2l-quick-eval-dismissed-activities.js
```diff
@@ -163,30 +163,37 @@ class D2LQuickEvalDismissedActivities extends mixinBehaviors(
...
+		if (evt.detail.action !== 'done') {
+			return;
+		}
+
+		const activitiesToRestore = this._getSelectedActivities();
+		this.selectedCount = activitiesToRestore.length;
+		const failures = [];
+
+		await Promise.all(
+			activitiesToRestore.map(async act => {
```
maboelsoudd2l: no wayyyy 😱 😱 😱 , is that how you batch start requests? does it work?

AlexBedley: I used this https://www.taniarascia.com/promise-all-with-async-await/

PR #581, opened 2020-01-03
US105733 - Fix for Quick Eval 1 result page return filter bug

dmackenz:
The reason why this is happening is that there is an exception/race condition occurring inside of the d2l-hm-filter component. When quick eval loads with a filter and no results, an entity with the class 'empty' is loaded. This triggers quick eval to clear the filter. This causes an exception because a subsequent entity is required before the filter can be cleared. This subsequent entity contains the 'clear' action. This bug was occurring before because the subsequent entity had not been received before the filter clearing was being called (thus resulting in a race condition).

The way that I proposed to fix this is when the 'empty' entity is received, add an event listener that listens for the subsequent entity to arrive and load the necessary data into the filter. On the event listener being triggered, clear the filter and remove the one-time event listener.


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -280,15 +285,12 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
 
-		if (this._initialLoad &&
-			entity.hasClass('empty') &&
-			(this.searchApplied || this.filterApplied)
-		) {
-			await this._clearFilterAndSearch();
-			this._initialLoad = false;
-			return;
+			if (entity.hasClass('empty') && (this.searchApplied || this.filterApplied)) {
+				this.addEventListener('d2l-hm-filter-filters-loaded', this._handleFilterLoadedNoResultsOnInitialLoad);
```
maboelsoudd2l: what's the difference between the two approaches? where before we would immediately clear the filters, now we wait for the filters to load, then we clear?

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -271,6 +271,11 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		];
 	}
 
+	async _handleFilterLoadedNoResultsOnInitialLoad(e) {
+		this.removeEventListener(e.type, this._handleFilterLoadedNoResultsOnInitialLoad);
```
maboelsoudd2l: why not just put `this.removeEventListener('d2l-hm-filter-filters-loaded',...)`? I think it's more readable and less prone for bugs.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -280,15 +285,12 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
-			entity.hasClass('empty') &&
-			(this.searchApplied || this.filterApplied)
-		) {
-			await this._clearFilterAndSearch();
-			this._initialLoad = false;
-			return;
+			if (entity.hasClass('empty') && (this.searchApplied || this.filterApplied)) {
+				this.addEventListener('d2l-hm-filter-filters-loaded', this._handleFilterLoadedNoResultsOnInitialLoad);
+				this._initialLoad = false;
+				return Promise.resolve();
```
maboelsoudd2l: I thought you said with async functions, we were supposed to just return rather than `Promise.resolve()`?

dmackenz: I'm just testing this right now, for some reason the unit test that calls this is timing out and I can't figure out why. Also can't get the tests to run locally so temporarily committing this

components/d2l-quick-eval/d2l-quick-eval-submissions.js
dmackenz: Does the spread operator work for that? I've never seen that before.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
dmackenz: Read the description, just updated it.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
maboelsoudd2l: sorry I mean `this.removeEventListener('d2l-hm-filter-filters-loaded', this._handleFilterLoadedNoResultsOnInitialLoad);`

components/d2l-quick-eval/d2l-quick-eval-submissions.js
maboelsoudd2l: 👍

components/d2l-quick-eval/d2l-quick-eval-submissions.js
maboelsoudd2l: If the `return;` was there before its surprising that it would now fail the tests that were passing before.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
dmackenz: I just did it this way because if you ever want to change the event that this function is triggered on, you only need to change it in two places, not one

test/d2l-quick-eval/d2l-quick-eval-clear-on-intial-load.html
```diff
@@ -62,7 +62,6 @@
 							module._loadData(entity)
 								.then(() => {
 									assert.isFalse(module._initialLoad);
-									assert.isTrue(clearSpy.calledOnce);
```
AlexBedley: you should probably delete the whole thing instead of just this single line

AlexBedley:
oh you already merged it lol

AlexBedley:
Do you have a gif of this working?

AlexBedley:
What about for activities?

maboelsoudd2l:
i asked that same question, apparently Dylan tested it on activities and found nothing. 🤷‍♂

AlexBedley:
> i asked that same question, apparently Dylan tested it on activities and found nothing.

tested how? it's a race condition after all

dmackenz:
> > i asked that same question, apparently Dylan tested it on activities and found nothing.
> 
> tested how? it's a race condition after all

Turns out I wasn't testing it properly and I just figured out how to reproduce it on activities as well. I will reopen this and do another pull request for activities.

dmackenz:
> What about for activities?

When I get my BSI working I will post GIFs of the solutions.

PR #584, opened 2020-01-06
Use Intl instead of LocalizeMixin for dates

dlockhart:
I've made some improvements to our Intl library such that it's a lot easier to call into its date/time/number formatting and parsing functions.

Since these aren't free to load on every page, I want to move them out of the `LocalizeMixin` (which many many components extend) and just have components import them only when they need to. So this change just calls into Intl directly.

Once the 4 places that are using `LocalizeMixin`'s Intl stuff (of which this is one) are calling Intl directly, I'm going to remove the functions from `LocalizeMixin`.


dlockhart:
They're methods on the `LocalizeMixin` that [I'll be deleting](https://github.com/BrightspaceUI/core/pull/323) once this merges.

jeroach:
> They're methods on the `LocalizeMixin` that [I'll be deleting](https://github.com/BrightspaceUI/core/pull/323) once this merges.

Sweet, thanks!

PR #585, opened 2020-01-06
Separating toggle button for visdiff tests.

cbusmand2l:


test/d2l-quick-eval/perceptual/d2l-quick-eval-view-toggle-button.visual-diff.js
```diff
@@ -0,0 +1,51 @@
...
+
+	['basic', 'right', 'left', 'selected'].forEach((name) => {
+		describe(name, () => {
+			[ 'normal', 'hover', 'focus', 'click' ].forEach((state) => {
+				it(state, async function() {
+					const selector = `#${name} d2l-quick-eval-view-toggle-button`;
+					switch (state) {
+						case 'hover':
+							await page.hover(selector);
+							await page.waitFor(1000);
```
AlexBedley: Is this just for insurance or is it because the animations take time?

cbusmand2l: The `box-shadow` render has a delay on it for some reason, despite having no transitions (which iirc, means I can't hook into `transitionend` to find out when it's done), so the latter unless there's a better way to do it.

dbatiste: Looks like there's a transition of `0.2s` on the `box-shadow` for the view-toggle-button.  Note: this is something that Dave L. recently removed from our core components here : https://github.com/BrightspaceUI/core/commit/466ccf4f732d375eefa046831772bb7c185d2d3b#diff-bc544adff72049fba29949f9d33aeba9

_Probably_ want to be consistent.

dbatiste: I have an idea on how to handle animations in a more holistic way but there some other higher priority things for me - may a Dragon Slayer task for me.

cbusmand2l: Ah! There is! I missed it because it was in the base style, not on the hover/focus. I'll update.

components/d2l-quick-eval/view-toggle/d2l-quick-eval-view-toggle-button.js
```diff
@@ -0,0 +1,93 @@
...
+
+	constructor() {
+		super();
+		this.selected = false;
+		this.side = 'middle';
+	}
+
+	static get properties() {
+		return {
+			// left, right
```
jeroach: nit: is this supposed to indicate what are valid values?

we set the value to `middle` in the constructor - should that be included in this comment

cbusmand2l: Maybe? If you want it to be a `middle` button, you just don't include either left or right so the consumer shouldn't be setting it to anything other than `left` or `right`. We're not using `middle` right now anyway (since we only have two options), but I figured having it as a default value (without rounded corners) made sense in case anything was added later.

jeroach: Makes sense to me!

PR #588, opened 2020-01-07
US111489 fix always initially showing ungraded, fix remove from grade…

mdulat:
…s setting back to ungraded
Fixes:
https://trello.com/c/NT7eSjPE/45-remove-from-grades-goes-back-to-ungraded-state 
and
https://trello.com/c/tcsMnAE3/44-initial-page-load-face-always-shows-ungraded

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -111,6 +111,7 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 			}
 			this._inGrades = entity.inGrades();
 			this._gradeType = entity.gradeType() || 'Points';
+			this._isUngraded = !this._inGrades && this._scoreOutOf.length === 0;
 		}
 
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -111,6 +111,7 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 			}
 			this._inGrades = entity.inGrades();
 			this._gradeType = entity.gradeType() || 'Points';
+			this._isUngraded = !this._inGrades && this._scoreOutOf.length === 0;
```
ryantmer: Is the `_scoreOutOf` always a string? If not, `&& !this._scoreOutOf` would work for both the string and null cases.

mdulat: I think `this._scoreOutOf` is always a string, as it's initialized in the constructor, but it does feel better to handle the null case, since in the LMS scoreOutOf could be null. I'll update this in a following PR.

PR #591, opened 2020-01-07
Add course level properties to qe

AlexBedley:
Just adding the properties, not doing anything with them yet

backend: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/6806/overview

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -132,6 +132,16 @@ class D2LQuickEval extends
 			returningToQuickEval: {
 				type: Boolean,
 				value: false
+			},
+			courseLevel: {
+				type: Boolean,
+				value: false
+			},
+			submissionFilters: {
+				type: Array
+			},
+			activityFilters: {
+				type: Array
 			}
 		};
 	}
--- a/components/d2l-quick-eval/d2l-quick-eval.js
+++ b/components/d2l-quick-eval/d2l-quick-eval.js
@@ -60,16 +60,27 @@ class D2LQuickEval extends
 					<h1 class="d2l-quick-eval-header-with-toggle" hidden$="[[!activitiesViewEnabled]]">[[headerText]]</h1>
 					<h1 class="d2l-quick-eval-header" hidden$="[[activitiesViewEnabled]]">[[headerText]]</h1>
 				</template>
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -132,6 +132,16 @@ class D2LQuickEval extends
 			returningToQuickEval: {
 				type: Boolean,
 				value: false
+			},
+			courseLevel: {
+				type: Boolean,
+				value: false
+			},
+			submissionFilters: {
+				type: Array
```
dmackenz: Do these arrays need a default value?

cbusmand2l: I would say that we don't need a default yet until we know what's coming back from the API (do we know that yet?). If `[]` is a valid response that we want to differentiate against no response/error/not making a request, then undefined as our default value might make sense. We can add in a default (if necessary) once we know how we want to deal with it.

AlexBedley: Hmm good point Carol. To me empty array `[]` suggests filtering is disabled, but `undefined` suggests maybe a version mismatch or something, which in that case maybe we want a default set of filters on the front end? Or maybe we should blow up on `undefined`? I'm not sure tbh

AlexBedley: Ok I'm gonna set the default to `[]` for now, and we can revert if we change our minds

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -60,16 +60,27 @@ class D2LQuickEval extends
...
+			<d2l-quick-eval-submissions
+				href="[[_lazySubmissionsHref]]"
+				token="[[token]]"
+				logging-endpoint="[[loggingEndpoint]]"
+				data-telemetry-endpoint="[[dataTelemetryEndpoint]]"
+				hidden$="[[_displayActivities(toggleState, activitiesViewEnabled)]]"
+				master-teacher="[[masterTeacher]]"
+				returning-to-quick-eval="[[returningToQuickEval]]"
+				course-level="[[courseLevel]]"
+				submission-filters="[[submissionFilters]]"></d2l-quick-eval-submissions>
```
AlexBedley: I just reformatted these to be on a new line for ease-of-use

AlexBedley:
Since I'm doing this to avoid merge conflicts I figured I might as well pass them down. I tested this against a real LMS and found a bug with the filters (they aren't stringifying properly). But `courseLevel` works fine.

PR #594, opened 2020-01-07
Fix listener for `d2l-pending-resolved`

wongvincent:
It was listening for the `d2l-pending-resolved` event from `d2l-activity-editor`, but the event is fired from the `PendingContainerMixin` which is included in `d2l-activity-assignment-editor` and not `d2l-activity-editor`.


dbatiste:
It would be amazing if we could switch this to use [AsyncContainerMixin](https://github.com/BrightspaceUI/core/tree/master/mixins/async-container).  It's on my list of todos - would like to use it instead of PendingContainerMixin, and remove it before anything else does. Need to choose event name though - the original from the proposal was `pending-state`.

PR #596, opened 2020-01-08
Override default button font

jeroach:
[US112217](https://rally1.rallydev.com/#/detail/userstory/354134948928?fdp=true): investigate activity card regression font issues

What it looks like now:

US112217_render

Why the problem existed, and where the fix is applied:

US112217


jeroach:
I'll regen the goldens once the vdiff tests fail :)

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
```diff
@@ -33,6 +33,7 @@ class D2LQuickEvalActivityCardActionButton extends mixinBehaviors(
 					text-align: center;
 					align-content: center;
 					justify-content: space-between;
+					font-family: 'Lato', 'Lucida Sans Unicode', 'Lucida Grande', sans-serif;
```
cbusmand2l: Is there anyway we can get this from typography (as a css variable maybe)? If it ever changes, there would have to be a manual update to this component.

dbatiste: I'm sorta surprised this is needed.  Yeah, we should not need to be setting Lato and such throughout the code.  In button, we inherit the font.

jeroach: Looking at the CSS in Chrome, the `font` from `button` appears to be Arial.

I can look into importing from typography

dbatiste: Even if you set `font-family: inherit;`?  That's what we do in core [here](https://github.com/BrightspaceUI/core/blob/master/components/button/button.js#L27).

jeroach: Thanks Dave! `font-family: inherit;` works

I really don't understand where that `Arial` font is coming from that I'm seeing in the CSS.

jeroach:
Updated image

updated

components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card-action-button.js
dbatiste: Yeah it's coming from the "User Agent Stylesheet"... i.e. the default styles of the browser.  Most elements don't have explicit font property values in the user agent stylesheet, but there are a couple... `button` is one.

jeroach: TIL :) Thanks for the info!

PR #598, opened 2020-01-08
Dmackenzie/US105733 fix one search result bugs quick eval

dmackenz:
This is the bug fix for activities as well.


dmackenz:
quick-eval-bug-fix-activities
quick-eval-bug-fix-submissions

components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -272,15 +277,12 @@ class D2LQuickEvalActivities extends mixinBehaviors(
...
-		}
 
-		if (this._initialLoad &&
-			entity.hasClass('empty') &&
-			(this.searchApplied || this.filterApplied)
-		) {
-			await this._clearFilterAndSearch();
-			this._initialLoad = false;
-			return;
+			if (entity.hasClass('empty') && (this.searchApplied || this.filterApplied)) {
```
AlexBedley: should this also be gated by `this._initialLoad` ?

dmackenz: it is by the wrapping if statement. it's just a nested version of the if statement before.

AlexBedley: ah then should be move `this._initialLoad = false;` outside of the nested if and delete it from the `finally` below? I think it would make your intentions more clear.

dmackenz: `this._initialLoad = false;` has to be inside of the if because of the return statement. It also has to be in the finally block because if `entity.hasClass('empty')` is `false` then this if block won't run. The reason why it is in multiple places is because of the `return` statement. If `entity.hasClass('empty') === true` then we don't want the `try`/`catch `block to run

AlexBedley: I'm suggesting

```js
if (initialLoad) {
    initialLoad = false;
    if (entity.hasClass('empty') {
        // ... etc
        return;
    }
}

File: components/d2l-quick-eval/d2l-quick-eval-submissions.js [hunk 2]
```diff
@@ -283,11 +283,11 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		this._loading = true;
 
 		if (this._initialLoad) {
+			this._initialLoad = false;
```
AlexBedley: needs to be deleted from the finally in this block, and then it's good to go

PR #602, opened 2020-01-09
US111908 Add Record Video button

ryantmer:
Per title, this adds the "Record Video" button to the attachments picker, and hooks up the dialog opener and dialog callback. This workflow is very similar to the workflow for adding files.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -6,15 +6,17 @@ import { AttachmentCollectionEntity } from 'siren-sdk/src/activities/AttachmentC
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 import { getLocalizeResources } from '../localization';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
 import { SaveStatusMixin } from '../save-status-mixin';
 
-class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeMixin(LitElement))) {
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -57,14 +68,18 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
...
 		D2L.ActivityEditor.FileUploadDialogCallback = files => {
 			for (const file of files) {
 				const fileSystemType = file.m_fileSystemType;
 				const fileId = file.m_id;
 				this.wrapSaveAction(super._entity.addFileAttachment(fileSystemType, fileId));
 			}
 		};
+		// Referenced by the server-side ActivitiesView renderer
+		D2L.ActivityEditor.RecordVideoDialogCallback = file => {
+			this.wrapSaveAction(super._entity.addVideoNoteAttachment(file.FileSystemType, file.FileId));
```
mdulat: Can the server call a client-side function like this? I guess if it works then the answer is yes, but that puzzles me.

ryantmer: Yeah, this is the remaining gross magic basically. The server sets up the `D2L.ActivityEditor.RecordVideoDialogOpener` that's referred to below, and part of that server-side setup is adding a callback, which it does by addressing the function.

This does introduce a magic dependency between the client and server, but most of the code I removed was magic anyway 😄 Now, the hard server-client magic is reduced to just these function names, and the `ObjectRepository.TryGet` keys below.

ryantmer: Er, to clarify though - the server doesn't actually _call_ this function, it just creates an `IJavascriptFunction` pointing to "D2L.ActivityEditor.RecordVideoDialogCallback". If this function doesn't end up being present for whatever reason, it just means that when the dialog closes it won't have a callback to call, and will just not do anything.

PR #603, opened 2020-01-09
US112479 - hide course column

AlexBedley:
Problem: In Course Level QE, it doens't make sense to show the course column in submissions
Solution: If we're in course mode, then just hide the course column (and don't make the web request)

image


components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -570,6 +548,41 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
...
+				key: 'submissionDate',
+				headers: [{ key: 'submissionDate', sortClass: 'completion-date', canSort: false, sorted: false, desc: false }]
+			},
+			{
+				key: 'masterTeacher',
+				headers: [{ key: 'masterTeacher', sortClass: 'primary-facilitator', canSort: false, sorted: false, desc: false }]
+			}
+		];
+
+		if (courseLevel) {
```
cbusmand2l: Should we be doing this in the same way we show/hide the `masterTeacher` column? I think that would involve updating `_getWidthCssClass` and `_shouldDisplayColumn` in the `submissions-table`?

AlexBedley: hrmmm yes i'm not a huge fan of the table making these kind of decisions but since it already exists we should probably use it.

cbusmand2l: I agree on both counts. Alternatively, could pull out the master teacher logic from the table and do it here, since I think that consistency is the most important (we already have two optional columns, how many more in the future?).

AlexBedley:
I decided not to touch the CSS because it looks the way I want it with no changes (tested on Chrome, Firefox)

PR #605, opened 2020-01-10
Cbusman/visdiff for submissions table

cbusmand2l:


cbusmand2l:
I tried stubbing `d2lfetch` in order to minimize digging into the activity name and user profile components, but I could get neither `window` nor a stub (to restore later) outside of `page.evaluate`, and I couldn't get `sinon` into `page.evaluate`.
Given that there were only three properties that would need to be set in order to get the minimum required for a visdiff test, I ended up just drilling into the sub components to set them manually.

AlexBedley:
The first golden `basic` is just the loading skeleton. My understanding is that it shimmers. Will that test be flakey?

PR #606, opened 2020-01-10
US112483 - Added course level you are all caught up message

dmackenz:
[Course Level QE][UI] Make new version of "you are all caught up" message (includes link to multi-course QE)

image


components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -0,0 +1,48 @@
...
+import {QuickEvalLogging} from './QuickEvalLogging.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
+import 'd2l-table/d2l-table.js';
+
+/**
+ * @customElement
+ * @polymer
+ */
+
+class D2LQuickEvalNoSubmissionsText extends QuickEvalLogging(QuickEvalLocalize(PolymerElement)) {
```
cbusmand2l: As this is a new component, should it be Lit instead of Polymer?

cbusmand2l:
visdiff tests will be added afterward?

AlexBedley:
You'll need to merge master into this branch to pass the vdiff tests in CI

components/d2l-quick-eval/build/lang/ar.js
```diff
@@ -7,6 +7,7 @@ export const LangAr = {
 			'cancel': 'إلغاء',
 			'caughtUp': 'لقد انتهيت من مهامك!',
 			'checkBackOften': 'تحقق باستمرار من احتمال توفر عمليات إرسال جديدة.',
+			'checkBackOftenCourseLevel': 'Check the Multi-Course Quick Eval to see if there are new submissions in your other courses.',
```
AlexBedley: please don't manually edit these files. You can add them to the corresponding json files and run `npm run build` to generate these. Your changes will be blown away the next time someone adds a langterm otherwise.

dmackenz: So what is the workflow? Do you add it to en.json then run `npm run build` and all of the langterms are updated?

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
dmackenz: I am not sure about this. I thought to do it in polymer because the parent components are also in polymer.

dmackenz:
> visdiff tests will be added afterward?

Good point @cbusmand2l I forgot that this is an entirely new screen essentially and will need new tests.

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
cbusmand2l: I don't see that we use the polymer logging behaviour in this component and the localization has been reproduced in lit, so we should probably be doing this one in lit. There are multiple lit components being used in polymer parents in QE if you need an example to look at; e.g. `d2l-quick-eval-activity-card-action-button-more` is a lit element that uses localization that is used in a polymer parent (`d2l-quick-eval-activity-card`) that incidentally also has a polymer child (`d2l-quick-eval-activity-card-action-button`).

components/d2l-quick-eval/build/lang/ar.js
AlexBedley: you need to add it to every langterm json file

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -0,0 +1,48 @@
...
+	}
+	static get properties() {
+		return {
+			courseLevel: {
+				type: Boolean,
+				value: false
+			}
+		};
+	}
+	static get observers() {
```
AlexBedley: this should just be deleted since it's empty

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -0,0 +1,48 @@
...
+					@apply --d2l-heading-2;
+					margin: 0;
+				}
+			</style>
+			<h2 class="d2l-quick-eval-no-submissions-heading">[[localize('caughtUp')]]</h2>
+			<template is="dom-if" if="[[courseLevel]]">
+				<p class="d2l-body-standard">[[localize('noSubmissionsCourseLevel')]]</p>
+				<p class="d2l-body-standard">[[localize('checkBackOftenCourseLevel')]]</p>
+			</template>
+			<template is="dom-if" if="[[!courseLevel]]">
```
AlexBedley: instead of using `dom-if` could you just use the langterm id as a property and set that as needed (since the html is otherwise exactly the same)?

dmackenz: Yeah, I guess that you could, I just did it this way because it makes the code that calls the component simpler as you only need the one boolean.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -271,6 +272,9 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			submissionFilters: {
 				type: Array,
 				value: []
+			},
+			multiCourseHref: {
```
dmackenz: We just need to make sure that there isn't a conflict between this and https://github.com/BrightspaceHypermediaComponents/activities/pull/620/files in terms of the properties.

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -0,0 +1,55 @@
+import {html, css, LitElement} from 'lit-element/lit-element.js';
+import {LitQuickEvalLocalize} from './LitQuickEvalLocalize.js';
+import {unsafeHTML} from 'lit-html/directives/unsafe-html.js';
+import 'd2l-link/d2l-link.js';
+
+/**
+ * @customElement
+ * @polymer
```
AlexBedley: not sure if it matters but this comment is incorrect

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -267,6 +268,14 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 			courseLevelName: {
 				type: String,
 				value: ''
+			},
+			submissionFilters: {
```
AlexBedley: bad merge? this is not used

components/d2l-quick-eval/d2l-quick-eval.js
```diff
@@ -81,7 +82,8 @@ class D2LQuickEval extends
 				hidden$="[[!_displayActivities(toggleState, activitiesViewEnabled)]]"
 				dismiss-enabled="[[dismissEnabled]]"
 				course-level="[[courseLevel]]"
-				filter-ids="[[activityFilters]]"></d2l-quick-eval-activities>
+				filter-ids="[[activityFilters]]"
+				multi-course-quick-eval-href="[[multiCourseQuickEvalHref]]"></d2l-quick-eval-activities>
```
AlexBedley: did you test this in unbundled bsi?

dmackenz: Yup, I developed it unbundled.

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
dmackenz: This has been removed.

components/d2l-quick-eval/d2l-quick-eval-submissions.js
dmackenz: This has been removed.

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -0,0 +1,52 @@
+import {html, css, LitElement} from 'lit-element/lit-element.js';
+import {LitQuickEvalLocalize} from './LitQuickEvalLocalize.js';
+import {unsafeHTML} from 'lit-html/directives/unsafe-html.js';
```
AlexBedley: unused import now. how come we abandoned this approach and switched to manual parsing?

cbusmand2l: I suggested the alternative to avoid arbitrarily trusting the text we get back from the localization. I know that in this case we probably won't be injecting anything malicious into our own code using the translation files, but I could easily see someone wanting to solve this same problem in the future and maybe their langterm includes user-provided content.

Also, anything labelled `unsafe` just makes me feel... not safe, you know? :)

AlexBedley: The current approach is still subject to XSS. While `unsafeHTML` also has the same weakness, it does [some checking](https://github.com/Polymer/lit-html/blob/master/src/directives/unsafe-html.ts#L39) to make sure it can only be rendered as text. It's also a huge warning to future developers that this code needs to be handled carefully. I would feel better if we reverted this change.

PR #607, opened 2020-01-10
Us111489 fix positioning

mdulat:
Fixes:
https://trello.com/c/hGI7A0Gu/65-ui-hop-on-focus-state-for-in-grades-button
https://trello.com/c/OjJN6HSd/66-points-text-should-share-baseline-with-numerical-score-entry


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -67,12 +67,15 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 			}
 			#score-info-container {
 				flex-wrap: wrap;
+				-webkit-align-items: center;
```
ryantmer: Huh - which browser is this for? [caniuse](https://caniuse.com/#search=align-items) indicates that everything but IE11 supports it natively 🤔

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -123,11 +126,13 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 			.grade-info:hover,
 			.grade-info[active] {
 				border-color: var(--d2l-color-mica);
+				border-width: 1px;
+				padding: calc(.5rem + 1px) calc(.6rem + 1px);
```
ryantmer: Totally a personal preference thing, but I like to comment these with what the math here is - so that if e.g. whatever the 1px is accounting for changes to 2px, this is somewhat-tied to it.

components/d2l-activity-editor/d2l-activity-score-editor.js
mdulat: Firefox on Windows, and everything (chrome, firefix, safari) on OSX. This fix is not yet tested on OSX, need to check on a quad site, so this may come back to haunt us.

mdulat: Note: it's not that it's not supported, it just behaves differently...

components/d2l-activity-editor/d2l-activity-score-editor.js
mdulat: Sure, will add this in for next PR.

components/d2l-activity-editor/d2l-activity-score-editor.js
ryantmer: Huh, awkward 😞 But as with all things web UI, whatever works! 😄

PR #608, opened 2020-01-10
Polish the Add Activity Dialogue

m6jones:
- Add Activities Dialogue: Already added courses are misaligned from those that haven't already been added to the LP.
- Add Activities Dialogue: The Clear Selection link is not accessible when using keyboard navigation. It also isn't read out by a screen reader.
- Add Activities Dialogue: Clicking on the Add button with nothing actually selected results in a 400 error being sent. This button should probably be disabled until something is selected.
- Add Activities Dialogue > Search for Activities: Selecting activities, search for more, select more activities, click Add > The first set of activities doesn't get added to the LP.
- Add Activities Dialogue > The 'Load More' button highlight (While using keyboard navigation) has the bottom part of the border cut off:
- Loading styles have been added to the Add Activities Dialogue.

### DIFF
--- a/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
+++ b/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
@@ -197,8 +197,11 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		this.clearAllSelected();
 	}
 
-	loadMore() {
-		this.getCandidates(this._actionCollectionEntity.getNextAction());
+	async loadMore() {
+		const lastItem = this.shadowRoot.querySelector('d2l-dialog d2l-list d2l-list-item:last-of-type');
+		await this.getCandidates(this._actionCollectionEntity.getNextAction());
+		await this.updateComplete;
+		lastItem.nextElementSibling.focus();
 	}
 
 	async open() {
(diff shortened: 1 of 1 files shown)


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -77,26 +84,41 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
 			collection.subEntitiesLoaded().then(() => {
 				this._items = items;
 				itemsLoadedOnce = true;
+				this._loadedImages[imageChunk].total = totalInLoadingChunk;
 			});
-			this._addExistingAction = collection._entity.getActionByName('start-add-existing-activity');
 		});
 
 		this._mainPageLoad = usage.subEntitiesLoaded().then(() => {
+			if (!this._loaded) {
```
AllanKerr: I might be missing it but where / when is `_loaded` actually set to `true`? Won't it always be `undefined`?

m6jones: Just below. So do this if it hasn't loaded.

https://github.com/BrightspaceHypermediaComponents/activities/pull/608/files#diff-081877bce5411bb06fa1b3634647445eL104

AllanKerr: I see... It would probably help if I expanded the collapsed section 😢. In that case we probably want to define `_loaded ` so that it starts as `false` rather than `undefined`?

m6jones: Sure.

PR #609, opened 2020-01-10
Move anonymous marking to secondary panel

wongvincent:


jchapmanD2L:
Yeah, sorry this was me.  User story never mentioned to place it in the secondary panel.  The page template came together real quick over the holidays!

PR #613, opened 2020-01-13
Use select styles from core

ryantmer:
Since adding this in, the `select` styles have become available in core, so we can now use those.

This also includes a whitespace change to fix line endings (based on `.editorconfig`).


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -1,225 +1,227 @@
...
-				</d2l-input-checkbox>
-				<d2l-input-checkbox-spacer ?hidden="${!this._anonymousMarkingHelpText}">
-					<span class="d2l-body-small">${this._anonymousMarkingHelpText}</span>
-				</d2l-input-checkbox-spacer>
-			</div>
-		`;
-	}
-}
-customElements.define('d2l-activity-assignment-editor-secondary', AssignmentEditorSecondary);
+import './d2l-assignment-turnitin-editor';
```
ryantmer: FYI @wongvincent looks like your editor may not be respecting the `.editorconfig`, as this file was CRLF rather than just LF

PR #614, opened 2020-01-13
Visdiff for activities table

cbusmand2l:


AlexBedley:
The basic test doesn't seem super useful - it's just an empty box. or is that not intentional?

cbusmand2l:
> The basic test doesn't seem super useful - it's just an empty box. or is that not intentional?

That's intentional - I wanted to mirror the submissions table with a test for the empty state. In this case, there should be nothing displayed if there is no data in the list, so we'll fail the test if we ever accidentally display something when the list should be empty.

AlexBedley:
> > The basic test doesn't seem super useful - it's just an empty box. or is that not intentional?
> 
> That's intentional - I wanted to mirror the submissions table with a test for the empty state. In this case, there should be nothing displayed if there is no data in the list, so we'll fail the test if we ever accidentally display something when the list should be empty.

Fair enough! :shipit:

PR #616, opened 2020-01-13
US112481- Get filter whitelist from the LMS

AlexBedley:
Problem: In course mode it doesn't really make sense to have a course filter (since activities are sourced from 1 course by design)
Solution: Get the valid filters from the LMS and use that as the whitelist

### DIFF
--- a/components/d2l-quick-eval/d2l-quick-eval-activities.js
+++ b/components/d2l-quick-eval/d2l-quick-eval-activities.js
@@ -128,7 +128,7 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 				<d2l-hm-filter
 					href="[[filterHref]]"
 					token="[[token]]"
-					category-whitelist="[[_filterIds]]"
+					category-whitelist="[[filterIds]]"
 					lazy-load-options>
 				</d2l-hm-filter>
 				<d2l-hm-search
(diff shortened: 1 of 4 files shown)


components/d2l-quick-eval/d2l-quick-eval-activities.js
```diff
@@ -255,10 +255,6 @@ class D2LQuickEvalActivities extends mixinBehaviors(
 			courseLevel: {
 				type: Boolean,
 				value: false
-			},
-			activityFilters: {
```
AlexBedley: I added this without using it in the last PR. Upon further thought it is not needed, let's just use `filterIds` instead (ditto for submissions)

components/d2l-quick-eval/d2l-quick-eval-submissions.js
```diff
@@ -481,15 +477,6 @@ class D2LQuickEvalSubmissions extends mixinBehaviors(
 		}
 	}
 
-	_getFilterIds(masterTeacher) {
-		// [ 'activity-name', 'enrollments', 'completion-date' ]
-		let filters = [ 'c806bbc6-cfb3-4b6b-ae74-d5e4e319183d', 'f2b32f03-556a-4368-945a-2614b9f41f76', '05de346e-c94d-4e4b-b887-9c86c9a80351' ];
-		if (masterTeacher) {
```
AlexBedley: the LMS will decide if master teacher is on (or not) for us

PR #612, opened 2020-01-11
Update the gulpfile to export from en.js

maboelsoudd2l:
Since json lang files are being phased out, I updated the `gulpfile` to fill in any missing lang terms in the other js files from `en.js`. It is supposed replace the need for https://github.com/BrightspaceHypermediaComponents/activities/pull/451 but this is meant for js files.

This PR should be merged first though: https://github.com/BrightspaceHypermediaComponents/activities/pull/611

PR #619, opened 2020-01-13
adding new lang term script

dmackenz:
This is a solution around the issues of adding a lang term and `npm run build` not working properly. gulp will not run because of the es6 keywords in the lang term js files. The way to solve this is to delete all of the js files for lang terms temporarily and then restore them after you have completed the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "test:polymer:local": "polymer test --skip-plugin sauce",
     "test:polymer:sauce": "polymer test --skip-plugin local",
     "test": "npm run lint && npm run test:polymer:local",
-    "serve": "polymer serve"
+    "serve": "polymer serve",
+    "clean-lang": "rm -rf components/d2l-activity-collection-editor/lang/* && rm -rf components/d2l-activity-editor/d2l-activity-assignment-editor/lang/* && rm -rf components/d2l-activity-editor/d2l-activity-attachments/lang/* && rm -rf components/d2l-activity-editor/lang/* && rm -rf components/d2l-activity-evaluation-icon/build/lang/* && rm -rf components/d2l-activity-list-item/build/lang/* && rm -rf components/d2l-quick-eval/build/lang/*"
   },
   "author": "D2L Corporation",
   "license": "Apache-2.0",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "test:polymer:sauce": "polymer test --skip-plugin local",
     "test": "npm run lint && npm run test:polymer:local",
     "serve": "polymer serve",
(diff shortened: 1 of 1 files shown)


dmackenz:
The npm script works but it is pretty messy admittedly. This PR is more of a POC.

I was thinking we could implement 3 scripts:

`npm run clean-lang`: deletes all of the lang terms that cause gulp to error
`npm run reset-lang`: calls `git checkout` on all of those same paths in `npm run clean-lang` in order to reset the repo back to its previous state
`npm run build-no-lang`: `npm run clean-lang && npm run build && npm run reset-lang`

package.json
```diff
@@ -16,7 +16,10 @@
     "test:polymer:local": "polymer test --skip-plugin sauce",
     "test:polymer:sauce": "polymer test --skip-plugin local",
     "test": "npm run lint && npm run test:polymer:local",
-    "serve": "polymer serve"
+    "serve": "polymer serve",
+    "clean-lang": "rm -rf components/d2l-activity-collection-editor/lang/* components/d2l-activity-editor/d2l-activity-assignment-editor/lang/* components/d2l-activity-editor/d2l-activity-attachments/lang/* components/d2l-activity-editor/lang/* components/d2l-activity-evaluation-icon/build/lang/* components/d2l-activity-list-item/build/lang/* components/d2l-quick-eval/build/lang/*",
```
AlexBedley: does this work on windows?

package.json
```diff
@@ -16,7 +16,10 @@
...
+    "reset-lang": "git checkout components/d2l-activity-collection-editor/lang/* components/d2l-activity-editor/d2l-activity-assignment-editor/lang/* components/d2l-activity-editor/d2l-activity-attachments/lang/* components/d2l-activity-editor/lang/* components/d2l-activity-evaluation-icon/build/lang/* components/d2l-activity-list-item/build/lang/* components/d2l-quick-eval/build/lang/*",
+    "build-no-lang": "npm run clean-lang && npm run build && npm run reset-lang"
```
AlexBedley: why do we reset after building them? wouldn't that destroy your changes?

maboelsoudd2l:
i have a better idea, i went to the gulpfile to understand why we are having problems with the "es6 keywords in the lang term js files", and realized it has to do with `requireDir = require('require-dir');` and the fact that in nodejs, it doesnt like importing es6 stuff.

So I solved it by activating using `babel/preset-env` to fix it. I did all of this in this PR https://github.com/BrightspaceHypermediaComponents/activities/pull/612

personally I think this PR is a roundabout way to do what is easier done by using babel (or you can just use my pr for this #612)

If you want to learn more about this just look at this tutorial:
https://timonweb.com/posts/how-to-enable-es6-imports-in-nodejs/

package.json
dmackenz: Good point, this is incorrect and unnecessary.

package.json
dmackenz: No it doesn't, I just did it this way to get the point across. If you wanted to make it cross platform you could put the paths in an `Array` in a `.js` then call `fs.rmdir()` or `fs.unlink()` on all of those directories, then just call the node script from `npm run clean-lang`.

PR #620, opened 2020-01-14
Dmackenzie/US112480 multi course qe option in ellipsis

dmackenz:
Adding a link from course-level quick eval back to multi-course quick eval in ellipsis menu on top right.


dmackenz:
Note that the multi-course-quick-eval-href property was added by @jmurray-d2l  in https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/7170/overview

PR #622, opened 2020-01-14
add label to ellipsis button

vix633:
FYI term came from Erin


components/d2l-quick-eval/lang/en.json
```diff
@@ -53,6 +53,7 @@
    "noDismissedActivities" : "You currently have no dismissed activities. Dismissed activities can be viewed and restored here.",
    "noResults" : "No results here.",
    "noSubmissions" : "There are no submissions that need your attention.",
+   "optionsForQuickEval" : "Options for Quick Eval",
```
jeroach: Unless the lang term process has changed, I think we need to add this to all of the different lang files

vix633: My b, didn't commit them

PR #626, opened 2020-01-15
Vy/label ellipsis

vix633:
This is the right way of doing this in lit.


components/d2l-quick-eval/dismiss/d2l-quick-eval-ellipsis-menu.js
```diff
@@ -39,7 +39,7 @@ class D2LQuickEvalEllipsisMenu extends LitQuickEvalLocalize(LitElement) {
 	}
 	render() {
 		return html`
-		<d2l-dropdown-more>
+		<d2l-dropdown-more text="${this.localize('optionsForQuickEval')}">
 			<d2l-dropdown-menu>
 				<d2l-menu>
```
dbatiste: The menu should have a `label` for a11y.  Only nested menus don't need a label (since their label is automatically applied using the menu-item that contains it.

vix633: Would this not just repeat the same thing via the screen reader as the top level label i added?

dbatiste: The `text` on `d2l-dropdown-more` would be announced when the user focuses on the opener.  In contrast, the `label` on `d2l-menu` is (sometimes) announced when the user moves focus into the menu - it may announce the label and the number of menu-items.  The browser/AT doesn't always seem to announce it though.  

The text of the opener may or may not be the same as the label for the menu - i.e. if the opener is "Open...", that wouldn't make sense as the menu label.

PR #628, opened 2020-01-15
Correcting pluralization.

cbusmand2l:


AlexBedley:
Is the assumption here that the already-translated terms have the correct grammar?

cbusmand2l:
> Is the assumption here that the already-translated terms have the correct grammar?

My assumption is that changing the `en` file will flag the term as needing a re-translation. I think that's how the translations work?

PR #630, opened 2020-01-16
Changed rendering logic of course level all caught up message

dmackenz:
This PR is to fix a potential issue when the backend is not able to provide a multi-course quick eval link due to permissions. The component will now no longer render the "Multi-Course Quick Eval" text as a link if the link is not given.


components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -39,8 +39,11 @@ class D2LQuickEvalNoSubmissionsText extends LitQuickEvalLocalize(LitElement) {
 		}
 		return this.localize('noSubmissions');
 	}
-	_computeCheckBackOftenText(courseLevel) {
-		if (courseLevel) {
+	_computeCheckBackOftenText(courseLevel, multiCourseQuickEvalHref) {
+		const isLinkPresent = (multiCourseQuickEvalHref && multiCourseQuickEvalHref.length !== 0) ? true : false;
```
dmackenz: This was coded as a ternary instead of a boolean expression to ensure that `isLinkPresent` is a boolean. If you do not use a ternary it is possible to return an `undefined`.

cbusmand2l: Could also do `!!(multiCourseQuickEvalHref && multiCourseQuickEvalHref.length !== 0)` or `Boolean(multiCourseQuickEvalHref && multiCourseQuickEvalHref.length !== 0)` I believe.

cbusmand2l: Hmmm, just realized how this was being used. Why do we not just check the href directly below? `if (courseLevel && !multiCourseQuickEvalHref)` and `} else if (courseLevel && multiCourseQuickEvalHref) {`?

dmackenz: I guess you could. I just assigned it to a variable to make it easier to read and if you wanted to add more conditions in the future you just need to change the one variable.

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
```diff
@@ -39,8 +39,11 @@ class D2LQuickEvalNoSubmissionsText extends LitQuickEvalLocalize(LitElement) {
 		}
 		return this.localize('noSubmissions');
 	}
-	_computeCheckBackOftenText(courseLevel) {
-		if (courseLevel) {
+	_computeCheckBackOftenText(courseLevel, multiCourseQuickEvalHref) {
+		const isLinkPresent = (multiCourseQuickEvalHref && multiCourseQuickEvalHref.length !== 0) ? true : false;
+		if (courseLevel && !isLinkPresent) {
+			return this.localize('checkBackOftenCourseLevel', {startTag: '', endTag: ''});
```
AlexBedley: This seems a bit weird.. wouldn't this prompt users to look for multi-course quick eval? Shouldn't we just be displaying the normal `there are no submissions` message?

Or maybe a third message that just says to check your other courses?

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
AlexBedley: If `multiCourseQuickEvalHref` is a string, isn't `&& multiCourseQuickEvalHref.length !== 0` redundant?

dmackenz: Will make this change

components/d2l-quick-eval/d2l-quick-eval-no-submissions-text.js
dmackenz: Dan suggested that we just go with the original message. Will make this change.

PR #632, opened 2020-01-16
Us111000 grade associations

mdulat:
Adds ability to associate existing grade item. Initially I had an individual grade-candidate entity in a separate component, but that didn't work because the `<select>` requires `<option>` to be directly under it, so I couldn't have `<select><d2l-grade-candidate><option>...`. 
Note the term for the menu item and dialog title are not localized yet as the term is not yet finalized.


components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-candidates.js
```diff
@@ -0,0 +1,63 @@
...
+			>
+				${repeat(this._gradeCandidateEntities, entity => entity, (entity, index) => html`
+					<option value="${index}">${entity.name()}</option>
+				`)}
+			</select>
+		`;
+	}
+}
+
+customElements.define('d2l-activity-grade-candidates', ActivityGradeCandidates);
```
ryantmer: nit/discussion point more than anything, but would `d2l-activity-grade-candidate-select` (or `-selector`) make sense? Just since this isn't a static list. Not too bothered either way, though.

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-candidates.js
```diff
@@ -0,0 +1,63 @@
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { GradeCandidateCollectionEntity } from 'siren-sdk/src/activities/GradeCandidateCollectionEntity';
+import { GradeCandidateEntity } from 'siren-sdk/src/activities/GradeCandidateEntity';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { repeat } from 'lit-html/directives/repeat';
+import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
```
ryantmer: mega-nit: this is the only import using `.js` (I've been steadily removing file extensions in imports since I think they're kinda ugly, but I could also be convinced very easily that we should always include them)

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-candidates.js
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		if (entity) {
+			this._gradeCandidateEntities = entity.getGradeCandidateEntities();
+		}
+
+		super._entity = entity;
+	}
+
+	_setSelected(e) {
```
ryantmer: `eslint` will complain about this I think, but `e` is unused here.

That said, won't `e` have the `value` in it here, which would remove the need for the `querySelector` below? I could be totally wrong though, I always forget what events contain for input elements...

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -267,6 +272,16 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 		this.wrapSaveAction(super._entity.setUngraded());
 	}
 
+	_setGradeItem() {
+		this.shadowRoot.querySelector('d2l-dialog').open()
+			.then((action) => {
+				if (action === "done") {
+					const selectedGradeCandidate = this.shadowRoot.querySelector('d2l-activity-grade-candidates').selected;
```
ryantmer: I'm not familiar with how `d2l-dialog` works, so I could be totally wrong here, but this seems like something that would normally be done with events, i.e. the selected value would be emitted by `d2l-activity-grade-candidates` when it's selected. Is that not a possibility due to how `d2l-dialog` is intended to work?

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -324,16 +339,28 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
...
+								></d2l-menu-item>
 							</d2l-menu>
 						</d2l-dropdown-menu>
 					</d2l-dropdown>
+					<d2l-dialog title-text="Edit Grade Item Link">
+						<d2l-activity-grade-candidates
+							href="${this._gradeCandidatesHref}"
+							.token="${this.token}">
+						</d2l-activity-grade-candidates>
+						<d2l-button slot="footer" primary dialog-action="done">Ok</d2l-button>
```
ryantmer: This and the Cancel text should be localized 🗺 (unless they're not final wording, in which case nvm!)

components/d2l-activity-editor/d2l-activity-score-editor.js
dbatiste: Technically I think that would be possible, however I suspect we would only want this code to run when the user clicks the primary dialog button.  If the user selects something and then aborts by clicking the Cancel or [x] (top right corner) then it should be a no-op from the user's perspective.

Aside, the dialog can be opened/closed in a couple of different ways.  In this case, the `open` method opens the dialog and returns a promise that will be resolved with the dialog action.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -267,6 +272,16 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 		this.wrapSaveAction(super._entity.setUngraded());
 	}
 
+	_setGradeItem() {
```
dbatiste: This is sort of a style thing, but the nesting could be avoided here by...
```
async _setGradeItem() {
    const dialog = this.shadowRoot.querySelector('d2l-dialog');
    const action = await dialog.open();
    if (action !== 'done') return;
    const selectedGradeCandidate = dialog.querySelector('d2l-activity-grade-candidates').selected;
    this.wrapSaveAction(selectedGradeCandidate.associateGrade());
}
```

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -324,16 +339,28 @@ class ActivityScoreEditor extends ErrorHandlingMixin(SaveStatusMixin(EntityMixin
 								></d2l-menu-item>
 								<d2l-menu-item
 									text="${this.localize('removeFromGrades')}"
-									?hidden="${!this._inGrades || !this._canEditGrades}"
+									?hidden="${!this._inGrades}"
```
dbatiste: Tip: Lit is great for conditional rendering, which reduces the amount of DOM processing the browser has to do. i.e. it is possible to factor this out and only render if necessary.
```
const someMenuItem = (this._inGrades ? html`<d2l-menu-item ...></d2l-menu-item>` : null);
...
return html`
    ...
    ${someMenuItem}
    ...
`;
```
It looks like there's lots of use of `?hidden` here that could be optimized. Maybe not something you want to change here/now, but it is definitely preferred to not render nodes unless they are needed in the DOM. The constructor and many of the life cycle methods still get run, and associated styles still get processed even though the element is hidden.  Dropdowns, menus, and dialogs attempt to defer as much as possible, but they are still kinda heavy.

components/d2l-activity-editor/d2l-activity-score-editor.js
ryantmer: Right! I misunderstood the flow here - it's `d2l-menu-item-select` -> open dialog, not vice versa.

PR #635, opened 2020-01-20
Fixing language for empty activities search

zinabat:
**[DE37459](https://rally1.rallydev.com/#/detail/defect/362134185372?fdp=true)**

Fix for language issue when activity search results are empty.


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -624,7 +624,7 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 
 	_renderCandidateItems() {
 		if (this._candidateItems.length <= 0) {
-			return html`<div class="d2l-activity-collection-no-activity d2l-body-standard">${this.localize('noActivitiesInLearningPath')}</div>`;
```
iFun: Do we know if the activities are empty after searching or before?

zinabat: `this._candidateItems` is initialized to an empty array, but `_renderCandidateItems` is not called until the promise from the activity fetch is resolved in `_handleFirstLoad` after the search. At least..I think that's what's happening.

PR #641, opened 2020-01-20
add labels to attachment buttons, and disable focus lock on tooltips

mdulat:
Fixes https://trello.com/c/K6rYX68v/71-a11y-attachments. Note that in chrome/nvda this issue was not noticeable because the tooltips are read out. However in Firefox/nvda the tooltips are not read out so the buttons need to be labeled for the screenreader. 
This PR also disables focus-lock on the tooltips so that if you hover over one tooltip while another has focus you don't see both tooltips overlapping.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -193,11 +193,13 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 				<d2l-button-icon
 					id="add-file-button"
 					icon="d2l-tier1:upload"
+					text="${this.localize('addFile')}"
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -193,59 +193,69 @@ class ActivityAttachmentsPicker extends SaveStatusMixin(EntityMixinLit(LocalizeM
 				<d2l-button-icon
 					id="add-file-button"
 					icon="d2l-tier1:upload"
+					text="${this.localize('addFile')}"
 					?hidden="${!this._canAddFile}"
 					@click="${this._launchAddFileDialog}">
 				</d2l-button-icon>
 				<d2l-tooltip
 					for="add-file-button"
+					disable-focus-lock
```
ryantmer: aha! Good catch, this always bugged me but I admittedly never took the time to figure it out...

PR #646, opened 2020-01-22
simplify initial activity render and move PendingContainerMixin to ac…

mdulat:
…tivity editor.
In addition to simplifying rendering of "Loading..." versus editor page after fixing tabbie position issue in LMS, I moved the PendingContainerMixin to the activity editor, as that is what receives the events anyway, seemed cleaner to have those events handled directly in there. This fixes same issue as https://github.com/BrightspaceHypermediaComponents/activities/pull/594 (I didn't realize that PR was there when I did this fix).


dbatiste:
If we need to move this again or in any other component, I think we should switch it to use `AsyncContainerMixin` and remove the `PendingContainerMixin`. They effectively duplicate, but with slightly different API.

PR #647, opened 2020-01-22
show editor nav once activity page is ready

mdulat:
The problem is that when we load the activity page we see the editor nav, then we see "Loading...", then we see the full page. We don't want to see the editor nav initially. 
This fix goes along with LMS change that initially hides the editor nav. This way we show it only when the activity page is ready.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -70,6 +70,8 @@ class AssignmentEditor extends EntityMixinLit(LitElement) {
 		this.addEventListener('d2l-siren-entity-save-error', () => {
 			this.shadowRoot.querySelector('#save-status').error();
 		});
+		const nav = this.shadowRoot.querySelector('#nav').assignedNodes()[0];
+		nav.removeAttribute('hidden');
 	}
 
 	_onAssignmentActivityUsageChange(assignmentActivityUsage) {
@@ -138,7 +140,7 @@ class AssignmentEditor extends EntityMixinLit(LitElement) {
 				@d2l-request-provider="${this._onRequestProvider}">
 
 				<d2l-template-primary-secondary slot="editor">
-					<slot name="editor-nav" slot="header"></slot>
+					<slot id="nav" name="editor-nav" slot="header"></slot>
 					<d2l-activity-assignment-editor-detail
 						href="${this._assignmentHref}"
 						.token="${this.token}"


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -70,6 +70,8 @@ class AssignmentEditor extends EntityMixinLit(LitElement) {
 		this.addEventListener('d2l-siren-entity-save-error', () => {
 			this.shadowRoot.querySelector('#save-status').error();
 		});
+		const nav = this.shadowRoot.querySelector('#nav').assignedNodes()[0];
+		nav.removeAttribute('hidden');
```
mdulat: I realize it is bad form to update the DOM in this way (outside of render function), however there is no way to initially hide the nav from within here, because it exists (created in LMS) before activity is loaded. I am open to other suggestions on how this can be done.

ryantmer: One thing that I thought about - do we want to hide the nav as things are loading? The existing pattern of replacing the entire page with "Loading" seems a little clunky to me, as it hides the "Back to X" button, so a user would have to wait until the whole page loads to be able to go back (or, I mean, use the browser's back button I guess)

mdulat: yeah, just chatted with Joseph about this, he's ok with "Loading..." replacing the entire page, with no back button. Rather he's ok with either option, but prefers the "no changes" option.

mdgbayly: Yeh this works but it is kind of ugly, introduces dependencies between stuff that's in the LMS and the web components, and would need to be applied to future editors too like a discussions editor. It's only a small thing but dunno ... i would think there should be some other way :-(. Is it because there is a delay before our top level editor shows the loading state related to how the pending stuff works. Could we instead look at initializing it in a loading state.

mdulat: I tried initializing it in a loading state, it didn't make a difference because that slotted nav tag is created in the LMS, and exists before it is actually slotted into the activity. Having said/done all this, I noticed that the problem only shows up locally for me, I don't see this happening on quad, so maybe it's ok to not get this fix in at all (since it's an ugly fix, and probably isn't needed in production)?

mdgbayly: ok - yeh - it's a weird one - possibly it is happening on quad too but maybe it is timing related, or maybe the nature of the build order on quad is affecting it.

I think the reason why we are seeing it locally is due to the order in which the web components are being upgraded by the browser. It seems like the nav is getting upgraded first so between the nav being upgraded and the editor being upgraded i think we effectively have a dom structure that looks something like this:

```
<d2l-activity-assignment-editor><!-- browser doesn't know what this is yet -->
    <div slot="editor-nav">
        <!-- maybe this is fully rendered before it get's slotted -->
        <d2l-immersive-nav>full nav content</d2l-immersive-nav>
    </div> 
</d2l-activity-assignment-editor>
```

So until the assignment editor web component gets upgraded, anything we do in there such as trying to hide either the component itself or anything that it contains is basically ignored...

I've seen references to issues with FOUC with web components before but not really in this context. There is a `:defined` pseudo-class which can be used to give an element styles before it is upgraded:

https://developers.google.com/web/fundamentals/web-components/customelements#prestyle

So possibly we do something like this somewhere to prevent the flash:

d2l-activity-assignment-editor:not(:defined) {
  display: none;
}

But that's also pretty ugly and means we'd have to replicate for each tool top level component...also no edge support.

mdulat: Cool, I didn't know about `:defined`. I updated the LMS solution (https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/7597/diff) to add that css instead of using the hidden attribute, and it works on my local. A much nicer solution, and doesn't need any changes to this repo. Still has the dependency as you said, but putting that nav into the slot itself is a dependency already!

mdulat:
Replaced with LMS-only solution: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/7597/overview

PR #648, opened 2020-01-22
Activity Editor mixins

mdgbayly:
These two mixins are intended to allow us to start co-ordinating the save behavior in FACE.

The `ActivityEditorContainerMixin` will be applied to the top level editor component (e.g. `d2l-activity-assignment-editor`) for Assignments. It handles listening on the `d2l-activity-editor-save` event which will be dispatched from the Save button where ever that happens to be rendered. It also includes an implementation to allow child editors to be registered so that when the Save button is clicked, child editors can be notified.

The `ActivityEditorMixin` will be applied to each component that supports editing capabilities.
It includes a `save()` method that will be called by the `ActivityEditorContainerMixin` when a Save is initiated. It's unlikely that every editor component will need to implement this `save()`. I anticipate that we will only implement the `save()` in the top level assignment and activity components and they will co-ordinate with the relevant MobX state to actually execute the save of the two main slices of state. However, by having every component use this mixin it gives us some flexibility in building components that could continue to operate in an autosave pattern if they detect they are not embedded by an editor container.

### DIFF
--- a/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
+++ b/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
@@ -1,0 +1,56 @@
+export const ActivityEditorMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			/**
+			 * Href for the component
+			 */
+			href: {
+				type: String,
+				reflect: true
+			},
+			/**
+			 * Token JWT Token for brightspace | a function that returns a JWT token for brightspace | null (defaults to cookie authentication in a browser)
+			 */
+			token: { type: String },
+		};
+	}
+
+	constructor() {
+		super();
+		this._container = null;
+	}
+
+	async save() {}
+
+	_dispatchActivityEditorEvent() {
+		const event = new CustomEvent('d2l-activity-editor-connected', {
+			detail: { editor: this },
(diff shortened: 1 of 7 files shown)


test/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -0,0 +1,44 @@
...
+const editor = defineCE(
+	class extends HTMLElement {
+		save() {
+			this.saveCalled = true;
+		}
+	}
+);
+
+function connectedEvent(editor) {
+	return  new CustomEvent('d2l-activity-editor-connected', {
```
wongvincent: nit. space

components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
```diff
@@ -0,0 +1,56 @@
...
+
+	_dispatchActivityEditorEvent() {
+		const event = new CustomEvent('d2l-activity-editor-connected', {
+			detail: { editor: this },
+			bubbles: true,
+			composed: true,
+			cancelable: true
+		});
+		this.dispatchEvent(event);
+		if (event.detail.container) {
```
wongvincent: cool~!

PR #650, opened 2020-01-23
hide tooltip from screenreader since it already has button descriptions

mdulat:
Fixes https://trello.com/c/OH0fyVUE/82-a11y-previous-tooltip-is-announced-when-tabbing-through-attachments


mdulat:
> Huh - the screenreader still gets confused with the `aria-hidden`? I figured screenreaders would just ignore any elements (and their content) with that

That would be a logical thing to assume, but alas...

PR #651, opened 2020-01-23
DE37554 - use default name for untitled learning path

jlai-d2l:


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -741,7 +741,8 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 	}
 
 	_titleChanged(e) {
-		this._specialization.setName && this._specialization.setName(e.target.value);
+		const newName = e.target.value !== '' ? e.target.value : this.localize('untitledLearningPath');
+		this._specialization.setName && this._specialization.setName(newName);
```
AllanKerr: This feels like a change that should be made in the edit-in-place component. If the text is empty then it shouldn't allow saving.

jlai-d2l: But when you first create the LP, it already gets saved with "Untitled Learning Path". And in other situations (with Description), you may want to allow for empty fields.

AllanKerr: I still think it should be added to the edit-in-place component along with an `allow-empty` property. Changing it back to "Untitled Learning Path" feels intuitive if we could just prevent the user from saving when its empty.

PR #652, opened 2020-01-23
Added spacing and flex alignment to selection count

zinabat:
[DE37544](https://rally1.rallydev.com/#/detail/defect/363034010104?fdp=true)

Spacing and flex alignment has been added to the selection count in add activity dialogue to fix ugliness in mobile.

Screen Shot 2020-01-23 at 4 06 08 PM


dbatiste:
I think it would be good if we had a list-summary component that provided this layout with an optional slot, possibly named search.

PR #655, opened 2020-01-24
Fixed activity loading spinner being cut off

zinabat:
[DE37560](https://rally1.rallydev.com/#/detail/defect/363226148060?fdp=true)

Loading spinner was partially cut off when loading more activities for editing a learning path. Added a scroll that prevents this.

### DIFF
--- a/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
+++ b/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
@@ -144,6 +144,12 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		this._candidateItemsLoading = true;
 		if (clearList) {
 			this._candidateFirstLoad = false;
+		} else {
+			// wait for the spinner, then scroll to the bottom
+			await this.updateComplete;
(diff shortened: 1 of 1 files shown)


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -144,6 +144,12 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		this._candidateItemsLoading = true;
 		if (clearList) {
 			this._candidateFirstLoad = false;
+		} else {
+			// wait for the spinner, then scroll to the bottom
+			await this.updateComplete;
+			const dialogScroll = this.shadowRoot.querySelector('#dialog').shadowRoot.querySelector('.d2l-dialog-content');
```
zinabat: I was unsure if there was a cleaner way to traverse the DOM here - suggestions welcome.

dbatiste: Yeah we should never be querying DOM scopes out of context.  I may be able to help here but I'm not 100% sure the use-case.  We can possibly add something to the dialog API if needed.

zinabat:
It was decided that this issue should wait for a scroll method from the dialog component instead of using deep querying. This issue is being added to the backlog until a better solution is available.

PR #661, opened 2020-01-29
Enable Save/Cancel for assignments name and instructions

mdgbayly:
Sorry, had hoped to break this up into multiple PRs for easier review but ended up kept deferring whilst i was making sure it was all working together.

There are two main aspects to this PR:

1. Introducing MobX state for assignments
2. Using that MobX state in the assignment activity editor to manage the name and instructions properties.

All the other assignment properties and activity properties are still using autosave.

**MobX state**

I've introduced Jest to unit test the state management code outside the browser as I was finding it too hard to mock things in the browser, as you cannot mock out ES6 modules.

The MobX state management includes 3 separate classes:

`assignment-store.js`:
This acts as a cache for assignment and assignment activity usage entities so that the same MobX state instance can be referenced from multiple components. Currently it is just caching based on URL. At some point we may need to enhance that to cache on auth token too similar to what the `EntityStore` does, but this seemed like an unnecessary complexity at this time and is not something we depend upon for FACE.

`assignment-activity-usage.js`:
This loads the `AssignmentActivityUsageEntity` from the `EntityStore` and is mostly used to get access to the assignment folder url, so we can get the non-activity assignment properties. It is used by the top level assignment editor to populate the assignment URL on the sub components.

`assignment.js`:
This loads the `AssignmentEntity` from the `EntityStore` and is used to manage the assignment specific properties. So this is the class that contains the relevant MobX observable properties and actions that allow us to support Save/Cancel of assignment properties. It also implements the `save` method that will ultimately be called to save the assignment when Save is pressed. It delegates the save to the `AssignmentEntity` in the `siren-sdk`.

**MobX decorators**

Given the state of the ES decorator specification I decided to avoid using ES decorators in the MobX state code and instead just use the MobX decorate utility. You'll see calls to `decorate` at the end of each state file. The `decorate` utility allows you to specify which properties and methods in the MobX based class are observable/actions. Not quite as pretty as using `@observable` but avoids all the messy transpilation requirements of the decorator spec which I discovered is still undergoing major rework.

**Component changes**

`d2l-activity-assignment-editor` controls things and uses the `AssignmentStore` to fetch the `AssignmentActivityUsage` MobX class from which it gets the assignment URL to initialize the assignment detail and secondary detail components.

It also implements the save event handler by fetching the `Assignment` MobX entity for the current activity so that it can actually perform the save.

`d2l-activity-assignment-detail` uses the `Assignment` MobX class to load the name and instructions related fields and to push those values back to the MobX state whenever the values change.


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-activity-usage.js
```diff
@@ -0,0 +1,51 @@
...
+		this.assignmentHref = '';
+		this.loading = Promise.resolve();
+	}
+
+	fetch() {
+		dispose(this._entity);
+		this._entity = null;
+
+		let pendingResolve;
+		this.loading = new Promise((resolve) => {
```
mdgbayly: Added this `loading` property in case we have use cases where we want to wait on the object to be loaded from the entity store. Currently everything just happens dynamically by causing a re-render. But it's possible we might want to use this when clicking Save for example.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -0,0 +1,77 @@
...
+		this.autoSave = autoSave;
+		this.loading = Promise.resolve();
+	}
+
+	async fetch() {
+		dispose(this._entity);
+		this._entity = null;
+
+		let pendingResolve;
+		this.loading = new Promise(resolve => {
```
mdgbayly: Added this `loading` property in case we have use cases where we want to wait on the object to be loaded from the entity store. Currently everything just happens dynamically by causing a re-render. But it's possible we might want to use this when clicking Save for example.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -0,0 +1,77 @@
...
+	async save() {
+		await this._entity.save({
+			name: this.name,
+			instructions: this.instructions
+		});
+		this.fetch();
+	}
+}
+
+decorate(Assignment, {
```
mdgbayly: This is the MobX `decorate` utility that avoids having to get into `@decorator` type transpilation issues...

test/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.spec.js
```diff
@@ -0,0 +1,40 @@
+import { Assignment } from '../../../../components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js';
+import { AssignmentActivityUsage } from '../../../../components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-activity-usage.js';
+import { AssignmentStore} from '../../../../components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js';
+import { expect } from 'chai';
+import { expect as jexpect } from '../../../custom-jest.js';
+
+jest.mock('../../../../components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js');
```
mdgbayly: I was experimenting here with using Jest for mocking but I may try and switch this back to using just Sinon.
I found a way to mock things uses Sinon in the other two MobX classes.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -150,16 +144,24 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 	}
 
 	render() {
+		const {
+			name,
+			canEditName,
+			instructions,
+			canEditInstructions,
+			instructionsRichTextEditorConfig
+		} = this._assignment || {};
```
ppaskaris-d2l: I don't like `somethingAsync || {}`.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
ppaskaris-d2l: So they would call `fetch` and then wait on `this.loading`? Why not just wait on `fetch` and make it return a promise..?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -175,8 +163,18 @@ class AssignmentEditor extends PendingContainerMixin(ActivityEditorContainerMixi
 		`;
 	}
 
-	save() {
-		alert('Save coming soon. We are still autosaving!');
+	async save() {
+		const assignment = store.fetchAssignment(this._activity.assignmentHref, this.token);
```
ppaskaris-d2l: I don't understand this control flow:

`store.fetchAssignment` → returns a stub and calls the `assignment.fetch` async method
`assignment.save` → calls `this._entity.save({...})` but `this._entity` is null until it loads..?

mdgbayly: yeh - i wasn't too happy about that either. When I got to the point of adding in the save call I realized that there was a bit of an impedance mismatch, which was why i added the loading promise to the MobX entities. So one thought I had was to add in something like

```
await assignment.loading;
assignment.save()
```

But I wasn't sure I liked that either. So was still ruminating on it. Possibly just making the store.fetch* method return a promise even though other usages of it don't require it.

mdgbayly: Oh - that was what you were suggesting in your other comment  - so yeh - i think i'll just switch it to return a promise from the fetch calls.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
mdgbayly: Yeh - i'm not particularly happy with the _entity stuff either. This was mostly an attempt to hook into the existing siren-sdk `entityFactory` stuff. And that by itself isn't necessarily required. In some other spikes I did, I just used the EntityStore promise based methods directly and didn't bother trying to register any callbacks/listeners.

The reason for using the entityFactory here was experimenting with trying to set up components to support autosave workflows too. The idea being that if a component is configured for autosave, then it might want to register a listener and be notified of changes. e.g. if it is being used outside the context of an editor.

However, if a component isn't using autosave, then it should ignore any notifications after the first one.

But I agree it is not clear, and half baked, so maybe I'll just rip that out and use the EntityStore promise API instead and not bother registering any listeners.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: Can you clarify what you mean by `somethingAsync`. `this._assignment` is just an object reference that can be undefined. So this is an idiom I've seen before for destructuring something that might be undefined. It does of course mean that all the destructured properties will be undefined.

We don't want to skip rendering completely here even if `this._assignment` has not been initialized yet because the `d2l-activity-editor` component handles rendering the loading state.

Did you have an alternative suggestion for how to keep the destructuring approach, or were you suggesting using a different approach that destructuring?

mdgbayly:
I updated the PR to try and simplify the async nature and removed the experimental support for retaining autosave behaviour. 

So now the state classes just fetch entities from the EntityStore but do not register any listeners.
The `assignment-store.js` caches promises for each fetched entity so that it can return the same entity for subsequent requests for the same url. The stored promise resolves to the fetched entity.

The editor web components therefore now use `await` to fetch the state objects and also when loading the assignment on save.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
ppaskaris-d2l: :+1: to ripping out half-baked things until we're motivated by an explicit need

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ppaskaris-d2l: Can you eliminate the time that it can be undefined? The store is synchronous, so why can't you get a defined reference from it before render is called?

The approach is bad because it is a band-aid over the problem of late initialization, which I don't like.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
ppaskaris-d2l: Saving before something is loaded doesn't really make sense to me - what are you saving? Loading explicitly changes the state of the assignment object (in `assignment.load(entity)`).

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
ppaskaris-d2l: You've made the formerly synchronous part asynchronous now, so my previous comment is no longer relevant.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
```diff
@@ -3,30 +3,36 @@ import { AssignmentActivityUsage } from './assignment-activity-usage.js';
...
 export class AssignmentStore {
 	constructor() {
-		this.assignments = new Map();
-		this.activities = new Map();
+		this._assignments = new Map();
+		this._activities = new Map();
 	}
 
-	fetchAssignment(href, token, autoSave = false) {
+	async fetchAssignment(href, token) {
```
ppaskaris-d2l: Why is this async? It does not us await.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
```diff
@@ -3,30 +3,36 @@ import { AssignmentActivityUsage } from './assignment-activity-usage.js';
 
 export class AssignmentStore {
```
ppaskaris-d2l: At this point, this is less of a store and more of a cache. If each entity has to retain a reference to anything form this store, how does any of the observable MobX stuff really help us? Are promises observable?

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -1,42 +1,25 @@
 import { action, configure as configureMobx, decorate, observable } from 'mobx';
-import { dispose, entityFactory } from 'siren-sdk/src/es6/EntityFactory.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
+import { fetchEntity } from '../../state/fetch-entity.js';
 
 configureMobx({ enforceActions: 'observed' });
 
 export class Assignment {
```
ppaskaris-d2l: Oh.. ok.. so each entity is the MobX store and the Store _really is just a cache_. Sorry, this is not how I've seen MobX used in the past, so it is difficult to accept initially.

components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
```diff
@@ -36,6 +36,21 @@ export const ActivityEditorMixin = superclass => class extends superclass {
...
 	}
 
+	async _fetch(fetcher) {
+		const promise = fetcher();
+		this._sendPendingEvent(promise);
+		return await promise;
+	}
+
+	_sendPendingEvent(promise) {
+		const pendingEvent = new CustomEvent('d2l-pending-state', {
```
ppaskaris-d2l: Where do we even use this? What is it for?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mdgbayly: OK - but yeh agree, i"m not keen on how it can render with undefined things, and that can still happen. And my previous comment about wanting to still render i think is mistaken, so I think I'll probably just change it to return immediately from the render if the assignment/activity is undefined.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
mdgbayly: The `assignment-store` class is the cache and is not using MobX. It just ensures that different components loading the same href (e.g. showing different renderings of the assignment or activity usage), are given the same MobX enabled object. The `assignment` and `assignment-activity-usage` are not really caches. They are the classes with observable MobX properties. And the intention is that those classes can be used by components which are potentially rendering the same state. e.g. an editor component and a summarizer component, where the summarizer component is likely just rendering the updated view of some state that might be changing somewhere else.

components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
mdgbayly: We use it in the top level container component to (`PendingContainerMixin`) to control whether to show a loading page/image. It tracks whether it has any loading children and only renders the loaded view once it detects it has no more children loading. It's so we can avoid showing partially rendered children as a result of all the different API calls being made concurrently.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
mdgbayly: Yeh, just left over - will remove.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
mdgbayly: Put another way, the `assignment-store` class is just a way to give each component access to a MobX enabled object in such a way that components are still somewhat independent and are making as few assumptions as possible about how they are being composed. To the extent that MobX even has the concept of a store, it is the `assignment` and `assignment-activity-usage` instances that are the more like what would be considered a MobX store.

cpacey:
@mdgbayly , this is the first place we're using MobX "for real" and that's likely to set some precedents (intentionally or otherwise).  This leads me to have a desire for slightly-more-than-usual review of this PR.  I would like to have @dbatiste as a reviewer because he brings like a century of front-end expertise and also brings some perspective from outside FACE.

mdgbayly:
@cpacey Yeh, that sounds reasonable to me. I was planning on doing an adhoc code walkthrough with the some devs on our team later this morning to help them get a better sense of what it is doing an ask more questions, so I'll see if @dbatiste would like to attend that too. I'll invite @ppaskaris-d2l too.

PR #662, opened 2020-01-30
null checks for empty entities

vix633:


vix633:
> LGTM
> 
> any tests need updating/creating?

could be good to add tests for the false path, but not entirely necessary. Let me know if you want them

vix633:
> We are changing the experience now from hiding bad rows to displaying incomplete rows. I think you mentioned you didn't think this was occurring impactfully in prod because nobody complained about it. Well they couldn't have complained because we would hide the row from them (I don't think this was an intentional choice, probably a side-effect of making a single row failure not bork the entire app). Now they will see rows with incomplete information, I expect we will see customer complaints about it (for better or worse, it really depends on why the activity name can't be fetched).
> 
> Interestingly we did not add the failure isolation code to `dismissed-activities` so any failure there breaks the whole dialog. But nobody would have gotten that far, since the only way to dismiss is through `activities` and the row would be hidden from them due to the fail-loud code.
> 
> My thoughts are if the activity-name thing is failing because the activity has been deleted (or some weird thing like that) then this was actually a "feature" and users will start seeing lots of garbage. OR the activity-name thing is unrelated and we've been hiding data from users for awhile now. Neither result seems good :( Without knowing why the bug is occurring it'll be hard to get out messaging ahead of this change, but we could still alert users of what they might expect to see.

Ah yes, I had a feeling of Deja Vu when I pushed this code. We had this exact discussion many months ago. Unfortunately in this case the buck needs to keep going, rather than stop here. Meaning we need to know why the entity store is not giving us what we want.

AlexBedley:
> Ah yes, I had a feeling of Deja Vu when I pushed this code. We had this exact discussion many months ago. Unfortunately in this case the buck needs to keep going, rather than stop here. Meaning we need to know why the entity store is not giving us what we want.

Sure but it's not great if we rely on the users telling us why the bug is occurring. But maybe we can save some face by coming out ahead of this change and telling users they should expect to see changes

vix633:
> > Ah yes, I had a feeling of Deja Vu when I pushed this code. We had this exact discussion many months ago. Unfortunately in this case the buck needs to keep going, rather than stop here. Meaning we need to know why the entity store is not giving us what we want.
> 
> Sure but it's not great if we rely on the users telling us why the bug is occurring. But maybe we can save some face by coming out ahead of this change and telling users they should expect to see changes

I think this is a thing we won't ship in the current state. I need to get at the permissions behind activity name and start there.

PR #677, opened 2020-02-03
Integrates the activity due date editor with a MobX store to enable Save/Cancel

mdgbayly:
For comparison, this PR uses a slightly different approach to the MobX integration than in the previous PR where I added support for name/instructions to the editors.

(previous PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/661) 

The core of this PR are the changes to the `d2l-activity-editor` and the `d2l-activity-due-date-editor`.

`d2l-activity-due-date-editor` now integrates with a MobX enabled `ActivityUsage` object to load and set the due date, and no longer uses autosave.

`d2l-activity-editor` hooks into the `d2l-activity-save` event handling, and calls `ActivityUsage.save` to save the changes using the `ActivityUsageEntity` in the `siren-sdk`. Currently it just calls `setDueDate` but as we refactor our date handling to enable better validation, this will call an action that saves all dates together.

In this PR, I've also made the class that I had referred to as the `store` use MobX too. This class is used to ensure that each web component that needed to share state got access to the same MobX enabled object. In this PR we're dealing with `due-date` which is on the ActivityUsage, so there is an `activity-store.js` and an `activity-usage.js`. `activity-usage.js ActivityUsage` is the class that will encapsulate all the activity usage properties for an instance of an activity.

In the previous PR the store class just acted as a cache of MobX enabled objects and each web component would `fetch` from the store and set a local standard Lit property to the value of that object.

```
this._assignment = await store.fetchAssignment(this.href, this.token);
```

In this PR the cache of MobX enabled objects in the store is instead represented as an observable map. So now in order to render a specific activity instance, the web component makes a call to get the object from the store in it's render method. This sets up the reaction with MobX, so if the object has not been fetched yet it will skip rendering. When the object is fetched (by something else or this component outside the render), and set into the store, it will cause the component to re-render via a MobX reaction.

```
render() {
    const activity = store.get(this.href);
    if (!activity) {
        return html``;
    }

    const {
       dueDate,
       canEditDueDate
    } = activity;

    return html`
        // use properties to render DOM
    `
}
```

As we still want each component to be able to be used independently and not make any assumptions about something else actually fetching the activity, each component will still make a call to fetch the activity when it's href is updated, but this is distinct and independent from how it is rendered.

```
updated(changedProperites) {
   super._fetch(() => store.fetch(this.href, this.token));
}
```
 
I'm not sure which of these approaches is better. This latter approach makes more use of MobX and has the advantage that it decouples fetching and rendering.

It also means that the components that just handle the save event no longer need to also await the fetch of the object in order to try and save it. Instead they just get the object from the store and if it is not defined yet, then they skip the save.

```
d2l-activity-editor.js

aync save() {
    const activity = store.get(this.href);
    if (activity) {
        await activity.save();
    }
}   
```

I also took the opportunity to clean up some of the additional promise handling that was in the `assignment-store.js`. That was implemented so that if multiple components were concurrently trying to fetch the same activity, we'd only try to fetch it once, but I still wanted to return the actual activity object from the call to fetch. I realized that by just returning `this` from the `fetch*` call I could simplify the promise handling.  

I also realized that the `store` feature was basically identical across each different type of object ('Assignment', 'AssignmentActivityUsage', 'ActivityUsage') just parameterized by the class. So in this PR the `ActivityStore` is just using an `ObjectStore` helper. If we prefer this approach I'll refactor the assignments store to use the same helper to remove the duplication.


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
```diff
@@ -8,27 +8,27 @@ export class AssignmentStore {
 	}
 
 	fetchAssignment(href, token) {
-
 		let promise = this._assignments.get(href);
 		if (!promise) {
-			promise = new Promise(async(resolve) => {
+			promise = new Promise((resolve, reject) => {
```
mdgbayly: Cleaning up promise handling so that errors are not swallowed.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -32,36 +31,18 @@ class ActivityDueDateEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMixin
...
-		if (entity) {
-			this._date = entity.dueDate();
-		}
-
-		super._entity = entity;
-	}
-
 	_onDatetimePickerDatetimeCleared() {
-		this.wrapSaveAction(super._entity.setDueDate(''));
+		store.get(this.href).setDueDate('');
```
mdgbayly: As we no longer have a local component property representing the activity we have to get it from the store in event handlers in order to change it. We could check for null here I guess, but if the activity didn't exist then the component wouldn't render.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -80,5 +61,37 @@ class ActivityDueDateEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMixin
 		`;
 	}
 
+	render() {
+		const activity = store.get(this.href);
```
mdgbayly: This `get` call is a MobX reaction, so render will re-execute once that href exists in the store as a result of an async fetch.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -80,5 +61,37 @@ class ActivityDueDateEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMixin
...
+			${this.dateTemplate(dueDate)}
+		`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if ((changedProperties.has('href') || changedProperties.has('token')) &&
+			this.href && this.token) {
+			super._fetch(() => store.fetch(this.href, this.token));
```
mdgbayly: We still asynchronously fetch the activity in this component even though we don't store the result of the fetch directly in the component. The call to super is what matches the `d2l-pending-state` event handling.

components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -33,6 +35,13 @@ class ActivityEditor extends LocalizeMixin(LitElement) {
 		return getLocalizeResources(langs, import.meta.url);
 	}
 
+	async save() {
+		const activity = store.get(this.href);
```
mdgbayly: Here we just get the activity to save it rather than having to await on the fetch.

components/d2l-activity-editor/state/object-store.js
```diff
@@ -0,0 +1,37 @@
+import { action, decorate, observable } from 'mobx';
+
+export class ObjectStore {
+	constructor(Type) {
+		this.Type = Type;
+		this._fetches = new Map();
```
mdgbayly: Because we are splitting up the fetching and the fetched objects (so we can use MobX reactions on the fetched objects), we store the fetch promises independently of the fetched objects.

components/d2l-activity-editor/state/object-store.js
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	get(href) {
+		return this._objects.get(href);
+	}
+}
+
+decorate(ObjectStore, {
+	// properties
+	_objects: observable.shallow,
```
mdgbayly: And we observe the fetched `_objects` so that calls to `store.get(href)` will setup a MobX reaction when the object is eventually fetched.

test/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.spec.js
```diff
@@ -1,43 +1,83 @@
+import chai, { expect } from 'chai';
```
mdgbayly: Just added some more test conditions to the existing assignment tests.

test/d2l-activity-editor/state/activity-store.spec.js
```diff
@@ -0,0 +1,78 @@
...
+import { ActivityStore } from '../../../components/d2l-activity-editor/state/activity-store.js';
+import { ActivityUsage } from '../../../components/d2l-activity-editor/state/activity-usage.js';
+import chaiAsPromised from 'chai-as-promised';
+import { when } from 'mobx';
+
+chai.use(chaiAsPromised);
+
+jest.mock('../../../components/d2l-activity-editor/state/activity-usage.js');
+
+describe('Activity Store', function() {
```
mdgbayly: These tests are really just testing the `ObjectStore` helper class, so if we stick with this approach, I'll refactor it and remove a lot of the existing 'store' tests which are really just duplicating behavior across different types of activity classes.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
ppaskaris-d2l: Perhaps just do this:

```
async function doFetch() {

    const assignment = new Assignment(href, token);
    await assignment.fetch();
    return assignment;
}

promise = doFetch();
```

So you can keep async-await

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -80,5 +61,37 @@ class ActivityDueDateEditor extends SaveStatusMixin(EntityMixinLit(LocalizeMixin
 		`;
 	}
 
+	render() {
+		const activity = store.get(this.href);
+		if (!activity) {
```
ppaskaris-d2l: I like this much, much better than the weird falsey destructuring business :+1:

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -0,0 +1,61 @@
...
+		if (!this._entity) {
+			return;
+		}
+
+		// TODO - replace with aggregate save method
+		// await this._entity.save({
+		// 	dueDate: this.dueDate
+		// });
+
+		await this._entity.setDueDate(this.dueDate);
```
ppaskaris-d2l: I assume the "aggregate" method will handle change tracking as well?

components/d2l-activity-editor/state/object-store.js
```diff
@@ -0,0 +1,37 @@
...
+		let promise = this._fetches.get(href);
+		if (!promise) {
+			const object = new this.Type(href, token);
+			promise = object.fetch();
+			this._fetches.set(href, promise);
+
+			promise.then(action(() => {
+				this._objects.set(href, object);
+			}), () => {
+				this._fetches.delete(href, promise);
```
ppaskaris-d2l: nit: delete `map.delete(key)`

components/d2l-activity-editor/state/object-store.js
```diff
@@ -0,0 +1,37 @@
...
+
+export class ObjectStore {
+	constructor(Type) {
+		this.Type = Type;
+		this._fetches = new Map();
+		this._objects = new Map();
+	}
+
+	fetch(href, token) {
+		let promise = this._fetches.get(href);
```
ppaskaris-d2l: Doesn't this mean something only gets fetched once per href? What happens on `save` when it calls `fetch`?

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
mdgbayly: Yeh - that would be cleaner. 

If we go with the `ObjectStore` implementation with the observable map I'll be removing this anyway. But the `ObjectStore.fetch` has also ended up with some `then()` handling which could be refactored to use `await` too.

components/d2l-activity-editor/state/activity-usage.js
mdgbayly: Yes, I have another PR in the works in `siren-sdk` where I've started setting up an aggregate save method which does change tracking.

https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/128

components/d2l-activity-editor/state/object-store.js
mdgbayly: > Doesn't this mean something only gets fetched once per href? What happens on `save` when it calls `fetch`?

Yes, currently its setup so that objects always refresh in place, so the reference in the object store never changes.

https://github.com/BrightspaceHypermediaComponents/activities/pull/677/files#diff-71167fbaed272876b07ec82d005fed7dR49

Actually calling `save` on for example an `ActivityUsage` the `ActivityUsage` calls `fetch` on itself which just updates all the local data based on the update entity from the response.

ppaskaris-d2l: oh ok, the names are confusing me without an IDE to provide context.

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -0,0 +1,61 @@
...
+	constructor(href, token) {
+		this.href = href;
+		this.token = token;
+	}
+
+	async fetch() {
+		const sirenEntity = await fetchEntity(this.href, this.token);
+		if (sirenEntity) {
+			const entity = new ActivityUsageEntity(sirenEntity, this.token, { remove: () => { } });
+			this.load(entity);
```
mdgbayly: This should `await this.load(entity)`

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -0,0 +1,61 @@
...
+
+		const canEditDueDate = await entity.canEditDueDate();
+
+		runInAction(() => {
+			this.dueDate = entity.dueDate();
+			this.canEditDueDate = canEditDueDate;
+		});
+	}
+
+	async setDueDate(date) {
```
mdgbayly: Does not require `async`

mdgbayly:
Updated based on some of the PR comments.
Also refactored the assignment editor MobX integration to use the MobX enabled store.
A few other code cleanups.

My intention is to merge the previous assignment PR as is and then fix up via this PR.
You can look at the last 4 commits to see what changed since we had the PR walk through.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
kieranderson: Naming question: Should this be named "activityUsage"? Since this is coming from the "activityUsageHref" passed in and mapping to `AssignmentActivityUsageEntity` from siren-sdk

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -56,6 +58,7 @@ decorate(Assignment, {
 	instructions: observable,
 	canEditInstructions: observable,
 	instructionsRichTextEditorConfig: observable,
+	activityUsageHref: observable,
```
kieranderson: Trying to figure out why this is set as observable, Is this a value we expect to change? Or does it link up to the whole activityUsage object (Dates etc..)

components/d2l-activity-editor/d2l-activity-due-date-editor.js
mdgbayly: Possibly but I think in general we are a bit loose in our usage of activity vs activity-usage. In our components we've tended to stick with the term `activity` just because `activity-usage` can be a bit of a mouthful, hence `d2l-activity-due-date-editor` rather than `d2l-activity-usage-due-date-editor`. Also there could be an argument made that these components should be able to edit properties of both activities, activity usages, and user activity usages based on the URL they are initialized with. That would probably require enhancements to the store integration, so have not really worried about it for now, and just stuck with calling these activity editors.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
mdgbayly: It probably does never change, but it is referenced from the render method of the `d2l-activity-assignment-editor`. By making it observable we don't build in any assumptions about when it might be initialized. e.g. it could be empty on the initial render, and then be given a value later.  This was how these properties used to work in the initial integration approach where the `Assignment` was loaded initially as an empty object. So I think it's clearer to make anything that is referenced from a component render method `observable`.

PR #678, opened 2020-02-05
Fix activity due date initial render

mdgbayly:
This is a slight tweak to how we are rendering the date editors to account for some ugly behaviour in the legacy `d2l-datetime-picker`. As part of the Save/Cancel enhancements, a more logical way to render components is to only render them when they have some valid data. Currently we tend to just render everything with undefined values initially which doesn't seem very clean.

So as part of the Save/Cancel work I've been skipping rendering when the data for the component does not exist yet. However, this breaks the initial date rendering due to some weirdness in how the `d2l-datetime-picker` supports both a `date` property and a `datetime` property. Internally the component tries to keep these in sync, but it does it badly so if the first time you render the picker, you provide a `datetime` but no `date` then the 2 way databinding used internally in the component's integration with the vaadin date picker clears the `datetime` property you provided. 

Short of hacking into the `d2l-datetime-picker` which I am loathe to do as it's used in a few places (and who knows to what extent they depend on this weird behavior), the only way to make it work is to rely on the fact that we initially render the `d2l-datatime-picker` with an `undefined` `datetime` property. Yuck - eventually that component will be replaced but for now we have to work around it.

So this change sets up the due date editor, such that it allows an initial render of undefined values.


components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -63,24 +63,26 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
...
+		// We have to render with null values for dueDate initially due to issues with
+		// how the d2l-datetime-picker converts between the date & datetime attributes.
+		// If we delay rendering until we have a valid datetime, the d2l-datetime-picker
+		// overwrites the datetime attribute with an emptydate.
+		// Don't want to mess with the datetime picker...
+		// Tried passing an invalid date attribute to force it to use our datetime attribute
+		// but the 2-way data binding with the vaadin date picker always overrides it
+		// Will be able to fix when we have a new data time component.
+		if (!activity) {
+			dueDate = null;
```
emil-furniss: Yuck indeed, but at least you've found a workaround. Providing both a date _and_ a datetime on first render is not an option?

mdgbayly: Unfortunately not - if we supply a valid date and datetime attribute the date attribute without time takes precedence so we lose the time component.

I tried supplying an invalid date such as `date="nodate"` in the hope that would allow the `datetime` attribute to win out, but the 2 way databinding with Vaadin basically converts that invalid date property into the default empty date property again which causes the datetime prop to be set to null too.

PR #680, opened 2020-02-06
Hook up activity visibility editor component to save/cancel workflow

mdgbayly:
Includes adding isDraft state to our MobX enabled state class and connecting the `d2l-activity-visibility-editor` to the state store.

I've also taken this opportunity to try out some more unit testing of UI components (which is something we've kind of ignored up to now). It's made easier by the simplicity of using the MobX state components as opposed to having to wire up a Siren entity. 

This did mean I needed to add some methods to the store for putting objects, and also some actions to the `ActivityUsage` for things like `setCanEditDraft` to facilitate testing of the different states. These are features that are only required by the unit tests which I always find to be a bit smelly, but if it makes it easier to write tests and we think there is value in these tests, then I think it's acceptable.


mdgbayly:
Hmm, I just had a thought about this PR.

I vaguely recalled at one point there was talk about the learning paths team using this visibility editor.
Then I was told they were not using it.

But I just checked and they do in fact appear to be using it, which raises some issues for us as they are assuming it works in an autosave mode.

When I was doing the spikes, I experimented with some ideas for building the state integration in such a way that components could operate in an autosave fashion too, but I was not particularly happy with it, so pulled it from these first PRs. But that might need to be something we reconsider.

So I'm going to close this PR for now until we address that.

PR #681, opened 2020-02-07
Adding associated rubrics

kieranderson:


mdgbayly:
Wondering if this should instead be a `d2l-activity-rubrics-editor` rather than an assignment specific component, given that rubrics are likely something we want on other activities too?

Also was the link to the rubrics already on the assignment folder or was that something we added for this feature?

https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/129/files#diff-96de7224ed5a03f8bf7e9101a552a882R155

Again just wondering if this should be exposed via the `ActivityUsage` instead?

mdgbayly:
Also, with reference to how we are exposing the rubric links, does this mean that we are adding a link per associated rubric to the assignment siren entity links collection? I don't think that's wrong per se, but it is a little different to how we've modeled other things as collections, and it means we don't have the ability to encapsulate actions for adding new rubrics to that collection.

I believe Farnsworth (Anthony Wikkerink and others) have been implementing some activity-rubric hypermedia APIs and I think there was a hope that we could leverage some of the stuff they have been building to implement the rubric editing.

kieranderson:
@mdgbayly I think this will end up being on the `ActivityUsage`, and it might be already as well. Currently it does work as you mentioned, one link per rubric in the entity links collection.

I haven't added anything to the api yet, all of these links are currently part of the response for an assignment activity.

ppaskaris-d2l:
>Wondering if this should instead be a d2l-activity-rubrics-editor rather than an assignment specific component, given that rubrics are likely something we want on other activities too?

Given how the "siren sdk" is structured, I think this is the sensible wrapper for now. There's no top-level link to a rubrics collection, just a bunch of links to attached rubrics, so making it "duck type" the "siren sdk" "class" doesn't seem ideal.

>Again just wondering if this should be exposed via the ActivityUsage instead?

I have lost sight on what this distinction means at this point :/

>I believe Farnsworth (Anthony Wikkerink and others) have been implementing some activity-rubric hypermedia APIs and I think there was a hope that we could leverage some of the stuff they have been building to implement the rubric editing.

:+1: on getting their input on this. I think the link is coming from something named `AssignmentFolderRubricLinkProvider` in the mighty monolith, which looks like it was maybe also a bit of Nimbus (Miles) for their marking app?

kieranderson:
Switched over to using the associations, which is a collection that holds single and potential associations. Some of the components might seem a bit empty for now but will be filled up soon in subsequent PRs.

This is just adding the read only view of a rubric if it is already associated

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -221,6 +222,27 @@ class AssignmentEditorSecondary extends SaveStatusMixin(RtlMixin(EntityMixinLit(
 				</select>
 			</div>
 
+			<div id="availability-dates-container">
+				<d2l-activity-availability-dates-editor
```
ppaskaris-d2l: bad merge resolution here

kieranderson: Yeah, I'm seeing double

ppaskaris-d2l: yeah no it was this merge but the diff is whack

ppaskaris-d2l: I'm removing it in my PR #700

kieranderson: :+1: Thanks

PR #682, opened 2020-02-07
Refactor d2l-activity-visibility-editor to support save/cancel.

mdgbayly:
The visibility editor is also being used by the activity collections developed for learning paths where they expect it to support autosave currently.

I had spiked a few solutions whereby I tried to make the state management MobX code handle both save/cancel and autosave patterns but I didn't like any of them. 

So as an alternative I've gone back to the 'Presentational/Container' component split originally made popular by React/Redux (which subsequently became unpopular again...)

https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

There are two container components here that manage state using different strategies (MobX vs Autosave via EntityMixinLit), and they each delegate to a presentational component that actually handles the rendering. 

> MobX
`d2l-activity-visibility-editor` is the original component that is now integrated with our MobX state management but delegates rendering to the presentational component.

> Autosave
`d2l-activity-visibilility-auto-editor` which is a copy of the old autosave component but delegates rendering to the presentational component

> Presentational
`d2l-activity-visibility-editor-toggle` for the actual DOM rendering and is used by the two components above.

This seemed like the least of all evils and marginally favourable to just creating two duplicated components.

I'm told that there is a likelihood that Learning Paths will at some point move to a Publish/Cancel model too so it may be that this split can later be removed.

On the other hand it does give us a potential pattern to follow if we find a need for other components that we want to support both saving paradigms.

**UI Unit Testing**

In this PR, I've also made a renewed attempt to do some UI unit testing. Testing the MobX state management code is reasonably straightforward, but we've pretty much not done any web component testing. There is an ongoing debate about how useful this is, but I still see some value in it, at the least, just to encourage us to keep our components smaller and more self-contained and hence more maintainable.

I've tried out some of the open-wc testing tools before, and have introduced a few of them here as well to test the visibility editor.

https://open-wc.org/testing/testing.html#testing-helpers
https://open-wc.org/testing/testing-helpers.html

In particular I'm using their element fixture setup stuff from `@open-wc/testing-helpers`, `chai-a11y-axe` and `chai-dom`.

Please comment on whether or not you think there is value in these kinds of unit tests. The visibility component is almost too simple to bother with, but I think it shows what can be done.

Gaudi has just moved over all the @brightspace-ui/core testing to the `@open-wc` stack, and it would be nice if we could start thinking about doing the same in this repo at some point and get rid of polymer-cli completely. I recognize there are still a lot of Polymer components in here, but maybe we set up a different runner for the Lit components.


components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -0,0 +1,84 @@
...
+		this.addEventListener('click', this._onClick);
+	}
+
+	_onClick(e) {
+		if (!this.switchEnabled) {
+			e.stopImmediatePropagation();
+		}
+	}
+
+	get switchEnabled() {
```
ppaskaris-d2l: Not a fan of this change. Why extract a simple conditional out of the render function and make my eyes seek?

mdgbayly: Just because it was used in 2 places as a result of the refactoring (in the _onClick handler too).

ppaskaris-d2l: I see. Tragic.

test/d2l-activity-editor/d2l-activity-visibility-editor.js
```diff
@@ -0,0 +1,89 @@
...
+	describe('enabled', () => {
+
+		beforeEach(async() => {
+			el = await fixture(html`
+				<d2l-activity-visibility-editor-toggle canEditDraft isDraft></d2l-activity-visibility-editor-toggle>
+			`);
+		});
+
+		it('passes accessibility test', async() => {
+			await expect(el).to.be.accessible();
```
emil-furniss: Yes to unit tests, be it just for these a11y ones but also for the peace of mind that basic functionality is being checked.

PR #683, opened 2020-02-10
Summarizer for anonymous marking

pureooze:
Added the summarizer to the `d2l-activity-assignment-editor-secondary` component. Currently only the anonymous marking option has a message displaying in the summary but we can add the other options once we support MobX.

## Initial State:
image

## Opened:
image

## Closed with Anonymous Marking enabled:
image


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -154,73 +165,94 @@ class AssignmentEditorSecondary extends SaveStatusMixin(RtlMixin(EntityMixinLit(
...
+		`;
+	}
+
+	_getSummarizedContent() {
+		return this._getAnonymousGradingSummary();
+	}
+
 	render() {
 		return html`
-			<div id="assignment-submission-type-container">
```
pureooze: Classic diff, all I did was wrap this in `d2l-accordion-collapse`

ppaskaris-d2l:
Aside: your screenshots still don't work due to the malware on our network I think, but I can extract them from the source so it's ok.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -155,81 +166,102 @@ class AssignmentEditorSecondary extends SaveStatusMixin(RtlMixin(EntityMixinLit(
...
-					?checked="${this._isAnonymousMarkingEnabled}"
-					?disabled="${!this._canEditAnonymousMarking}"
-					ariaLabel="${this.localize('chkAnonymousMarking')}">
-					${this.localize('chkAnonymousMarking')}
-				</d2l-input-checkbox>
-				<d2l-input-checkbox-spacer ?hidden="${!this._anonymousMarkingHelpText}">
-					<span class="d2l-body-small">${this._anonymousMarkingHelpText}</span>
-				</d2l-input-checkbox-spacer>
-			</div>
+			<d2l-accordion-collapse class="accordion" flex header-border>
```
ppaskaris-d2l: Given Dave's recent remark, should we switch over to the "d2l labs" thing now?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -155,81 +166,102 @@ class AssignmentEditorSecondary extends SaveStatusMixin(RtlMixin(EntityMixinLit(
 		this.wrapSaveAction(super._entity.setAnonymousMarking(event.target.checked));
 	}
 
+	_getAnonymousGradingSummary() {
+		// TODO: replace with MobX once that is setup for this repo
+		return html`
+			<li class="d2l-body-compact" style=${this._isAnonymousMarkingEnabled ? '' : 'visibility: hidden'}>
```
ppaskaris-d2l: Prefer `hidden?=${this._isAnonymousMarkingEnabled}`, and either that will work normally, or you can add this rule:

```
li[hidden] { display: none; }
```

Do not use `visibility: hidden;` because it will continue to take up vertical space.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
pureooze: Yes I think we will need to do that now.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-secondary.js
```diff
@@ -155,81 +166,102 @@ class AssignmentEditorSecondary extends SaveStatusMixin(RtlMixin(EntityMixinLit(
...
+		`;
+	}
+
+	_getSummarizedContent() {
+		return this._getAnonymousGradingSummary();
+	}
+
 	render() {
 		return html`
-			<div id="assignment-type-container">
```
mdgbayly: Should the summarizer be wrapping everything? Or just the anonymous marking checkbox for now?

pureooze: So I talked to the designers, I originally misunderstood what they were asking for. The `Evaluation & Feedback` summarizer needs to wrap Rubrics, ePortfolio Artifacts, Anonymous Marking, Annotation Tools and Turn it In Integration. Some of these havent been added to the UI yet but I will push an update to wrap the ones that are already here (Anonymous Marking and Annotation Tools).

https://miro.com/app/board/o9J_kwb1k3k=/

mdgbayly: Ok - yeh - sounds good. 

I wonder if we should consider breaking up some of these components into sub-components, possibly along the lines of which summarizer they will be in. The detail and secondary components were originally intended as just a place to dump stuff whilst we were still working out the designs and where things would live.

I quite like the simplicity of components like the due date editor, availability dates and Turn it in editors. And it makes it easier to write tests if components are very single purpose.  I don't think we necessarily have to take this to extremes. e.g. maybe a component for the something like the anonymous marking checkbox would be overkill. But where ever there is extra UI logic for managing a property it doesn't hurt to put it in a self-contained component.

What do you guys think?

pureooze: I think breaking them up by the summarizer they will be in makes sense. It might make it easier to separate the state so that any summarizers that dont contain the anonymous marking checkbox dont have to worry/know about its state. Also testing them individually will probably be easier/cleaner.

ppaskaris-d2l: I think making each editor its own component, and each summarizer its own components, to be an ideal solution. Components are cheap, reading thousand line render functions is hard.

PR #684, opened 2020-02-11
Initial attachment collection MobX store. 

mdgbayly:
Just supports loading collections and attachments for now.
Integration with web components will follow in later PR.

The store supports caching both collections and attachments so that resource based web components for the attachment list and attachments can each reference their associated entities.

The collection object contains observable properties for:
* the list of entity hrefs (not to be confused with attachment target urls), for each attachment in the collection.
* permissions for each attachment type that will be used to determine what buttons are enabled in the picker.

The attachment object contains observable properties for:
* the status (editing, creating, deleted) - used to determine whether we have a remove button/undo button etc
* the name/url of the attachment used to determine how the attachment will render.

This PR depends on a new `siren-sdk` method in this PR:
https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/131

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
@@ -14,12 +14,12 @@ export class AttachmentCollection {
 		const sirenEntity = await fetchEntity(this.href, this.token);
 		if (sirenEntity) {
 			const entity = new AttachmentCollectionEntity(sirenEntity, this.token, { remove: () => { } });
-			await this.load(entity);
+			this.load(entity);
 		}
 		return this;
 	}
 
-	async load(entity) {
+	load(entity) {
 		this._entity = entity;
 
 		this.canAddAttachments = entity.canAddAttachments();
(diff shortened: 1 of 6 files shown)


components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
```diff
@@ -0,0 +1,49 @@
...
+	async fetch() {
+		const sirenEntity = await fetchEntity(this.href, this.token);
+		if (sirenEntity) {
+			const entity = new AttachmentCollectionEntity(sirenEntity, this.token, { remove: () => { } });
+			await this.load(entity);
+		}
+		return this;
+	}
+
+	async load(entity) {
```
emil-furniss: Does this need to be `async`?

components/d2l-activity-editor/d2l-activity-attachments/state/attachment.js
```diff
@@ -0,0 +1,45 @@
...
+	async fetch() {
+		const sirenEntity = await fetchEntity(this.href, this.token);
+		if (sirenEntity) {
+			const entity = new AttachmentEntity(sirenEntity, this.token, { remove: () => { } });
+			await this.load(entity);
+		}
+		return this;
+	}
+
+	async load(entity) {
```
emil-furniss: And how about this `async`?

mdgbayly: yeh - these ones don't really need it - just because i copy/pasted the template from another class.

PR #685, opened 2020-02-11
Truncating activity/course columns properly.

cbusmand2l:
Activity and course columns should now truncate properly: 
image

The columns were not being assigned the truncated class properly, and their widths were expanding if there was enough text in the column: 
image


components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -96,9 +96,7 @@ class D2LQuickEvalSubmissionsTable extends QuickEvalLogging(QuickEvalLocalize(Po
 					overflow: hidden;
 					text-overflow: ellipsis;
 					max-width: 24rem;
-					text-overflow: ellipsis;
```
cbusmand2l: `text-overflow` and `overflow` were being set twice in this selector for some reason. This is just clean up.

PR #686, opened 2020-02-11
Jroach/us113249

jeroach:
DO NOT MERGE THIS GARBAGE

with that out of the way ;)

This is one half of the POC work to show message passing from parent to iframe

Other half: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/8346/overview


components/d2l-evaluation-page/d2l-evaluation-page.js
```diff
@@ -0,0 +1,46 @@
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+
+class EvaluationPage extends LitElement {
```
jeroach: simple component that uses the `primary-secondary` page template

components/d2l-evaluation-page/d2l-evaluation-page.js
```diff
@@ -0,0 +1,46 @@
...
+		return html`
+			<d2l-template-primary-secondary>
+				<iframe
+					slot="primary"
+					id="d2l-evaluation-page-iframe"
+					src="${this.leftPanelUrl}">
+				</iframe>
+				<button
+					slot="secondary"
+					onclick="SendMessageFromParentToIframe()">Button in parent
```
jeroach: This button calls the method in here: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/8346/diff#le/evaluationhub/static/js/EvaluationPage.js

dmackenz: Can this logic be inside of the BSI repo instead of served up by the backend?

dmackenz: I've been thinking about how we can implement this with MobX. 

message from parent to iframe:
invoke action -> state change -> observer invokes SendMessageFromParentToIframe -> iframe receives message

message from iframe to parent:
event listener accepts message -> invokes action -> state updates

dmackenz:
Is there any logic currently for sending a message from iframe to parent?

components/d2l-evaluation-page/d2l-evaluation-page.js
AlexBedley: yeah I'm not a huge fan of relying on a global for this. Ideally a module or callback (or event name??) is passed in so it's testable

jeroach:
> Is there any logic currently for sending a message from iframe to parent?

Not yet. This is slightly tricky since there needs to be some JS logic inside the page we're going to iframe. Given we're going to be using existing evidence pages, we either have to:

- add logic to the existing page that handles the event firing logic for the case when it's in an iframe
- write our own wrapper for the existing evidence page, and include the event firing logic in there

Either way, the code to fire and receive the events is straightforward:

```
in iframe:

window.parent.postMessage(data, targetOrigin, false);

in page-template:

window.addEventListener('message', function(event) {
	if(event.origin === targetOrigin) {
		// handle message
	}
}, false);
```

components/d2l-evaluation-page/d2l-evaluation-page.js
jeroach: Yeah good call guys - I was mostly just hacking this together to prove events can be sent and received. I'll work on refining the POC

jeroach: @AlexBedley for the example you mentioned - do you mean:

this line: https://search.d2l.dev/xref/lms/le/evaluationhub/D2L.LE.EvaluationHub/Web/Desktop/Views/EvaluationHubView.Renderer.cs?r=27b02c9b#179

which references this function: https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web/UI/Html/Authentication/OAuth2/ClassInitializer.js?r=a9713b1a#2

PR #689, opened 2020-02-12
Integrate MobX state management into attachment editor components

mdgbayly:
Only supports rendering for now. Adding/deleting will come later.

It will be good to get this PR reviewed, but I will not be merging it until after I've added support for add & delete, as I don't want to break the existing functionality.

I've also extended the ActivityEditorMixin to handle the `updated` callback to initiate the fetch to avoid having to add this handler to every component.

I had avoided doing that previously because the `assignment-store.js` encapsulates two stores: one for the `Assignment` and one for the `AssignmentActivityUsage`, so it has methods like `fetchActivity` and `fetchAssignment`. But I realized as I was doing the attachments stuff (which has collections and attachments) that there is not really any requirement to encapsulate the stores in this way, and that by just keeping each store independent, we can just make every store expose a `fetch` method, which means we can now move the fetch handler to the mixin by passing the component's store to the mixin.

Sorry, this is a pretty big PR, because I needed to refactor all 4 of the components in order to get all the tests working. I guess if I'd started with the leaf components, I could have done 4 PRs - live & learn.


components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
```diff
@@ -68,4 +69,13 @@ export const ActivityEditorMixin = superclass => class extends superclass {
 			super.disconnectedCallback();
 		}
 	}
+
+	updated(changedProperties) {
```
emil-furniss: Liking this! My recent availability-dates-editor conversion #690 still used the old way, so that might need updating here.

mdgbayly: > Liking this! My recent availability-dates-editor conversion #690 still used the old way, so that might need updating here.

Yep - there are a few places where we can refactor to use this instead. The assignments store will need to be split up first too. But I won't be able to merge this until i've got most of the attachments stuff working.

mdgbayly:
The following 2 PRs were reviewed/approved independently prior to merging into this PR which will be merged into master. This was done to keep the changes distinct for easier review but to allow all changes to be deployed as a single change to master.

https://github.com/BrightspaceHypermediaComponents/activities/pull/692
https://github.com/BrightspaceHypermediaComponents/activities/pull/698

PR #690, opened 2020-02-12
Convert availability-dates-editor to save/cancel

emil-furniss:
- add start and end date properties and methods to the activity-usage mobx store
- convert editor to use store
- add editor ui and store tests

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
@@ -1,6 +1,6 @@
 import 'd2l-datetime-picker/d2l-datetime-picker';
-import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
 import { css, html } from 'lit-element/lit-element';
+import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
 import { getLocalizeResources } from './localization';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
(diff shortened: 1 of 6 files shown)


components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -1,125 +1,106 @@
...
+
+	_onEndDatetimePickerDatetimeChanged(e) {
+		store.get(this.href).setEndDate(e.detail.toISOString());
+	}
+
+	render() {
+		const activity = store.get(this.href);
+		let canEditStartDate, canEditEndDate, startDate, endDate;
+
+		if (!activity) {
```
emil-furniss: Ran into the same issue that came up in #678 in the due date editor, so had to set some default dates and permissions for the first render.

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -24,10 +24,16 @@ export class ActivityUsage {
 		this._entity = entity;
 
 		const canEditDueDate = await entity.canEditDueDate();
```
mdgbayly: Maybe in a separate follow on PR, we can refactor all this `canEdit*Date` stuff to just check for the new action you added to the siren-sdk for aggregate dates.  These existing functions are really cumbersome because they actually make an API call if there is no current date. Now we have our new aggregate dates we could add a simple `canEditDates` function to the `ActivityUsageEntity` which just looks for the new action instead.

PR #691, opened 2020-02-12
Use combined dates action to check edit dates permissions in editors

emil-furniss:
Replaces checking for `canEditStartDate`, `canEditDueDate`, and `canEditEndDate` with `canEditDates` based on the new aggregate save dates action.

Depends on a new helper methon in the siren-sdk: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/132


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -22,21 +22,12 @@ export class ActivityUsage {
 
 	async load(entity) {
 		this._entity = entity;
-
-		const canEditDueDate = await entity.canEditDueDate();
-		const canEditStartDate = await entity.canEditStartDate();
-		const canEditEndDate = await entity.canEditEndDate();
-
-		runInAction(() => {
```
emil-furniss: `canEdit*Date` are all async (can depend on request to start-add-new), but `canEditDates` is not.

PR #692, opened 2020-02-12
enable adding new activity editor attachments with save & cancel

mdgbayly:
Adds support for adding and deleting attachments to the Save/Cancel MobX state.
Note: Save is not supported yet. Just add/delete whilst editing.

The extra Attachment subclasses are required to enable us to delegate to the correct siren action once I actually implement Save in the next PR. For now they look a bit innocuous. But the FileAttachment/LinkAttachment do contain some `init*` helpers for setting up a new attachment. 

These subclasses will only be used when creating a new attachment before it is saved. They will not be used when loading existing attachments as we have no polymorphic behaviour that needs to be enforced once an attachment has been saved. eg. for files, FileAttachment encapsulates the `fileSystemType` and `fileId` that is needed to create a file from a temp file. Once the file is saved a file attachment is just another URL like any other attachment, hence the Attachment base class is sufficient.

I'd initially created some of the store classes with `CRLF` endings by mistake. Fixed here, but use the cog at the top of the PR to ignore whitespace so you can filter out the noise.


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
```diff
@@ -33,14 +34,20 @@ class ActivityAttachment extends ActivityEditorMixin(MobxLitElement) {
...
 		return html`
 			<d2l-labs-attachment
 				attachmentId="${id}"
 				.attachment="${attachment}"
+				?deleted="${deleted}"
+				?creating="${creating}"
 				?editing="${editing}"
-				@d2l-attachment-removed="${this._onAttachmentRemoved}">
+				@d2l-attachment-removed="${this._onAttachmentRemoved}"
+				@d2l-attachment-restored="${this._onAttachmentRemoved}"
```
emil-furniss: Is `restored` a new event? And just as a nit, could the handler be renamed to sound more toggle-esque, so `restore` is not being handled by `onRemoved`?

components/d2l-activity-editor/d2l-activity-attachments/state/attachment.js
```diff
@@ -1,49 +1,114 @@
...
+
+configureMobx({ enforceActions: 'observed' });
+
+export class Attachment {
+	constructor(href, token) {
+		this.href = href;
+		this.token = token;
+		this.editing = false;
+		this.creating = false;
+		this.deleted = false;
```
emil-furniss: Just out of curiosity, could these be initialized with `editing = true`, `creating = true`, and `deleted = false`, rather than having to set those in each `init*` methods? Or would that cause issues when loading existing attachments?

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
mdgbayly: No restore always existed but we couldn't use it in the autosave implementation. Yeh I guess the handler could be renamed. Will do on my newer branch.

components/d2l-activity-editor/d2l-activity-attachments/state/attachment.js
mdgbayly: Possibly - i think i only added them in here because i was trying to understand how MobX would handle setting props in subclasses. It wouldn't really affect the load. I guess i figured setting them as defaults for a readonly attachment was clearer for whatever reason in case a user doesn't have permissions but I don't think it matters that much. But I could probably clean it up so that we are not always setting all 3. e.g. we don't really need to be setting the deleted prop in the init/load methods if i set it here.

PR #696, opened 2020-02-13
Fix assignment due date editor blank render

emil-furniss:
The assignment activity is available on first render, which causes the picker to render blank (see #678 ).
This change forces the first render to happen with date set to null, and then immediately re-render with the actual date.

Yuck. Let me know if there are nicer ways of doing this.


mdgbayly:
So this started breaking again? I thought once I had seen this happen, but then couldn't reproduce. Was this happening on a quad build?

I guess the only danger of this is what happens if there isn't a second render required.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -86,6 +88,10 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
 		`;
 	}
 
+	firstUpdated() {
+		this._isFirstLoad = false;
```
mdgbayly: should probably call `super.firstUpdated()` too.

mdgbayly: > Yuck. Let me know if there are nicer ways of doing this.

Fixing the due date picker ? ;-)

PR #698, opened 2020-02-14
Implement save of attachments

mdgbayly:
This is the final PR for save/cancel of attachments. Once this is approved I will merge the 2 branches into the `mbayly/save-cancel-attachments-2` branch, and then request a final approval on that merged PR in order to push to master.

The bulk of this PR is adding the polymorphic save methods to each of the attachment types, and adding the attachment collection save method, which determines which attachments are new/deleted/discarded based on analyzing the `creating` and `deleted` flags. The attachment collection tells each changed attachment to either delete or save itself, passing in the current `AttachmentCollectionEntity` into the `save` method so the attachment can call the specific attachment save action (e.g. link, google, onedrive, file etc).

You'll see that I had to add a reference to the attachment store, to the attachment collection store and give each collection a reference to it's attachment collection store so that it can get access to the attachment store. This is needed during save to allow the attachment collection to get access to the actual attachments, as currently the collection only stores the hrefs of the associated attachments. With hindsight, I might have been able to avoid having to connect the stores in this way by storing the actual attachment references in the attachment collection as opposed to just the hrefs. Possibly a candidate for a future refactoring if we think that would be cleaner, but I'm happy enough with the implemented approach for now.

@emil-furniss I also renamed the delete attachment method and simplified the setting of the attachment status flags in this PR, based on your comments from the previous PR.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
@@ -5,10 +5,11 @@ import { fetchEntity } from '../../state/fetch-entity.js';
 configureMobx({ enforceActions: 'observed' });
 
 export class AttachmentCollection {
-	constructor(href, token) {
+	constructor(href, token, store) {
 		this.href = href;
 		this.token = token;
 		this.attachments = [];
+		this.store = store;
 	}
 
 	async fetch() {
@@ -68,6 +69,49 @@ export class AttachmentCollection {
 
 	addAttachment(attachment) {
 		this.attachments.push(attachment.href);
+	}
+
+	async save() {
+		const attachmentStore = this.store.getAttachmentStore();
+		if (!attachmentStore) {
+			throw new Error('No attachment store configured. Cannot save');
+		}
+
+		const discarded = [];
+		let hasChanged = false;
+
+		for (const href of this.attachments) {
+			// TODO - Should we run these concurrently using an array of promises?
+			// Siren action helper will still serialize them but we could setup the
+			// siren sdk methods to allow us to pass the immediate option.
+			const attachment = attachmentStore.get(href);
+			if (attachment.deleted && !attachment.creating) {
+				await attachment.delete();
+				hasChanged = true;
+			}
+			if (attachment.creating && !attachment.deleted) {
+				await attachment.save(this._entity);
+				hasChanged = true;
+			}
+
+			if (attachment.creating && attachment.deleted) {
+				discarded.push(href);
+			}
+
+			// Clean up store reference to temporary or deleted attachments
+			if (attachment.creating || attachment.deleted) {
+				attachmentStore.remove(href);
+			}
+		}
+
+		if (hasChanged) {
+			await this.fetch();
+		} else if (discarded.length > 0) {
+			// Clean up new attachments that were removed before save
+			for (const href of discarded) {
+				this.attachments.remove(href);
+			}
+		}
 	}
 }
 
--- a/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
@@ -135,5 +135,6 @@ decorate(AttachmentCollection, {
 	setCanRecordVideo: action,
 	setCanRecordAudio: action,
 	setAttachments: action,
-	addAttachment: action
+	addAttachment: action,
+	save: action
 });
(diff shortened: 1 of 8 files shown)


test/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.spec.js
```diff
@@ -65,7 +76,9 @@ describe('Attachment Collection', function() {
 
 	describe('updating', () => {
 		it('reacts to new attachment', async(done) => {
-			const attachment = new Attachment('http://attachment/1', 'token');
+			const attachment = {
```
mdgbayly: You'll see I'm using  a combination of sinon and jest mocking in these MobX tests.
It's unfortunate to have to grok two mocking frameworks, but unfortunately sinon doesn't support mocking out ES6 modules/classes, which is why we are having to use Jest mocking in some places. We could possibly just use jest, but i find its API to be pretty cumbersome compared to sinon. So generally I'm only using it where we need to mock something that is loaded by another file as an ES6 module or where another file is calling `new` on an ES6 class.

So you'll see in this test I switched to just using plain objects for the mock attachments, and using sinon to mock the `delete` and `save` method calls. This is sufficient because these tests never cause a new attachment to be created outside the test file.

components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
```diff
@@ -69,6 +70,49 @@ export class AttachmentCollection {
...
+			throw new Error('No attachment store configured. Cannot save');
+		}
+
+		const discarded = [];
+		let hasChanged = false;
+
+		for (const href of this.attachments) {
+			// TODO - Should we run these concurrently using an array of promises?
+			// Siren action helper will still serialize them but we could setup the
+			// siren sdk methods to allow us to pass the immediate option.
```
emil-furniss: Is there a benefit to doing that? Performance or something? Because if not, then I think this is perfectly good way to do it, and nice and easy to understand.

mdgbayly: Yeh - it's mostly about the time it takes to do a save.  Things like delete are particularly slow because they include a link header to the collection. And the fetches on those link headers are fetched in the context of the async delete call too. So for example if you delete two attachments then the save call will involve the following 'serialized/sequenced' API calls.

* delete attachment 1
* fetch attachments collection
* delete attachment 2
* fetch attachments collection
* save completed!

So I think there are potentially other optimizations we could employ as well. e.g. in the context of the save here, we expect to close the page so we don't really care about those intermediate fetch collection calls, as we are not going to re-render the updated collection. So maybe we could enhance the siren action integration to tell it to ignore link header processing.

Once we've done more of the Save/Cancel integrations and got it working with the existing API approach (which is very chatty) I think it would be useful to step back and think about how we could potentially improve it. Optimizations tend to lead to more complex/less consistent code though, so yes, it is nice and simple right now.

components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
```diff
@@ -69,6 +70,49 @@ export class AttachmentCollection {
 	addAttachment(attachment) {
 		this.attachments.push(attachment.href);
 	}
+
+	async save() {
+		const attachmentStore = this.store.getAttachmentStore();
```
emil-furniss: I might be completely missing something here, but: Whats the thinking behind getting the attachment store and saving each attachment individually from here, rather than following the pattern from `state/activity-usage.js` and calling save on the `AttachmentsCollectionEntity` via `this._entity.save()`? Wouldn't that avoid having to pass the state reference in the constructor?

mdgbayly: Do you mean pass the list of attachments to `AttachmentCollectionEntity.save`. And then do the looping in there? That could possibly work, but I think it would rely on storing the actual attachment objects in the collections `attachments` props (rather than just hrefs) as discussed in my other comment. Otherwise we'd still need to pass the attachment collection store to the entity save method.

Also, the collection still might need to have a reference to the attachment store for the case of discarded attachments (attachments that are added & deleted without ever being saved). I didn't like the idea of not cleaning up attachments from the attachments store as that represents a potential 'memory leak', even though generally on save, the whole thing is going to be thrown away. 

With hindsight, I'm not entirely happy with the implementation of the attachment-store. I originally modeled it in that way because it fit our idea of all components being independent and href based etc. So there was a store that just stored attachments by href. But I'm wondering now if it wouldn't have been better to make the attachment collection stand in for the attachment store instead. That would probably have solved many of these problems.

But this works so I think I'd like to go with it for now, but yeh I definitely might want to refactor this at some future point, and I might try a different approach if we find we have to model other collection based entities (e.g. grades).

PR #701, opened 2020-02-18
Using mobx to save/load state from submission summarizer

pureooze:


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -1,23 +1,18 @@
...
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
-import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
-import { SaveStatusMixin } from '../save-status-mixin.js';
+import { MobxLitElement } from '@adobe/lit-mobx';
 import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
+import { shared as store } from './state/assignment-store.js';
 
-class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(EntityMixinLit(LocalizeMixin(LitElement)))) {
-
+class AssignmentEditorSubmissionAndCompletion extends LocalizeMixin(MobxLitElement) {
```
ppaskaris-d2l: Looks like you should inherit from `ActivityEditorMixin`. Consider aligning your implementation with the availability dates editor that @mdgbayly created, and also naming it similarly: `ActivityAssignmentSubmissionAndCompletionEditor`

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -54,51 +49,27 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
...
-		}
-
-		this._submissionTypes = assignment.submissionTypeOptions();
-		this._canEditSubmissionType = assignment.canEditSubmissionType();
-		this._completionTypes = assignment.completionTypeOptions();
-		this._canEditCompletionType = assignment.canEditCompletionType();
-	}
-
 	_saveCompletionTypeOnChange() {
 		const completionType = this.shadowRoot.querySelector('select#assignment-completion-type').value;
```
ppaskaris-d2l: Consider taking an `event` argument and doing: `event.target.value` if possible.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -54,51 +49,27 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
...
-		this.wrapSaveAction(super._entity.setCompletionType(completionType));
+		store.getAssignment(this.href).setCompletionType(completionType);
 	}
 
-	_getSubmissionTypeOptions() {
+	_getSubmissionTypeOptions(assignment) {
+		const submissionTypes = assignment ? assignment.submissionTypeOptions : [];
```
ppaskaris-d2l: Consider:

```
if (!assignment) {
    return html``;
}
```

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -129,9 +101,9 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
 				<select
 					id="assignment-submission-type"
 					class="d2l-input-select block-select"
-					@change="${this._saveSubmissionTypeOnChange}"
-					?disabled="${!this._canEditSubmissionType}">
-						${this._getSubmissionTypeOptions()}
+					@change="${this._saveSubmissionTypeOnChange}}"
```
ppaskaris-d2l: How does this compile: double `}`

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -141,18 +113,20 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
 		return html``;
 	}
 
-	_renderAssignmentCompletionType() {
+	_renderAssignmentCompletionType(assignment) {
+		const completionTypeOptions = assignment ? assignment.completionTypeOptions : [];
+		const canEditCompletionType = assignment ? assignment.canEditCompletionType : false;
 		return html `
-			<div id="assignment-completion-type-container" ?hidden="${!this._completionTypes.length > 0}">
+			<div id="assignment-completion-type-container" ?hidden="${!completionTypeOptions.length > 0}">
```
ppaskaris-d2l: nit; I do not like `! binary_operator_that_makes_me > read_the_operator_precedence`. Consider parentheses (because in this case, I believe it does this: `(!completionOptions.length) > 0` instead of `!(completionOptions.length > 0)`, or:

```
const hidden = assignment ? assignment.completionTypeOptions.length < 0 : true;
```

pureooze: I like the last one best, will switch to that.

ppaskaris-d2l:
Adding Martin, because this is a fresh conversion from Siren-stuff to MobX, and I would appreciate your review.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -28,6 +28,28 @@ export class Assignment {
...
+		this.completionTypeOptions = entity.completionTypeOptions();
+		this.canEditSubmissionType = entity.canEditSubmissionType();
+		this.canEditCompletionType = entity.canEditCompletionType();
+		this.submissionType = entity.submissionType();
+		this.completionType = entity.completionType();
+	}
+
+	setSubmissionType(value) {
+		const type = this.submissionTypeOptions[value];
+		this.submissionType = {
```
ppaskaris-d2l: I think this should just be `this.submissionType = type.value`, no? What do you need the `title` for? And if you need the `title`, why drop the `selected` prop? Are you just trying to shallow-copy it? If so, consider `Object.assign({}, type)`.

pureooze: Discussed this offline, the API uses an object with a structure `{title: "File submission", value: 0}` so I wanted to keep it stored in that format. Also didnt use Object.assign because the submissionTypeOptions objects have additional properties on them.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -141,18 +108,21 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
 		return html``;
 	}
 
-	_renderAssignmentCompletionType() {
+	_renderAssignmentCompletionType(assignment) {
+		const canEditCompletionType = assignment ? assignment.canEditCompletionType : false;
+		const completionHidden = assignment ? assignment.completionTypeOptions.length <= 0 : false;
```
mdgbayly: Minor nit: but should the default here for a null assignment be hidden is true?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -54,51 +43,28 @@ class AssignmentEditorSubmissionAndCompletion extends SaveStatusMixin(RtlMixin(E
...
+	_saveSubmissionTypeOnChange(event) {
+		store.getAssignment(this.href).setSubmissionType(event.target.value);
 	}
 
-	_getCompletionTypeOptions() {
+	_getCompletionTypeOptions(assignment) {
+		const completionTypeOptions = assignment ? assignment.completionTypeOptions : [];
 		return html`
-			${this._completionTypes.map(option => html`<option value=${option.value} ?selected=${option.selected}>${option.title}</option>`)}
+			${completionTypeOptions.map(option => html`<option value=${option.value} ?selected=${option.selected}>${option.title}</option>`)}
```
mdgbayly: How is this updating the selected option when the UI changes the MobX state? I think it probably looks like it's working because it is just rendering the state of the HTML select. However, if something else were to call `set[Submission|Completion]Type` on the MobX state class, I don't think it would render the correct selected option, because it isn't using the observable submission/CompletionType props, and the store class is not updating the select prop on the options.

So I think it needs to compare the current option value with the current `assignment.completionType` value in order to determine what is selected.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -28,6 +28,28 @@ export class Assignment {
 		this.canEditInstructions = entity.canEditInstructions();
 		this.instructionsRichTextEditorConfig = entity.instructionsRichTextEditorConfig();
 		this.activityUsageHref = entity.activityUsageHref();
+		this.submissionTypeOptions = entity.submissionTypeOptions();
+		this.completionTypeOptions = entity.completionTypeOptions();
+		this.canEditSubmissionType = entity.canEditSubmissionType();
+		this.canEditCompletionType = entity.canEditCompletionType();
+		this.submissionType = entity.submissionType();
```
mdgbayly: I would probably just set `this.submissionType` to the `value` field of the `entity.submissionType()` given that's all that is really needed for managing the state. Then that avoids having to add back in the title when the type changes.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
mdgbayly: I recall seeing some issues with handling changes to the selected option due to a mismatch between the numeric value that was being used in the action to represent the options (e.g. in the `update-submission-type` siren action), and the fact that the HTML Option tag value was always being coerced to a string. Is that not happening with the way you are using `event.target.value` to handle the change? e.g. is `event.target.value` actually numeric?

mdgbayly: Also, this needs to take into account that different submission types have different valid completion types. Ideally if the submission type changes such that the currently selected completion type is still valid then the completion type should not change (the current autosave implementation always resets it).

mdgbayly: Are you planning on updating the `save` implementation to include these types? My recommendation would be to extend the siren-sdk `AssignmentEntity.save` method to support these new fields.

That implementation contains a bit of a hack at the moment to force it to use the 'PATCH' API endpoint. The original update action was incorrectly declared as a 'PUT' even though it isn't and not all fields are declared in the action even though they are supported by the backend. I'd like us to clean up the action declaration to make it correct reflect how it is actually used. But for now we can just explicitly add the fields. I think this is preferable to calling all the distinct siren actions that were originally created for the autosave implementations because they all call the same API endpoint anyway, and ideally we want to aggregate these API calls where possible.

ppaskaris-d2l: >Also, this needs to take into account that different submission types have different valid completion types. Ideally if the submission type changes such that the currently selected completion type is still valid then the completion type should not change (the current autosave implementation always resets it).

I think there's a lot more to chew through here than can get done in this pull request. What are we doing for data-dependent form configurations? In legacy, we have all these rules client-side in JavaScript provided by the server at page render.

mdgbayly: For the submissionType/completionType stuff there is support in the siren representations to do the validation in the UI. The team that originally added this to the APIs decided to include a list of valid completionTypes in each submissionType. It's not particularly 'standard' but I don't think there really are any standard ways to do this in Siren. I would probably have preferred to use some kind of JSON schema definition for this kind of thing, but this is what we have for now.
```
    {
      "href": "https://ac7b9194-5157-4af6-acdb-8270adaffc49.assignments.api.dev.brightspace.com/132655/folders/1098",
      "name": "update-submission-type",
      "method": "PATCH",
      "fields": [
        {
          "type": "radio",
          "title": "Submission Type",
          "name": "submissionType",
          "value": [
            {
              "title": "File submission",
              "value": 0,
              "completionTypes": null,
              "selected": true
            },
            {
              "title": "Text submission",
              "value": 1,
              "completionTypes": null,
              "selected": false
            },
            {
              "title": "On paper submission",
              "value": 2,
              "completionTypes": [
                1,
                2,
                3
              ],
              "selected": false
            },
            {
              "title": "Observed in person",
              "value": 3,
              "completionTypes": [
                1,
                2,
                3
              ],
              "selected": false
            }
          ]
        }
      ]
    }
```

It's also unfortunate that they chose to use numeric values instead of string values as I have a feeling it causes issues when comparing with the actual OPTION tag select value.

mdgbayly: Yeh, I agree this is a big PR so no problem if we want to break it up. It's just that if we merge this PR as is then I think we'll be breaking the save of submission type/completion type?

For some of the bigger changes I've been doing, I've broken it up over multiple PRs to make reviews easier, but then deferred merging them until all were ready. 

Up to you guys how you want to manage this.

pureooze: So I looked into this a bit and I'm wondering how we want to deal with the case where the initial load has a submission type that does not have any completion types associated to it but then the user changes to one that does. When fetching using `entity.completionTypeOptions();` on the initial load we will get an empty array. However if the user then selects a submission type that does have completion types, there doesn't seem to be an easy way to get a list of the valid completion types. As it is, I think we would need to save the changes to the submission type and then fetch the completion types again to get the list. Any idea if there is a cleaner way to do this?

mdgbayly: @pureooze Sorry - only just got back to this PR. If you have questions that you'd like a more timely response to, feel free to ask on #face-dev or just message me directly on slack.

Not sure if you already came up with a solution, but I think you just need to enhance the `AssignmentEntity` class in the `siren-sdk` to make this information available. I did something like that when I was working on the state management spikes. The set of valid completion types is defined in the action for updating the submission type as shown above.

e.g. I added the following helper methods (note: this was spike code so probably not perfect)

```
	/**
	 * @returns {Array} Set of completion type options for the submissionType
	 */
	validCompletionTypeOptions(submissionType) {

		submissionType = Number(submissionType);

		if (!this.canEditCompletionType()) {
			return [];
		}

		const action = this._entity.getActionByName(Actions.assignments.updateCompletionType);
		if (!action.hasFieldByName('completionType')) {
			return [];
		}

		const validCompletionTypes = this.validCompletionTypes(submissionType)

		return action.getFieldByName('completionType').value.filter(option => {
			return validCompletionTypes.indexOf(option.value) > -1;
		});
	}

	validCompletionTypes(submissionType) {
		submissionType = Number(submissionType);
		const validCompletionTypes = this.submissionTypeOptions()
			.find(option => option.value === submissionType)
			.completionTypes;
		if (validCompletionTypes === null) {
			return [];
		}
		return validCompletionTypes;
	}

	ensureValidCompletionType(submissionType, completionType) {
		submissionType = Number(submissionType);
		completionType = Number(completionType);

		const validCompletionTypes = this.validCompletionTypes(submissionType);

		if (validCompletionTypes.indexOf(completionType) === -1) {
			return validCompletionTypes.length > 0 ? validCompletionTypes[0] : '';
		}
		return completionType;
	}
```

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -53,7 +53,7 @@ class AssignmentEditorSubmissionAndCompletion extends ActivityEditorMixin(Locali
 		}
 
 		return html`
-			${assignment.submissionTypeOptions.map(option => html`<option value=${option.value} ?selected=${option.selected}>${option.title}</option>`)}
+			${assignment.submissionTypeOptions.map(option => html`<option value=${option.value} ?selected=${option.value === assignment.submissionType}>${option.title}</option>`)}
```
ppaskaris-d2l: So this one is a string?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -64,7 +64,7 @@ class AssignmentEditorSubmissionAndCompletion extends ActivityEditorMixin(Locali
 	_getCompletionTypeOptions(assignment) {
 		const completionTypeOptions = assignment ? assignment.completionTypeOptions : [];
 		return html`
-			${completionTypeOptions.map(option => html`<option value=${option.value} ?selected=${option.selected}>${option.title}</option>`)}
+			${completionTypeOptions.map(option => html`<option value=${option.value} ?selected=${option.value.toString() === assignment.completionType.toString()}>${option.title}</option>`)}
```
ppaskaris-d2l: Consider:

```
var needle = String(assignment.completionType);
/* other stuff */ ?selected="${String(option.value) === needle}"> /* other stuff */
```

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -20,6 +20,35 @@ export class Assignment {
 		return this;
 	}
 
+	_getValidCompletionTypes(currentSubmissionType) {
+		return this.submissionTypeOptions.find(
+			submissionType => submissionType.value.toString() === currentSubmissionType.toString()
```
ppaskaris-d2l: Consider extracting `currentSubmissionType.toString()` as `String(currentSubmissionType)` into a local const.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -20,6 +20,35 @@ export class Assignment {
...
+	}
+
+	_isCompletionTypeValid(completionType, validCompletionTypes) {
+		return validCompletionTypes.some(validCompletionType => completionType.toString() === validCompletionType.toString());
+	}
+
+	_setValidCompletionTypeForSubmissionType() {
+		const validCompletionTypes = this._getValidCompletionTypes(this.submissionType);
+
+		if (validCompletionTypes !== null && validCompletionTypes.length > 0) {
```
ppaskaris-d2l: Consider `!validCompletionTypes` to account for `undefined` as well. It sneaks up on you in reckless languages like JavaScript.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -20,6 +20,35 @@ export class Assignment {
...
+			this.canEditCompletionType = true;
```
ppaskaris-d2l: Justify this with a comment, e.g. "We know it must have been true to for this method to have been called" if that's an accurate statement (I assume it is).

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
pureooze: Good catch, its a number on first load but then it becomes a string if the selection is changed.

ppaskaris-d2l: Consider coercing it in the store. 100% avoid types changing due to spooky action at a distance imo.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -20,6 +20,38 @@ export class Assignment {
...
+	_isCompletionTypeValid(completionTypeId, validCompletionTypes) {
+		const completionType = String(completionTypeId);
+		return validCompletionTypes.some(validCompletionType => validCompletionType.toString() === completionType);
+	}
+
+	_setValidCompletionTypeForSubmissionType() {
+		const validCompletionTypes = this._getValidCompletionTypes(this.submissionType);
+
+		if (validCompletionTypes && validCompletionTypes.length > 0) {
+			// we allow editing completion types only if the submission type has more than one valid completion type
```
kieranderson: should you be checking >1 if you need more than one valid completion type?

mdgbayly: I think maybe it is just the comment that is misleading? As I think we'd still want to render the drop down even if there is only one valid completion type? Which we don't actually have an example of right now. It's more that we only want to show the completion type dropdown if there is at least one valid completion type.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
mdgbayly: I agree all the conversion from number to string is super confusing and leads to complex looking code.
As mentioned elsewhere, it was unfortunate that when they originally wrote the API they used numeric values rather than strings, as the UI/API doesn't really care that the option value is a number. It's just an identifier.

So, I was wondering whether we should clean this up by converting all the values to strings at the time we generate the valid submission/completion options. Then in the UI we would just deal with strings.

That could possibly even be done in the siren-sdk, given that nothing else apart from FACE is likely to be using that stuff. If we feel that is already too risky, then maybe we could just do it in the Assignment MobX state class.

E.g. similar to where in the `load(entity)` you co-erce the submission/completion type value to a string, also iterate the options we get from the siren sdk class and convert them into local representations that use strings for each value instead. Then initialize the UI with that representation of the options, so the UI only knows about strings.

Then when we save and call back the siren-sdk with the values we want to save we convert back to numeric at that point if necessary ( i think it already co-erces it back to a number anyway)

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -59,9 +107,14 @@ decorate(Assignment, {
 	canEditInstructions: observable,
 	instructionsRichTextEditorConfig: observable,
 	activityUsageHref: observable,
+	completionTypeOptions: observable,
```
mdgbayly: Probably want to make the `submissionType` and `completionType` observable too for completeness, although not strictly necessary to make it work.

mdgbayly:
@pureooze Generally this is looking pretty good. Thanks for taking the MobX-ification for this on!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
pureooze: Yeah even implementing it I found it a bit confusing. The tests helped me find a lot of the issues that were not immediately obvious. I think we could revisit it but maybe we can do that in a separate PR? This one is starting to have a lot of churn on it and its getting hard to track what changes are actually going through. I'd like to also discuss it with @ppaskaris-d2l but he won't be back until next week.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
pureooze: Yeah I'm going to reword the comment. I've also realized the `canEditCompletionType` actually works a bit differently that I thought. I think I shouldnt change its value here because this will make it difficulty to disable the select element when the assignment has submissions. Instead I think I want to just leave it as is when we first fetch the values.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
mdgbayly: yep - i'm totally fine with getting things working - and then improving/refactoring in follow up PRs.

ppaskaris-d2l: I originally advised against doing it in the siren-sdk because the need for it to be a string is strictly an issue with the DOM, but I don't super care. The store is close enough for me, and its indeed how I'm doing it for release conditions.

PR #703, opened 2020-02-20
Integrate validation requests with mobx state 

emil-furniss:
When saving, it will first run the new `validate` method on all editors that implement it. If any validations fail, the LMS will return a 400 with details on the type of error. The error props can be used by components to determine if the error is relevant to them, and which lang term to use for the error tooltip.

Depends on https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/134

The datetime pickers are in flux right now, so I just added some pretty ugly tooltips for now, just to demonstrate that the errors and langterms are working. They should ultimately be replaced by the incoming new datetime input component.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-due-date-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-due-date-editor.js
@@ -110,7 +110,7 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
 		if (!activity || this._isFirstLoad) {
 			dueDate = null;
 			canEditDates = false;
-			errorTerm = null
+			errorTerm = null;
 		} else {
 			dueDate = activity.dueDate;
 			canEditDates = activity.canEditDates;
--- a/components/d2l-activity-editor/d2l-activity-due-date-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-due-date-editor.js
@@ -75,26 +75,6 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
 		`;
 	}
 
-	_getErrorLangterm(errorType) {
-		if (!errorType) {
-			return null;
-		}
-
-		if (errorType.includes('end-due-start-date-error')) {
-			return this.localize('dueBetweenStartEndDate');
-		}
-
-		if (errorType.includes('start-after-end-date-error') || errorType.includes('start-after-due-date-error')) {
-			return this.localize('dueAfterStartDate');
-		}
-
-		if (errorType.includes('end-before-start-date-error') || errorType.includes('end-before-due-date-error')) {
-			return this.localize('dueBeforeEndDate');
-		}
-
-		return null;
-	}
-
 	render() {
 		const activity = store.get(this.href);
 		let dueDate, canEditDates, errorTerm;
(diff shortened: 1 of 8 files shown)


components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -1,28 +1,35 @@
...
+		this._editors.delete(editor);
+	}
+
+	async _save() {
+		const validations = [];
+		for (const editor of this._editors) {
+			validations.push(editor.validate());
+		}
+
+		await Promise.all(validations).then(async() => {
```
mdgbayly: minor nit: I find it a little confusing to mix `await` and `then` in this way. 
I would probably just add a try/catch around the `await Promise.all(...)` and then have the `await editor.save()` loop as the next statement after the `Promise.all()`

But up to you.

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -54,6 +56,29 @@ export class ActivityUsage {
...
+			endDate: this.endDate
+		}).catch(e => runInAction(() => {
+			this.isError = true;
+			if (e.json && e.json.properties && e.json.properties.type) {
+				this.errorType = e.json.properties.type;
+			}
+			throw e;
+		}));
+
+		runInAction(() => {
```
mdgbayly: Minor nit: if we reset the error state before we call validate, then we wouldn't need the `runInAction`. But maybe you decided it was better to not reset the error until after we validate?

components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
emil-furniss: Agreed. `catch`ing also saves us an ugly error message in the console.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -56,16 +58,46 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
...
+		if (!errorType) {
+			return null;
+		}
+
+		if (errorType.includes('end-due-start-date-error')) {
+			return this.localize('dueBetweenStartEndDate');
+		}
+
+		if (errorType.includes('start-after-end-date-error') || errorType.includes('start-after-due-date-error')) {
+			return this.localize('dueAfterStartDate');
```
mdgbayly: Did the designers specifically ask for a due date error when the error is just that the start date is after the end date?
Personally, I think I would have just added this error setting logic to the state class and had three separate date error terms, as it seems like there is some duplication and it might have made it easier to test and removed some of the duplication. Up to you if you want to change.

emil-furniss: Yeah they did. If either start or end dates are in error, the due date should be in error as well - and ideally display errors that are pertinent to the specific type of validation error. We even drew up a chart to list which fields should get which langterms under which circumstance:
https://d2lmail.sharepoint.com/:w:/s/uxteam/Eckt-gDP9JZCsIWU3q1H0YIBlyZhR4we0ePJyijX-9qEow?e=hjAAxH
With that in mind I did try to centralize this logic in the state, but couldn't quite figure out how since each  date editor field has unique conditions.

mdgbayly: Couldn't you just have a switch statement on the different errorTypes and for each errorType set up to 3 error lang terms? Is errorType an array or just a string?

emil-furniss: That's a good idea, let me try!

mdgbayly: i guess its the proverbial six of one, half a dozen of the other, but I find it difficult to even understand how many error states there are with the validation logic spread out like it is.

PR #707, opened 2020-02-21
using Lit styles instead of Polymer mixin

dlockhart:
This is what this looked like before any changes:

Screen Shot 2020-02-21 at 3 02 25 PM

This wasn't actually working -- you were just getting the default `<h2>` styles that the browser ships with. Firstly, `<style>` blocks aren't allowed in CSS literals. Second, mixins are a Polymer thing and don't work in Lit. Instead, you can just import our typography styles into your component and then apply the `d2l-heading-2` CSS class to the element you want to target.

This is what things look like after the change:
Screen Shot 2020-02-21 at 3 10 11 PM


dlockhart:
Visual diff results, which are expected so I'll re-generate goldens:
https://s3.ca-central-1.amazonaws.com/visualdiff.gaudi.d2l/screenshots/BrightspaceHypermediaComponents/activities/no-submissions-text/2020-02-21-20.29.53.309-939a724837dfb92f4f28700896275f470cdf0bff-report.html

PR #708, opened 2020-02-21
Add missing tokens

svanherk:


test/index.html
```diff
@@ -15,63 +15,36 @@
 	<script>
 		// Load and run all tests (.html, .js):
 		WCT.loadSuites([
-			'd2l-activity-list-item/d2l-activity-list-item-test.html?wc-shadydom=true&wc-ce=true',
```
svanherk: FYI @AlexBedley 
Now that shadow dom is on, we don't need to explicitly test shady dom and shadow on every browser... we can just test the default state on that browser.  This should help make activity tests a lot faster!

PR #716, opened 2020-02-26
Part 1: Activity Score/In grades State managment 

mdgbayly:
[US113166](https://rally1.rallydev.com/#/29180338367d/detail/userstory/362987616496)

* Add MobX state class for score and grade
* Connect MobX state to `d2l-activity-score-editor` for rendering state
* Maintain activity score/grades state when UI changes. (e.g Score, Ungraded, In Grades, Not In Grades)

The PR adds a new `ActivityScoreGrade` MobX enabled class for managing the score/in grades state, and the existing `ActivityUsage` MobX state class contains an instance of this new class. This was done this way mostly to avoid making the `ActivityUsage` MobX class a messy dumping ground for all activity properties. The `d2l-activity-score-editor` is still initialized with an activity-usage href, and uses this to get a reference to the `ActivityUsage` for the href, but then it pulls the `ActivityScoreGrade` from that instance and uses that in its rendering and event handlers.

TODO: Hook up activity score/grade to save implementation.


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -0,0 +1,62 @@
...
+
+configureMobx({ enforceActions: 'observed' });
+
+export class ActivityScoreGrade {
+
+	constructor(entity) {
+		this.scoreOutOf = entity.scoreOutOf().toString();
+		this.inGrades = entity.inGrades();
+		this.gradeType = (entity.gradeType() || 'Points').toLowerCase();
+		this.isUngraded = !this.inGrades && !this._scoreOutOf;
```
emil-furniss: `this._scoreOutOf` should probably be `this.scoreOutOf`, without the underscore?

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -33,6 +34,7 @@ export class ActivityUsage {
 		this.dueDateErrorTerm = null;
 		this.startDateErrorTerm = null;
 		this.endDateErrorTerm = null;
+		this.scoreAndGrade = new ActivityScoreGrade(entity);
```
emil-furniss: I like this pattern. It makes me want to extract some of the date and error lang term stuff as well.

mdgbayly: Yeh - i think as we end up putting more stuff in here, it makes it a bit saner to refactor some of it into helper classes.

components/d2l-activity-editor/state/activity-score-grade.js
mdgbayly: Yeh - i'm just looking into how the heck my unit test was passing with that!

mdgbayly: added more tests to catch it and fixed!

PR #719, opened 2020-02-27
US113175 Display alert on validation error

emil-furniss:
Adds a new generic `d2l-activity-editor-alert` component, which is toggled by `ActivityUsage.isError`.
When instantiated by `AssignmentEditorDetail`, we also give it the assignment-specific langterm.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-editor-alert.js
+++ b/components/d2l-activity-editor/d2l-activity-editor-alert.js
@@ -1,0 +1,43 @@
+import 'd2l-alert';
+import { css, html } from 'lit-element/lit-element.js';
+import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
+import { MobxLitElement } from '@adobe/lit-mobx';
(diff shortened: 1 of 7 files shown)


components/d2l-activity-editor/d2l-activity-editor-alert.js
```diff
@@ -0,0 +1,40 @@
...
+			 * Text to be displayed in the alert
+			 */
+			text: { type: String }
+		};
+	}
+
+	static get styles() {
+		return css`
+			.alert-container {
+				padding-bottom: 10px;
```
emil-furniss: Don't like having to set styling in the component. I would like to set this from within `AssignmentEditorDetail` but that component has no knowledge of `ActivityUsage.isError`, and by extension no knowledge if the alert is being displayed or not.

mdgbayly: I think that's fine.
minor nit: and not worth changing, but for adding styling like the padding, you could probably just style the alert and skip the div?

PR #720, opened 2020-02-27
Update secondary panel w/ styled summarizer + background color

stevenao:
Screenshot: https://user-images.githubusercontent.com/3785788/75496655-ae175880-5976-11ea-85e6-ac82ccf155eb.png

@pureooze: note the header and summary classes are used in summarizers (d2l-activity-assignment-availability-editor.js , d2l-activity-assignment-editor-submission-and-completion.js, and d2l-activity-assignment-evaluation-editor.js).

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -4,6 +4,8 @@ import '../d2l-activity-release-conditions-editor.js';
 import '@brightspace-ui-labs/accordion/accordion-collapse.js';
 import { bodySmallStyles, heading4Styles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html } from 'lit-element/lit-element.js';
+import { summarizerHeaderStyles, summarizerSummaryStyles } from './activity-summarizer-styles.js';
+
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { getLocalizeResources } from '../localization.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
@@ -45,13 +47,9 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 				.d2l-body-small {
 					margin: 0 0 0.3rem 0;
 				}
-
-				.summary {
-					list-style: none;
-					padding-left: 0.2rem;
-					color: var(--d2l-color-galena);
-				}
-			`
+			`,
+			summarizerHeaderStyles,
+			summarizerSummaryStyles,
 		];
 	}
 
@@ -127,13 +125,13 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 	render() {
 		return html`
 			<d2l-labs-accordion-collapse flex header-border ?opened=${this._errorInAccordion}>
-				<h4 class="header" slot="header">
+				<h4 class="d2l-heading-4 activity-summarizer-header" slot="header">
 					${this.localize('hdrAvailability')}
 				</h4>
-				<ul class="summary" slot="summary">
-					${this._renderAvailabilityDatesSummary()}
-					${this._renderReleaseConditionSummary()}
-					${this._renderSpecialAccessSummary()}
+				<ul class="d2l-body-small activity-summarizer-summary" slot="summary">
+					<li>${this._renderAvailabilityDatesSummary()}</li>
+					<li>${this._renderReleaseConditionSummary()}</li>
+					<li>${this._renderSpecialAccessSummary()}</li>
 				</ul>
 				${this._renderAvailabilityDatesEditor()}
 				${this._renderReleaseConditionEditor()}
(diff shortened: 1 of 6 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -1,183 +1,193 @@
-import '@brightspace-ui-labs/accordion/accordion-collapse.js';
```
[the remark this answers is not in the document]
mdgbayly: It's because it was originally created with `CRLF` line endings, and Steve has probably updated it to be `LF` which is our standard. This has caught me out a few times too as my editor (VS Code) seems to default to creating new files with `CRLF` even though we have an editor config with `end-of-line=lf`. Need to work out how to get the editor to respect that!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -133,12 +142,14 @@ class AssignmentEditor extends ActivityEditorContainerMixin(ActivityEditorMixin(
...
-					href="${assignmentHref}"
-					.token="${this.token}"
-					slot="secondary"
-					class="d2l-activity-assignment-editor-secondary-panel">
-				</d2l-activity-assignment-editor-secondary>
+				<div slot="secondary">
+					<d2l-activity-assignment-editor-secondary
+						href="${assignmentHref}"
+						.token="${this.token}"
+						slot="secondary"
```
mdgbayly: nit: If we are changing the slotted element to the div should we be removing the `slot` attribute here.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -42,10 +42,14 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(LitElement) {
 				.d2l-body-small {
 					margin: 0 0 0.3rem 0;
 				}
-
+				.header {
```
mdgbayly: nit: we seem to have these header and summary styles duplicated in quite a few places now.
Could we move them into a shared stylesheet and import them into each component using the summarizer.

Also, and this can probably wait, the styles that we're applying to the editor components to get the padding, rounded corners, background color etc, will eventually need to be used for other kinds of activities too, e.g. when we do discussions etc, so at that point it will be nice to extract those too so they are reusable. But less clear what is the best way to do that, so fine with deferring until we need it.

dbatiste: Is this really an accordion, or is this simply some expand/collapse?  The screenshot doesn't really look like an accordion to me.  We have some vanilla [expand/collapse](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-expand-collapse.js) logic that I/we could port to core.  It is not tied to Polymer in any way. Accordion carries with it some semantics that may or may not be what we want.  In the LMS we do have the concept of simple expand/collapse (aside from accordion), and this is something we may want to share - both logic and styles.  I am willing to help with this if need-be.

The other thing I wanted to mention is that we aren't free from shady-DOM/shadyCSS yet, meaning that this rendered DOM is potential target of CSS in the outer page, and these class names are **super** generic.  It would not be the first time that such DOM elements have been unexpected targeted by such CSS.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -1,183 +1,193 @@
...
+	}
+
+	render() {
+		return html`
+            <d2l-labs-accordion-collapse class="accordion" flex header-border>
+				<h4 class="header d2l-heading-4" slot="header">
+					${this.localize('submissionCompletionAndCategorization')}
+				</h4>
+				<ul class="summary d2l-body-small" slot="summary">
+					${this._renderAssignmentTypeSummary()}
```
dbatiste: It looks like this doesn't render valid HTML.  The only valid child of `ul` and `ol` is `li`.  Is it really a list, semantically speaking?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
stevenao: > The accordion is a graphical control element comprising a vertically stacked list of items, such as labels or thumbnails. ... An accordion is similar in purpose to a tabbed interface, a list of items where exactly one item is expanded into a panel (i.e. list items are shortcuts to access separate panels).

From Wiki, our UI fits the description visually. However, behavioral wise, we allow more than one item can be expanded at the same time. Personally, I feel that is a bug. I prefer a less clutter secondary panel with just 1 item expanded at a given time. (I will run the idea with Joseph).

I agree. The class names are not unique enough.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
stevenao: I question the same thing. Unless the original author has a reason for me not to change it, I will change all of them in other places to a `div`.  @pureooze may be you know something about it?

stevenao: https://files.slack.com/files-pri/T025EV233-FUBDF45CK/image.png

The summary should be in a list.. so, a `li` tag is probably missing.

pureooze: Yeah these summary methods return nothing at the moment because we are still working on getting the summary stuff working. I recently merged some changes to get MobX working with this component so we can now start replacing these methods.

For example this is what it will look like (note its still a WIP): https://github.com/BrightspaceHypermediaComponents/activities/blob/US112499/summarizer-for-groups/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js#L88-L95

PR #721, opened 2020-02-28
US113175 Open availability dates accordion on error

emil-furniss:
This wires the availability dates editor component with the accordion into the activity usage state, so that it can check if any errors that pertain to the accordion are set.
In this case it cares about the start and end dates, which are known to be in error state when the respective `*DateErrorTerm` property is truthy.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -115,7 +115,7 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 	}
 
 	// Returns true if any error states relevant to this accordion are set
-	get _openOnError() {
+	get _errorInAccordion() {
 		const activity = store.get(this.href);
 		if (!activity) {
 			return false;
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -111,10 +114,19 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(LitElement) {
...
+			return false;
+		}
+
+		return activity.endDateErrorTerm || activity.startDateErrorTerm;
+	}
 
+	render() {
 		return html`
-			<d2l-labs-accordion-collapse flex header-border>
+			<d2l-labs-accordion-collapse flex header-border ?opened=${this._errorInAccordion}>
```
ppaskaris-d2l: Won't this close the accordion when the error has been cleared?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -111,10 +114,19 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(LitElement) {
...
 
-	render() {
+	// Returns true if any error states relevant to this accordion are set
+	get _errorInAccordion() {
+		const activity = store.get(this.href);
+		if (!activity) {
+			return false;
+		}
+
+		return activity.endDateErrorTerm || activity.startDateErrorTerm;
```
ppaskaris-d2l: So the return type is `Bool | String`?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
emil-furniss: You're right, it will. That's currently not really an issue since the error is only cleared on save & close. But it def feels wrong to have that effect, so I'll make sure to rework the logic in another PR.
Also, if we ever had client-side errors that clear without saving, it would become a problem.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
emil-furniss: Good point - it's truthy, but not `true`. Will address that as well.

PR #722, opened 2020-02-28
Part 2: Activity Score/In Grades - Refactor validation to support Save/Cancel

mdgbayly:
[US113166](https://rally1.rallydev.com/#/29180338367d/detail/userstory/362987616496)

This builds on 
https://github.com/BrightspaceHypermediaComponents/activities/pull/716

It moves the validation from the web component to the MobX state class so that it can also be called when the Save button is clicked. Some of the validation is still done immediately when the state of the score out of field changes. e.g. if someone enters in a non-numeric value, or if someone clears the score out of field when 'In Grades' is selected. But we also need to do the validation when Save is clicked in case they have not resolved the errors, or for the scenario where they select 'Graded' but never actually enter a score out of.

The MobX state class actions for setting the score out of and the various grade states now set/clear a validation error observable to the appropriate lang term, and the `d2l-activity-score-editor` UI component's render method observes that error property, and includes the tooltip when necessary.

There is also a new `ActivityScoreGrade.validate` method that runs the validation on demand, eg. when Save is clicked. This hooks into the existing validation that had been added for the dates.
It's slightly different to the dates validation in that currently the validation is implemented only in the MobX state class `ActivityScoreGrade` and the `ActivityUsage` state class calls `ActivityScoreGrade.validate()` as part of it's own `validate` implementation.

If the score/grade validation fails, it sets the `ActivityUsage.isError` property that was previously only being used for dates. As discussed yesterday with @emil-furniss , at some point we may need to work out how to integrate Assignments error handling with this Activity error handling.

The existing `ActivityUsage` validation delegates to the siren-sdk `ActivityUsageEntity` to do it's validation because it needs to make an API call.

So I find it a little bit inconsistent that the validation for score/grade is handled by the `ActivityScoreGrade` state class, but when it comes to saving, we delegate the save of the activity score and grade to the siren-sdk `ActivityUsageEntity` too.

As mentioned in the PR, the validation was originally done in the UI component which is why I've moved it to the MobX state class, but it might have been better to have been put into the `ActivityUsageEntity`, so that we could call `ActivityUsageEntity.validate` for both dates and score/grade making it more consistent with how save is implemented. But that would have required us to come up with some way to return all the validation errors and then work out how to map those back to the individual error observables required by the different state classes. `ActivityScoreGrade` needs to track error state for its fields independently of the error state fields being used to manage the date errors. So having it all handled by the state class makes that easier.

I suggest we stick with this for now and monitor going forward.

@emil-furniss As you mentioned, it's nice having the `ActivityUsage` state broken up between different classes and it would be nice at some point to consider refactoring the date stuff into a helper class and that might give us some more insight into how to manage the aggregate save/validation behavior.

### DIFF
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -6,6 +6,8 @@ export class ActivityScoreGrade {
 
 	constructor(entity) {
 		this.scoreOutOf = entity.scoreOutOf().toString();
+		this.scoreOutOfError = null;
+
 		this.inGrades = entity.inGrades();
 		this.gradeType = (entity.gradeType() || 'Points').toLowerCase();
 		this.isUngraded = !this.inGrades && !this.scoreOutOf;
@@ -16,12 +18,14 @@ export class ActivityScoreGrade {
 
 	setScoreOutOf(value) {
 		this.scoreOutOf = value;
+		this.scoreOutOfError = null;
+		this.validate();
 	}
 
 	setUngraded() {
-		this.scoreOutOf = '';
 		this.inGrades = false;
 		this.isUngraded = true;
+		this.setScoreOutOf('');
 	}
 
 	setGraded() {
@@ -31,16 +35,35 @@ export class ActivityScoreGrade {
 
 	removeFromGrades() {
 		this.inGrades = false;
+		if (this.scoreOutOfError === 'emptyScoreOutOfError') {
+			this.scoreOutOfError = null;
+		}
 	}
 
 	addToGrades() {
 		this.inGrades = true;
+	}
+
+	validate() {
+		// This validation was hardcoded in the original UI implementation.
+		// It might have been better to come up with a way to represent this in the Siren representation
+		// to avoid duplicating business logic, but just copying the rules to the MobX state for now.
+		if (this.inGrades && (this.scoreOutOf || '').trim().length === 0) {
+			this.scoreOutOfError = 'emptyScoreOutOfError';
+		}
+
+		if (this.scoreOutOf && this.scoreOutOf.length !== 0 &&
+			(isNaN(this.scoreOutOf) || this.scoreOutOf < 0.01 || this.scoreOutOf > 9999999999)) {
+			this.scoreOutOfError = 'invalidScoreOutOfError';
+		}
+		return !this.scoreOutOfError;
 	}
 }
 
 decorate(ActivityScoreGrade, {
 	// props
 	scoreOutOf: observable,
+	scoreOutOfError: observable,
 	inGrades: observable,
 	gradeType: observable,
 	isUngraded: observable,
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -125,8 +129,11 @@ export class ActivityUsage {
 				this.errorType = e.json.properties.type;
 				this.setErrorLangTerms(this.errorType);
 			}
-			throw e;
 		}));
+
+		if (this.isError) {
+			throw new Error('Activity Usage validation failed');
```
mdgbayly: With the way validation is handled, the specifics of this error are not really relevant. Previously the dates validation was throwing the specific date error, but the error is ignored by the activity editor container mixin that controls the save/validation workflow, so i think it's fine here to just throw a generic error so that a failed validation allows the save to be skipped.

test/d2l-activity-editor/state/activity-score-grade.spec.js
```diff
@@ -2,6 +2,17 @@ import { ActivityScoreGrade} from '../../../components/d2l-activity-editor/state
 import { expect } from 'chai';
 import { when } from 'mobx';
 
+function catchErrors(done, callback) {
```
mdgbayly: I noticed that when I started adding expectations to the reactive mobx tests (the ones using `when`), that if an expectation failed, the expectation wasn't reported. It just caused the test to timeout.

I'd initially picked up this pattern of testing reactions from this MobX github discussion:
https://github.com/mobxjs/mobx/issues/494#issuecomment-262810772

In there it also talks about issues with failed expectations causing the jest runner to fail. There is a link to a long discussion about this in Jest.
https://github.com/facebook/jest/issues/2059

Apparently the crashing issue has been fixed, but I was still finding it didn't report the failed expectation correctly (instead just times out), which some others on that jest thread also mentioned.

So this pattern allows you to explicitly fail the test with the failed expectation, which means it is reported much more nicely in the console when running tests. 

I should probably have put this in a helper file, but the next person who wants to use this can do that.

components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -31,16 +35,35 @@ export class ActivityScoreGrade {
...
+		// to avoid duplicating business logic, but just copying the rules to the MobX state for now.
+		if (this.inGrades && (this.scoreOutOf || '').trim().length === 0) {
+			this.scoreOutOfError = 'emptyScoreOutOfError';
+		}
+
+		if (this.scoreOutOf && this.scoreOutOf.length !== 0 &&
+			(isNaN(this.scoreOutOf) || this.scoreOutOf < 0.01 || this.scoreOutOf > 9999999999)) {
+			this.scoreOutOfError = 'invalidScoreOutOfError';
+		}
+		return !this.scoreOutOfError;
```
wongvincent: whawt's the value of `scoreOutOfError` before we negate it?

mdgbayly: it's set to null initially and whenever we update the `scoreOutOf`,so this is really just saying return true if `scoreOutOfError` is false i.e. has no value? Is that confusing? I kind of agree, that it makes me double check it to understand what it is doing.

PR #725, opened 2020-02-28
Make alert full width

emil-furniss:


components/d2l-activity-editor/d2l-activity-editor-alert.js
```diff
@@ -17,8 +17,9 @@ class ActivityEditorAlert extends ActivityEditorMixin(MobxLitElement) {
 
 	static get styles() {
 		return css`
-			.alert-container {
-				padding-bottom: 10px;
+			d2l-alert {
+				margin-bottom: 10px;
+				max-width: 100%;
```
stevenao: i rarely see usage of `max-width: 100%`, but I see a lot of `width: 100%`. Are you using it to constraint the max size of d2l-alert to not go over its parent container?

emil-furniss: No, it's just overriding the max-width that is set in the component. `width: 100%` is already set (https://github.com/BrightspaceUI/alert/blob/master/d2l-alert.js#L33)

PR #726, opened 2020-03-02
Fix submission and completion types saving

pureooze:
Fixes a bug where the submission and completion types were not being saved correctly

Depends on https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/140


pureooze:
This change produces the same behavior as this part of the siren-sdk when we used auto save https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L386-L390

My hope is to update the API to not always require the completion type and submission type but this might break stuff for other people so need to evaluate that. For now this works.

PR #727, opened 2020-03-02
Extract activity dates mobx state from activity-usage

emil-furniss:
Adds a new `ActivityDates` class that encapsulates most of the dates-related functions and properties.

Relies on https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/141

Also implements focusing on first element in error after a validation error.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -121,7 +121,7 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 			return false;
 		}
 
-		return activity.endDateErrorTerm || activity.startDateErrorTerm;
+		return activity.dates.endDateErrorTerm || activity.dates.startDateErrorTerm;
 	}
 
 	render() {
--- a/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
@@ -29,37 +29,37 @@ class ActivityAvailabilityDatesEditor extends (ActivityEditorMixin(LocalizeMixin
 	}
 
 	_onStartDatetimePickerDatetimeCleared() {
-		store.get(this.href).setStartDate('');
+		store.get(this.href).dates.setStartDate('');
 	}
 
 	_onStartDatetimePickerDatetimeChanged(e) {
-		store.get(this.href).setStartDate(e.detail.toISOString());
+		store.get(this.href).dates.setStartDate(e.detail.toISOString());
 	}
 
 	_onEndDatetimePickerDatetimeCleared() {
-		store.get(this.href).setEndDate('');
+		store.get(this.href).dates.setEndDate('');
 	}
 
 	_onEndDatetimePickerDatetimeChanged(e) {
-		store.get(this.href).setEndDate(e.detail.toISOString());
+		store.get(this.href).dates.setEndDate(e.detail.toISOString());
 	}
 
 	render() {
-		const activity = store.get(this.href);
+		const dates = store.get(this.href) ? store.get(this.href).dates : null;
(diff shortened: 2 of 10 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -121,7 +121,7 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 			return false;
 		}
 
-		return activity.endDateErrorTerm || activity.startDateErrorTerm;
+		return activity.dates.endDateErrorTerm || activity.dates.startDateErrorTerm;
```
mdgbayly: @ppaskaris-d2l raised good questions on the earlier PR regarding the return value of this method. It's actually returning strings not a boolean.
He also raised a good question on whether clearing the error will close the accordion?

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -57,59 +40,22 @@ export class ActivityUsage {
...
-			this.dueDateErrorTerm = 'dueBeforeEndDate';
-			this.startDateErrorTerm = null;
-			this.endDateErrorTerm = 'dueBeforeEndDate';
-			return;
-		}
-
-		this.dueDateErrorTerm = null;
-		this.startDateErrorTerm = null;
-		this.endDateErrorTerm = null;
+		this.dates.setErrorLangTerms(errorType);
```
mdgbayly: OK with this for now, but it still sort of assumes that any ActivityUsage validation error is a dates error, and it's likely that if we had more API validation happening, that we'd want to reflect all errors in the UI, so we'd need to come up with a way to associated any errors with the correct part of the state. Also I don't think `this.errorType` is actually used anywhere?

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -29,37 +29,37 @@ class ActivityAvailabilityDatesEditor extends (ActivityEditorMixin(LocalizeMixin
...
 	}
 
 	_onEndDatetimePickerDatetimeChanged(e) {
-		store.get(this.href).setEndDate(e.detail.toISOString());
+		store.get(this.href).dates.setEndDate(e.detail.toISOString());
 	}
 
 	render() {
-		const activity = store.get(this.href);
+		const dates = store.get(this.href) ? store.get(this.href).dates : null;
```
ppaskaris-d2l: consider:

```
const dates = entity ? entity.dates : null;
```

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -68,7 +68,7 @@ class ActivityDueDateEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitEle
 	}
 
 	render() {
-		const activity = store.get(this.href);
+		const dates = store.get(this.href) ? store.get(this.href).dates : null;
```
ppaskaris-d2l: same as above, don't call store.get twice

components/d2l-activity-editor/state/activity-usage.js
ppaskaris-d2l: the dates guy is filtering the error, presumably the strategy here is "visitor-style"

emil-furniss: My thinking was that the `_entity.validate` should stay in one piece, so we can only divest some of the logic. In this case the dates need to set lang terms based on the errorType, so we hook them into it here. Other sub classes can do the same if they need to.
And no, I guess `this.errorType` is no longer used by anything.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
emil-furniss: Responded to those comments in situ, will address them in a separate PR.

PR #728, opened 2020-03-03
Fix availability dates editor accordion closing on error clear

emil-furniss:
Previously the accordion would open on error, but then close when the error is cleared. Now it will leave it open when the error is cleared.

To do this we need to know if the accordion is currently open: Either track the open state in a prop via the accordion state change events, or check the dom via `this.shadowRoot.querySelector('d2l-labs-accordion-collapse').opened`. I went for the former, using `this._opened`.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -63,16 +63,6 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
 		this._opened = false;
 	}
 
-	connectedCallback() {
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -57,6 +58,25 @@ class ActivityAssignmentAvailabilityEditor extends LocalizeMixin(ActivityEditorM
...
 	}
 
+	constructor() {
+		super();
+		this._opened = false;
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('d2l-labs-accordion-collapse-state-changed', this._onAccordionStateChange);
```
ppaskaris-d2l: `state` is an internal property that you should not rely on.

PR #732, opened 2020-03-04
Vwong/us111000 render grade candidates

wongvincent:
Will add tests for the MobX files at the end since I am still working on this story and there will be more changes to these MobX objects.


components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate.js
```diff
@@ -0,0 +1,62 @@
...
+		const gradeCandidate = new GradeCandidate(gradeCandidateEntity, this.token);
+		await gradeCandidate.fetch();
+		return gradeCandidate;
+	}
+
+	async load(entity) {
+		this._entity = entity;
+		this.name = entity.name();
+		this.isCategory = this.gradeCandidateEntity.isCategory();
+		if (!this.isCategory) {
```
mdgbayly: Looks good - looking at this now, I'm wondering if it would have been cleaner to have a different MobX state class for `GradeCandidateGrade` and `GradeCandidateCategory`. (possibly even for the siren-sdk stuff too), but this works.

mdgbayly:
All the grade candidate/category/grade stuff is definitely a bit of a head spin, so i can see why we kept getting confused about how to represent things, but I think this looks pretty good. Let's see what Emil thinks as he is coming into it with fresh eyes.

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate.js
wongvincent: Had the same thought actually. Might end up doing it if I have some more time later.

wongvincent:
> All the grade candidate/category/grade stuff is definitely a bit of a head spin, so i can see why we kept getting confused about how to represent things, but I think this looks pretty good. Let's see what Emil thinks as he is coming into it with fresh eyes.

@emil-furniss 👁 👁

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-candidate-selector.js
```diff
@@ -24,36 +23,53 @@ class ActivityGradeCandidateSelector extends EntityMixinLit(LitElement) {
...
 		return html`
 			<select
 				id="grade-candidates"
 				class="d2l-input-select"
 				@change="${this._setSelected}"
 			>
-				${repeat(this._gradeCandidateEntities, entity => entity, (entity, index) => html`
-					<option value="${index}">${entity.name()}</option>
-				`)}
+				${this._renderGradeCandidateTemplates(gradeCandidates)}
```
emil-furniss: Loving these chained renderers - very clean!

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate.js
emil-furniss: Wouldn't that just move the logic to distinguish between the two into the `grade-candidate-collection`? But it does looks like it might be less logic.
Where are the `baseWeight` and `maxPoints` being used?

wongvincent: They will be used by the `d2l-activity-grade-candidate-selector` to display the points and weight below the dropdown.
image

wongvincent: I was thinking `GradeCandidateCollection` -- > `GradeCandidate` --> `GradeCandidateGrade` or `GradeCandidateCategory`.
So that the logic stays in `GradeCandidate` But I'm not going to worry about it for now.

PR #733, opened 2020-03-04
Added summarizer for groups and individual submission

pureooze:
[Screenshot of summarizer closed](https://user-images.githubusercontent.com/2337524/76351340-aabf8d80-62e3-11ea-961c-0ae1fb9fb128.png)

[Screenshot of summarizer opened](https://user-images.githubusercontent.com/2337524/76351366-b7dc7c80-62e3-11ea-871c-a06ae118e593.png)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -10,7 +11,7 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
 import { shared as store } from './state/assignment-store.js';
 
-class AssignmentEditorSubmissionAndCompletion extends ActivityEditorMixin(LocalizeMixin(MobxLitElement)) {
+class ActivityAssignmentSubmissionAndCompletionEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitElement)) {
```
pureooze: Making the naming consistent with the other editors

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/en.js
```diff
@@ -28,7 +28,11 @@ export default {
 	"txtIndividual": "Individual Assignment", // Label for individual assignment type
 	"txtGroup": "Group Assignment", // Label for group assignment type,
 	"txtGroupCategoryWithName": "Group Category: {groupCategory}", //Label for the group category {groupCategory} is the name of the group category
-	"txtGroupCategory": "Group Category", // Label for group category
+	"txtGroupCategory": "Group Category", // Label for group category,
+	"txtGroupAssignmentSummary": "Group assignment", // Summary message for accordion when assignment type is set to group
```
pureooze: Sorry about the weird placement, there were new terms added to master that got put below this one

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -86,7 +88,12 @@ class AssignmentEditorSubmissionAndCompletion extends ActivityEditorMixin(Locali
 	}
 
 	_renderAssignmentTypeSummary() {
-		return html``;
+		return html`
+			<d2l-activity-assignment-type-summary-editor
```
ppaskaris-d2l: nit: I like `-summary` and `-editor` but not `-summary-editor`

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -73,111 +60,102 @@ class AssignmentTypeEditor extends SaveStatusMixin(RtlMixin(EntityMixinLit(Local
...
-		this.wrapSaveAction(super._entity.setAssignmentTypeGroupCategory(event.target.value));
+		if (!isIndividualAssignmentType) {
+			return this.localize('folderTypeCreateGroups');
+		}
+
+		return;
 	}
 
 	render() {
+		const assignment = store.getAssignment(this.href);
```
ppaskaris-d2l: Consider rendering an empty template if assignment is falsey.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -73,111 +60,102 @@ class AssignmentTypeEditor extends SaveStatusMixin(RtlMixin(EntityMixinLit(Local
...
+		const isIndividualType = assignment ? assignment.isIndividualAssignmentType : true;
```
ppaskaris-d2l: Then you don't need all these conditions, and there will be a natural place to add the "skeleton" thing when it comes time.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -1,27 +1,14 @@
...
-			_folderTypeText: { type: String },
-			_groupCategories: { type: Array },
-			_groupTypeDisabled: { type: Boolean },
-			_groupTypeText: { type: String },
-			_infoText: { type: String },
-			_isIndividualType: { type: Boolean },
-			_isReadOnly: { type: String }
-		};
-	}
+class AssignmentTypeEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitElement)) {
```
ppaskaris-d2l: Note: I think this works only because other elements are triggering the "fetch" for the assignment.

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/en.js
ppaskaris-d2l: I add my terms at a random line. It is the best way to minimize merge conflicts, as we are using a bad format (unsorted JSON).

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -64,6 +64,16 @@ export class Assignment {
...
 		this.submissionType = String(entity.submissionType().value);
 		this.completionType = String(entity.completionType().value);
+		this.isGroupAssignmentTypeDisabled = entity.isGroupAssignmentTypeDisabled();
+		this.isIndividualAssignmentType = entity.isIndividualAssignmentType();
+		this.groupCategories = entity.getAssignmentTypeGroupCategoryOptions();
+		this.isReadOnly = entity.isAssignmentTypeReadOnly();
+
+		if (!this.isIndividualAssignmentType) {
+			this.selectedGroupCategoryId = String(
+				this.groupCategories.find(category => category.selected === true).value
```
ppaskaris-d2l: Consider making this more defensive by checking the return value of `Array#find` before dereferencing it. We expect this to yield a value, of course, but the product is complicated and the unexpected can happen.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -111,11 +139,17 @@ decorate(Assignment, {
 	canEditCompletionType: observable,
 	submissionType: observable,
 	completionType: observable,
+	isIndividualAssignmentType: observable,
```
ppaskaris-d2l: You're missing two properties here: `isGroupAssignmentTypeDisabled` and `isReadOnly`. While these are not mutable from the perspective of the editor, they do change from `undefined` to having a value asynchronously. Currently, you are rendering everything in one component that references the other properties, so it's propagating through them by proxy.

I think.

pureooze: Interesting. [I actually did have them as observables originally](https://github.com/BrightspaceHypermediaComponents/activities/pull/733/commits/ada471b7090f657616249a6682a8af25a1bcd64c#diff-1adff5808f8c62b73da9a3b0daf9e881L143) but thought we wouldn't need them since we get the value from the API and then never modify it again. 

If they do change from undefined to whatever is fetched and MobX doesn't propagate that change because they aren't observable (I'm not sure if this is the case, will look into it), maybe this is something to be mindful of going forward. 

Also I don't think we should rely on save forcing page navigation, its not too much work to make them observable 😄

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
pureooze: I'm not sure what you mean by this. Do you mean that the store is only populated because something else called fetch for the mixin?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
pureooze: Lol this was a copy paste error 😆

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
ppaskaris-d2l: I think I made a mistake similar to this in the conditions store. I might have to go revisit it.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
ppaskaris-d2l: Yes. If you don't [pass in a store to the constructor](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js#L19-L22), then [it doesn't do a fetch](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js#L80)

pureooze: Huh. So do we need to be calling the constructor in every component that we use this in then? I dont think I did this for the submission and completion type editors 🤔

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js
```diff
@@ -26,3 +26,4 @@ export class AssignmentStore {
 }
 
 export const shared = new AssignmentStore();
+export const { _assignments: assignments } = shared;
```
ppaskaris-d2l: 🤔 are you trying to conflict with me?

😂

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -70,9 +70,11 @@ export class Assignment {
 		this.isReadOnly = entity.isAssignmentTypeReadOnly();
 
 		if (!this.isIndividualAssignmentType) {
-			this.selectedGroupCategoryId = String(
-				this.groupCategories.find(category => category.selected === true).value
-			);
+			const category = this.groupCategories.find(category => category.selected === true);
+
+			if (category) {
+				this.selectedGroupCategoryId = String(category.value);
```
ppaskaris-d2l: Not saying that this _will_ happen, but if `load` is called multiple times, this needs to get "unset" I think. Consider assigning a default value up above, or doing two `else` blocks.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -74,110 +61,110 @@ class AssignmentTypeEditor extends SaveStatusMixin(RtlMixin(EntityMixinLit(Local
...
+
+		const isIndividualType = assignment.isIndividualAssignmentType;
+		const infoText = this._getInformationText(assignment);
+		const isReadOnly = assignment.isReadOnly;
+		const groupTypeDisabled = assignment.isGroupAssignmentTypeDisabled;
+
 		return html`
-			<div ?hidden=${!this._isReadOnly} id="read-only-assignment-type-container">
+			<div ?hidden=${!isReadOnly} id="read-only-assignment-type-container">
 				<div class="d2l-body-standard">${this._folderTypeText}</div>
```
ppaskaris-d2l: Undefined properties: `this._folderTypeText`, `this._groupTypeText`.

pureooze: I really wish javascript would throw errors in these situations 😢

ppaskaris-d2l: Don't you want to `move fast and break things`? :joy:

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -74,110 +61,110 @@ class AssignmentTypeEditor extends SaveStatusMixin(RtlMixin(EntityMixinLit(Local
...
 	}
 
 	render() {
+		const assignment = store.get(this.href);
+
+		if (!assignment) {
+			return html``;
+		}
+
+		const isIndividualType = assignment.isIndividualAssignmentType;
```
ppaskaris-d2l: I'm working on the other story now, where we're doing permissions, and this implementation is totally broken when the instructor doesn't have `AddEditDropbox` permission. Not something you need to fix here, but I'd like to iterate on it before merging the permission fix in the LMS.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -74,110 +61,114 @@ class AssignmentTypeEditor extends SaveStatusMixin(RtlMixin(EntityMixinLit(Local
...
+		const infoText = this._getInformationText(assignment);
+		const isReadOnly = assignment.isReadOnly;
+		const groupTypeDisabled = assignment.isGroupAssignmentTypeDisabled;
+		const folderTypeText =	isIndividualType ? this.localize('txtIndividual') : this.localize('txtGroup');
```
pureooze: Converted from this: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js#L104-L112

PR #734, opened 2020-03-05
US113175 Refactor save error alert

emil-furniss:
We were previously using a mobx wrapper component `d2l-activity-editor-alert` and the activity state to display the save error alert when needed.

This didn't work for the assignment errors, so this refactoring gets rid of the mobx wrapper component and just uses basic conditional rendering, triggered by the `d2l-activity-editor-container-mixin` (which knows if there are any errors on the page, no matter where they come from).
For the container-mixin to be able to set the error prop, the alert also needed to be moved from `d2l-activity-editor-assignment-editor-detail` to `d2l-activity-assignment-editor.js`.

We do _not_ need to close the alert if a client-side validation field error is cleared.


emil-furniss:
@mdgbayly Will merge this for now, but have made a note to update this when the name field is hooked into the state validation.

PR #738, opened 2020-03-05
Vwong/us111000 track selected grade candidate

wongvincent:
- This tracks what the selected grade candidate is and keeps the GradeCandidate MobX object stored in the `selected` property of `GradeCandidateCollection`.
(Whenever there is a `change` event on the `select` element, it takes the value of the select `option` which is an `href`, and finds the `GradeCandidate` object from the `GradeCandidateCollection` that matches that `href`, and stores that object in the `selected` property of `GradeCandidateCollection`.)

- This also computes what should be the default selected item when the dialog is loaded for the first time.
(It does this by iterating over the list of grade candidates to find the grade candidate that has the `current-association` class, which signals that it is associated to the current assignment. If it is not associated to anything, then the first `GradeCandidate` that is a grade from the `GradeCandidateCollection` is stored in the `selected` property and shown as the default selected item in the `select` dropdown. )


components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
```diff
@@ -35,12 +35,52 @@ export class GradeCandidateCollection {
...
 		this.gradeCandidates = await Promise.all(gradeCandidatePromises);
+		this.selected = this._findCurrentAssociation(this.gradeCandidates) || this._findFirstGradeItemFromCandidates(this.gradeCandidates);
+	}
+
+	setSelected(href) {
+		this.selected = this._findGradeCandidate(href, this.gradeCandidates);
+	}
+
+	_findGradeCandidate(href, gradeCandidates) {
+		for (const gc of gradeCandidates) {
```
mdgbayly: Possibly this could have been made simpler by using `array.find(() => {})` although the nesting makes it a bit more complex, but i'm ok with this.

wongvincent: > Looks good - tests coming in one of those follow up PRs? :-)

yup

PR #739, opened 2020-03-05
Vwong/us111000 display points and weights

wongvincent:
This displays Points and Weights below the grade candidate dropdown.
The weights text is only rendered if the grade candidate has a weight, which it only does if the grade book is a weighted one.
image


components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-candidate-selector.js
```diff
@@ -60,17 +85,26 @@ class ActivityGradeCandidateSelector extends (ActivityEditorMixin(MobxLitElement
...
 
 		const {
-			gradeCandidates
+			gradeCandidates,
+			selected
 		} = collection;
 
+		const formatNumberOptions = { maximumFractionDigits: 2 };
+		const formattedPoints = formatNumber(selected.maxPoints, formatNumberOptions);
+		const formattedWeight = selected.baseWeight !== undefined ? formatPercent(selected.baseWeight / 100, formatNumberOptions) : '';
```
wongvincent: The `baseWeight / 100` here is because `FormatPercent` multiplies the value by 100.

I am also rounding to 2 decimal places as that is what is wanted for now. May be changed to round to more or less digits after more discussion.

PR #741, opened 2020-03-06
Vwong/us111000 update assignment out of

wongvincent:
When the "Edit Grades Link" dialog closes. We check to see if the `selected` item to link to has changed. If it has changed, update the `scoreOutOf` value for the Assignment with the value from the grade item.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -11,6 +11,7 @@ import { bodyCompactStyles, labelStyles } from '@brightspace-ui/core/components/
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
 import { getLocalizeResources } from './localization';
+import { shared as gradeCandidateCollectionStore } from './d2l-activity-grades/state/grade-candidate-collection-store.js';
 import { inputStyles } from '@brightspace-ui/core/components/inputs/input-styles.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -207,14 +208,26 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeMixin(RtlMixin(Mob
...
-		this.wrapSaveAction(selectedGradeCandidate.associateGrade());
+
+		const newGradeSelected = gradeCandidateCollection.selected;
+		if (oldGradeSelected.href === newGradeSelected.href) {
+			return;
+		}
+
+		this._setGraded();
+		if (newGradeSelected.maxPoints !== undefined) {
+			scoreAndGrade.setScoreOutOf(newGradeSelected.maxPoints.toString());
```
mdgbayly: This is probably OK for now, but I think in order to hook up with the save, we may want to be calling something like:

`scoreAndGrade.selectedGrade(newGradeSelected)` here, so that when we get to saving we can manage the saving of the associated grade.

Looking at this again, I'm not completely convinced we have the separation between the grade candidate collection store, and the activity usage/activity usage score and grade aspects of the state correct.

It's sort of similar to the stuff I did with attachments where I had distinct stores for the attachment collection and the attachments, but I was left thinking at the end that i should have made the attachment collection represent the store for the attachments.

Here it sort of feels like maybe the `ActivityScoreGrade` MobX class should have been used as the `store` for the grade candidates. That way they are much more connected.

I'm not entirely sure how that would get setup though. But I'm going to muse on it. But this works for now.

wongvincent: My next task is to call to associate grade on save, which means `scoreAndGrade` will need to know the item that is selected. So I'm probably going to be thinking about this now.. and see what challenges I run into.

PR #746, opened 2020-03-09
Set scoreAndGrade's selected grade candidate and use it when saving a…

wongvincent:
…n activityUsage

### DIFF
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -96,5 +96,5 @@ decorate(ActivityScoreGrade, {
 	removeFromGrades: action,
 	addToGrades: action,
 	validate: action,
-	setSelectedGrade: action
+	setAssociatedGrade: action
 });
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -15,6 +15,7 @@ export class ActivityScoreGrade {
 		this.canSeeGrades = entity.canSeeGrades();
 		this.canEditGrades = entity.canEditGrades();
 		this.associatedGrade = null;
+		this.gradeHref = entity.gradeHref();
 		this.gradeCandidatesHref = entity.gradeCandidatesHref();
 	}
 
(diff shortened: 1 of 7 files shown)


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -37,6 +40,8 @@ export class ActivityScoreGrade {
 
 	removeFromGrades() {
 		this.inGrades = false;
+		this.associatedGrade = null;
```
mdgbayly: Could update tests in `activity-score-grade.spec.js` to account for this.

components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -60,6 +65,15 @@ export class ActivityScoreGrade {
 		}
 		return !this.scoreOutOfError;
 	}
+
+	setAssociatedGrade(gradeCandidate) {
```
mdgbayly: Could update tests in `activity-score-grade.spec.js` to account for this.

wongvincent:
> One change that might make things a little clearer is to give the ActivityScoreGrade.associatedGrade a more descriptive name, as I think it only has a value when the associated grade is changing?

It won't have a value if it is currently associated to something and we don't ever change it to associate to any other grade. If we change it to associate to something else, then back to the original association, then it will have a value.

components/d2l-activity-editor/state/activity-score-grade.js
wongvincent: Today will be a day of writing unit tests. Including for the MobX grade objects I added!

wongvincent:
This visual diff was failing: https://s3.ca-central-1.amazonaws.com/visualdiff.gaudi.d2l/screenshots/BrightspaceHypermediaComponents/activities/action-dismiss-dialog/2020-03-11-06.50.52.750-602ef8483b3d171cb6b643acc2457b3784114508-report.html

components/d2l-activity-editor/state/activity-score-grade.js
mdgbayly: nit: I think we are still missing a test for this action?

wongvincent: next pr :)

wongvincent: https://github.com/BrightspaceHypermediaComponents/activities/pull/763/commits/c8be395cdb56525286c6dee5e504e0867562922d

PR #755, opened 2020-03-11
Cancel: Connect button to navigation

stevenao:


components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -68,4 +81,21 @@ export const ActivityEditorContainerMixin = superclass => class extends supercla
 		this.dispatchEvent(this.saveCompleteEvent);
 	}
 
+	async _cancel() {
+		const hasPendingChanges = Array.from(this._editors).some(editor => editor.hasPendingChanges());
+
+		if (hasPendingChanges) {
+			const dialog = this.shadowRoot.querySelector('d2l-dialog');
```
mdgbayly: nit: I wonder if this selector should be a  bit more specific? Components might include other dialogs.
It also makes me wonder if we should be trying to include more of this shared save/cancel logic in the mixin. We're likely to need the same cancel dialog when we do discussions/quizzes too. So we either take the YAGNI approach and defer until we have a second implementation or we try and do it now. I'm OK to defer given it's already implemented and we are at the end of the iteration, but I think we should be looking for opportunities to not make everything so assignment specific.

There is definitely a lot going on with the interactions between the assignment/activity specific components and it's definitely not always clear on how to seamlessly integrate them. Up to now we've avoided having any overridable/reusable rendering, but I noticed yesterday that the `d2l-dialogs` implementation uses mixins which contribute to the rendering, so that's possibly an approach we could consider within this project too to try and get more reusability.

https://github.com/BrightspaceUI/core/blob/master/components/dialog/dialog-mixin.js#L281

dbatiste: Yeah I think we've done that in a few places to get some reuse.  It adds some complexity, but in the dialog case, it allowed us to roll a couple flavors of dialog (generic and confirm, and likely full page / overlay in the future) with minimal code.  In it, the sub class's `render` calls helper render methods to compose the required DOM.  An alternative approach is to have the mixin define `render`, and call overridden helpers defined by the sub class.  In either case, the mixin should be careful making assumptions about sub class rendering.

components/d2l-activity-editor/d2l-activity-editor-buttons.js
```diff
@@ -1,55 +1,64 @@
...
+
+	connectedCallback() {
+		super.connectedCallback();
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+	}
+
+	async _save() {
```
dbatiste: Nit: Looks like `_save` and `_cancel` don't need to be `async`.

components/d2l-activity-editor/d2l-activity-editor-buttons.js
```diff
@@ -1,55 +1,64 @@
...
+				margin-right: 0;
+			}
+		`;
+	}
+
+	static async getLocalizeResources(langs) {
+		return getLocalizeResources(langs, import.meta.url);
+	}
+
+	connectedCallback() {
```
dbatiste: Nit: should not need these callbacks unless something more is going to be done here.

PR #767, opened 2020-03-12
do not refresh attachments on save to prevent flicker

mdgbayly:
Currently we refetch attachments on save and also clean up removed/temporary attachments.
However, this causes components referencing those attachments to re-render which is causing the flicker. So until we need to support in-place saves, I am removing the extra handling that managed the refetching and attachment cleanup.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
@@ -102,9 +102,6 @@ export class AttachmentCollection {
 			throw new Error('No attachment store configured. Cannot save');
 		}
 
-		const discarded = [];
-		let hasChanged = false;
-
 		for (const href of this.attachments) {
 			// TODO - Should we run these concurrently using an array of promises?
 			// Siren action helper will still serialize them but we could setup the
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
```diff
@@ -102,39 +102,16 @@ export class AttachmentCollection {
...
 				await attachment.save(this._entity);
-				hasChanged = true;
-			}
-
-			if (attachment.creating && attachment.deleted) {
-				discarded.push(href);
-			}
-
-			// Clean up store reference to temporary or deleted attachments
-			if (attachment.creating || attachment.deleted) {
```
wongvincent: Are there any issues if we re-add an attachment we deleted? Would it just re-reference the item in the store?

mdgbayly: Not entirely clear what you mean by 
> re-add an attachment we deleted?

Do you mean via `Undo`, or if you just add an attachment which is the same as an existing attachment that you deleted.

I don't think we do anything currently to prevent you from adding an attachment twice. But they would be using a different internal identifier, so it should not matter.

PR #775, opened 2020-03-16
Remove usage completionTypeOptions method

pureooze:
For https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/157 we need to remove these usages of the completionTypeOptions method.

The variable is used to hold the current visible completionTypeOptions in the UI. So we now populate it using the `_setValidCompletionTypeForSubmissionType` method.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
@@ -23,29 +23,49 @@ export class Assignment {
 	_getValidCompletionTypes(currentSubmissionType) {
 		const selectedSubmissionType = String(currentSubmissionType);
 
-		return this.submissionTypeOptions.find(
+		const submissionType = this.submissionTypeOptions.find(
 			submissionType => submissionType.value.toString() === selectedSubmissionType
-		).completionTypes;
+		);
+
+		if (!submissionType) {
+			return [];
+		}
+
+		return submissionType.completionTypes;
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -76,6 +75,8 @@ export class Assignment {
 		this.isReadOnly = entity.isAssignmentTypeReadOnly();
 		this.selectedGroupCategoryName = entity.getAssignmentTypeSelectedGroupCategoryName();
 
+		this._setValidCompletionTypeForSubmissionType();
```
ppaskaris-d2l: Danger: this can overwrite `this.completionType` if the client-side logic drifts; the server is canon.

ppaskaris-d2l: Aside: there are some lines in `_setValidCompletionTypeForSubmissionType` and `_getValidCompletionTypes` that would benefit from being more defensive.

pureooze: I added some checks in my latest commit that made sense to me, let me know if there are any more checks we should be doing.

ppaskaris-d2l: The assignment thing is fixed now, so it's ok.

PR #798, opened 2020-03-26
Replace usage of PendingContainerMixin with AsyncContainerMixin

emil-furniss:
These two mixins are functionally basically identical, and there has been the desire to retire `PendingContainerMixin` and use `AsyncContainerMixin` from core across the board.

In this case waiting for `d2l-pending-resolved` from `PendingContainerMixin` and then displaying the editor is equivalent to waiting for the `asyncState` from `AsyncContainerMixin` to be `complete`. This also requires the editors to fire `pending-state` events instead of `d2l-pending-state`.


test/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
```diff
@@ -27,6 +27,6 @@ describe('d2l-activity-editor-mixin', function() {
 
 		setTimeout(() => el._fetch(() => Promise.resolve()));
 
-		await oneEvent(el, 'd2l-pending-state');
+		await oneEvent(el, 'pending-state');
```
dbatiste: I have mixed feelings about this test.  While it's great to have the test, it's not super great that it needs to check the event name, which we've basically made an internal detail of core.  If we decide to change the event name for some reason, this test will eventually break, which will be an easy fix, but we wouldn't want this to happen in lots of repos.

dbatiste: Perhaps `AsyncStateEvent` could be imported here, and grab the name from `(new AsyncStateEvent()).type` for the check?  Unless there is a better way to extract the name...

emil-furniss: Fair point. And yeah, I guess that's a decent way of checking.
Alternatively I could go back and expose the `pending-state` type through a static getter on the class as `AsyncStateEvent.type` or something like that.

PR #802, opened 2020-03-27
detect URLs in Instructions field and add as attachment in FACE assignmnts

spandanabaruah:
- doesn't currently name the attachment based on the URL (that will come in the next commits)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -9,6 +9,8 @@ import { css, html } from 'lit-element/lit-element.js';
 
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { AssignmentEntity } from 'siren-sdk/src/activities/assignments/AssignmentEntity.js';
+import { shared as attachmentCollectionStore } from '../d2l-activity-attachments/state/attachment-collections-store.js';
+import { shared as attachmentStore } from '../d2l-activity-attachments/state/attachment-store.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
 import { ErrorHandlingMixin } from '../error-handling-mixin.js';
@@ -19,8 +21,6 @@ import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from '../save-status-mixin.js';
-import { shared as attachmentCollectionStore } from '../d2l-activity-attachments/state/attachment-collections-store.js';
-import { shared as attachmentStore } from '../d2l-activity-attachments/state/attachment-store.js';
 import { shared as store } from './state/assignment-store.js';
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -78,6 +82,23 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		}
 	}
 
+	_addToCollection(attachment) {
```
mdgbayly: minor nit: Does having this as a separate method add much value. I guess most of the time we are just adding a single link, but if we do add multiple links, e.g. through pasting, then it means that we are finding the attachment collection for every link. So we could just load the collection directly in the addLinks method?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -78,6 +82,23 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
...
+		const collection = attachmentCollectionStore.get(this._attachmentsHref);
+		collection.addAttachment(attachment);
+	}
+
+	addLinks(links) {
```
mdgbayly: nit: should be a private method `_addLinks`?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -4,28 +4,33 @@ import '../d2l-activity-due-date-editor.js';
...
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 
 class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMixinLit(LocalizeMixin(RtlMixin(ActivityEditorMixin(MobxLitElement)))))) {
 
 	static get properties() {
 		return {
 			_nameError: { type: String },
-			_attachmentsHref: { type: String }
+			_attachmentsHref: { type: String },
+			_linksProcessor: { type: Object }
```
mdgbayly: nit: doesn't need to be a Lit property. It's just an internal property. We don't expect it's value to affect rendering.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
spandanabaruah: good point! I'll just load the collection in the addLinks method :)

spandanabaruah: Fixed it in the next PR for naming the attachments

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -45,8 +50,6 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 				#score-and-duedate-container {
 					display: flex;
 					flex-wrap: wrap;
-					min-height: 90px;  /* Hack to force a consistent the height for the old */
-					padding-bottom: 0; /* datetime picker. Can hopefully be removed when the new picker is used.*/
 				}
```
jchapmanD2L, line 53: @spandanabaruah was this change intentional?  I think it might have reverted some stuff @emil-furniss did to make the date picker stay in place.

spandanabaruah: oops! - not sure how I ended up deleting those lines. Will add them back.

spandanabaruah: merged the PR - updating activities in BSI now

PR #807, opened 2020-03-31
use unfurl event to set attachment name

spandanabaruah:
also contains some nit fixes from previous PR (https://github.com/BrightspaceHypermediaComponents/activities/pull/802)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -82,15 +82,11 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 		}
 	}
 
-	_addToCollection(attachment) {
-		const collection = attachmentCollectionStore.get(this._attachmentsHref);
-		collection.addAttachment(attachment);
-	}
-
 	addLinks(links) {
+		const collection = attachmentCollectionStore.get(this._attachmentsHref);
 		links = links || [];
 		links.forEach(element => {
-			this._addToCollection(attachmentStore.createLink(element.name, element.url));
+			collection.addAttachment(attachmentStore.createLink(element.name, element.url));
 		});
 	}
 
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachment.js
@@ -18,6 +18,13 @@ class ActivityAttachment extends ActivityEditorMixin(MobxLitElement) {
 		super(store);
 	}
 
+	_onAttachmentUnfurled(e) {
+		const attachment = store.get(this.href);
+		if (!attachment.attachment.name) {
+			attachment.setName(e.detail.title);
+		}
+	}
+
 	_onToggleRemoved() {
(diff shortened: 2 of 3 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -92,7 +92,11 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(EntityMi
 
 	_saveInstructions(value) {
 		store.getAssignment(this.href).setInstructions(value);
-		this._linksProcessor.process(value, linkAttachments => this.addLinks(linkAttachments));
+		this._debounceJobs.value = Debouncer.debounce(
```
mdgbayly, line 95: What behaviour were you seeing that caused you to add this debounce? As it looks like the call to `_saveInstructions` is already debouncing?

spandanabaruah: When I was testing it out locally, I noticed that the URLs from the instructions field were being detected and added as attachments to the FACE assignment before I could finish typing out the full URL in the instructions. For example, I was trying to type "http://www.google.com" but it just added "http://www.go" as an attachment before I could finish typing out the full URL. So my intention was to cause a delay (enough to let the user type out the full URL) before adding the URL as an attachment. 

Is there a better way for me to achieve this behaviour?

mdgbayly: I think you may be seeing this more because of the way the debouncing of the instructions has been set up for FACE and also because of the way event handling has been setup for FACE with the HTML editor.

In the Activity Feed assignments-ui, it looks like the instructions change debounce was set to be 1000ms instead of 500ms. Also, in assignments-ui the change event wasn't even being debounced when the HTML editor was being used. It was only debounced when the user had disabled the HTML editor, which doesn't make total sense to me, as I think when using a regular text area, we still only processed changes onChange rather than onInput (ie. only when the text area lost focus rather than on every keystroke).

https://github.com/Brightspace/assignments-ui/blob/master/d2l-assignment-editor.js#L722
vs.
https://github.com/BrightspaceHypermediaComponents/activities/blob/082336b25e727ddf25d5b2dc894d5cabc2bf7462/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js#L143

In FACE I think the HTML editor was setup to send change events on every keystroke using onUpdate. Normal usage of the HTML editor only has it send change events when tinymce generates an 'undo' point or loses focus.

https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-html-editor.js#L150

I was not keen on that additional onInput event handling. It's not really even getting the input event from tinymce but from the underlying 'contenteditable' which is only supported in some browsers. I think it was originally set up that way for FACE to try and limit missing changes when autosave was being used. But to be honest, now that we don't have autosave, I don't think we really need that anymore, and we can probably go back to only sending change events onChange again, which would mean that we probably don't even need to debounce anything.

Let's chat tomorrow and we can discuss how to go about changing/testing this behaviour if we switch the text editors to just use onChange events.

Of if we do decide that we want to keep the onInput handling then maybe we can just change the existing debouncer to use 1000ms rather than 500ms.

PR #810, opened 2020-04-01
add Files Allowed Per Submission (file submission)

stevenao:


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -97,6 +113,62 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends ActivityEditorMixi
...
+		const assignment = store.getAssignment(this.href);
+		const data = e.target.value;
+		assignment && assignment.setFilesSubmissionLimit(data);
+	}
+
+	_renderAssignmentFilesSubmissionLimit(assignment) {
+		if (!assignment) {
+			return html ``;
+		}
+		const isFileSubmission = assignment.submissionTypeOptions
```
mdgbayly, line 126: Hmm, yuck. This works but highlights again how difficult it is to represent these kind of relationships in Siren responses. 
It's an ongoing discussion in things like JSON schema on how to add support for these kinds of cross field validations, and if that existed yet I'd want to see us moving in that direction maybe to avoid baking these things into the UI. But without that I'm not sure we have much choice.
https://github.com/json-schema-org/json-schema-spec/issues/541

However, I do think things like this should be in the state classes rather than being baked into the UI components.

E.g. in the actions that set/change the submission type we could set state that determines whether the fileSubmissionLimit should be rendered.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -93,6 +93,10 @@ export class Assignment {
 		this.canEditCompletionType = entity.canEditCompletionType();
 		this.submissionType = String(entity.submissionType().value);
 		this.completionType = String(entity.completionType().value);
+
+		this.canEditFilesSubmissionLimit = entity.canEditFilesSubmissionLimit();
```
mdgbayly, line 97: Technically these properties should be observable, which will be required if we switch to managing the submissionType/filesSubmissionLimit in this state class.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -117,6 +121,10 @@ export class Assignment {
 		this._setValidCompletionTypeForSubmissionType();
 	}
 
+	setFilesSubmissionLimit(value) {
```
mdgbayly, line 124: Technically this should be decorated as an action which will be required once you make the `filesSubmissionLimit` observable.

PR #811, opened 2020-04-01
DE38379  Disable assignment type when user does not have edit permission

pureooze:
Continuing work from https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/171

Disables the assignment type select when the user is missing edit permission or the assignment has submissions


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -120,7 +120,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitElem
 
 		const isIndividualType = assignment.isIndividualAssignmentType;
 		const infoText = this._getInformationText(assignment);
-		const isReadOnly = assignment.isReadOnly;
+		const isReadOnly = assignment.isReadOnly || assignment.hasSubmissions;
```
pureooze, line 123: Unfortunately at the moment we don't have a specific message to display for the user when they are missing permissions 😢

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -96,7 +96,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(LocalizeMixin(MobxLitElem
 
 		const isIndividualAssignmentType = assignment.isIndividualAssignmentType;
 
-		if (assignment.isReadOnly) {
+		if (assignment.hasSubmissions) {
```
kieranderson, line 99: I think you have it set to `assignmentHasSubmissions` instead of just `hasSubmissions`

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
ppaskaris-d2l: Why will it be confusing?

ppaskaris-d2l: Also, doesn't hasSubmissions imply readOnly? Why is this change needed o_o

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
pureooze: ugh this week sucks

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
pureooze: Hmm yeah you are right since readOnly relies on the actions now we don't need this

pureooze: I made this PR before updating the siren-sdk so that would explain this lol

PR #819, opened 2020-04-03
US113746 Add grade category selector in grades dialog

emil-furniss:
Adds a new `ActivityGradeCategorySelector` to the "Create and link to a new grade item" radio option in the `ActivityGradesDialog`.

The new selector gets its data from an instance of `GradeCandidateCollection` which is added to the activity usage as `ScoreAndGrade.newGradeCandidatesCollection`. The collection elements are `GradeCandidate`s using the siren-sdk `GradeCategory`, based on the category link provided in the grade candidate sub-entities.
Since the first grade candidate of the collection does not have a category link, most of the changes to `GradeCandidateCollection` and `GradeCandidate` are to allow the instantiation without an href.

Relies on https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/174


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -29,6 +31,15 @@ export class ActivityScoreGrade {
 		await this.gradeCandidateCollection.fetch();
 	}
 
+	async fetchGradeCategories() {
```
wongvincent, line 34: can we rename this function to `fetchNewGradeCandidates` since it's specific to fetching new grade candidates.

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
```diff
@@ -50,9 +50,20 @@ export class GradeCandidateCollection {
...
+		return this.gradeCandidates.some(gc => gc.isNewGradeCandidateWithCategory());
+	}
+
 	_findGradeCandidate(href, gradeCandidates) {
+		if (!gradeCandidates) {
+			return;
+		}
 		for (const gc of gradeCandidates) {
-			if (href === gc.href) {
+			if (href === gc.href || (typeof gc.href === 'undefined' && href === 'undefined')) {
```
wongvincent, line 66: do we have to specifically check if they're undefined? would `href === gc.href || (!gc.href && !href)` work?

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-category-selector.js
```diff
@@ -0,0 +1,86 @@
...
+		return html`
+			<div id="d2l-activity-grade-category-selector">
+				<label class="d2l-label-text">${this.localize('newGradeItemCategory')}</label>
+				<select
+					id="grade-categories"
+					class="d2l-input-select"
+					@change="${this._setSelected}"
+					>
+					${gradeCandidates.map(gc => html`
+						<option value="${gc.href}" .selected="${selected && gc.href === selected.href}">
```
wongvincent, line 76: is the `gc.href` for the category-less candidate, the string `undefined`?

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
emil-furniss: Half yes, half no. We can check for `!gc.href`, but an undefined href is actually the _string_ `undefined`, so it would evaluate to true.

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-category-selector.js
emil-furniss: yes

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grade-category-selector.js
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	render() {
+		const activity = store.get(this.href);
+
+		if (!activity || !activity.scoreAndGrade.newGradeCandidatesCollection || !activity.scoreAndGrade.newGradeCandidatesCollection.hasNewGradeCandidateWithCategory()) {
+			return html``;
+		}
+
+		this.newGradeCandidatesCollection = activity.scoreAndGrade.newGradeCandidatesCollection;
```
mdgbayly, line 60: It seems a little bit unorthodox to be storing this mobx property as an element property here.
It should probably just fetch it when it needs it?

emil-furniss: It's just needed for the change event handler, so it seemed wasteful to get it again just for that. On the other hand it's probably not a big deal to get again if it's preferred.

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate.js
```diff
@@ -11,14 +11,21 @@ export class GradeCandidate {
...
 		const href = this.gradeCandidateEntity.href();
-		const sirenEntity = await fetchEntity(href, this.token);
+		let sirenEntity;
+		if (href) {
+			sirenEntity = await fetchEntity(href, this.token);
+		}
 		if (sirenEntity) {
 			let entity;
-			if (this.gradeCandidateEntity.isCategory()) {
+			if (this.gradeCandidateEntity.isCategory() || this.gradeCandidateEntity.isNewGradeCandidate()) {
```
mdgbayly, line 28: Not entirely clear to me why we would represent the entity as a `GradeCategoryEntity` for a new grade candidate?
I guess it loads the name property with the name of the category? Are we using that somewhere in the UI?

emil-furniss: Yeah, ultimately the entity represents a category with a category name in the category selector, so that's what I went with:
https://github.com/BrightspaceHypermediaComponents/activities/pull/819/files#diff-ba04341997b40d006e6258c694016841R77

mdgbayly: ah, ok yeh.
I think my initial assumption was that we would be using a grade category selector that was built from the grade categories collection that Vincent had initially created for this use case, before we started to get into the complexity of how all this was going to hang together.  
But I can see how you ended up with building the grade categories selector from our new grade candidates collection instead.
It might have been more slightly more reusable if it only depended on categories and then maybe used events to indicate the chosen grade category.
But this works for now.

PR #821, opened 2020-04-06
US113746 Call action to associate new grade with a category on assignment save

emil-furniss:
Gets the save action from the currently selected entity, and passes it into the data for the save.

Tests are coming.

### DIFF
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -111,6 +111,19 @@ export class ActivityScoreGrade {
 	setNewGradeName(name) {
 		this.newGradeName = name;
 	}
+
+	getAssociateNewGradeAction() {
+		let newGradeCandidateEntity;
+		if (this.newGradeCandidatesCollection && this.newGradeCandidatesCollection.selected) {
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -111,6 +111,19 @@ export class ActivityScoreGrade {
...
 	}
+
+	getAssociateNewGradeAction() {
+		let newGradeCandidateEntity;
+		if (this.newGradeCandidatesCollection && this.newGradeCandidatesCollection.selected) {
+			newGradeCandidateEntity = this.newGradeCandidatesCollection.selected.gradeCandidateEntity;
+		}
+
+		if (!newGradeCandidateEntity) {
+			return;
```
wongvincent, line 122: How are we getting the `save` action on the category-less entity?

emil-furniss: Still like this. The `gradeCandidateEntity` on the `selected` item is entire grade candidate entity incl action. The only difference is a lack of category href, which doesn't matter when getting the action.

wongvincent: Oh, makes sense. For some reason, I was thinking of `gradeCategoryEntity`

PR #824, opened 2020-04-07
Create mobx-enabled state layer for ActivityCollectionEditor

zinabat:
**NOTE**
This is a merge to the `polaris/explicit-save-cancel` feature branch, not `master`.

**Context**
[US114989](https://rally1.rallydev.com/#/detail/userstory/378717782596?fdp=true)

This pull request introduces MobX, a state management library, to the previously existing `ActivityCollectionEditor`. The component now interacts with a new class, `Collection`, which in turn interacts with `siren-sdk`. Adding this new layer enables two things:

1. Separation of complex `siren-sdk`/hypermedia-related logic from the front-end component.
2. Ability to alter a shared state from multiple sub components.

The second point will become important for the explicit save/cancel feature, as changes to the state must be kept track of in order to send the changes in a batch request to the server.

Note that this PR **does not change the user's experience** and should appear nearly identical to before the change.

**Other Important Changes**
* `EntityLitMixin` has been replaced with `MobxMixin`
  - This mixin leverages a pointer to the state and disposes of the siren entity if the state is no longer being referenced
* `LitElement` has been replaced with `MobxLitElement`
  - This is a library which tracks changes to observables in the state and runs an `update()` on the LitElement component if the observable is referenced in `render()`

**Before/After**
Screenshots from before and after should be identical:
**Before**
Screen Shot 2020-04-08 at 1 00 22 PM

**After**
Screen Shot 2020-04-08 at 12 59 46 PM

**Quality**

- New unit tests for `Collection` and `MobxMixin`
- New integration tests for `activity-collection-editor`
- Existing unit tests
- Tested with local BSI and local LE instances

**To Be Added**

- [x] Convert MobxMixin tests to use `open-wc/testing` library helpers


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -28,169 +20,36 @@ import 'd2l-alert/d2l-alert-toast.js';
...
 		this.role = 'main';
 		this._currentDeleteItemName = '';
 		this._dialogOpen = false;
 		this._isLoadingMore = false;
-		this._candidateItemsLoading = false;
-		this._setEntityType(ActivityUsageEntity);
+
+		// any observables in the this state accessed
+		// in render() will trigger updates
+		this._setStateType(Collection);
```
zinabat, line 40: Similar to the old `_setEntityType`, this type is used in a factory in the `MobxMixin`

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -28,169 +20,36 @@ import 'd2l-alert/d2l-alert-toast.js';
...
-				this._candidateLoad = this.getCandidates(this._addExistingAction, null, true);
-			}
-			if (!hasACollection) {
-				this._items = [];
-			}
-			this.ariaBusy = 'false';
-			this._loaded = true;
-		});
-	}
-
```
zinabat, line 143: This logic has been moved and refactored in the `Collection` state class

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -214,31 +73,24 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
 		this._isLoadingMore = true;
 		const lastItem = this.shadowRoot.querySelector('d2l-dialog d2l-list d2l-list-item:last-of-type');
-		await this.getCandidates(this._actionCollectionEntity.getNextAction());
-		await this.updateComplete;
+		await this.getCandidates(this._state._actionCollectionEntity.getNextAction());
 		this._isLoadingMore = false;
-		lastItem.nextElementSibling.focus();
+		await this.updateComplete;
+		// TODO: fix the scrolling to top
+		lastItem.focus();
```
zinabat, line 80: Loading more activities in the add activities dialog now results in the cursor resetting to the top - I could not figure out why in a timely matter.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -317,7 +169,7 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 				grid-template-areas: "." "list" ".";
 				grid-template-columns: 100%;
 				min-height: 500px;
-  				grid-template-rows: auto auto auto;
+				grid-template-rows: auto auto auto;
```
zinabat, line 172: Weird extra spaces were here, caught by the linter.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -469,35 +321,21 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 	}
 
 	render() {
-		if (!this._mainPageLoad) {
-			return null;
-		}
```
zinabat, line 474: These types of values are no longer needed, as we do not need to wait for promises to render the page. Instead, the component will automatically update via the state layer's `.isLoaded` property.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -469,35 +321,21 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
-
-		const learningPathTitle = this._mainPageLoad.then(() => {
-			return html`
-				<h1 class="d2l-heading-1 d2l-activity-collection-title-header">
-					<d2l-labs-edit-in-place size="49" placeholder="${this.localize('untitledLearningPath')}" maxlength="128" value="${this._name}" @change=${this._titleChanged}></d2l-labs-edit-in-place>
-				</h1>
-			`;
-		});
-		const learningPathTitleSketeton = html`
+		const learningPathTitleSkeleton = html`
```
zinabat, line 324: Typo fix.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -469,35 +321,21 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
-					<d2l-labs-edit-in-place size="49" placeholder="${this.localize('enterADescription')}" maxlength="280" value="${this._description}" @change=${this._descriptionChanged}></d2l-labs-edit-in-place>
-				</div>
-			`;
-		});
-		const learningPathDescriptionSketeton = html`
+		const learningPathTitle = this._state.isLoaded ? html`
+				<h1 class="d2l-heading-1 d2l-activity-collection-title-header">
+					<d2l-labs-edit-in-place size="49" placeholder="${this.localize('untitledLearningPath')}" maxlength="128" value="${this._state.name}" @change=${this._titleChanged}></d2l-labs-edit-in-place>
+				</h1>
+		` : learningPathTitleSkeleton;
```
zinabat, line 336: The template parts now load similarly to one another. If the component's state `isLoaded` is false or undefined, the template will render the skeleton. When `isLoaded` is set to true, the component automatically re-renders because `isLoaded` is referenced in these parts.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -506,64 +344,71 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
+			</div>
+		`;
+
+		const learningPathVisibilityToggle = this._state.isLoaded ? html`
+			<d2l-activity-visibility-auto-editor class="d2l-activity-collection-toggle-container" ?disabled="${!this._state.activities.length}" .href="${this.href}" .token="${this.token}"></d2l-activity-visibility-auto-editor>
+			<d2l-button-icon
+				class="d2l-activity-collection-toggle-container-button"
+				?disabled="${!this._state.canEditDraft || this.disabled}"
+				@click="${() => this._state.setIsVisible(!this._state.isVisible)}"
+				icon=${this._state.isVisible ? 'tier1:visibility-show' : 'tier1:visibility-hide'}>
```
zinabat, line 379: Expanded for readability

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -506,64 +344,71 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
 				<div class="d2l-activity-collection-header-content">
 					<div class="d2l-heading-4 d2l-activity-collection-sub-header">${this.localize('editLearningPath')}</div>
 					<div class="d2l-activity-collection-base-info">
 						<div class="d2l-activity-collection-header-col1" style="position: relative">
-							${guard([this._loaded], () => html`
-								${until(learningPathTitle, learningPathTitleSketeton)}
-								${until(learningPathDescription, learningPathDescriptionSketeton)}
+							${guard([this._state.isLoaded], () => html`
+								${learningPathTitle}
+								${learningPathDescription}
```
zinabat, line 411: Waiting for promises no longer necessary.

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -684,21 +527,18 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		this.updateComplete.then(() => {
 			this._currentCandidateElement = this.shadowRoot.querySelector('.d2l-add-activity-dialog d2l-list');
 		});
-		const candidates = this._handleFirstLoad(this._renderCandidateItems.bind(this),
-			() => {
+		const candidates = this._state.candidatesAreLoaded ? this._renderCandidateItems() :
+			(() => {
```
zinabat, line 531: Removing promise dependency

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -735,28 +575,52 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
...
+		await this._state.addActivities(keys);
 	}
 
-	_descriptionChanged(e) {
-		this._specialization.setDescription && this._specialization.setDescription(e.target.value);
+	removeItem(item) {
+		this._reloadOnOpen = true;
+		this._currentDeleteItemName = item.name();
+		this._state.removeActivity(item);
+		this.shadowRoot.querySelector('#delete-succeeded-toast').open = true;
```
zinabat, line 623: This is now a regular handler on the component instead of being attached directly to the item.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+	constructor(href, token) {
+		this._href = href;
+		this._token = token;
+
+		this._loadedImages = [];
+		this._organizationImageChunk = {};
+		this._candidateItemsLoading = false;
+		this._candidateFirstLoad = false;
+		this.activities = [];
+		this.candidates = [];
```
zinabat, line 30: These are new names for `items` and `candidateItems` respectively.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+
+		entityFactory(ActivityUsageEntity, href, token, this._onServerResponse.bind(this));
```
zinabat, line 32: Note that the siren entity is bound to the state here.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+	setIsLoaded(value) {
+		this.isLoaded = value;
+	}
+
+	setName(value) {
+		const oldValue = this.name;
+		this.name = value.trim();
+		if (oldValue !== value) {
+			this._specialization.setName(value);
+		}
```
zinabat, line 184: Prevent a needless change to the entity.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+	_makeState() {
+		if (sharedState) {
+			this._state = sharedState.ref;
+			sharedState.addRef();
+			return;
+		}
+		if (typeof this._stateType !== 'function') {
+			throw Error('State creation failed - state type has no constructor');
+		}
+		this._state = stateFactory(this._stateType, this.href, this.token);
```
zinabat, line 318: This is where the state object is actually created.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	/**
+	 * LitElement lifecycle event. Connects the component to the state
+	 *
+	 * @memberof MobxMixin
+	 */
+	connectedCallback() {
+		super.connectedCallback();
+		this._makeState();
```
zinabat, line 332: Note that the state is only created the first time the component is connected to the DOM

components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
m6jones: This is probably a rendering issue and losing focus. Let mark this as a known regression and make a new story to fix it.

components/d2l-activity-collection-editor/state/Collection.js
m6jones: Much better

components/d2l-activity-collection-editor/state/Collection.js
m6jones: Comment about your comment and haven't gone through the code.
If the component is disconencted and the state is cleaned up. Then re connected then it will remake the state? 

It doesn't make a state if it already made but will make one if it isn't made?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+		addRef: () => ++refCount,
+		removeRef: () => --refCount,
+		refCount: () => refCount,
+		ref: ref,
+	};
+})();
+
+// Global shared state smart pointer. Begins as null so we may dynamically
+// create the state type
+export let sharedState = null;
```
m6jones, line 375: Why is this being exported?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,375 @@
...
+	setCandidatesAreLoaded: action
+});
+
+/**
+ * Handles the connection between the LitElement component and the shared
+ * state smart pointer, as well as cleanup
+ *
+ * @export
+ */
+export const MobxMixin = superclass => class extends superclass {
```
m6jones, line 260: It would be awesome if sharedState wasn't directly handled by this class. 

This is something that should be a smaller change so I'd say this is out of scope and something for a new story.

components/d2l-activity-collection-editor/state/Collection.js
zinabat: my thinking was that we will need to share the same state across multiple components potentially, but right now this isn't being used

components/d2l-activity-collection-editor/state/Collection.js
zinabat: correct

components/d2l-activity-collection-editor/state/Collection.js
zinabat: I was thinking this too, but having trouble wrapping my brain around how to manage a global state without doing this.

AKobets:
could you please provide screenshots with before/after change?

test/d2l-activity-collection-editor/mixins/mobx-mixin-tests.js
```diff
@@ -0,0 +1,60 @@
...
+			component._state.extra = 'Added property';
+			component.href = 'http://1';
+			component.token = 'someToken';
+			await component.updateComplete;
+			expect(component._state.extra).to.be.undefined;
+		});
+
+		it('throws an error when state type is not a class', () => {
+			expect(function() { fixture('mobx-bad'); }).to.throw(Error, 'State creation failed - state type has no constructor');
+		});
```
zinabat, line 27: This test fails and I don't know why. The correct error is through, but Chai is not catching it properly and it's instead an "uncaught exception". Any tips?

test/d2l-activity-collection-editor/mixins/mobx-mixin-tests.js
```diff
@@ -0,0 +1,60 @@
+import { expect, assert } from "@open-wc/testing";
```
AntonBazhal: nit: looks like `assert` is not used

test/d2l-activity-collection-editor/mixins/mobx-mixin-tests.js
```diff
@@ -0,0 +1,60 @@
...
+			expect(component._state.extra).to.be.undefined;
+		});
+
+		it('throws an error when state type is not a class', () => {
+			expect(function() { fixture('mobx-bad'); }).to.throw(Error, 'State creation failed - state type has no constructor');
+		});
+	});
+
+	describe('Multiple components sharing a state', () => {
+		let firstComp, secondComp;
```
AntonBazhal, line 33: It's, probably, a preference thing, but I find defining variables this way to be not the most elegant. Mocha allows you to use `this` for such purposes, like:
```js
beforeEach(function() {
  this.firstComp = fixture('mobx-basic');
  this.secondComp = fixture('mobx-uses-shared');
});

it('changes the state for two components', function() {
  this.firstComp._state.item = true;
  expect(this.secondComp._state.item).to.be.true;
});
```

You need to use `function` for this approach to work, but mocha discourages arrow function use anyways.

test/d2l-activity-collection-editor/mixins/mobx-mixin-tests.js
AntonBazhal: this doc suggests that `ficture` is an async function and returns Promise https://www.npmjs.com/package/@open-wc/testing#chai

test/d2l-activity-collection-editor/mixins/mobx-mixin-tests.js
zinabat: @m6jones had a preference for arrow functions, so I guess it's whichever one we want to commit to RE style. I have no real preference.

zinabat: BrightspaceUI/core uses arrow functions, so to stay consistent we will leave this as is. :)

zinabat:
> this will not return boolean, it will return undefined if one of the properties not set, and value of this.token if both properties are set.

I have changed this so that it will always return boolean. The original LitElement function returns boolean, so might as well stay consistent.

components/d2l-activity-collection-editor/mixins/MobxMixin.js
```diff
@@ -0,0 +1,136 @@
+
+import { dispose } from 'siren-sdk/src/es6/EntityFactory.js';
+/**
+ * Handles the connection between the LitElement component and the shared
+ * state smart pointer, as well as cleanup
+ *
+ * @export
+ */
+export const MobxMixin = superclass => class extends superclass {
+	static get properties() {
```
AKobets, line 10: Where this static method is used? I could not find calls to it in PR.

components/d2l-activity-collection-editor/mixins/MobxMixin.js
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		}
+		if (typeof this._stateType !== 'function') {
+			throw new StateError('State type has no constructor');
+		}
+		this._state = stateFactory(this._stateType, this.href, this.token);
+	}
+
+	_setStateType(type) {
+		this._stateType = type;
```
AKobets, line 76: should it perform validation of `type is a function` here as well?
Is there a reason to not init type in constructor ?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+		this._setSirenProvider({
+			entityFactory: entityFactory,
+			performAction: performSirenAction,
+			createActionCollection: (parent, entity) => new ActionCollectionEntity(parent, entity)
+		});
+		this._sirenProvider.entityFactory(ActivityUsageEntity, href, token, this._onServerResponse.bind(this));
+	}
+
+	/**
+	 * Set the siren provider. Allows for easier dependency injection
```
AKobets, line 41: Is this method for testing only or it has other usages too

test/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -1,15 +1,42 @@
...
 
 describe('d2l-activity-collection-editor', () => {
 	let element;
 
-	beforeEach(async() => {
-		element = fixture('collection-editor');
-		await element.updateComplete;
+	describe('Null Collection', () => {
+		it('Passes A11y aXe tests', async() => {
+			element = await fixture(html`
```
AKobets, line 8: is it running tests for empty collection editor? would it be better test coverage if collection has some activities in it. If yes, then I'm OK to add those tests later.

zinabat: So, this is a test that looks new but was actually adapted from the existing old test. aXe test doesn't pass the non-empty one yet (and never did), so this is out of scope for this PR

components/d2l-activity-collection-editor/state/Collection.js
zinabat: It is for testing only but theoretically could have other usages. This was my "cleanest" way of getting around installing `rewire`, but if you have other suggestions I will take them

components/d2l-activity-collection-editor/mixins/MobxMixin.js
zinabat: It is a LitElement-specific function. Properties specified in this way will trigger the `update` lifecycle event when they are changed. I can add a function header to make this more obvious like I've done elsewhere. https://lit-element.polymer-project.org/guide/properties#declare

components/d2l-activity-collection-editor/mixins/MobxMixin.js
zinabat: Good catch - I thought it might be overkill since this check exists elsewhere but you are right

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+		// in theory this will later send a single "publish" request
+		// to the new 'draft' state API
+		//this._specialization.setName(this.name);
+		//this._specialization.setDescription(this.description);
+		//usage.setDraftStatus(draftStatus)
+	}
+
+	// to be called whenever the user changes an input
+	validate() {
+
```
AKobets, line 199: empty method?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+	/**
+	 * Removes an activity from the collection.
+	 * Sets the visibility to hidden if this results in an empty
+	 * collection
+	 *
+	 * @param {*} activity
+	 */
+	removeActivity(activity) {
+		this._collection.removeItem(activity.itemSelf);
+		if (this.activities.length - 1 === 0) {
```
AKobets, line 277: does not look right. why `this.activities.length - 1 === 0` ?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+			this.setIsVisible(false);
+		}
+	}
+
+	// reorderActivity(activityToMove, activityBefore) {
```
AKobets, line 282: comments cleanup needed.

test/index.html
```diff
@@ -42,9 +42,11 @@
 			'd2l-quick-eval/d2l-quick-eval-clear-on-intial-load.html',
 			'd2l-activity-editor/index.html',
 			'd2l-activity-admin-list/d2l-activity-admin-list.html',
+			'd2l-activity-collection-editor/mixins/mobx-mixin.html',
+			'd2l-activity-collection-editor/state/collection.html',
 			'd2l-activity-collection-editor/d2l-activity-collection-editor.html',
 			'd2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.html',
-			'd2l-quick-eval/behaviors/d2l-quick-eval-refresh-behavior.html'
+			'd2l-quick-eval/behaviors/d2l-quick-eval-refresh-behavior.html',
```
AKobets, line 49: last `,` could be removed

components/d2l-activity-collection-editor/mixins/MobxMixin.js
AKobets: thanks, please add documentation comments

components/d2l-activity-collection-editor/state/Collection.js
AKobets: In constructor it does `this._sirenProvider.entityFactory(ActivityUsageEntity, href, token, this._onServerResponse.bind(this))`
Should it be moved into `_setSirenProvider` ?

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+	 * @param {*} value
+	 * @memberof Collection
+	 */
+	async searchCandidates(value) {
+		const searchAction = this._actionCollectionEntity.getSearchAction();
+		const fields = [{ name: 'collectionSearch', value: value }];
+		await this.fetchCandidates(searchAction, fields, true);
+	}
+
+	save() {
```
AKobets, line 189: If you see value of having non implemented methods, may be it worth to have doc comments with description that the method is not implemented. In IDE it may help someone who is trying to use it

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,311 @@
...
+
+	/**
+	 * Sets the name if the new one is different
+	 *
+	 * @param {*} value New name to set
+	 * @param {*} autosave Save this change to the entity
+	 * @memberof Collection
+	 */
+	setName(value, autosave) {
+		const oldValue = this.name;
```
AKobets, line 232: Is setName could be called with same old value as a side-effect of something else changed?
If setName is only could be called explicitly,  then may be there is no point to add complexity to code and just call `this._specialization.setName(value);` with whatever value.

components/d2l-activity-collection-editor/state/Collection.js
zinabat: In this case, the removal from the `activities` array occurs  when `_onServerResponse` is called. It's likely that this is actually the wrong place for this check but is a result of a previous solution to a defect. When a user deletes the last activity, the collection should be set to hidden.

I can investigate if I can move this code elsewhere.

components/d2l-activity-collection-editor/state/Collection.js
zinabat: The addition of the autosave param I added later prevents this, so you're right - this is not needed

components/d2l-activity-collection-editor/state/Collection.js
zinabat: I've added function headers to these two methods that state they are not implemented - in this case I think these two empty methods are useful skeletons for later work

components/d2l-activity-collection-editor/state/Collection.js
zinabat: I've moved this code to the `setActivities` method. This should also prevent forcing a hidden state if removing an activity fails.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,330 @@
...
+
+	/**
+	 * Callback function which occurs when we receive the entity
+	 * from the server
+	 *
+	 * @param {*} usage
+	 * @param {*} error
+	 */
+	async _onServerResponse(usage, error) {
+		if (error) {
```
m6jones, line 57: Also check if usage is null and fail on both.

components/d2l-activity-collection-editor/state/Collection.js
```diff
@@ -0,0 +1,330 @@
...
+			this.setIsVisible(false);
+		}
+	}
+
+	/**
+	 * Add activities to the collection
+	 *
+	 * @param {*} activityKeys Array of activity keys
+	 */
+	async addActivities(activityKeys) {
```
m6jones, line 284: This isnt refreshing the page. Adding doesn't add components until a refresh.

Note my instance is throwing this error:
image
So maybe it is just my instance broken.

zinabat: I will double check

zinabat: I haven't been able to reproduce this - can you provide more information about the instance so I can make sure we're not missing something?

As an aside, while investigating, I found and fixed a separate regression :)

PR #830, opened 2020-04-13
Update referenced core version

wongvincent:
To pick up latest changes to the primary-secondary template


package.json
```diff
@@ -72,7 +72,7 @@
     "@adobe/lit-mobx": "0.0.3",
     "@brightspace-ui-labs/accordion": "^2.0.2",
     "@brightspace-ui-labs/edit-in-place": "^1.0.11",
-    "@brightspace-ui/core": "^1.39.0",
+    "@brightspace-ui/core": "^1.42.0",
```
emil-furniss, line 75: Shouldn't this just pick up any new minor versions automatically?

wongvincent: oh, true

PR #834, opened 2020-04-14
US115407 Initial Learning Outcomes integration

emil-furniss:
- new `d2l-activity-outcomes` component which can be integrated into a given activity editor. This will render the opener button if there are no alignments yet, or the tags for any existing aligned outcomes.
- we need to respect the lang term set by the config variable `d2l.Languages.Terminology.LearningOutcomes` in the LMS, which allows admins to set the terminology for learning outcomes: "Outcomes" vs "Standards" vs "Objectives" vs "Expectations" vs "Competencies". We do this via a prop that is set on the tag in the LMS renderer, and then access the term when required via the `requestProvider` event handler in `d2l-activity-assignment-editor.js`.
- conditional rendering within `d2l-activity-outcomes` is dependent on a couple of new props in the `ActivityUsage` mobx class: If `alignmentsHref` is defined on the activity usage entity, then outcomes are enabled in general. We fetch the `AlignmentsCollectionEntity` from that href, and that tells us if the user has permissions to modify the alignments (`canUpdateAlignments`) and which, if any, alignments are present (`hasAlignments`).

Required additions to the `siren-sdk`, incl new `AlignmentsCollectionEntity`:
https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/179/files

Passing config lang term to activities renderer in the LMS:
https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/10546/diff

**EDIT:** This work needs to be flagged behind the M3 milestone, so the component now also checks the milestone via the `ActivityEditorFeaturesMixin`. I also added a `hidden` attribute to the component for when it is not being displayed, so we can add/hide the 20px padding in the parent component `AssignmentEditorDetail`. Previously the padding was always there, even if the outcomes component was not showing anything.

https://rally1.rallydev.com/#/29180338367d/detail/userstory/382756597864

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-outcomes.js
+++ b/components/d2l-activity-editor/d2l-activity-outcomes.js
@@ -1,6 +1,7 @@
 import '@brightspace-ui/core/components/button/button.js';
 import '@brightspace-ui/core/components/dialog/dialog.js';
 import 'd2l-activity-alignments/d2l-select-outcomes.js';
+import { ActivityEditorFeaturesMixin, Milestones } from './mixins/d2l-activity-editor-features-mixin.js';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
 import { getLocalizeResources } from './localization';
 import { html } from 'lit-element/lit-element';
@@ -10,10 +11,12 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { shared as store } from './state/activity-store.js';
 
-class ActivityOutcomes extends ActivityEditorMixin(LocalizeMixin(RtlMixin(MobxLitElement))) {
+class ActivityOutcomes extends ActivityEditorFeaturesMixin(ActivityEditorMixin(LocalizeMixin(RtlMixin(MobxLitElement)))) {
 
 	static get properties() {
 		return {
+			hidden: { type: Boolean, reflect: true },
+			_featureEnabled: { type: Boolean },
 			_opened: { type: Boolean },
 			_outcomesTerm: { type: String }
 		};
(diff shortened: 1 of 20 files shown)


components/d2l-activity-editor/d2l-activity-outcomes.js
```diff
@@ -29,20 +32,34 @@ class ActivityOutcomes extends ActivityEditorMixin(LocalizeMixin(RtlMixin(MobxLi
...
 
 		this.addEventListener('d2l-alignment-list-added', this._onDialogAdd);
 		this.addEventListener('d2l-activity-alignment-outcomes-updated', this._onOutcomeTagDeleted);
 		this.addEventListener('d2l-alignment-list-cancelled', this._onDialogCancel);
 	}
 
+	disconnectedCallback() {
+		super.disconnectedCallback();
+
+		this.removeEventListener('d2l-alignment-list-added', this._onDialogAdd);
```
mdgbayly, line 50: nit: ? Are these events coming from the outcomes components? If so could we just add declarative event handlers to those components?

components/d2l-activity-editor/d2l-activity-outcomes.js
```diff
@@ -99,10 +121,12 @@ class ActivityOutcomes extends ActivityEditorMixin(LocalizeMixin(RtlMixin(MobxLi
 		} = activity;
 
 		if (!canUpdateAlignments) {
+			this.hidden = true;
```
mdgbayly, line 124: Does rendering nothing still cause some space to be taken up in the host component? Just wondering why we need to explicitly hide this component?

components/d2l-activity-editor/d2l-activity-outcomes.js
emil-furniss: yes they are, so yes we can! I'll move them over in the next PR.

components/d2l-activity-editor/d2l-activity-outcomes.js
emil-furniss: No, it doesn't take up any space, but we want the parent to be able to apply padding when this component is displaying something. So this is just a way to signal that to the parent. I feel like there might be a much simpler way of doing that, but couldn't find it.

PR #837, opened 2020-04-15
when focus is on 'ungraded' button, screen reader announcement will b…

spandanabaruah:
when focus is on 'ungraded' button, screen reader announcement will be 'add a grade button'
a11y FACE Trello card: https://trello.com/c/DD6GiKEb/4-face-score-out-of-field-is-labeled-as-ungraded-should-it-be-add-a-grade


wongvincent:
nice branch name :D

PR #839, opened 2020-04-16
Vwong/us114943 telemetry save event

wongvincent:
https://rally1.rallydev.com/#/29180338367d/detail/userstory/378525822800

This PR is to track the Save.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -98,8 +98,8 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 	constructor() {
 		super(store);
 
-		this._type = "assignment";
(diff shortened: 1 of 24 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -219,7 +228,9 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 					href="${assignmentHref}"
 					.token="${this.token}"
 					slot="footer"
-					class="d2l-activity-assignment-editor-footer">
+					class="d2l-activity-assignment-editor-footer"
+					.type="${this._type}"
+					.telemetryId="${this._telemetryId}">
```
mdgbayly, line 233: Don't totally get why we have to pass this telemetryId all the way down the hierarchy and then pass it back up in the save event? Couldn't the `ActivityEditorContainerMixin` get the props from this component through abstract/overridable properties?

PR #848, opened 2020-04-28
Clear out the rubric href after editing and update render

kieranderson:
When closing the overlay or using the floating buttons, the href for the current rubric is removed. This along with rendering empty html when there is no href set, fixes the issues we were seeing when switching rubric types.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
+++ b/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
@@ -90,8 +90,10 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 	}
 
 	_closeEditNewAssociationOverlay() {
-		const editNewAssociationOverlay = this.shadowRoot.querySelector('#create-new-association-dialog');
+		const editNewAssociationOverlay =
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -214,11 +232,9 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 			<d2l-simple-overlay
 				id="create-new-association-dialog"
 				no-cancel-on-outside-click
+				@d2l-simple-overlay-close-button-clicked="${this._clearNewRubricHref}"
```
pureooze, line 235: Does this also work if you press `Esc` to close the overlay? (Do we care about users using `Esc`?)

kieranderson: It does not cover ESC. I will look into it.

dbatiste: My current thoughts are that we should replace the internals of `d2l-simple-overlay` (to get this off of Polymer and the iron components) with an implementation of the dialog mixin, which automatically handles ESC and other things.  I know Jeff G. has been looking into usage of the simple overlay very recently.

PR #851, opened 2020-04-29
Set the rubric href when closing the overlay with esc

kieranderson:
Same setting of rubric href, just on using ESC. Follow up from Uzair's comment in last PR.


components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -233,6 +233,7 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 				id="create-new-association-dialog"
 				no-cancel-on-outside-click
 				@d2l-simple-overlay-close-button-clicked="${this._clearNewRubricHref}"
+				@iron-overlay-canceled="${this._clearNewRubricHref}"
```
dbatiste, line 236: Is it possible to wire up to `d2l-simple-overlay-canceled` instead?  That way this code isn't tied to the underlying Polymer iron-overlay implementation.

PR #875, opened 2020-05-07
Mobile view page template to render Right Hand Panel as Bottom Panel 

stevenao:


stevenao:
image

It is the drop shadow at the top of the secondary panel. I tried a few things to add it to the divider in the template, but that would be covered by the panel and disappeared. 

I feel like I should push back on this one. The drop shadow will be gone as soon as the secondary is scroll down.

wongvincent:
Yeah, it will definitely be temporary since the mobile implementation of the template will have some sort of "thick" divider there for the drawer to go up and down.

PR #886, opened 2020-05-13
DE38983 - fix Choose Grades dialog not showing assignment title on fi…

wongvincent:
…rst open

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -155,7 +155,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeMixin(RtlMixin(Mob
 		if ((changedProperties.has('href') || changedProperties.has('token')) &&
 			this.href && this.token) {
 			this.store && this._fetch(() => {
-				let fetch = this.store.fetch(this.href, this.token)
+				const fetch = this.store.fetch(this.href, this.token);
 				fetch.then(() => {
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -152,21 +152,35 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeMixin(RtlMixin(Mob
 	updated(changedProperties) {
 		super.updated(changedProperties);
 
+		if ((changedProperties.has('href') || changedProperties.has('token')) &&
+			this.href && this.token) {
```
emil-furniss, line 156: Rather than duplicating this here, could we add async/await to the `updated` implementation in `ActivityEditorMixin` and then just `await` the return of `super.updated(changedProperties);` here before calling `_setNewGradeName`?
That might be useful elsewhere as well if we ever want to wait for the store fetch to complete before doing other component-specific work.

wongvincent: I gave it a try, but `super` can only appear inside methods and not async functions.
There's also 6 other places in the activities repo that is doing the 
```if ((changedProperties.has('href') || changedProperties.has('token')) &&
			this.href && this.token) ``` 
stuff, so that's where I copied it from.

emil-furniss: Ah well, thanks for trying it out. I know that there's precedence, just wondering if we could clean it up a bit :)

PR #890, opened 2020-05-14
Replaced d2l-dialog for discard changes with d2l-dialog-confirm

pureooze:
For [DE39076](https://rally1.rallydev.com/#/110294864140d/detail/defect/391026725736)

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -5,6 +5,7 @@ import './d2l-activity-assignment-editor-footer.js';
 import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
 import 'd2l-save-status/d2l-save-status.js';
 import '@brightspace-ui/core/components/colors/colors.js';
+import '@brightspace-ui/core/components/dialog/dialog-confirm.js';
 
 import { css, html } from 'lit-element/lit-element.js';
 import { ActivityEditorContainerMixin } from '../mixins/d2l-activity-editor-container-mixin.js';
@@ -248,11 +249,10 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 
 			</d2l-activity-editor>
 
-			<d2l-dialog title-text="${this.localize('discardChangesTitle')}">
-				<span>${this.localize('discardChangesQuestion')}</span>
+			<d2l-dialog-confirm title-text="${this.localize('discardChangesTitle')}" text=${this.localize('discardChangesQuestion')}>
 				<d2l-button slot="footer" primary dialog-action="confirm">${this.localize('yesLabel')}</d2l-button>
 				<d2l-button slot="footer" dialog-action="cancel">${this.localize('noLabel')}</d2l-button>
-			</d2l-dialog>
+			</d2l-dialog-confirm>
 		`;
 	}
 
--- a/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
+++ b/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
@@ -98,17 +98,20 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
 		const hasPendingChanges = Array.from(this._editors).some(editor => editor.hasPendingChanges());
 
 		if (hasPendingChanges) {
-			const dialog = this.shadowRoot.querySelector('d2l-dialog');
-			const action = await dialog.open();
-			if (action === 'cancel') {
-				return;
-			}
-		}
+			const dialog = this.shadowRoot.querySelector('d2l-dialog-confirm');
+			dialog.opened = true;
+			
+			dialog.addEventListener('d2l-dialog-close', async (e) => {
+				if (e.detail.action === 'cancel') {
+					return;
+				}
 
-		if (this.isNew) {
-			await this.delete();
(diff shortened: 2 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -248,11 +249,10 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 
 			</d2l-activity-editor>
 
-			<d2l-dialog title-text="${this.localize('discardChangesTitle')}">
-				<span>${this.localize('discardChangesQuestion')}</span>
+			<d2l-dialog-confirm title-text="${this.localize('discardChangesTitle')}" text=${this.localize('discardChangesQuestion')}>
```
pureooze, line 252: Using the dialog confirm removes the close button on the top right of the dialog and it fixes the issue where pressing escape caused the user to lose their changes.

components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -98,17 +98,20 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
...
-			const dialog = this.shadowRoot.querySelector('d2l-dialog');
-			const action = await dialog.open();
-			if (action === 'cancel') {
-				return;
-			}
-		}
-
-		if (this.isNew) {
-			await this.delete();
+			const dialog = this.shadowRoot.querySelector('d2l-dialog-confirm');
```
pureooze, line 101: see https://github.com/BrightspaceUI/core/tree/master/components/dialog#d2l-dialog-confirm

components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -98,17 +98,26 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
...
+					return;
+				}
+
+				if (this.isNew) {
+					await this.delete();
+				}
+
+				this.dispatchEvent(this.cancelCompleteEvent);
+			});
+		} else {
```
pureooze, line 115: This makes sure that even if there are no pending changes we still fire the cancel event to take the user back to Manage Assignments

dbatiste: Why the change from using the promise?  One of the things to be careful with, if listening for the event, is that if the same dialog is launched a 2nd time, the old handler will still be wire-up unless it is removed each time or the `once` options is specified for the handler wire-up.  The `open` method handles this for you.  Both approaches are valid, but extra care is required if wiring up to the event.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
dbatiste: I am curious to know more about losing changes when the user presses [escape].  The treatment of [escape] should be the same for `d2l-dialog` and `d2l-dialog-confirm`.  I am wondering/concerned that something additional should be done in the dialog component.

components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
pureooze: Ah yeah good point. I tried using it yesterday but I was running into some issues with it. I just tried it though and it work so I'm going to make use of the `open` method, its a lot cleaner. Thanks!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
pureooze: I just tried using the `open` method with the confirm and it looks like the issue is back. So its probably a bug with the code in the `cancel` method in mixin not dealing with the 'abort' case and not an issue with the dialog 🤔

dbatiste: What's the behavior that you are seeing (i.e. how can i repro this)?

pureooze: If you go into manage assignments and edit an assignment (or create one and then edit it) you just need to make a change to a field like the name and then click on "Back to Manage Assignments". If you press `esc` when the dialog pops up the browser will still go back to manage assignments.

This is in FACE so you might need to enable the `d2l.Tools.Dropbox.CreateExperienceOptIn` flag on your instance. Feel free to message me on slack I can show you how to reproduce it.

astadelbauer: If you just want to see it quick and not debug it, go to a quad site, [USA] FACE - USAFACE course ( /d2l/home/132655).  Assignments > Edit  (or New Assignment). Edit a field, hit Cancel. The dialog should open for you, then if you hit X or Esc it loses your changes and exits assignments instead of just closing the dialog.

dbatiste: I am a bit surprised that we don't have the Escape issue either way, since the logic for handling that is shared between them.  It would seem that something else higher in the DOM is also listening for that key, and dialog should be stopping propagation of the event.

Aside, I think that `d2l-dialog-confirm` is more appropriate for this use case anyway.

PR #893, opened 2020-05-19
Unskip qe vdiff tests

AlexBedley:


dbatiste:
Note: there is a helper available in the latest visual-diff lib for creating the page that will emulate the OS preference for reduced motion, which the dialog (and other core components now supports).  Besides having some a11y benefit, this also makes visual-diff testing somewhat easier/faster.

```
page = await visualDiff.createPage(browser, {viewport: {width: 800, height: 600}});
```
The 2nd argument is optional, the default being 800x800.

Waiting for the `d2l-dialog-open` event is still recommended though, due to the async nature of Lit rendering, placement of focus, etc.  Ex. [d2l-dialog-confirm visual-diff test](https://github.com/BrightspaceUI/core/blob/master/components/dialog/test/dialog-confirm.visual-diff.js#L48)

PR #894, opened 2020-05-19
submission and completion stuff

spandanabaruah:
https://trello.com/c/aHg7U2Ua/184-submission-dropdown-should-be-disabled-when-submission-present

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
@@ -92,7 +92,8 @@ export class Assignment {
 		this.canEditSubmissionType = entity.canEditSubmissionType();
 		this.canEditCompletionType = entity.canEditCompletionType();
 		this.submissionType = String(entity.submissionType().value);
-		this.completionType = String(entity.completionType().value);
+		const completionType = entity.completionType();
+		this.completionType = completionType ? String(completionType.value) : null;
 
 		this.canEditSubmissionsRule = entity.canEditSubmissionsRule();
 		this.submissionsRule = entity.submissionsRule() || 'keepall';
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -109,7 +110,12 @@ export class Assignment {
 		this.selectedGroupCategoryName = entity.getAssignmentTypeSelectedGroupCategoryName();
 
 		const validCompletionTypes = this._getValidCompletionTypes(this.submissionType);
-		this.completionTypeOptions = this._getCompletionTypeOptions(validCompletionTypes);
+		if (entity.canEditCompletionType()) {
+			this.completionTypeOptions =  this._getCompletionTypeOptions(validCompletionTypes);
+		} else {
+			const readOnlyCompletionType = entity.completionType();
```
emil-furniss, line 116: This `const readOnlyCompletionType` looks like a duplicate of `const completionType` in line 95?

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -109,7 +110,12 @@ export class Assignment {
...
+			this.completionTypeOptions = readOnlyCompletionType ? [readOnlyCompletionType] : [];
```
emil-furniss, line 117: Are there any concerns around completion types not being valid in this case? `_getCompletionTypeOptions` does a check against `validCompletionTypes`.

wongvincent: good question. Nope, there shouldn't be. When we can't edit the completion type (i.e. read-only), the actions that contain information about the completion type options and the valid completion types for that submission type aren't present. So we use the `completionType` property instead. And that `completionType` property should be valid.

PR #908, opened 2020-05-25
DE39243 - make orgUnitId available to be DI by d2l-labs-attachment

wongvincent:
We replace {orgUnitId} that is present in QuickLink hrefs with the actual orgUnitId.
The orgUnitId is provided by activities.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -152,6 +152,17 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 			return;
 		}
 
+		// Provides orgUnitId for d2l-labs-attachment
+		// https://github.com/Brightspace/attachment/blob/74a66e85f03790aa9f4e6ec5025cd3c62cfb5264/mixins/attachment-mixin.js#L19
+		if (e.detail.key === 'd2l-provider-org-unit-id') {
+			const activity = store.getActivity(this.href);
+			const assignment = activity && store.getAssignment(activity.assignmentHref);
+			const richTextEditorConfig = assignment && assignment.instructionsRichTextEditorConfig;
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -152,6 +152,17 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 			return;
 		}
 
+		// Provides orgUnitId for d2l-labs-attachment
+		// https://github.com/Brightspace/attachment/blob/74a66e85f03790aa9f4e6ec5025cd3c62cfb5264/mixins/attachment-mixin.js#L19
+		if (e.detail.key === 'd2l-provider-org-unit-id') {
+			const activity = store.getActivity(this.href);
+			const assignment = activity && store.getAssignment(activity.assignmentHref);
+			const richTextEditorConfig = assignment && assignment.instructionsRichTextEditorConfig;
```
emil-furniss, line 160: So we're relying on the html editor config to get the orgUnitId for attachments? Is that information still available if the html editor is disabled?

wongvincent: Yes, we're using the html editor config. And yes, it is available.

mdgbayly: yeh - i felt a little bit icky suggesting this as a solution - but the whole rich text editor config is already kind of icky... 
But yeh - good point about whether it's included if rich text editing is off - but i guess we didn't join those dots.

But as the main point is to allow the host component to inject an org unit id, i guess it's up to the host to decide how to get the org unit id.

This is also yet more stuff we're adding to assignments which will have to be hooked up again for the next activity type. But as of yet, we don't really have any way to include the rich text editor in the activity usages as i think it's a child entity of the rich text instructions entity.

Maybe if we find another need for injecting org unit ids, we can consider getting it another way.

PR #917, opened 2020-05-29
visibility toggle text removed in mobile view

spandanabaruah:
https://rally1.rallydev.com/#/29180338367d/detail/userstory/388899934252?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-editor-buttons.js
+++ b/components/d2l-activity-editor/d2l-activity-editor-buttons.js
@@ -8,7 +8,7 @@ class ActivityEditorButtons extends RtlMixin(LocalizeMixin(LitElement)) {
 	static get styles() {
 		return css`
 			:host {
-				display: inline-flex;
+				display: inline-block;
 			}
 			:host([hidden]) {
 				display: none;
@@ -28,7 +28,7 @@ class ActivityEditorButtons extends RtlMixin(LocalizeMixin(LitElement)) {
 					display: none;
 				}
 				.mobile {
-					display: inline-flex;
+					display: inline-block;
 				}
 			}
 		`;
--- a/components/d2l-activity-editor/d2l-activity-editor-buttons.js
+++ b/components/d2l-activity-editor/d2l-activity-editor-buttons.js
@@ -8,7 +8,7 @@ class ActivityEditorButtons extends RtlMixin(LocalizeMixin(LitElement)) {
 	static get styles() {
 		return css`
 			:host {
-				display: inline-block;
+				display: inline-flex;
 			}
 			:host([hidden]) {
 				display: none;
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -67,14 +72,14 @@ class ActivityVisibilityEditorToggle extends LocalizeMixin(LitElement) {
 				>
 					<div class="d2l-label-text">
 						<d2l-icon icon=${icon}></d2l-icon>
-						${switchVisibilityText}
+						<div class="mobile"> ${switchVisibilityText} </div>
```
emil-furniss, line 75: Are there spaces here now?

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -67,14 +72,14 @@ class ActivityVisibilityEditorToggle extends LocalizeMixin(LitElement) {
 				>
 					<div class="d2l-label-text">
 						<d2l-icon icon=${icon}></d2l-icon>
-						${switchVisibilityText}
+						<span class="mobile"> ${switchVisibilityText} </span>
```
emil-furniss, line 75: the extra spaces between the tags and the text are still here, and in the other span below

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
spandanabaruah: Nope I just have tabs to indent the <div class="mobile"> on the same level as the <d2l-icon> element and one level deeper than the <div class="d2l-label-text"> element

components/d2l-activity-editor/d2l-activity-editor-buttons.js
```diff
@@ -8,7 +8,7 @@ class ActivityEditorButtons extends RtlMixin(LocalizeMixin(LitElement)) {
 	static get styles() {
 		return css`
 			:host {
-				display: inline-block;
+				display: inline-flex;
```
emil-furniss, line 11: This might be problematic as it stops the row from ever braking, so the visibility toggle will be pushed out of sight if the screen is too narrow and/or the buttons too wide.

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
spandanabaruah: oops sorry! changing that in a sec

components/d2l-activity-editor/d2l-activity-editor-buttons.js
spandanabaruah: hmm... shall I add another media query here such that if the screen width goes below some specific pixel mark then the footer buttons should be set back to display: inline-block?

PR #920, opened 2020-06-01
remove extra bottom padding

spandanabaruah:
https://trello.com/c/ckeakpjS/196-mobile-reduce-space-between-attachment-bar-and-secondary-panel


emil-furniss:
Any chance we could achieve this by changing the way the padding is being set here: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js#L21
Looks like that is adding extra padding to the bottom of the attachments picker bar and the attachments list component, and I can't see why that would be required.

PR #939, opened 2020-06-08
M2 features in the evaluation accordion should check M2 flag

wongvincent:
https://trello.com/c/7Z8dtPG8/207-ld-flag-turning-off-m2-flag-with-some-m3-feature-flags-on-wont-hide-all-m2-features


wongvincent:
Yes, it's because we want competencies to still be rendered when `m3competencies` is enabled but `m2` is disabled.

PR #943, opened 2020-06-11
US116960 Add saving shim to activity-editor

emil-furniss:
Uses the `d2l-background` component as a shim to obscure the editor while saving is in progress. The shim will enable after 0.8s and fade in over the following 1.2s. It will disable instantly when a validation or save error occurs.

Due to the multitude of editors, the save action is coordinated in `d2l-activity-editor-container-mixin`. This is used in `d2l-activity-assignment-editor`, which in turn embeds the assignment-independent `d2l-activity-editor`. For the backdrop to remain independent of assignments, I added it to `d2l-activity-editor`, but in return need to tell that component when a save action is in progress. For that purpose I added a new prop, `isSaving`, to the `d2l-activity-editor-container-mixin.`, which is passed to `d2l-activity-editor` via the `d2l-activity-assignment-editor`.
We then simply display the backdrop based `isSaving`, with a little extra logic to enforce the 0.8s delay and avoid duplicate timers.

https://rally1.rallydev.com/#/29180338367d/detail/userstory/394462027544

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -31,10 +31,6 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 			 */
 			trustedSitesEndpoint: { type: String },
 			/**
-			 * If there is an error on the page. Is used to toggle the d2l-alert.
-			 */
-			isError: { type: Boolean },
-			/**
 			* based on the LaunchDarkly flag face-assignments-milestone-2
 			*/
 			milestoneTwoEnabled: { type: Boolean },
@@ -247,7 +243,7 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 			<d2l-template-primary-secondary slot="editor" width-type="${this.widthType}">
 				<slot name="editor-nav" slot="header"></slot>
 				<div slot="primary" class="d2l-activity-assignment-editor-primary-panel">
-					${this.isError ? html`<d2l-alert type="error">${this.localize('assignmentSaveError')}</d2l-alert>` : null}
+					<d2l-alert type="error" ?hidden=${!this.isError}>${this.localize('assignmentSaveError')}</d2l-alert>
 					<d2l-activity-assignment-editor-detail
 						href="${assignmentHref}"
 						.token="${this.token}">
@@ -279,7 +275,8 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 				.token=${this.token}
 				unfurlEndpoint="${this.unfurlEndpoint}"
 				trustedSitesEndpoint="${this.trustedSitesEndpoint}"
-				@d2l-request-provider="${this._onRequestProvider}">
+				@d2l-request-provider="${this._onRequestProvider}"
+				?isSaving=${this.isSaving}>
 
 				${this._editorTemplate}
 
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -30,10 +30,6 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 			 * API endpoint for determining whether a domain is trusted
 			 */
 			trustedSitesEndpoint: { type: String },
-			/**
-			 * If there is an error on the page. Is used to toggle the d2l-alert.
-			 */
-			isError: { type: Boolean },
```
emil-furniss, line 36: Cleanup: This doesn't break either way, but this should always have been declared in `d2l-activity-editor-container-mixin.js`. I think this was a remnant from when I was messing around during the original implementation.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -279,7 +275,8 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 				.token=${this.token}
 				unfurlEndpoint="${this.unfurlEndpoint}"
 				trustedSitesEndpoint="${this.trustedSitesEndpoint}"
-				@d2l-request-provider="${this._onRequestProvider}">
+				@d2l-request-provider="${this._onRequestProvider}"
+				?isSaving=${this.isSaving}>
```
mdgbayly, line 279: Does this `isSaving` attribute work with Edge? Just remembering those camel-cased attribute name issues with Edge. 

Not sure what those `unfurlEndpoint` and `trustedSitesEndpoint` attributes are doing, but probably they would have issues too. But I don't see them being used in the component, so maybe they are not an issues.

mdgbayly: I"m still not completely clear i totally grok this camel cased issue with Edge though, we do seem to have some camel cased attributes that work...

components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -45,6 +62,21 @@ class ActivityEditor extends ActivityEditorTelemetryMixin(AsyncContainerMixin(Ac
 		if (changedProperties.has('asyncState') && this.asyncState === asyncStates.complete) {
 			this.logLoadEvent(this.href, this.type, this.telemetryId);
 		}
+		if (changedProperties.has('isSaving')) {
```
mdgbayly, line 65: I'm also a bit unclear on why we are overriding the `update` method here rather than `updated`. Normally we override the `updated` method. The Lit documents imply you should not normally need to override this `update` method, but if you do make sure you call `super` which we are. The main difference between this and `updated` is that property changes in `update` are not supposed to cause a `render`, but i'm assuming this is causing a `render` for you possibly because we are doing the changes in a `setTimeout`, although clearing the timeout when `show` is false is being done directly?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
emil-furniss: Good catch. It did _not_ work on Edge, so I switched it over to an attribute and now it works.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -273,10 +273,10 @@ class AssignmentEditor extends ActivityEditorContainerMixin(LocalizeMixin(Activi
 				telemetryId="${this.telemetryId}"
 				.href=${this.href}
 				.token=${this.token}
+				.isSaving=${this.isSaving}
```
mdgbayly, line 276: I think this works - but it might not be ideal as a solution, because other people who come along to use this component might not realise they have to set that prop directly using the prop setter syntax.

Dave L was recommending that we instead use the lit attribute binding syntax and explicitly defined a hash based attribute name.

https://lit-element.polymer-project.org/guide/properties#observed-attributes

// Observed attribute will be called my-prop
isSaving: { attribute: 'is-saving' }

That way if the client/host uses `isSaving` it will break in all browsers.

But I'm still totally confused about why these attribute names are causing issues in Edge given we seem to have so many camelCased attribute names that don't cause problems.

Just playing around with a sample project to see if I can understand when it breaks.

PR #945, opened 2020-06-12
Fix `d2l-activity-outcomes` taking up space when nothing is rendered

wongvincent:
Missed a case where `this.hidden` should be set to `true`. Without it, `d2l-activity-outcomes` does not have the `hidden` attribute set, and that causes the element to take up space due to padding.
`hidden` needs to be `true` when there are no alignments and we cannot update/edit alignments.
https://trello.com/c/RhZ3MjRM/211-outcomes-tag-takes-up-ui-space-when-not-visible


wongvincent:
Don't merge until after CERT, so that it's easier for CERT to stay on Master instead of branching from before this change.

PR #948, opened 2020-06-17
Add support for Karma

pureooze:
Proof of concept for inspiration, trying to use Karma for our tests. For this PR it only applies to test files ending in `.test.js`. I want to get it working for the Submission and Completion component first to get an idea of how this works but in the next PR I would really like to look into ditching `polymer test` for the other tests.

PR #949, opened 2020-06-19
Fix for failing color-contrast test

pureooze:
Running accessibility tests in https://github.com/BrightspaceHypermediaComponents/activities/pull/948 revealed a failure on an accessibility test case:

```
  d2l-activity-assignment-editor-submission-and-completion
    × passes accessibility test
      Chrome Headless 83.0.4103.106 (Windows 10)
    Error: Accessibility Violations
    ---
    Rule: color-contrast
    Impact: serious
    Elements must have sufficient color contrast (https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=axeAPI)
    
    Issue target: d2l-activity-assignment-editor-submission-and-completion-editor,li:nth-child(2)
    Context: <li><!----><!---->first submission type<!----><!----></li>
    Fix any of the following:
      Element has insufficient color contrast of 3.4 (foreground color: #868c8f, background color: #ffffff, font size: 8.4pt (11.2px), font weight: normal). Expected contrast ratio of 4.5:1
    ---
```

See https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=axeAPI

Before:
image

After:
image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/activity-summarizer-styles.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/activity-summarizer-styles.js
@@ -14,7 +14,7 @@ export const summarizerSummaryStyles = css`
 		padding: 0;
 		margin-top: 5px;
 		min-height: 20px;
-		color: var(--d2l-color-galena);
+		color: var(--d2l-color-tungsten);
 	}
 
 	ul.activity-summarizer-summary > li {


components/d2l-activity-editor/d2l-activity-assignment-editor/activity-summarizer-styles.js
```diff
@@ -14,7 +14,7 @@ export const summarizerSummaryStyles = css`
 		padding: 0;
 		margin-top: 5px;
 		min-height: 20px;
-		color: var(--d2l-color-galena);
+		color: var(--d2l-color-tungsten);
```
kieranderson, line 17: If we set no colour specifically, does the one set with the d2l-dropdown pass the test?

pureooze: So the bug wasnt actually in the dropdown like I originally thought, it was with the summary text for the summarizer which is only styled in activities (since the accordion requires components using it to provide the style for the summary)

pureooze:
@JordanCorner can't thread on github but:

- How did we end up in this state in the first place?
We eyeballed the color and it looked fine to us but we didnt check this type of thing

- Was this preventable without doing this testing here?
Doubt it, we would need automation of some kind to detect these. Can't really expect people to mentally calculate a contrast number in their head 🤣 

- Are there guidelines anywhere that we should have followed?
Hmm good question, I'm not too sure. These tests followed a lot of the stuff in `core` so we can at the very least do similar types of tests in FACE https://github.com/BrightspaceUI/core/blob/092ef3ab931918c2ecf55368343e55b861e7ca71/components/button/test/button.axe.js#L9-L12

PR #951, opened 2020-06-22
Update visibility-editor-toggle to use d2l-switch from core.

dbatiste:
This PR updated the toggle impl to use the new core `d2l-switch-visibility` component from core.

It removes the aria messages - there is further discussion on the text to be used, since previously the text changed when the state of the switch changed - this is not recommended because it is confusing.  If needed, we will apply a new tooltip description (supported by core switch).

### Before
Kapture 2020-06-23 at 11 12 44

### After
Kapture 2020-06-23 at 11 15 27


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
```diff
@@ -28,6 +28,7 @@ class AssignmentEditorFooter extends SaveStatusMixin(EntityMixinLit(RtlMixin(Loc
 				display: inline-block;
 			}
 			.d2l-activity-assignment-editor-footer-left {
+				align-items: baseline;
```
dbatiste, line 31: This fixes vertical alignment - an existing issue.

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -44,46 +37,52 @@ class ActivityVisibilityEditorToggle extends LocalizeMixin(LitElement) {
...
-		// the event when disabled
-		this.addEventListener('click', this._onClick);
+		this._textHidden = false;
 	}
 
-	_onClick(e) {
-		if (!this.switchEnabled) {
-			e.stopImmediatePropagation();
-		}
+	_onChange() {
```
dbatiste, line 43: This will only get dispatched if the switch showing... the above conditional (from previous impl) is not necessary.

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -44,46 +37,52 @@ class ActivityVisibilityEditorToggle extends LocalizeMixin(LitElement) {
...
+		);
 	}
 
 	get switchEnabled() {
 		return this.canEditDraft && !this.disabled;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this._narrowViewportQuery = matchMedia('only screen and (max-width: 615px)');
```
dbatiste, line 55: Contemplating moving this into a reusable helper in core.  Just don't want this repeated in a bunch of places.

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
```diff
@@ -44,46 +37,52 @@ class ActivityVisibilityEditorToggle extends LocalizeMixin(LitElement) {
...
+					text-position="${this._textHidden ? 'hidden' : 'end'}">
+				</d2l-switch-visibility>
+			`;
+		} else {
+			return html`
 				<div class="d2l-label-text">
-					<d2l-icon icon=${icon}></d2l-icon>
-					<span class="mobile">${switchVisibilityText}</span>
+					<d2l-icon icon="${this.isDraft ? 'tier1:visibility-hide' : 'tier1:visibility-show'}"></d2l-icon>
+					<span class="${classMap({'d2l-offscreen': this._textHidden})}">
```
dbatiste, line 79: Previously this was `display: none;` (from the mobile class) which would not have been announced by screen readers.  This needs to be offscreen so that it will be meaningful to non-sighted users.

components/d2l-activity-editor/d2l-activity-visibility-editor-toggle.js
zinabat: I think that's a good idea for the future.

m6jones:
I like the new component. It has a disabled state? 

In ours use case I switch to just a button with the eye. Will there be a small version without the word visibility?

dbatiste:
> It has a disabled state?
Yep, new component does have a disabled state, although it is not used here.  Design could opt into the disabled state instead of not using the switch if they wanted.  The UX would be slightly different.  Either approach would be valid I think.

> Will there be a small version without the word visibility?
The `d2l-switch` and `d2l-switch-visibility` has a property `text-position` with valid values being `start`, `end` (default), and `hidden`.  In this case, this property is toggle depending on the media-query - hiding the next with narrow view-port.

PR #965, opened 2020-07-07
US112250: updating FACE RC styles

astadelbauer:
[US112250](https://rally1.rallydev.com/#/110294864140d/detail/userstory/354235320832?fdp=true)
Note: I don't actually know what's considered "bad practice" these days for CSS, so let me know if anything looks bad...

Updating the styles of the RC pane in FACE to look more similar to the Rubrics pane. Screenshots after the changes:
RCFace
RCFaceSummary

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-conditions-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-conditions-editor.js
@@ -64,32 +64,38 @@ class ActivityConditionsEditor
 			css`
 			.d2l-list-item {
 				display: flex;
-				border: 1px solid var(--d2l-color-gypsum);
-				border-radius: 6px;
 				margin-top: 0.5rem;
 				margin-bottom: 0.5rem;
+				align-items: center;
+			}
+
+			.d2l-list-item-body{
+				flex-grow: 1;
+				border: 1px solid var(--d2l-color-chromite);
+				border-radius: 6px;
+				padding: 12px;
 			}
 
 			.d2l-list-item-decoration {
 				flex: 0 0 auto;
-				padding-top: 18px;
-				padding-left: 15px;
-				padding-right: 18px;
-				padding-bottom: 18px;
+				display: flex;
+				margin-right: 12px;
+				margin-bottom: 2px;
+				float: left;	
 			}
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -64,32 +64,38 @@ class ActivityConditionsEditor
...
-				border: 1px solid var(--d2l-color-gypsum);
-				border-radius: 6px;
 				margin-top: 0.5rem;
 				margin-bottom: 0.5rem;
+				align-items: center;
+			}
+
+			.d2l-list-item-body{
+				flex-grow: 1;
+				border: 1px solid var(--d2l-color-chromite);
```
pureooze, line 74: Given [our past experience with the contrast issues](https://github.com/BrightspaceHypermediaComponents/activities/pull/949), is chromite fine or should we use a [darker shade like tungsten](https://github.com/BrightspaceUI/core/tree/master/components/colors) (which is what the updated summary color is)

pureooze:
>Note: I don't actually know what's considered "bad practice" these days for CSS, so let me know if anything looks bad...

I found a [CSSLinter](https://github.com/CSSLint/csslint) which might help? I'm not super great at CSS either so.... 🤷

components/d2l-activity-editor/d2l-activity-conditions-editor.js
astadelbauer: Chromite is what the rubrics one is currently using (intentionally matching it) and what Patrick requested in the design, so hopefully it's fine?

pureooze: Yeah but is rubrics right? 😆

astadelbauer: OK, Patrick's checking on it but I think I can treat that as a separate PR later where I update whatever else is "chromite" as well if needed.

astadelbauer: update: they decided the contrast doesn't matter much in this case, so we're leaving it for now

PR #968, opened 2020-07-10
Consolidate lang files to one for d2l-activity-editor to fix importing

kieranderson:
[DE39632](https://rally1.rallydev.com/#/detail/defect/400393435268?fdp=true): Convert BSI builds to be unbundled > FACE lang terms fail to load in legacy-Edge

Since the BSI switched to be unbundled, the dynamic loading of the lang files was causing issues for roll up to find the lang files. This was also causing the legacy-edge version to not get the lang files

Solution:
- Move all the `d2l-activity-editor` lang terms into one file, and have one mixin that points to a specific lang  `import('en.js');`
- Namespace all the lang terms in the file so we know where they are being used
- Update all the components to point to the new mixin
- Remove the old `getLocalizeResources` call

This will also reduce our lang calls to 1 from 4, however it's possible we will retrieve more lang terms than needed if some components are not used together.

Translations files incoming, this is just the `en.js` and the components pointing to the new mixin.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
@@ -1,14 +1,13 @@
 import { css, html } from 'lit-element/lit-element.js';
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
-import { getLocalizeResources } from '../localization.js';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
-import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { LocalizeActivityEditor } from '../mixins/d2l-activity-editor-lang-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { assignments as store } from './state/assignment-store.js';
 
 class ActivityAssignmentAnnotationsEditor
-	extends ActivityEditorMixin(RtlMixin(LocalizeMixin(MobxLitElement))) {
+	extends ActivityEditorMixin(RtlMixin(LocalizeActivityEditor(MobxLitElement))) {
 
 	static get styles() {
 
@@ -36,11 +35,6 @@ class ActivityAssignmentAnnotationsEditor
 		];
 	}
 
-	static async getLocalizeResources(langs) {
-
-		return getLocalizeResources(langs, import.meta.url);
-	}
-
 	constructor() {
 
 		super(store);
@@ -66,13 +60,13 @@ class ActivityAssignmentAnnotationsEditor
 
 		return html`
 			<label class="d2l-label-text">
-				${this.localize('annotationTools')}
+				${this.localize('d2l-activity-assignment-editor.annotationTools')}
 			</label>
 			<d2l-input-checkbox
 				@change="${this._toggleAnnotationToolsAvailability}"
 				?checked="${entity.annotationToolsAvailable}"
-				ariaLabel="${this.localize('annotationToolDescription')}">
-				${this.localize('annotationToolDescription')}
+				ariaLabel="${this.localize('d2l-activity-assignment-editor.annotationToolDescription')}">
+				${this.localize('d2l-activity-assignment-editor.annotationToolDescription')}
 			</d2l-input-checkbox>
 		`;
 	}
--- a/components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js
+++ b/components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js
@@ -1,0 +1,65 @@
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export const LocalizeActivityEditor = superclass => class extends LocalizeMixin(superclass) {
+
+	static async getLocalizeResources(langs) {
+		let translations;
+		for await (const lang of langs) {
+			switch (lang) {
+				case 'ar':
+					translations = await import('../lang/ar.js');
+                    break;
+                case 'da-dk':
+					translations = await import('../lang/da-dk.js');
+					break;
+				case 'de':
+					translations = await import('../lang/de.js');
+					break;
+				case 'en':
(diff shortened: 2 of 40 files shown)


components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js
```diff
@@ -0,0 +1,65 @@
...
+export const LocalizeActivityEditor = superclass => class extends LocalizeMixin(superclass) {
+
+	static async getLocalizeResources(langs) {
+		let translations;
+		for await (const lang of langs) {
+			switch (lang) {
+				case 'ar':
+					translations = await import('../lang/ar.js');
+                    break;
+                case 'da-dk':
```
pureooze, line 12: nit: Weird indent here, seems like it switched to spaces?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
```diff
@@ -66,13 +60,13 @@ class ActivityAssignmentAnnotationsEditor
 
 		return html`
 			<label class="d2l-label-text">
-				${this.localize('annotationTools')}
+				${this.localize('d2l-activity-assignment-editor.annotationTools')}
```
dlockhart, line 63: Not a big deal, but since all the lang terms are "activity" related, and I'm not expecting any component will have multiple localized things mixed in, you could likely drop the `d2l-activity` from the lang namespace. Making this particular term `assignment-editor.annotationTools`.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
```diff
@@ -1,14 +1,13 @@
...
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
-import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { LocalizeActivityEditor } from '../mixins/d2l-activity-editor-lang-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { assignments as store } from './state/assignment-store.js';
 
 class ActivityAssignmentAnnotationsEditor
-	extends ActivityEditorMixin(RtlMixin(LocalizeMixin(MobxLitElement))) {
+	extends ActivityEditorMixin(RtlMixin(LocalizeActivityEditor(MobxLitElement))) {
```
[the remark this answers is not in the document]
emil-furniss: Yeah, there are a handful of components that extend one but not the other, and a handful where it's vice versa.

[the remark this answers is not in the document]
dlockhart: Ok cool, so probably better to keep them separate and be explicit about which behaviours to mix in. Thx.

components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js
```diff
@@ -0,0 +1,65 @@
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export const LocalizeActivityEditor = superclass => class extends LocalizeMixin(superclass) {
+
+	static async getLocalizeResources(langs) {
+		let translations;
+		for await (const lang of langs) {
```
mdgbayly, line 7: Does this implementation only return terms for a single language? I think the original implementation would iterate the requested languages and build up the terms such that if the list of langs was something like ['zh-tw', 'zh', 'en'] it would progressively build up the list of terms ensuring that if a term was missing from one lang file it would default to a less specific locale.

Maybe `LocalizeMixin` has support for that now with it's `_getAllLocalizeResources`. Not sure if that is the same thing, but I think we'd still need to return an array of resources.

This was so that language files for a country/region override didn't need to duplicate all the terms.

dlockhart: Correct, this will only return a single language.

I'm pretty sure the way Serge.io is configured, it's going to write all the terms to the provided language files anyway -- it'll never put a subset of terms in `zh-tw` that are the ones that deviate from `zh` for example. The way this is working is also better for performance anyway, since this way users only ever make a single request to get their lang terms, instead of having to build them up with 2-3 requests.

The way we deal with this in core is for new terms to just drop the English values into all the translation files initially. Then when Serge.io kicks in a few weeks later, the pull request overwrites them with the translated values.

dlockhart:
Yeah like if you want many copies of basically the same mixin, but each one just loading a particular component's lang terms that's totally fine -- the main thing I care about is that the strings are static so that Rollup can find them and generate the legacy-Edge versions (which is broken currently).

The reason I suggested combining these into a single language file was because it seemed like all these components are typically always used together anyway. There'll always be a balance that needs to be reached in terms of combining things that are used together into a single file vs. having many many requests.

For core, we decided to combine everything into a single language file. It meant that yes, the user was downloading a few translations for things that weren't on the page, but it also meant that loading the homepage didn't add 30 extra requests. One day if that language file starts getting large, we may end up splitting out a few of the more rarely used components into their own language files, but that's an informed choice we can make at that point.

I think you could do something similar. Once you know which FACE components are truly going to be broken out and reused in isolation, then yeah you could start to break things apart. 🤷

mdgbayly:
> I think you could do something similar. Once you know which FACE components are truly going to be broken out and reused in isolation, then yeah you could start to break things apart. 🤷

Yeh - it's a tough call - i think we know that the assignment stuff is not likely to be used with the upcoming (at some point) quizzing FACE components or discussion FACE components, which was the original reason for splitting out activity and assignments terms. 

But maybe we didn't need to go as far as distinct lang term modules for `attachments`, `grades`, `rubrics`, others if we have them.
I think I was initially working on the lines of splitting tool specific things into their own lang bundle, and then all activity concept related things into a single bundle.

So maybe that would be a reasonably way to slice things now.
Just two sets of lang terms. Assignments and Activity.

But @kieranderson, you have already put a lot of work into this, so your call.

kieranderson:
I'm not opposed to changing anything. If we think that one file is too much I can split it into two.

Does attachments fall into the activity or assignments bucket? Right now it lives outside the assignment-editor folder but it seems rather tied to assignments.

PR #969, opened 2020-07-10
Launch Manage Special Access Dialog and Handle Save/Cancel/Close scen…

wongvincent:
…arios

https://rally1.rallydev.com/#/29180338367d/detail/userstory/397388004180/tasks

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-special-access-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-special-access-editor.js
@@ -1,6 +1,5 @@
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/button/button-subtle.js';
-import '@brightspace-ui/core/components/button/floating-buttons.js';
 import { bodySmallStyles, labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-special-access-editor.js
```diff
@@ -1,6 +1,6 @@
-import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/button/button-subtle.js';
+import '@brightspace-ui/core/components/button/floating-buttons.js';
```
mdgbayly, line 3: is this being used? Is it somehow used by the LMS dialog?

components/d2l-activity-editor/d2l-activity-special-access-editor.js
```diff
@@ -78,11 +78,67 @@ class ActivitySpecialAccessEditor extends ActivityEditorMixin(RtlMixin(LocalizeM
...
+		];
+
+		// Launch into our "friend", the LMS, to do the thing.
+		const delayedResult = D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
+			/*               opener: */ document.body,
+			/*             location: */ location,
+			/*          srcCallback: */ 'SrcCallback',
+			/*       resizeCallback: */ '',
+			/*      responseDataKey: */ 'result',
+			/*                width: */ 1920,
```
mdgbayly, line 120: Would be awesome if just setting a really big width/height forced the dialog into a mobile rendering but I guess that is too much to expect?

components/d2l-activity-editor/d2l-activity-special-access-editor.js
wongvincent: sorry, that was leftover from when I was experimenting with the simple-overlay and adding custom floating buttons to that!

components/d2l-activity-editor/d2l-activity-special-access-editor.js
wongvincent: that's too much to ask! 😆 
we'll figure something out when we get to the fullscreen dialog attribute and styling!

PR #970, opened 2020-07-13
Update slot name to supporting-info

ciss1995:
context:

BrightspaceUI/core#681
This updates the slot name secondary so supporting-info


components/d2l-activity-collection-editor/d2l-activity-collection-editor.js
```diff
@@ -350,7 +350,7 @@ class CollectionEditor extends LocalizeMixin(EntityMixinLit(LitElement)) {
 				margin-left: 0.5rem;
     			align-self: center;
 			}
-			.d2l-list-item-secondary {
+			.d2l-list-item-supporting-info {
```
dbatiste, line 353: Your change here is correct, however it seems like this class ought not to be necessary since [d2l-list-item-content](https://github.com/BrightspaceUI/core/blob/master/components/list/list-item-content.js) specifies `color` and `font-size` at a lower level, via typography [d2l-body-small](https://github.com/BrightspaceUI/core/blob/master/components/typography/typography.js#L49).

So it seems like this olivine color is not doing anything, but would need to confirm, since it would be odd for the original author to leave this here if it were not working.  

Aside from the color, the font-size should already be `14px`, and if we did need to override it for some reason, I think `!important` would actually be necessary in this case, and we'd really need to be specifying `rem` units instead (we should never really be using `px` for `font-size`.

It's fair if you want to just to the name change though, but this appears to need a little cleanup.  `16px` is used immediately above as well - we shouldn't be specifying typography in app components, and also not with absolute units.

PR #971, opened 2020-07-13
Vy/fire event on files attach

vix633:
Hey guys, we need to fire an event on file attach so we can catch it and call save on our end. If this already exists somewhere or there is a better way of doing this, we are definitely looking for that.

For context this is where this is needed: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/39/files#diff-ee6ed211fab74583eb3a6218fa2d49a0L106


AlexBedley:
Once https://github.com/BrightspaceHypermediaComponents/activities/pull/970 is merged, you can pull master in and get the updated goldens. Or if we're ready first we can do it ourselves

vix633:
> Looks like there is still a forgotten `.DS_Store` file in this diff. Otherwise, looks good to me.

This file wont go away

PR #972, opened 2020-07-13
Update lang structure into two lang folders, one for activity-editor and one for activity-assignment-editor

kieranderson:
[DE39632](https://rally1.rallydev.com/#/detail/defect/400393435268?fdp=true): Convert BSI builds to be unbundled > FACE lang terms fail to load in legacy-Edge

Since the BSI switched to be unbundled, the dynamic loading of the lang files was causing issues for roll up to find the lang files. This was also causing the legacy-edge version to not get the lang files

Solution:
- Move all the `d2l-activity-editor` lang terms into one file, and have one mixin that points to a specific lang  `import('en.js');`
- Have a seperate mixin for the activity-assignment-editor.
- Namespace all the lang terms in the file so we know where they are being used
- Update all the components to point to the new mixins
- Remove the old `getLocalizeResources` call

This will also reduce our lang calls to 1 from 2, however it's possible we will retrieve more lang terms than needed if some components are not used together.


kieranderson:
Separate PR based on feedback from: https://github.com/BrightspaceHypermediaComponents/activities/pull/968

PR #973, opened 2020-07-13
Fix outcomes dialog not re-opening after closing via 'X' button

emil-furniss:
The `d2l-dialog-close` event was not being handled, so while clicking the `X` button closes the dialog, the `_opened` prop is never set to false. When the user tries to re-open the dialog, the dialog does not re-open since the prop it thinks that it is already open and the page does not update.

https://trello.com/c/xpi0o2E5/222-cannot-re-open-outcomes-dialog-after-closing-via-x-button


dbatiste:
Note: user can also hit `[Esc]` key, which would also cause this issue.

PR #979, opened 2020-07-15
Vy/fire events on audio video uploads

vix633:
Emitting events here to consistent eval nows when to save the uploaded files.


vix633:
> No issues here.
> 
> Is a future PR coming where we catch and handle these events?

You're on it!

PR #987, opened 2020-07-20
Fix lang term keys based on changes to lang term loading for activity editor

emil-furniss:
With the new lang term loading, terms in files that are using the `d2l-activity-editor-lang-mixin.js` for i18n are referencing [this lang file](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/lang/en.js) (or a respective translation). The terms in that file are prefixed, and some prefixes were missed in the original conversion.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -274,7 +274,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeActivityEditorMixi
 							showing
 							align="start"
 						>
-							${scoreOutOfError ? html`<span>${this.localize(scoreOutOfError)}</span>` : null}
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -274,7 +274,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeActivityEditorMixi
 							showing
 							align="start"
 						>
-							${scoreOutOfError ? html`<span>${this.localize(scoreOutOfError)}</span>` : null}
+							${scoreOutOfError ? html`<span>${this.localize(`editor.${scoreOutOfError}`)}</span>` : null}
```
wongvincent, line 277: I guess the other way to do this would be to change the `scoreOutOfError` property in the MobX object, but I guess this is fine too.

PR #992, opened 2020-07-21
check to see how to go forward before conditions are defined

spandanabaruah:


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -248,6 +249,17 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends ActivityEditorMixi
 		return html``;
 	}
 
+	// _renderSubmissionEmailNotificationSummary() {
```
spandanabaruah, line 252: I've commented all of these out because I'm not yet sure of what the conditions for the rendering are. Would this be dependent on the implementation of US114739 https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F376732834212&fdp=true and US114740 https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F376732844524&fdp=true ?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-submission-email-notification-summary.js
```diff
@@ -0,0 +1,38 @@
...
+			}
+		`;
+	}
+
+	constructor() {
+		super(store);
+	}
+
+	render() {
+		// const assignment = store.get(this.href);
```
spandanabaruah, line 26: Commented this stuff out again here because I'm not sure if something like isSubmissionEmailNotificationEnabled is something that will be defined in previous user stories such as US114739 https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F376732834212&fdp=true and US114740 https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F376732844524&fdp=true (which is what I should use here?)

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -248,6 +249,17 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends ActivityEditorMixi
...
+	// 	if (!this.SubmissionEmailNotificationEnabled) {
```
wongvincent, line 253: I think the this editor part is going to be added by @stevenao , and it's worth checking in with him.

wongvincent: nevermind, this is the summary

wongvincent, line 253: I believe this should show if its enabled and there is a value in the notification email input field.

Should check in with @stevenao to see if the property in the assignment MobX object will be called `SubmissionEmailNotificationEnabled` or something else. 
And this should also only be shown if there is a value inputted into the email input and
Assignment submission Type = File OR Text AND when LD flag for email submission is ON.
I'm not sure if `SubmissionEmailNotificationEnabled` will already take the submissionType into account or not, probably check with Steve.

You'll need to add this to the render function on line 262 of this file and have it conditionally show if the email flag is on. I think more will become clear after Steve adds the notification email editor.

stevenao: const isNotificationEnabled = this._isMilestoneEnabled(Milestones.M4EmailSubmission);
		if (!isNotificationEnabled || !assignment || !assignment.showNotificationEmail) {
			return html ``;
		}

Basically, check the LD flag and there will be a Mobx state..

components/d2l-activity-editor/d2l-activity-assignment-editor/lang/en.js
```diff
@@ -33,6 +33,7 @@ export default {
 	"txtGroupCategoryWithName": "Group Category: {groupCategory}", //Label for the group category {groupCategory} is the name of the group category
 	"txtGroupCategory": "Group Category", // Label for group category,
 	"txtGroupAssignmentSummary": "Group assignment", // Summary message for accordion when assignment type is set to group
+	"txtSubmissionEmailNotificationSummary": "Submission notification  by email", // Summary message when submission notification by email option is enabled
```
wongvincent, line 36: extra space

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-submission-email-notification-summary.js
```diff
@@ -0,0 +1,39 @@
...
+
+	render() {
+
+		const assignment = store.get(this.href);
+		if (assignment && assignment.notificationEmail) {
+			return html`${this.localize('txtSubmissionEmailNotificationSummary')}`;
+		}
+
+		return html``;
+	}
```
wongvincent, line 33: make sure this file is set to LF line endings and the spacing is set to tab size 4

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -235,6 +244,19 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends ActivityEditorFeat
 		assignment && assignment.setNotificationEmail(data);
 	}
 
+	_renderSubmissionEmailNotificationSummary(assignment) {
+		const isNotificationEnabled = this._isMilestoneEnabled(Milestones.M4EmailSubmission);
```
wongvincent, line 248: Would be better to set `this._m4EmailSubmissionEnabled` in the `connectedCallback()` function, similar to what is done in `d2l-activity-assignment-evaluation-editor` and `d2l-activity-assignment-availability-editor`. So that you can just reference `this._m4EmailSubmissionsEnabled` here and in `_renderAssignmentSubmissionNotificationEmail`

stevenao: _isMilestoneEnabled and showNotificationEmail are being called twice. Let's refactor it into a new function.

PR #994, opened 2020-07-22
Kapatel/user/de39250 modify href

patel-kajal:
[DE28769](https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/13680/overview)
Updated some instances of `href` to `.href` to prevent undefined href from being coerced into a string and hopefully stop some chatty logging of the following error:

`The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.`

I updated the href's related to dialog's and all href's that are set using the following:
	- assignmentHRef
	- activityUsageHref
	- entity.associationsHref
	- this._newlyCreatedPotentialAssociationHref
	- rubricHref

I tested these things on Edge and Chrome to make sure nothing breaks:
- the main editor area and opening/editing the grades dialog
- all links on the side bar that open external dialog's (focusing on adding dates, adding a rubric and managing learning objectives)


patel-kajal:
20.20.9.23794

PR #995, opened 2020-07-22
DE39843: announcing when RCs are added or removed from FACE for screenreaders

astadelbauer:
Similar to what I did for Rubrics before... but RCs seems to work much nicer.  Tested in FF and Chrome with NVDA so far, and both seem to work reliably.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-conditions-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-conditions-editor.js
@@ -5,6 +5,7 @@ import 'd2l-dropdown/d2l-dropdown-menu.js';
 import { bodyCompactStyles, bodySmallStyles, labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
+import { announce } from '@brightspace-ui/core/helpers/announce.js';
 import { LocalizeActivityEditorMixin } from './mixins/d2l-activity-editor-lang-mixin.js';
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -181,7 +182,24 @@ class ActivityConditionsEditor
...
+		entity.remove(key);
+
+		if (condition) {
+			// don't want <strong> tags in screenreader text
+			const title = condition.replace(/<strong>|<\/strong>/g, '');
+			announce(`${this.localize('editor.txtConditionRemoved', {title})}`);
+		}
+	}
+
+	_conditions({conditions}) {
```
astadelbauer, line 201: this function seems a bit silly, but I wasn't sure how else to get the conditions off the entity. If you know a better way, please let me know :)

components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -181,7 +182,20 @@ class ActivityConditionsEditor
 			return;
 		}
 
-		entity.remove(event.target.dataset.key);
+		const key = event.target.dataset.key;
+		const conditions = entity.conditions.reduce((map, x) => {
```
kieranderson, line 186: If we end up doing more actions against conditions as a dictionary instead of an array. Would make sense to move this to the entity itself, but for this one case not really needed

astadelbauer: ..no idea why this was erroring on my local... oh well :shrug: thanks for the fix @pureooze  ;)

PR #999, opened 2020-07-23
US111737 remove warning text indicating folder type cannot change whe…

mdulat:
…n assignemnt has submissions

I also removed extraneous entries from `d2l-activity-editor/lang/da-dk.js`, which didn't exist in other lang files (and are repeated from corresponding `d2l-activity-assignment-editor/lang` file.

### DIFF
--- a/components/d2l-activity-editor/lang/da-dk.js
+++ b/components/d2l-activity-editor/lang/da-dk.js
@@ -71,7 +71,7 @@ export default {
 	"rubrics.txtDeleteRubric": "Slet rubrik", // Text for deleting rubric icon
 	"rubrics.btnClose": "Close", // X button for exiting the create new rubric overlay
 	"rubrics.txtRubricAdded": "Rubric added", // Text for notifying screenreader rubric was added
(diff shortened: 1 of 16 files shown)


mdulat:
@astadelbauer I've included you to this PR because it looks like you've done some work with lang files recently, and I want to check with you the changes I've made to `d2l-activity-editor/lang/da-dk.js`. It looks like a remnant from something else, as those entries were only present in that file, not other langs, and also they are a repeat from corresponding `d2l-activity-assignment-editor/lang` file. Please let me know if there is a concern removing those entries, thanks.

components/d2l-activity-editor/lang/da-dk.js
```diff
@@ -95,50 +95,4 @@ export default {
 	"attachments.addLinkMenu": "Weblink", // Attach menu item text
 	"attachments.addOneDriveLinkMenu": "OneDrive", // Attach menu item text
 	"attachments.addQuicklinkMenu": "Eksisterende aktivitet", // Attach menu item text
-
```
astadelbauer, line 98: These must have got missed during some back and forth in our changes, it looks like you're right and they should be deleted. Thanks for noticing and letting me know!

PR #1003, opened 2020-07-27
Hide the rubrics label when the user doesn't not have any permissions or associations

patel-kajal:

Hide the rubrics label when the user doesn't have any permissions for rubrics and there is no existing association

Tested:
- edited an assignment with an instructor that only has "See Rubrics" permission and there is no existing rubric association, the rubrics header is correctly hidden
- edited an assignment with an instructor that only has "See Rubrics" permission but there is an existing rubric association, the rubrics header is correctly shown
- edited an assignment with an instructor that has no permissions, the rubrics header is correctly hidden
- edited an assignment with an instructor that has all permission, the rubrics header is correctly shown


components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -177,6 +177,30 @@ class ActivityRubricsListContainer extends ActivityEditorFeaturesMixin(ActivityE
 		this._newlyCreatedPotentialAssociationHref = '';
 	}
 
+	_renderRubricsList(entity) {
+		const canCreatePotentialAssociation = entity.canCreatePotentialAssociation();
+		const canCreateAssociation = entity.canCreateAssociation();
+		const totalAssociations = entity.fetchAssociations().length;
```
kieranderson, line 183: Should this be checking for the associating/associated and not deleted? If I remember correctly fetchAssociations includes potential associations.

patel-kajal: yeah thanks, your right, since those are the only ones being rendered. I updated it to use `fetchAttachedAssociationsCount()`

components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -177,6 +177,30 @@ class ActivityRubricsListContainer extends ActivityEditorFeaturesMixin(ActivityE
 		this._newlyCreatedPotentialAssociationHref = '';
 	}
 
+	_renderRubricsList(entity) {
+		const canCreatePotentialAssociation = entity.canCreatePotentialAssociation();
+		const canCreateAssociation = entity.canCreateAssociation();
+		const associationsCount = entity.fetchAttachedAssociationsCount();
+
+		if (!canCreatePotentialAssociation && !canCreateAssociation && associationsCount === 0) {
```
pureooze, line 185: Would we want to make this a computed value in the mobx store instead?
https://mobx.js.org/refguide/computed-decorator.html

patel-kajal: You mean create a computed value from this expression?:
` (!canCreatePotentialAssociation && !canCreateAssociation && associationsCount === 0)`

I'm not really sure, because its only used here

patel-kajal:
PR related to same defect (hiding Rubrics summary when there is no permissions):
https://github.com/BrightspaceHypermediaComponents/activities/pull/1004

components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
patel-kajal: > Would we want to make this a computed value in the mobx store instead?
> https://mobx.js.org/refguide/computed-decorator.html
> 
> (this is a question because I think it might help but it might not) 😄

Hmm, I'm not sure if it will add any benefit to use a computed value because the associationCount won't really be changing if the user doesn't have any permissions 🤔

astadelbauer:
well, that was a majorly failed merge 👎

PR #1004, opened 2020-07-27
hide the rubrics summary if the user doesn't have any permissions for rubrics

patel-kajal:

Hide the summary for rubrics if the user doesn't have any permission to see rubrics 

Testing:
- Try to edit an assignment with and without an existing rubric attached but with a user that has no permissions for rubrics. The rubrics summary is hidden
- Try to edit an assignment with and without an existing rubric but with a user with full permissions for rubrics, the rubrics summary is shown

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-summary.js
+++ b/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-summary.js
@@ -20,7 +20,7 @@ class ActivityRubricsSummary
 		const canCreatePotentialAssociation = entity.canCreatePotentialAssociation();
 		const canCreateAssociation = entity.canCreateAssociation();
 		const associationsCount = entity.fetchAttachedAssociationsCount();
-		if(!canCreatePotentialAssociation && !canCreateAssociation){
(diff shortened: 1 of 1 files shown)


patel-kajal:
PR related to the same defect (Hiding the rubrics label when there is no permissions): https://github.com/BrightspaceHypermediaComponents/activities/pull/1003

components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-summary.js
```diff
@@ -17,8 +17,12 @@ class ActivityRubricsSummary
 			return html``;
 		}
 
+		const canCreatePotentialAssociation = entity.canCreatePotentialAssociation();
+		const canCreateAssociation = entity.canCreateAssociation();
 		const associationsCount = entity.fetchAttachedAssociationsCount();
-		if (associationsCount <= 0) {
+		if (!canCreatePotentialAssociation && !canCreateAssociation) {
+			return html``;
+		} else if (associationsCount <= 0) {
```
astadelbauer, line 25: nitpick: this could just be a subsequent if statement rather than an else if

PR #1008, opened 2020-07-27
US118882 Hide anonymous marking options when submission type is in-person or on-paper

emil-furniss:
Applies to the checkbox and the summary text.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F407977501928


emil-furniss:
@wongvincent Good point. This is now excluding the isAnonymous prop if the submission type does not support it, and there's a new PR in siren-sdk to deal with the dirty-check on cancel.
https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/205

PR #1015, opened 2020-07-30
Update lang mixin to fetch lang overrides

kieranderson:
Call into the core helper to fetch lang overrides if they are there.


components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js
```diff
@@ -1,8 +1,15 @@
+import { getLocalizeOverrideResources } from '@brightspace-ui/core/helpers/getLocalizeResources.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
 
 export const LocalizeActivityEditorMixin = superclass => class extends LocalizeMixin(superclass) {
 
 	static async getLocalizeResources(langs) {
+
+		function resolveOverridesFunc() {
+			return resolveUrl('../lang/overrides.js', import.meta.url);
```
pureooze, line 10: Is this path relative to this JS file or `@brightspace-ui/core/helpers/getLocalizeResources.js`?

kieranderson: this file

pureooze: Interesting, I was curious if resolveUrl did something weird since `getLocalizeResources` is what ends up calling it.

components/d2l-activity-editor/d2l-activity-assignment-editor/mixins/d2l-activity-assignment-lang-mixin.js
```diff
@@ -1,8 +1,15 @@
+import { getLocalizeOverrideResources } from '@brightspace-ui/core/helpers/getLocalizeResources.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
 
 export const LocalizeActivityAssignmentEditorMixin = superclass => class extends LocalizeMixin(superclass) {
 
 	static async getLocalizeResources(langs) {
+
+		function resolveOverridesFunc() {
+			return resolveUrl('../lang/overrides.js', import.meta.url);
```
JordanCorner, line 10: Should we be doing all of this work more generally at the `LocalizeMixin` level? Is that something we should be putting on the roadmap for next release too?

astadelbauer: Maybe if/when we're confident it works well and want it everywhere. I'm also not sure if the lang folder location is consistent everywhere so it could be interesting.

PR #1018, opened 2020-07-31
hide the RC header when the user doesn't have manage delivery permission

patel-kajal:
[DE339958](https://rally1.rallydev.com/#/110294864140d/custom/110851118712?detail=%2Fdefect%2F408302156092&fdp=true)

Hide the RC header when the user doesn't have manage delivery permissions

Testing:
- remove manage delivery permissions and see that the RC header doesn't show under Availability on FACE
- give manage delivery permission and see that the RC header does show under Availability on FACE

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -112,7 +112,7 @@ class ActivityAssignmentAvailabilityEditor extends ActivityEditorFeaturesMixin(L
 
 	_renderReleaseConditionEditor() {
 		const activity = store.get(this.href);
-		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.conditionsHref) {
+		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.canEditReleaseConditions) {
 			return html``;
 		}
 
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -111,7 +111,8 @@ class ActivityAssignmentAvailabilityEditor extends ActivityEditorFeaturesMixin(L
 	}
 
 	_renderReleaseConditionEditor() {
-		if (!this._m3ReleaseConditionsEnabled) {
+		const activity = store.get(this.href);
+		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.conditionsHref) {
 			return html``;
 		}
 
--- a/components/d2l-activity-editor/state/activity-usage.js
+++ b/components/d2l-activity-editor/state/activity-usage.js
@@ -28,6 +28,7 @@ export class ActivityUsage {
 	async load(entity) {
 		this._entity = entity;
 		this.conditionsHref = entity.conditionsHref();
+		this.canEditReleaseConditions = entity.canEditReleaseConditions();
 		this.isDraft = entity.isDraft();
 		this.canEditDraft = entity.canEditDraft();
 		this.isError = false;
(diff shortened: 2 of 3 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -111,7 +111,8 @@ class ActivityAssignmentAvailabilityEditor extends ActivityEditorFeaturesMixin(L
 	}
 
 	_renderReleaseConditionEditor() {
-		if (!this._m3ReleaseConditionsEnabled) {
+		const activity = store.get(this.href);
+		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.conditionsHref) {
```
pureooze, line 115: Instead of checking `activity.conditionsHref` should we be using something like `activity.CanEditReleaseConditions`?

patel-kajal: Hmm, so I'm not sure if I'm looking in the right place, but I don't see `CanEditReleaseConditions` in the store:
https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/state/activity-usage.js?r=c01347f5

patel-kajal: Also, I used the same thing that was used to hide the editor (`d2l-activity-usage-conditions-editor`):

https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-usage-conditions-editor.js?r=33702da8

pureooze: Sorry I was really value lol 😅 I meant making a new method in siren-sdk to check that permission and then store that in the store. Not sure if its worth it/needed though, I'm not super familiar with release conditions. Just in general I think if we do permission checks we should do it siren-sdk side not in the UI.

patel-kajal: Oh okay, yeah I think your right, so there's already a method in siren-sdk. I can use it to add to the store and update both these files, it will probably make more sense

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -111,7 +111,8 @@ class ActivityAssignmentAvailabilityEditor extends ActivityEditorFeaturesMixin(L
 	}
 
 	_renderReleaseConditionEditor() {
-		if (!this._m3ReleaseConditionsEnabled) {
+		const activity = store.get(this.href);
+		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.canEditReleaseConditions) {
```
pureooze, line 115: Are the manage delivery permissions different than this? Or is this the same thing?

pureooze: The `canEditReleaseConditions` was just an example on the spot so not sure if the permission mapping is actually right 🤔

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -28,6 +28,7 @@ export class ActivityUsage {
 	async load(entity) {
 		this._entity = entity;
 		this.conditionsHref = entity.conditionsHref();
+		this.canEditReleaseConditions = entity.canEditReleaseConditions();
```
pureooze, line 31: Also need to add this as an observable at the bottom of this file

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
patel-kajal: Yeah, it looks like that API will call this function which will make sure it has manage delivery permissions
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox/Utilities/Access/DropboxAccessChecker.cs?r=0bc6b66e#55

PR #1024, opened 2020-08-04
Add stylelint integration

emil-furniss:
Adds the new [`@brightspaceUI/stylelint-config`](https://github.com/BrightspaceUI/stylelint-config) integration and makes the existing files compliant. More info in the [blog post](https://blog.d2l.dev/2020/08/04/style-linting/).

Most of the changes are reordering styles to be in alphabetical order, and renaming classes to have the required `d2l-` prefix.

One other change is to use the shared core styles for the html editor. We were previously using a copy of them from before their conversion to Lit.

There are a couple of files with the eol char has changed, so ignoring white space might make the diff more manageable.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-html-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-editor.js
@@ -2,6 +2,7 @@ import 'd2l-html-editor/d2l-html-editor.js';
 import 'd2l-html-editor/d2l-html-editor-client.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
+import { inputStyles } from '@brightspace-ui/core/components/inputs/input-styles';
 import { LocalizeActivityEditorMixin } from './mixins/d2l-activity-editor-lang-mixin.js';
 import { resolveUrl } from '@polymer/polymer/lib/utils/resolve-url.js';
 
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -75,6 +76,7 @@
     "@brightspace-ui-labs/list-item-accumulator": "^1.0.0",
     "@brightspace-ui/core": "^1.55.0",
     "@brightspace-ui/intl": "^3.0.1",
+    "@brightspace-ui/stylelint-config": "^0.0.1",
```
dbatiste, line 79: I think these two new deps should be moved up to `devDependencies`.

pureooze:
Other than what Dave said, LGTM we should probably also do this for `eslint-config-brightspace` at some point, there was an update that required a bunch of things to be sorted. It would be nice to get that in too 😄  https://github.com/BrightspaceHypermediaComponents/activities/pull/935

components/d2l-activity-editor/d2l-activity-html-editor.js
```diff
@@ -150,7 +79,7 @@ class ActivityHtmlEditor extends LocalizeActivityEditorMixin(LitElement) {
 
 				<div id="toolbar-shortcut-${this._htmlEditorUniqueId}" hidden="">${this.localize('editor.ariaToolbarShortcutInstructions')}</div>
 				<div
-					class="d2l-html-editor-container"
+					class="d2l-html-editor-container d2l-input"
```
emil-furniss, line 82: @dbatiste Just noticed during some more regression testing that the disabled editor styles are not being applied here, though the editing capability is disabled correctly. I thought it was because I didn't add the `disabled` attribute to this div, but that doesn't appear to make a difference. Any ideas?

dbatiste: I think `d2l-input` styles that are being included here have some selectors that look for the `:disabled` pseudo-class.  However, because this is not a form element, perhaps this element does not get selected by that class.  If that is the case, two options come to mind: 1) we add the attribute selector to the `d2l-input` styles (i.e. `[disabled].d2l-input`), or 2) you'll need to specify a couple of custom styles for your case here.  Would not be against option 1 since this could come up in a couple of instances, but I'd probably also want to validate this first, and perhaps get a 2nd opinion.

emil-furniss: Interesting, thanks! To get this PR moving along I've copied over the disabled styles for now which works fine. And without looking into it much more, I tried adding the attribute selector styles to the `inputStyles` but didn't get very far.

PR #1026, opened 2020-08-05
fix padding issue in choose-from-grades

spandanabaruah:


components/d2l-activity-editor/d2l-activity-grades-dialog.js
```diff
@@ -55,6 +55,9 @@ class ActivityGradesDialog extends ActivityEditorMixin(LocalizeActivityEditorMix
 				padding-top: 6px;
 				padding-bottom: 6px;
 			}
+			.d2l-input-radio-label {
```
wongvincent, line 58: wouldn't this add padding to the first radio button as well?

spandanabaruah: oh gosh yes you are right

PR #1028, opened 2020-08-05
US119006 Disable annotations editor if action is not available

emil-furniss:
Previously the annotations editor was being hidden completely if the edit actions is not available. With the new read-only design, this should change to show a disabled editor.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
@@ -53,11 +53,6 @@ class ActivityAssignmentAnnotationsEditor
 			return html``;
 		}
 
-		const shouldRenderEditor = entity.canSeeAnnotations;
-		if (!shouldRenderEditor) {
-			return html``;
-		}
-
 		return html`
 			<label class="d2l-label-text">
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-annotations-editor.js
```diff
@@ -53,19 +53,15 @@ class ActivityAssignmentAnnotationsEditor
...
 		return html`
 			<label class="d2l-label-text">
 				${this.localize('annotationTools')}
 			</label>
 			<d2l-input-checkbox
 				@change="${this._toggleAnnotationToolsAvailability}"
 				?checked="${entity.annotationToolsAvailable}"
-				ariaLabel="${this.localize('annotationToolDescription')}">
+				ariaLabel="${this.localize('annotationToolDescription')}"
+				?disabled="${!entity.canSeeAnnotations}">
```
mdulat, line 64: Should this be called `canEditAnnotations` then? I realize that requires changing the siren-sdk counterpart as well.

emil-furniss: Yeah, might as well. siren-sdk PR is here: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/210

PR #1029, opened 2020-08-06
US119007 - FACE view when Edit Evaluation permission is missing

wongvincent:
User is required to have See and Manage ASFs and Edit Evaluation Properties permissions to modify score out of, grades, and rubrics create/add existing, and Default Scoring Rubric.

When permission is missing, this PR makes it so that out-of input is disabled, 'ungraded' button is disabled, In Grades / Not In Grades dropdown is disabled, rubrics Create New / Add Existing dropdown is disabled.
I did not need to make any changes to make remove 'x' rubric hidden or make Default Scoring Rubric dropdown disabled as that was already the behavior when the permission is missing.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F408759753408

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -53,6 +53,11 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeActivityEditorMixi
 				--d2l-input-padding: 0.4rem 0.75rem 0.4rem 1.65rem;
 				--d2l-input-padding-focus: calc(0.4rem - 1px) calc(0.75rem - 1px) calc(0.4rem - 1px) calc(1.65rem - 1px);
 			}
+			#ungraded:disabled,
+			.d2l-grade-info:disabled {
+				cursor: default;
+				opacity: 0.5;
+			}
 			#score-info-container,
 			#score-out-of-container,
 			#grade-info-container {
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -282,8 +298,8 @@ class ActivityScoreEditor extends ActivityEditorMixin(LocalizeActivityEditorMixi
 				${canSeeGrades ? html`
 					<div id="grade-info-container">
 						<div id="divider"></div>
-						<d2l-dropdown>
-							<button class="d2l-label-text d2l-grade-info d2l-dropdown-opener">
+						<d2l-dropdown ?disabled="${!canEditScoreOutOf}">
+							<button class="d2l-label-text d2l-grade-info d2l-dropdown-opener" ?disabled="${!canEditScoreOutOf}">
```
emil-furniss, line 302: It would be nice if we ccould use the `d2l-dropdown-button-subtle` for this, then we'd get all the consistent styles for free, but I don't think it supports adding the custom grades icon.

dbatiste: It should be possible to use `d2l-button-subtle` instead of `button`.  You would still need to indicate the opener using the `d2l-dropdown-opener` class, but that should give the ability to set a different icon whilst using the subtle button styles.

wongvincent: The button is supposed to look like this 

image
Note the non-blue text color.

And when hovered, 
image
Note the transparent background-color and gray border.

It doesn't seem like button-subtle would work unless I could override those styles.

dbatiste: Ah I did not realize this was special, beyond the icon.  I would really question whether we need a special variant of subtle button like this though, given that we already have a few buttons well defined in the design patterns and implemented in core.  i.e. why is this a one-off, and if it's not, then it should be in core.

PR #1031, opened 2020-08-06
Create new folder for new face content creation pages

imcdonald-d2l:
- Added a simple template page with no information or properties for now. Just adding so we can hook it up on the back-end and test the opt-in feature with the new component. 
- Added a demo page that renders the basic template page.

Let me know if anything should be moved/renamed or if anything needs to be added to the basic template for testing.

image


pbrown-d2l:
Looks good once you fix the Lint issues that are failing that build

PR #1033, opened 2020-08-07
DE39814 - Make sure new grade candidates are fetched when doing an A…

wongvincent:
…ctivity-Usage PATCH for all create new grade scenarios

The `fetchNewGradeCandidates()` was only being called if the Choose From Grades dialog was opened once. But now, if "In Grades" is selected and it was previously ungraded or not in grades, we should still be fetching the new grade candidates in order to use its fields as part of the activity-usage PATCH.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fdefect%2F405157847512

**Associated PRs**
https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/14588/overview
https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/211

### DIFF
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -124,6 +124,12 @@ export class ActivityScoreGrade {
 
 		return newGradeCandidateEntity.getSaveAction();
 	}
+
+	async save() {
+		if (this.inGrades && this.createNewGrade) {
+			await this.fetchNewGradeCandidates();
+		}
+	}
 }
 
 decorate(ActivityScoreGrade, {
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -124,6 +124,12 @@ export class ActivityScoreGrade {
 
 		return newGradeCandidateEntity.getSaveAction();
 	}
+
+	async save() {
```
mdgbayly, line 128: This is weirdly named? As it is not 'saving' anything. Given it is just a helper method called explicitly from the activity-usage state, does it need to be called save? could it be called something more expressive?

PR #1035, opened 2020-08-11
Kapatel/user/de39269 hide rubrics label when the 

patel-kajal:

Hide the rubrics label when the user doesn't have any permissions for rubrics and there is no existing association

Tested:
- edited an assignment with an instructor that only has "See Rubrics" permission and there is no existing rubric association, the rubrics header is correctly hidden
- edited an assignment with an instructor that only has "See Rubrics" permission but there is an existing rubric association, the rubrics header is correctly shown
- edited an assignment with an instructor that has no permissions, the rubrics header is correctly hidden
- edited an assignment with an instructor that has all permission, the rubrics header is correctly shown


patel-kajal:
moved [this pr](https://github.com/BrightspaceHypermediaComponents/activities/pull/1003)

PR #957, opened 2020-06-26
Moved submission mobx stuff into its own class

pureooze:
This will allow us to call a setter in UI tests to setup the store for test scenarios without having to stub Hypermedia APIs.

PR #1042, opened 2020-08-18
Move MobX completion type stuff into a shared store with submission state

pureooze:
Continuing the work in #957 and moving the completion properties into the store with the submission stuff. These two are closely related to each other so it makes sense to keep them in the same store.


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-submission-and-completion.js
```diff
@@ -0,0 +1,131 @@
...
+		this.allCompletionTypeOptions = entity.allCompletionTypeOptions;
+		this.canEditCompletionType = entity.canEditCompletionType;
+		this.completionType = entity.completionType;
+
+		const validCompletionTypes = this._getValidCompletionTypes(this.submissionType);
+		if (this.canEditCompletionType) {
+			this.completionTypeOptions =  this._getCompletionTypeOptions(validCompletionTypes);
+		} else {
+			const completionType = entity.completionType();
+			this.completionTypeOptions = completionType ? [completionType] : [];
```
patel-kajal, line 27: I'm not sure if I remember this correctly. But I think File/Text submission only have one valid completion type which is why it can't be edited, right? Do we need to check "canEditCompletionType"? 🤔 

I know this is also how it was originally implemented, but just asking out of curiosity

pureooze: Its been a while but I think this was for the case when they didn't have permission but it was set to a submission type where completion type is visible but disabled? So you would see the one that was already selected but couldn't change it? I think it might have been because the API didn't send the full list when you don't have permission?

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js
```diff
@@ -21,58 +21,9 @@ export class Assignment {
...
-		if (this.completionType === null || !this._isCompletionTypeValid(this.completionType, validCompletionTypes)) {
-			if (validCompletionTypes && validCompletionTypes.length > 0) {
-				this.completionType = String(validCompletionTypes[0]);
-			} else {
-				this.completionType = null;
-			}
-		}
-	}
-
 	_isSubmissionTypeWithAnonMarking() {
```
kieranderson, line 24: Debatable to move this in there as well? Depends if "WithAnonMarking" is part of the submission type or separate that only works with x,y,z types.

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-submission-and-completion.js
patel-kajal: Ooh okay , yeah I think I remember it wasn't sending the completion type list when you couldn't actually edit the option

PR #1046, opened 2020-08-21
Move MobX assignment type stuff into its own store

pureooze:
Continuing the work from #1043 this PR moves the assignment type stuff into its own store.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
@@ -66,13 +66,13 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 			return html``;
 		}
 
-		const isIndividualType = assignment.isIndividualAssignmentType;
+		const isIndividualType = assignment.assignmentTypeProps.isIndividualAssignmentType;
 		const infoText = this._getInformationText(assignment);
-		const canEditAssignmentType = assignment.canEditAssignmentType;
-		const groupTypeDisabled = assignment.isGroupAssignmentTypeDisabled;
+		const canEditAssignmentType = assignment.assignmentTypeProps.canEditAssignmentType;
+		const groupTypeDisabled = assignment.assignmentTypeProps.isGroupAssignmentTypeDisabled;
 		const folderTypeText =	isIndividualType ? this.localize('txtIndividual') : this.localize('txtGroup');
-		const groupTypeText = !isIndividualType && assignment.selectedGroupCategoryName
-			? this.localize('txtGroupCategoryWithName', 'groupCategory', assignment.selectedGroupCategoryName)
+		const groupTypeText = !isIndividualType && assignment.assignmentTypeProps.selectedGroupCategoryName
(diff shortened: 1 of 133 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -136,14 +136,14 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 	}
 
 	_getInformationText(assignment) {
-		if (!assignment) {
+		if (!assignment || !assignment.assignmentTypeProps) {
```
kieranderson, line 139: Should check `submissionAndCompletionProps` here as well

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-summary.js
```diff
@@ -24,7 +24,7 @@ class AssignmentTypeSummary extends ActivityEditorMixin(RtlMixin(LocalizeActivit
 
 	render() {
 		const assignment = store.get(this.href);
-		if (assignment && !assignment.isIndividualAssignmentType) {
+		if (assignment && !assignment.assignmentTypeProps.isIndividualAssignmentType) {
```
kieranderson, line 27: check for `assignmentTypeProps` as well? Similar to the component above,

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-type.js
```diff
@@ -0,0 +1,51 @@
...
+	setAssignmentTypeGroupCategory(value) {
+		this.selectedGroupCategoryId = value;
+	}
+
+	setToGroupAssignmentType() {
+		this.isIndividualAssignmentType = false;
+		this.selectedGroupCategoryId =
+			this.selectedGroupCategoryId
+				? String(this.selectedGroupCategoryId)
+				: String(this.groupCategories[0].value);
```
kieranderson, line 33: Is this array guaranteed to have values?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
pureooze: Good catch!

components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-type.js
pureooze: Group type seems to be disabled when there are no categories so it should not be possible to call this method when that is the case https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=ec193248#760

pureooze: And siren will set the group type field to disabled in the case that class is present https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/c4bd360263cd1ae2183c10bdc0fd2c94dd30f73a/src/activities/assignments/AssignmentEntity.js#L249-L258

PR #1048, opened 2020-08-24
Bugfix: Retrieve completion type from property instead of accessing method that does not exist

pureooze:


pureooze:
Fix for bug introduced in https://github.com/BrightspaceHypermediaComponents/activities/pull/1042

PR #1067, opened 2020-09-02
US116271 Use new d2l-input-date-time for FACE due and availability dates

emil-furniss:
Replaces the old `d2l-datetime-picker` with the new `d2l-input-date-time` for the due, start, and end dates. Yay!! The new picker looks so much nicer!

Since our workflow relies on server-side validation and uses 3 dates which are all interdependent, this does not (yet?) make use of any of the pre-built range components. In order to customize the error and tooltip behavior, I've hacked into `invalid` and `validationError` properties defined by the FormElementMixin. That doesn't appear to be an intended way to use the component, but am unsure how else to realize our design. Might @margaree and/or @AllanKerr have any suggestions? Maybe it's just a case of officially supporting those attributes, but maybe there is a different "officially supported" approach that could be utilized here?

@mdgbayly Let me know if this fulfills the requirements for early loading for the skeleton loading stuff. There is no longer the need to initialize the pickers with null values on first load, but this is still being a little defensive when it comes to loading when the entity doesn't exist yet (but will load if the dates subentity is not present yet). I've had some issues loading with dummy values immediately (by leaving the dates empty and assuming that `canEditDates` is true), specifically getting it to refresh with existing dates when the entity finally loads in.


components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -8,99 +8,41 @@ import { shared as store } from './state/activity-store.js';
...
+		if (!dates.canEditDates) {
+			return html``;
 		}
 
-		return html`
-			<label class="d2l-label-text" ?hidden="${!canEditDates}">${this.localize('editor.dueDate')}</label>
-			${this.dateTemplate(dueDate, canEditDates, errorTerm)}
-		`;
-	}
-	updated(changedProperties) {
```
mdgbayly, line 67: I think with this override removed, technically you have to implement a constructor that passes the store object to the base `ActivityEditorMixin` otherwise this component won't initiate an activity fetch. It will work without it, but the intention is to make every component mostly standalone, so that if they are used in isolation, they will still work.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -8,99 +8,41 @@ import { shared as store } from './state/activity-store.js';
...
-
-	firstUpdated() {
-		super.firstUpdated();
-		this._isFirstLoad = false;
-	}
 	render() {
 		const entity = store.get(this.href);
-		const dates = entity ? entity.dates : null;
-		let dueDate, canEditDates, errorTerm;
+		if (!entity) {
```
mdgbayly, line 21: Yeah - when we do the skeleton implementation we'll probably want to still allow the render even if the entity doesn't exist yet, but we can address that when we get to it. It's possible that we'll be wanting to make the date picker **skeleton** aware as part of that anyway, so that we can skeletize the label which is now embedded in the the date picker component. We have some other places where we are using external labels, but from speaking with the Dave's this is not ideal from an a11y perspective even if we use an `aria-label` on the inputs.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
```diff
@@ -8,99 +8,41 @@ import { shared as store } from './state/activity-store.js';
...
 		}
 
-		return html`
-			<label class="d2l-label-text" ?hidden="${!canEditDates}">${this.localize('editor.dueDate')}</label>
-			${this.dateTemplate(dueDate, canEditDates, errorTerm)}
-		`;
-	}
-	updated(changedProperties) {
-		super.updated(changedProperties);
+		const errorTerm = this.localize(dates.dueDateErrorTerm);
```
mdgbayly, line 30: Don't totally follow what's going on here? So if there is a dates error then `dates.dueDateErrorTerm` is a lang term otherwise it is empty? And if it's empty calling `localize` on an empty term returns an empty value? 

If so might just be a nit, but it seems kind of hacky and unclear. Couldn't we just use a conditional and skip calling `localize` if there is no date error?

components/d2l-activity-editor/d2l-activity-due-date-editor.js
emil-furniss: Ah yes, I'll add the constructor back in. That's probably why I was having issues with the refresh, in which case I'll also make the change to allow rendering without the entity.

components/d2l-activity-editor/d2l-activity-due-date-editor.js
emil-furniss: Yeah, you're reading that right but I'll change it. Just cause it was like that before doesn't mean it should stay like that ^^

components/d2l-activity-editor/d2l-activity-due-date-editor.js
mdgbayly: i'd be surprised if that was causing the refresh issues, because it just determines whether something initiates the fetch on the activity, and in FACE various components initiate that fetch. The refresh issue does sound like something we'll want to address at some point but OK for you to move ahead for now without it, because it only becomes a problem for us when we want to skeletize this.

emil-furniss:
@mdgbayly Updated in response to comments, and also enabled loading without the entity being present (found my refreshing bug).

mdgbayly (approved):

PR #1068, opened 2020-09-02
Rfegan/us120114 refactor assignment editor footer

ronan-f:
Changes based on [this ticket.](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F423721882656)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -328,7 +328,7 @@ class AssignmentEditor extends ActivityEditorContainerMixin(RtlMixin(LocalizeAct
 						options
 					);
 				}).then(filteredContent => {
-					const matchSrc = function (str) {
+					const matchSrc = function(str) {
 						// excludes matching query string as filterHtml may modify the query string
 						return str.match(/src=["']([^?"']+)/i);
 					};
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -183,6 +183,8 @@ class AssignmentEditor extends ActivityEditorContainerMixin(RtlMixin(LocalizeAct
 			return html``;
 		}
 
+		const activityUsageHref = activity.href;
+
 		const {
 			assignmentHref
 		} = activity;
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -214,7 +216,7 @@ class AssignmentEditor extends ActivityEditorContainerMixin(RtlMixin(LocalizeAct
 					</d2l-activity-assignment-editor-secondary>
 				</div>
 				<d2l-activity-assignment-editor-footer
-					.href="${assignmentHref}"
+					.href="${activityUsageHref}"
```
mdgbayly, line 219: The href of this component `d2l-activity-assignment-editor` is the same as `activity.href` so we don't need to wait on the activity to load in order to set the href on the footer. We can just set the footer `href` attribute directly to `${this.href}`.

When we move further with this feature and start adding the skeleton states, we'll be removing the null check on the activity and just allowing things to render OK to leave that in for now.

```
		if (!activity) {
			return html``
		}
```

mdgbayly (commented):
Changes to the footer component look good. So just suggest we make the change to how we set it's href from the activity assignment editor as it will make things a bit cleaner, particularly later when we're allowing things to just render. But what you have would work for now.

mostlyfabulous (commented):
LGTM

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -328,7 +328,7 @@ class AssignmentEditor extends ActivityEditorContainerMixin(RtlMixin(LocalizeAct
 						options
 					);
 				}).then(filteredContent => {
-					const matchSrc = function (str) {
+					const matchSrc = function(str) {
```
mostlyfabulous: @emil-furniss & @wongvincent My VSCode's ESLint adds a space after a `function` and @ronan-f has the same problem. I'm wondering what in the config is going wrong if the `npm lint:js` which runs eslint too, isn't happy with.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
ronan-f: Thanks @mdgbayly. I've changed the href being passed through to use the value of this.href in `d2l-activity-assignment-editor`

mdgbayly (approved):

PR #1073, opened 2020-09-04
US120106 Implement skeleton re-usable styles and mixin

mostlyfabulous:
https://rally1.rallydev.com/#/detail/userstory/423713941720?fdp=true

Need some help fixing linting issues my VS Code isn't helping with

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
@@ -5,9 +5,13 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityAssignmentEditorMixin } from './mixins/d2l-activity-assignment-lang-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from '../save-status-mixin.js';
-import { SkeltizeMixin } from '../mixins/d2l-skeletize-mixin';
+import { SkeletizeMixin } from '../mixins/d2l-skeletize-mixin';
 
-class AssignmentEditorFooter extends SkeltizeMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement))))) {
+class AssignmentEditorFooter extends SkeletizeMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement))))) {
+
+	static get properties() {
+		return { skeleton: super.properties.skeleton };
+	}
 
 	static get styles() {
 		return [super.styles, css`
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
@@ -5,11 +5,12 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityAssignmentEditorMixin } from './mixins/d2l-activity-assignment-lang-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from '../save-status-mixin.js';
+import { SkeltizeMixin } from '../mixins/d2l-skeletize-mixin';
 
-class AssignmentEditorFooter extends SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement)))) {
+class AssignmentEditorFooter extends SkeltizeMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement))))) {
 
 	static get styles() {
-		return css`
+		return [super.styles, css`
 			:host {
 				display: flex;
 			}
@@ -34,12 +35,12 @@ class AssignmentEditorFooter extends SaveStatusMixin(ActivityEditorMixin(RtlMixi
 					justify-content: space-between;
 				}
 			}
-		`;
+		`];
 	}
 
 	render() {
 		return html`
-			<div class="d2l-activity-assignment-editor-footer-left">
+			<div class="skeletize d2l-activity-assignment-editor-footer-left">
 				<d2l-activity-visibility-editor
 					.href="${this.href}"
 					.token="${this.token}">
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
@@ -40,7 +40,7 @@ class AssignmentEditorFooter extends SkeltizeMixin(SaveStatusMixin(ActivityEdito
 
 	render() {
 		return html`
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
```diff
@@ -34,12 +35,12 @@ class AssignmentEditorFooter extends SaveStatusMixin(ActivityEditorMixin(RtlMixi
...
 				}
 			}
-		`;
+		`];
 	}
 
 	render() {
 		return html`
-			<div class="d2l-activity-assignment-editor-footer-left">
+			<div class="skeletize d2l-activity-assignment-editor-footer-left">
```
zinabat, line 43: The linter is complaining because it expects classes that are applied to components to start with the name of the class. This is why all of our component classes start with `d2l-`. If we change this class to `d2l-skeletize` everybody is happy 🍬

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
```diff
@@ -0,0 +1,71 @@
+import '@brightspace-ui/core/components/colors/colors.js';
+import { css } from 'lit-element/lit-element.js';
+
+export const SkeltizeMixin = superclass => class extends superclass {
+	static get properties() {
+		return {
+			skeleton: { attribute: 'skeleton', reflect: true, type: Boolean }
```
zinabat, line 7: Nit: You don't need to specify the attribute here - this only needs to be done if the attribute is different than the property, as is the case with our multi-word properties.

zinabat (commented):
image

Sorry, I couldn't resist 😂

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
```diff
@@ -0,0 +1,71 @@
+import '@brightspace-ui/core/components/colors/colors.js';
+import { css } from 'lit-element/lit-element.js';
+
+export const SkeltizeMixin = superclass => class extends superclass {
```
mdgbayly, line 4: I think I prefer the name `SkeletonMixin`. That also matches the name Dave is using in the spikes he is doing in core, so on the off chance that this actually becomes something we run with in core, we should use that name i think.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
mdgbayly: Let's go with `d2l-skeletize`!

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
mdgbayly: Although does that break the naming convention of our `d2l-skeletize` class then? If we make the mixin `SkeletonMixin` does the class need to become `d2l-skeleton-skeletize`....???

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
mostlyfabulous: Good to know and thank you for explaining!

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
mostlyfabulous: ✔️

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
mostlyfabulous: Too many skeletons in the class 😆  I think we could go with `SkeletizeMixin` until others strongly say otherwise.

mdgbayly: i'm fine with whatever for now - it's a simple search replace if we decide to change it later if/when we promote to core.

mdgbayly (dismissed):

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
```diff
@@ -5,11 +5,12 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityAssignmentEditorMixin } from './mixins/d2l-activity-assignment-lang-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from '../save-status-mixin.js';
+import { SkeltizeMixin } from '../mixins/d2l-skeletize-mixin';
 
-class AssignmentEditorFooter extends SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement)))) {
+class AssignmentEditorFooter extends SkeltizeMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement))))) {
```
mdgbayly, line 10: Oh, guess you need to fix the typo in this name too :-)

mostlyfabulous: Ooops thanks

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+			z-index: 2000;
+		}
+		:host([skeleton]) .d2l-skeletize {
+			border: none;
+			box-shadow: none;
+			color: transparent;
+			position: relative;
+		}
+
+		@keyframes loadingPulse {
```
mdgbayly, line 38: sorry - should have seen this earlier - these keyframes are duplicated. Also we don't really need the non-host styles below. Possibly Ronan copied those from my two style sheets in the draft PR - but that was just to make the demo page work.

mostlyfabulous: Understood, wasn't sure I understood the big picture/long term plan entirely so I left it in

mostlyfabulous: Got it will tidy it up

mdgbayly (approved):

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
```diff
@@ -5,11 +5,16 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityAssignmentEditorMixin } from './mixins/d2l-activity-assignment-lang-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from '../save-status-mixin.js';
+import { SkeletizeMixin } from '../mixins/d2l-skeletize-mixin';
 
-class AssignmentEditorFooter extends SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement)))) {
+class AssignmentEditorFooter extends SkeletizeMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(LitElement))))) {
+
+	static get properties() {
```
mdgbayly, line 12: Not needed to do this. the footer will pick up the properties from the mixin.

mostlyfabulous: Understood

PR #1077, opened 2020-09-08
US120104 AsyncContainerMixin

mostlyfabulous:
## Add AsyncContainerMixin to detail, secondary and footer components
https://rally1.rallydev.com/#/detail/userstory/423713247156?fdp=true

**Resolved: ~~Lint issues causing CI test failures~~**
The ordering rule for imports wants multiple import statements to preceed single import statements but this tends to conflict with the alphabetical imports rule.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -70,8 +70,8 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 	constructor() {
 		super();
 		this._debounceJobs = {};
-
 		this._linksProcessor = new LinksInMessageProcessor();
+		this.skeleton = true;
 	}
 
 	render() {
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -9,6 +9,7 @@ import '../d2l-activity-attachments/d2l-activity-attachments-editor.js';
 import { css, html } from 'lit-element/lit-element.js';
 
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
+import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { shared as attachmentCollectionStore } from '../d2l-activity-attachments/state/attachment-collections-store.js';
 import { shared as attachmentStore } from '../d2l-activity-attachments/state/attachment-store.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -9,6 +9,7 @@ import '../d2l-activity-attachments/d2l-activity-attachments-editor.js';
 import { css, html } from 'lit-element/lit-element.js';
 
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
+import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
```
emil-furniss, line 12: The multi-before-single takes precedence over the alphabetical-order rule, so this import should go at the very top.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-footer.js
```diff
@@ -1,13 +1,14 @@
 import '../d2l-activity-editor-buttons.js';
 import '../d2l-activity-visibility-editor.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
```
emil-furniss, line 4: And this import needs to be moved up by one line

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
mostlyfabulous: Thanks @emil-furniss

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
```diff
@@ -14,7 +15,7 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { shared as store } from './state/assignment-store.js';
 
-class AssignmentEditor extends ActivityEditorContainerMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(ActivityEditorMixin(MobxLitElement)))) {
+class AssignmentEditor extends ActivityEditorContainerMixin(AsyncContainerMixin(RtlMixin(LocalizeActivityAssignmentEditorMixin(ActivityEditorMixin(MobxLitElement))))) {
```
mdgbayly, line 19: I think as you're implementing the two mixins in the footer, you don't need to implement the `AsyncContainerMixin` on the `AssignmentEditor`. As it isn't using the `skeleton` property anywhere anyway?

Depending on what rendering behaviour we want for the footer it might make sense to have the two mixins on the `AsignmentEditor` rather than the footer, but it would depend on when we want the footer to show. 

If we want it to show as soon as the components being rendered by the footer are ready then we would just implement the mixins on the footer.

If, however, we want the footer to be delayed until both the left and right panels are ready, then we could make the footer just implement the `SkeletizeMixin` and have the `AssignmentEditor` implement both mixins and also set the `skeleton` attribute on the footer. That way the `AssignmentEditor` would be waiting on all APIs being called by the left, right panels and the footer and the footer would not render until the whole thing is ready.

So it shouldn't break anything to have it like this with the `AssignmentEditor` implementing the `AsyncContainerMixin` but it's not really necessary unless we go with the second approach. So OK to merge this, but we should revisit when we get to the footer.

mdgbayly (approved):
Just about the comment about reconciling how we manage the rendering of the footer skeleton, but I think we can leave it until we get to that story and we can chat with Joseph and the designers about how they'd like the footer rendered, because it wasn't really included in the initial designs.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
mostlyfabulous: Re-reading the US gives me the opinion that we will probably want to remove that `AsyncContainerMixin` from `AssignmentEditor` as it makes more sense (to me) to have the footer skeletize as early as possible instead of waiting for the detail and secondary components complete their API calls.
Thanks for the explanation!

PR #1078, opened 2020-09-08
US120108 Implement skeleton behavior on score out of

stevenao:
https://rally1.rallydev.com/#/detail/userstory/423713964032?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -124,8 +124,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 
 			<div id="score-and-duedate-container">
 				<div id="score-container">
-					<label class="d2l-label-text">${this.localize('scoreOutOf')}</label>
+					<label class="d2l-label-text d2l-skeletize">${this.localize('scoreOutOf')}</label>
 					<d2l-activity-score-editor
+						.skeleton="${this.skeleton}"
 						.href="${this.activityUsageHref}"
 						.token="${this.token}"
 						.activityName="${name}">
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -126,7 +126,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 				<div id="score-container">
 					<label class="d2l-label-text d2l-skeletize">${this.localize('scoreOutOf')}</label>
 					<d2l-activity-score-editor
-						.skeleton="${this.skeleton}"
+						?skeleton="${this.skeleton}"
 						.href="${this.activityUsageHref}"
 						.token="${this.token}"
 						.activityName="${name}">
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -117,8 +117,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(Localize
 
 			<div id="score-and-duedate-container">
 				<div id="score-container">
-					<label class="d2l-label-text">${this.localize('scoreOutOf')}</label>
+					<label class="d2l-label-text d2l-skeletize">${this.localize('scoreOutOf')}</label>
 					<d2l-activity-score-editor
+						.skeleton="${this.skeleton}"
```
mdgbayly, line 122: Couldn't we set this as an attribute rather than using the property setter syntax? The property setter syntax always feels a bit hacky to me. 

Or is it causing the same old 'undefined' string issue. If so, maybe it should wait until the property is actually defined by the detail component.

Out of interest, is the fact that the `SkeletizeMixin` is defaulting `this.skeleton` to true causing any issues when the user of the `d2l-activity-score-editor` component dosn't provide the attribute. I'm hoping in that case that the lack of the attribute overrides the default true value. If not then we should possibly default the value to false in the mixin.

stevenao: Defaulting the `skeleton` property in the `SkeletizeMixin` is causing trouble here. I just can't set `?skeleton` to false.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -117,8 +117,9 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(SaveStatusMixin(Localize
 
 			<div id="score-and-duedate-container">
 				<div id="score-container">
-					<label class="d2l-label-text">${this.localize('scoreOutOf')}</label>
+					<label class="d2l-label-text d2l-skeletize">${this.localize('scoreOutOf')}</label>
 					<d2l-activity-score-editor
+						?skeleton="${this.skeleton}"
```
mostlyfabulous, line 122: I presume the `?` in front of `skeleton` lets us optionally set the value provided `this.skeleton` is not undefined?

mdgbayly: The question mark identifies it to LitElement as a boolean attribute, such that if the value of `this.skeleton` is falsy then the attribute will be omitted.

components/d2l-activity-editor/mixins/d2l-skeletize-mixin.js
```diff
@@ -36,9 +36,4 @@ export const SkeletizeMixin = superclass => class extends superclass {
 		}
 		`;
 	}
-
-	constructor() {
-		super();
-		this.skeleton = true;
```
mdgbayly, line 42: Do we want to remove this completely or would it be better to initialize to `false`.

mostlyfabulous: If the default value of `true` from the mixin prohibits `?skeleton` from being set to `false` then I think we should.

mdgbayly (approved):

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
dbatiste: FYI, for string attribute values that could be `undefined`, there is an [ifDefined](https://lit-html.polymer-project.org/guide/template-reference#ifdefined) directive that can be used to accomplish similar behavior - i.e. only render the attribute if the value is defined.

PR #1083, opened 2020-09-09
US120110 skeletize attachments

ronan-f:
- Based on comments in the [user story ](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F423713966144), I made `<d2l-activity-attachments-list>` display conditionally depending on the value of `this.skeleton`. 

- In `<d2l-activity-attachments-picker>` I had to move the padding from :host to `.d2l-button-container` to get the skeleton view to take up all of the attachments container

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -156,7 +156,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 
 			<div id="assignment-attachments-editor-container" ?hidden="${!attachmentsHref}">
 				<d2l-activity-attachments-editor
-					.skeleton="${this.skeleton}"
+					?skeleton="${this.skeleton}"
 					href="${attachmentsHref}"
 					.token="${this.token}">
 				</d2l-activity-attachments-editor>
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -153,6 +153,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 
 			<div id="assignment-attachments-editor-container" ?hidden="${!attachmentsHref}">
 				<d2l-activity-attachments-editor
+					.skeleton="${this.skeleton}"
 					href="${attachmentsHref}"
 					.token="${this.token}">
 				</d2l-activity-attachments-editor>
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -153,6 +153,7 @@ class AssignmentEditorDetail extends ErrorHandlingMixin(AsyncContainerMixin(Skel
 
 			<div id="assignment-attachments-editor-container" ?hidden="${!attachmentsHref}">
 				<d2l-activity-attachments-editor
+					.skeleton="${this.skeleton}"
```
mdgbayly, line 156: It should be possible to pass this as an attribute rather than a property (so should not need the dot property setter syntax).

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
```diff
@@ -8,7 +8,8 @@ import { shared as store } from './state/attachment-collections-store.js';
 class ActivityAttachmentsEditor extends ActivityEditorMixin(MobxLitElement) {
 	static get properties() {
 		return {
-			_canAddAttachments: { type: Boolean }
+			_canAddAttachments: { type: Boolean },
+			skeleton: { type: Boolean }
```
mdgbayly, line 12: This could implement the SkeletizeMixin to get the skeleton property definition.

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
```diff
@@ -34,19 +35,21 @@ class ActivityAttachmentsEditor extends ActivityEditorMixin(MobxLitElement) {
 			canAddAttachments,
 		} = collection;
 
-		return html`
-			<d2l-activity-attachments-list
+		return html`${!this.skeleton ?
```
mdgbayly, line 38: Just thinking about this again, rather than deferring the initalization of the attachments list until after skeleton mode is disabled, it might be better to render the list as hidden.  Otherwise we are potentially holding up the loading of the attachments, and the loading of the attachments involves other API calls like unfurling, loading thumbnail images etc. So it would be best to get that going as soon as possible.

Ultimately might not make much difference, as I think the attachments are currently one of the last things to finish loading anyway, but every bit might help.

mdgbayly, line 38: We want to render the attachments picker even before we have the attachments collection - basically we want every skeleton to render immediately on first render before any data is loaded.

To accomplish this we need to make this handle the scenario where `store.get(this.href)` returns null and still set the `canAddAttachments` variable to true so that the picker renders.

This could be accomplished with something like:
```
const {
			canAddAttachments,
		} = collection || {
			canAddAttachments: true
		};
```

You may find that when you add this that you run into the issue I was describing in the story notes, where I found that the picker originally renders with no height because there are no buttons.

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
mostlyfabulous: Once the SkeletizeMixin is implemented, the call to super() in the constructor will apply the `skeleton` property to the component. That should be all you need

mdgbayly (approved):
Looks good - i still think we might need to do more to allow the skeleton to render immediately before the assignments/attachments APIs have returned, but OK to get it merged so we can see it all working together and then we can do follow up PRs if we need to tweak things.

PR #1084, opened 2020-09-09
fix skeleton label spacing

stevenao:


mostlyfabulous:
LGTM @mdgbayly thoughts?
image

mdgbayly:
I think we need to replace the `display: block` or other labels will be rendered at 100% width? I think it's sort of working of the score label to the fact that the score component is probably rendered i some kind of flex box?

mdgbayly:
If we don't want to try to use `d2l-input-label` yet (which we probably shouldn't given it's late in the day) maybe we should just remove these label overrides added by Emil. He added them mostly to make the internal date label and other labels consistent. But as we're moving the label outside the date for now, that won't be a problem anymore. The left label panel label bottom margins will still be inconsistent with some of the right panel label bottom margins - but they were not completely consistent with all the right panel labels anyway.

stevenao:
image

stevenao:
> If we don't want to try to use `d2l-input-label` yet (which we probably shouldn't given it's late in the day) maybe we should just remove these label overrides added by Emil. He added them mostly to make the internal date label and other labels consistent. But as we're moving the label outside the date for now, that won't be a problem anymore. The left label panel label bottom margins will still be inconsistent with some of the right panel label bottom margins - but they were not completely consistent with all the right panel labels anyway.
> 
> Did not mean to close ! :-)

yes, the due date label will need to be tweaked.

mdgbayly (approved):
ok to get this merged and to do any remaining label fix up on follow up PR

mdgbayly (approved):

PR #1085, opened 2020-09-09
Fix menu imports

emil-furniss:
The core menu components were [recently unbundled](https://github.com/Brightspace/brightspace-integration/pull/2435), and we were not explicitly importing them so the menu components in the grades dropdown were not being rendered properly.


wongvincent (approved):

wongvincent (commented):
There are other places where the dropdown-menu is being used, are there going to be other PRs to fix those?

emil-furniss:
@wongvincent Importing them in one place actually fixes all locations.

mostlyfabulous (dismissed):

mostlyfabulous:
Good find @emil-furniss

emil-furniss:
Decided to import them in the other locations that might need it as well. Technically a single import within the page will do, but if that component is not loaded (for whatever reason) then the others would be in trouble.

components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -2,6 +2,8 @@ import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import 'd2l-dropdown/d2l-dropdown-button-subtle.js';
 import 'd2l-dropdown/d2l-dropdown-menu.js';
```
wongvincent, line 4: think you need to replace these

emil-furniss: Nah, technically not. Replaced them in the other location out of habit, but the dropdown imports are not the problem.

emil-furniss: It's one of the legacy imports that technically still work. We could go around and redirect all of them at the new core versions, but I think that's out of scope for this.

wongvincent (approved):

components/d2l-activity-editor/d2l-activity-conditions-editor.js
dbatiste: The old legacy imports (referenced here) were updated to point to core when we did the migration, which is why they still work, but it results in an extra HTTP request in the unbundled scenario.  We definitely would like to see this switched eventually.

PR #1093, opened 2020-09-10
US118988 Disable assignment loading shim to see skeleton loading

mostlyfabulous:
https://rally1.rallydev.com/#/detail/userstory/408693111172?fdp=true

**This removes the logic that shows the "Loading..." text and shim** to observe the `skeleton` behaviour

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-editor.js
@@ -41,6 +41,13 @@ class ActivityEditor extends ActivityEditorTelemetryMixin(AsyncContainerMixin(Ac
 			<div id="editor-container">
 				<slot name="editor"></slot>
 			</div>
+			<d2l-backdrop
+				for-target="editor-container"
+				?shown="${this._backdropShown}"
+				no-animate-hide
+				delay-transition
+				slow-transition>
+			</d2l-backdrop>
 		`;
 	}
 	update(changedProperties) {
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -38,17 +38,9 @@ class ActivityEditor extends ActivityEditorTelemetryMixin(AsyncContainerMixin(Ac
...
+			<div id="editor-container">
 				<slot name="editor"></slot>
 			</div>
-			<d2l-backdrop
-				for-target="editor-container"
-				?shown="${this._backdropShown}"
-				no-animate-hide
-				delay-transition
-				slow-transition>
-			</d2l-backdrop>
```
emil-furniss, line 51: We won't want to get rid of this backdrop component since it's unrelated to loading. Instead it provides a semi-transparent shim while the editor saves.

mostlyfabulous: I'll add that back

components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -38,17 +38,9 @@ class ActivityEditor extends ActivityEditorTelemetryMixin(AsyncContainerMixin(Ac
 
 	render() {
 		return html`
-			<div ?hidden="${this.asyncState === asyncStates.complete}" class="d2l-activity-editor-loading">${this.localize('editor.loading')}</div>
-			<div id="editor-container" ?hidden="${this.asyncState !== asyncStates.complete}">
```
emil-furniss, line 42: Is this the permanent removal? In which case it might be nice to clean out the now-unused imports, styling, etc

mostlyfabulous: I think we want it till we're sure the loading text is going to be removed

emil-furniss (approved):

PR #1095, opened 2020-09-10
no longer notify me via email

AlexBedley:
it was there before so i could get a pulse on how the build was going


zinabat (approved):
But won't you miss the spam tho? 💌

PR #1098, opened 2020-09-10
Revert to old date pickers for 20.10 release

emil-furniss:
Pretty straight revert of the main switch PR #1067, as well as the smaller changes in  #1087 and #1094.

Also includes hiding the due date picker while the `skeleton` prop is true.


wongvincent (approved):
And then after branching, we can revert the revert!!

mdgbayly (approved):

PR #1102, opened 2020-09-11
stop attachments row skeleton flicker

emil-furniss:


mdgbayly (approved):

debug.log
```diff
@@ -0,0 +1 @@
+[0910/120138.835:ERROR:prune_crash_reports.cc(35)] PruneCrashReportDatabase: Failed to get pending reports
```
mostlyfabulous, line 1: What created this?

emil-furniss: Shoot good catch, I'll remove that. I think node sometimes gives me one of those.

wongvincent (approved):

PR #1104, opened 2020-09-11
Revert "Revert to old date pickers for 20.10 release (#1098)"

emil-furniss:
This reverts commit b8c0d07dcca0834ca7b1fa2b344e490f713e8d2c. from #1098


wongvincent (approved):
Nice. Revert the revert of the revert of the revert of the revert of the revert of the revert of the revert of the revert

PR #1113, opened 2020-09-17
Creating activity-content-editor to align with pre-existing activity editors

imcdonald-d2l:
In summary, this PR starts the work to align our face create experience with the approach utilized by assignments. It adds the name field to the UI and a MobX store to keep track of this value and content activities.

Specifically the work done in this PR:
- Created `<d2l-activity-content-editor>` this is the parent component/editor that all future components will be added to. It contains the `<d2l-activity-editor>` and template component which we fill with our custom components for the primary, secondary and footer slots.
- Created `d2l-activity-content-editor-detail`. This is where the title input field exists. Mobx hooks have been added here to get  the title from state when rendering and updating the state with any updates made to this field. UPDATE: also added description field but this input is not hooked up to the MobX store. 
- Created `d2l-activity-content-editor-secondary`. This is right hand panel drop-downs exist. No functionality here right now. Just HTML and CSS to get basic component shells ready for future work.
- Created `d2l-activity-content-editor-footer`. This is where the save/cancel buttons exist. Right now they don't function but I have added the UI components for future tweaking when we properly setup the save/cancel functionality.
- Created a MobX store with `content-store.js` and `content.js `. These two files allow us to load the content activity items into the store while accessing and editing the `title` property. 

This is what the UI looks like right now:
image

Description edit options:
description

**Note:** you won't be able to test this just yet as back end requires a bit more work. Putting this PR up early so you can ask any questions you may have.


components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,111 @@
...
+		const isTitleEmpty = (title || '').trim().length === 0;
+
+		const errorProperty = '_titleError';
+		const emptyNameErrorLangterm = 'content.emptyNameField';
+		const tooltipId = 'title-tooltip';
+
+		if (isTitleEmpty) {
+			this.setError(errorProperty, emptyNameErrorLangterm, tooltipId);
+		} else {
+			this.clearError(errorProperty);
```
imcdonald-d2l, line 102: `this.setError` and `this.clearError` come from `ErrorHandlingMixin`

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+
+	hasPendingChanges() {
+		// TODO
+		return false;
+	}
+
+	async save() {
+		// TODO
+		return;
+	}
```
imcdonald-d2l, line 80: These two functions will be used by `ActivityEditorContainerMixin` when we get around to implementing the save/cancel functionality

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+			</d2l-activity-editor>
+		`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if ((changedProperties.has('href') || changedProperties.has('token')) &&
+			this.href && this.token) {
+			super._fetch(() => store.fetchContentActivity(this.href, this.token));
```
imcdonald-d2l, line 68: this gets the content activity and adds it to the store

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -0,0 +1,43 @@
...
+		if (sirenEntity) {
+			const entity = new ContentEntity(sirenEntity, this.token, { remove: () => { } });
+			this.load(entity);
+		}
+		return this;
+	}
+
+	load(entity) {
+		this._entity = entity;
+		this.title = entity.title();
```
imcdonald-d2l, line 26: `title()` function comes from the defined function in siren-sdk repo [here](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/content/ContentEntity.js#L10).

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,111 @@
+import 'd2l-inputs/d2l-input-text.js';
+import 'd2l-tooltip/d2l-tooltip';
+
+import { css, html } from 'lit-element/lit-element.js';
+import { ContentEntity } from 'siren-sdk/src/activities/content/ContentEntity.js';
+import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
+import { ErrorHandlingMixin } from '../error-handling-mixin.js';
```
ianlaidlaw, line 8: small nit that we could get in the habit of doing: it will look better if we group the imports with all of our absolute (dependency) imports above all the relative imports. It will help give an idea of what each file is using at a glance.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,111 @@
...
+					prevent-submit
+					novalidate
+				>
+				</d2l-input-text>
+				${this._getTitleTooltip()}
+			</div>
+		`;
+	}
+
+	_getTitleTooltip() {
```
ianlaidlaw, line 69: I might be thinking in React, but should this function be named `renderTitleTooltip` ?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,111 @@
...
+	_saveTitle(value) {
+		store.getContentActivity(this.href).setTitle(value);
+	}
+
+	_saveTitleOnInput(e) {
+		const title = e.target.value;
+		const isTitleEmpty = (title || '').trim().length === 0;
+
+		const errorProperty = '_titleError';
+		const emptyNameErrorLangterm = 'content.emptyNameField';
```
ianlaidlaw, line 96: nit: redeclaring these on every save feels a little redundant, maybe we can have them defined somewhere else or do them inline?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-footer.js
```diff
@@ -0,0 +1,53 @@
...
+				display: none;
+			}
+			d2l-activity-visibility-editor {
+				display: inline-block;
+			}
+			.d2l-activity-content-editor-footer-left {
+				align-items: baseline;
+				display: flex;
+				flex: 1;
+				flex-direction: row-reverse;
```
ianlaidlaw, line 24: any reason to use `row-reverse` and not just row and have the elements in the right order?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { shared as store } from './state/content-store.js';
+
+class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditorMixin(MobxLitElement))) {
+
+	static get properties() {
+		return {
+			widthType: { type: String, attribute: 'width-type' }
```
ianlaidlaw, line 17: can you explain what's going on with this and why it's needed?

ianlaidlaw: Like why it's a property

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+			:host([dir="rtl"]) d2l-icon {
+				padding-left: 1rem;
+				padding-right: 0;
+			}
+		`;
+	}
+
+	constructor() {
+		super(store);
+		this.type = 'content';
```
ianlaidlaw, line 46: nit: can this just be inlined? seems to only be used once

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+				.href=${this.href}
+				.token=${this.token}
+			>
+				${this._editorTemplate}
+			</d2l-activity-editor>
+		`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
```
ianlaidlaw, line 64: what does the parent `updated` do and does it need to be inside your check below?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: I think we may be limited to ordering imports alphabetically in the activities repo. Every file I'm seeing has them ordered that way 😞

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-footer.js
imcdonald-d2l: That's a good point. Just copied the styles from the `assignment-editor-footer` but will clean these up so it is more logical.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ianlaidlaw: ah, that's fair. we dont need to stray from the existing rules

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: Yeah, would probably make more sense to be named that way. 👍

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: That's a good point, inline would work nicely!

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+		return css`
+			:host {
+				display: block;
+			}
+			div[slot="primary"] {
+				padding: 20px;
+			}
+			div[slot="secondary"] {
+				background: var(--d2l-color-gypsum);
+				height: calc(100% - 20px);
```
ianlaidlaw, line 31: This kinda stuff makes me nervous, LX had a lot of these calcs with seemingly "random" pixel offsets. I would suggest finding a different way or at least putting these into CSS vars

ianlaidlaw: Like if the 20px in the primary slot div changes and this doesn't, things are gonna look weird and it might be hard to spot why

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+				padding: 10px;
+			}
+			d2l-icon {
+				padding-right: 1rem;
+			}
+			:host([dir="rtl"]) d2l-icon {
```
ianlaidlaw, line 37: kinda sucks that we have to duplicate & inverse the css for rtl stuff... I think Lessons used some library that handled that for us, I wonder if this repo has anything like that? is this what other files are doing?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
imcdonald-d2l: Good point, it's a result of the `padding` property right below this one. Will swap out value for a variable 👍

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
imcdonald-d2l: This was a copy/paste from `assignment-editor-detail` component. Looks like some files are taking a manual approach for tweaking `rtl` stuff. Not seeing any immediate libraries being used.

dbatiste: 👍 This is the standard approach that needs to be applied.  Typically, the RTL assignments are pretty trivial, but eventually the browsers will all support CSS properties that are agnostic to direction. ex. start/end vs left/right.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
imcdonald-d2l: good catch, the `super.updated()` isn't actually needed here. `super` is referencing  `ActivityEditorMixin` in this file. Debugging led me to find that `super.updated()` which goes to [this line](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js#L64) would execute the same thing as the block below but doesn't execute properly. Since `store` ends up being `undefined`.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
imcdonald-d2l: this property is used on line 84 within this file and set on the BE [here](https://search.d2l.dev/xref/lms/le/content/D2L.LE.Content/Web/Desktop/Views/Lessons/ContentCreationViewFactory.cs?r=0ccdf8d6#41). It sets the width type of the template component. Arguably it will never change from `fullscreen` but doesn't hurt to have the ability to change that in the future.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				--d2l-secondary-padding: 10px;
+				display: block;
+			}
+			div[slot="primary"] {
+				padding: 20px;
```
ianlaidlaw, line 28: is this supposed to be `2 * --d2l-secondary-padding` ?

imcdonald-d2l: I can add a new variable for primary padding and set the height accordingly 👍

components/d2l-activity-editor/d2l-activity-content-editor/state/content-store.js
```diff
@@ -0,0 +1,23 @@
...
+		return this._contents.get(href);
+	}
+
+	put(href, object) {
+		this._contents.put(href, object);
+	}
+}
+
+export const shared = new ContentStore();
+export const { _contents: contents } = shared;
```
ianlaidlaw, line 23: is this used anywhere? seems like only the `shared` is being used. i might be wrong though

imcdonald-d2l: haha nothing gets passed you! No, it is not being used right now. I removed it. Good catch 👍

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,107 @@
...
+	render() {
+		const contentEntity = store.getContentActivity(this.href);
+		if (!contentEntity) {
+			return html``;
+		}
+		const {	title } = contentEntity;
+
+		return html`
+			<div id="content-title-container">
+				<label class="d2l-label-text" for="content-title">${this.localize('content.name')}*</label>
```
ChrisKraljevicD2L, line 53: Why is there an asterix after `${this.localize('content.name')}`?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	_renderTitleTooltip() {
+		if (this._titleError) {
+			return html`
+				<d2l-tooltip
+					id="title-tooltip"
+					for="content-title"
+					position="bottom"
+					?showing="${this._titleError}">
```
ChrisKraljevicD2L, line 94: Does `_titleError` work as a falsy statement? No `!!` needed?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,107 @@
...
+		const title = e.target.value;
+		const isTitleEmpty = (title || '').trim().length === 0;
+
+		if (isTitleEmpty) {
+			this.setError('_titleError', 'content.emptyNameField', 'title-tooltip');
+		} else {
+			this.clearError('_titleError');
+			this._debounceJobs.title = Debouncer.debounce(
+				this._debounceJobs.title,
+				timeOut.after(500),
```
ChrisKraljevicD2L, line 119: I'm pretty consistent with being nitty about magic numbers and this is no different :P Where does this 500ms timeout come from?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: So it shows up as `Name*` instead of just `Name`. UX thing to help users recognize it's mandatory. Assignments does this but we don't have to follow suit if we don't want.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: haha agreed, this logic was copied from `assignments-detail` component. Just kept their debounce value for the title field. Assumed it was safe if we followed suit. Can make this a more legitimate variable if we feel it's necessary?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ChrisKraljevicD2L: Aaaah gotcha. Seemed a little weird with the asterix not being part of a langterm but with the explanation it makes sense :+1:

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ChrisKraljevicD2L: I think making the timeout value a constant somewhere here in `activities` would be a good idea.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: it seems to be working properly so I assume so ... will add the `!!` to be on the extra safe side though 👍

ChrisKraljevicD2L: https://travis-ci.com/github/BrightspaceHypermediaComponents/activities/jobs/387124504#L506 this isn't related to the `!!` change is it? Seems like a failing test in a seemingly unrelated component 🤔

imcdonald-d2l: Yeah, seems completely unrelated. Re-running the tests seems to make it happy.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ianlaidlaw: Yeah this is fine temporarily for now, but imo there should be something handled at the component level, like a `<d2l-label />` component that can take in a `required` property and then add anything it needs to accordingly. Whether it's adding an *, adding colour, etc.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	_onRichtextChange(e) {
+		// TODO - properly handle description content
+		// eslint-disable-next-line no-unused-vars
+		const content = e.detail.content;
+	}
+
+	_renderTitleTooltip() {
+		if (this._titleError) {
```
ianlaidlaw, line 90: Lit question, if this is false, nothing gets returned here. Is that alright, or do we need to return html`` ?

imcdonald-d2l: It works just fine as is so I see no reason to return empty `html`.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -0,0 +1,127 @@
...
+			return html``;
+		}
+		const {	title } = contentEntity;
+
+		return html`
+			<div id="content-title-container">
+				<label class="d2l-label-text" for="content-title">${this.localize('content.name')}*</label>
+				<d2l-input-text
+					id="content-title"
+					maxlength="128"
```
CodeBaboon, line 58: Why 128? I wonder if this should be some kind of shared constant? I wonder if this should (eventually) be coming back from the hypermedia update/create action in some way?

CodeBaboon (dismissed):
I think this seems good.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: This was the same value used by the `assignments-editor-detail` component. Was duplicating their logic. Totally open to other approaches if we feel this should be improved.

ChrisKraljevicD2L: Is there a max length for titles of units in the db? We could use that instead. I'm anticipating clients complaining that they can't have obscenely long names for units anymore.

ChrisKraljevicD2L (approved):
LGTM, the title max length comment is more of a nit.

ChrisKraljevicD2L (approved):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
pbrown-d2l: It is set to 150 in the DB

`[Title] [nvarchar](150) NOT NULL,`

pbrown-d2l (changes requested):
W should have the title length match the DB column,

CodeBaboon (approved):
Sorry thought I approved with my earlier review.

pbrown-d2l (approved):

PR #1115, opened 2020-09-18
Strings that are too long for a card now wrap

ChrisLabattD2L:
For DE39598

https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fdefect%2F399579266472

Card Metadata now centers because the overflow caused by long strings now was solved by wrapping the content to the next line.

See difference below

Before:
Before

After:
After

### DIFF
--- a/components/d2l-activity-card/d2l-activity-card.js
+++ b/components/d2l-activity-card/d2l-activity-card.js
@@ -44,6 +44,7 @@ class D2lActivityCard extends ActivityCardLocalize(PolymerElement) {
 
 				.d2l-activity-card-content-organization-info {
 					display: block;
+					word-break: break-word;
 				}
 
 				.d2l-activity-card-activity-information {


BhamelinD2L (approved):
Good first PR! I'm confident this will solve the issue. I believe this repo will automatically handle versioning, so you should be all set to merge this. 💯

Some additional things to note -
Because your change is limited to `d2l-activity-card-content-organization-info` class, it won't affect the 'course' element, and it seems to still be possible to cause this issue by changing 'course' to a massive long word. However, we're in control of that word, rather than being user defined like semester name, so it's not really a concern.

If you look at the css of the parent class, `d2l-activity-card-content-container`, you'll see its already making an attempt to wrap all the text using overflow-wrap: break-word;, but it doesn't work. Display: flex seems to prevent `overflow-wrap: break-word` from happening. It doesn't stop word-break: `break-word;` though. css is fun 😄

PR #1116, opened 2020-09-18
US120338 - Group filter, sort and search extra params inside one single query string param

amauryperez:
This changes affects the information that is send back from the evaluation pages when they return to Quick Eval, allowing Quick Eval to maintain the previous selected sort, filter and search.

It will allow the new evaluation routes to handle query string params easily.

Before:
?filter=eyJmMmIzMmYwMy01NTZ&sort=eyJTb3J0cyI6W3s0&collectionSearch=eddie

After
?cfi=filter_eyJmMmIzMmYwMy01NTZ-sort_eyJTb3J0cyI6W3s0-collectionSearch_eddie

Legacy LMS pages are been changed at the same time to handle the new single param

### DIFF
--- a/components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
+++ b/components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
@@ -90,7 +90,7 @@ D2L.PolymerBehaviors.Siren.D2LSirenHelperBehaviorImpl = {
 			extraParams.push(
 				{
 					name: 'collectionSearch',
-					value: searchVal
+					value: window.btoa(searchVal)	// base64 encoding of the search text
 				}
 			);
 		}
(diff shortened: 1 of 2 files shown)


AlexBedley (approved):

components/d2l-quick-eval/behaviors/d2l-siren-helper-behavior.js
```diff
@@ -90,7 +90,7 @@ D2L.PolymerBehaviors.Siren.D2LSirenHelperBehaviorImpl = {
 			extraParams.push(
 				{
 					name: 'collectionSearch',
-					value: searchVal
+					value: window.btoa(searchVal)	// base64 encoding of the search text
```
amauryperez, line 93: this is to avoid issues if the search text includes `-` or `_`

amauryperez: I think you need to re-approve :-)

AlexBedley (approved):

PR #1121, opened 2020-09-23
Add key identifier for component to update

JamieTran:
Related to [US120640](https://rally1.rallydev.com/#/388869855380d/dashboard?detail=%2Fuserstory%2F436813346276)

`d2l-activity-text-editor` does not respond to value changes because the html editor maintains it's own state. 
With the addition of the unique `key` property, we can use it to track whether the component should be updated or not.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-text-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-text-editor.js
@@ -10,6 +10,7 @@ class ActivityTextEditor extends LitElement {
 			richtextEditorConfig: { type: Object },
 			disabled: { type: Boolean },
 			ariaLabel: { type: String },
+			key: { type: String },
 		};
 	}
 
@@ -28,6 +29,7 @@ class ActivityTextEditor extends LitElement {
 			return html`
 				<d2l-activity-html-editor
 					ariaLabel="${this.ariaLabel}"
+					.key="${this.key}"
 					.value="${this.value}"
 					?disabled="${this.disabled}"
 					@d2l-activity-html-editor-change="${this._onRichtextChange}"
--- a/components/d2l-activity-editor/d2l-activity-html-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-editor.js
@@ -14,6 +14,7 @@ class ActivityHtmlEditor extends LocalizeActivityEditorMixin(LitElement) {
 			value: { type: String },
 			ariaLabel: { type: String },
 			disabled: { type: Boolean },
+			key: { type: String },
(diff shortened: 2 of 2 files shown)


AlexBedley:
Hey Jamie, do you mind explaining a bit more?

JamieTran:
@mdgbayly expained this well: 

The main reason why we don't respond to value changes in the html editor after initialization is because the html editor maintains it's own state, such as selections for example. So if we treat it like a standard component whose state always reflects it's properties like `value`, then you run into situations like the following:
1. User is editing the text and selects a word and chooses bold
2. Component sends a change event with the updated text
3. Host application updates it's local state with the latest text value and this causes a re-render.
4. Re-render causes html editor to be updated with latest text value, so the value is correct but any 'selection' object in the HTML editor is discarded, so the editor loses it's selection state which is not desired.

You'd see this behavior if you remove the extra `&& typeof changedProperties.get('value') === 'undefined'` in your local build.

This hasn't been a problem for us with this implementation so far as we're generally working with the contents of the editor as is and it's ok to allow it to manage it's own state. However, we've experienced similar issues to you in other scenarios where we've had to integrate tinymce into web components.

e.g. here in a Polymer implementation of the html editor in rubrics
https://github.com/Brightspace/d2l-rubric/blob/master/editor/d2l-rubric-html-editor.js#L208

The solution we used there is to give the text editor/html editor an additional `key` property that we could use to track whether to update the editor's internal value e.g. because the part of the state it is rendering changed. So I think we could do something similar here too.

You could add a key property to the `d2l-activity-text-editor` and `d2l-activity-html-editor` and then change the `updated` implementation to have `|| changedProperties.has('key')`

Then you would just need to set the `d2l-activity-text-editor` `key` property/attribute to some value that was unique enough to change whenever the student/feedback that you want to render changes. e.g. it could be the student identifier or feedback identifier.

components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -28,6 +29,7 @@ class ActivityTextEditor extends LitElement {
 			return html`
 				<d2l-activity-html-editor
 					ariaLabel="${this.ariaLabel}"
+					.key="${this.key}"
```
mdgbayly, line 32: Looks good to me - not entirely sure we need to use the property setter here. I think we were using the property setter on the value for other reasons related to handling undefined and it may not be an issue for your key, but it's not a big deal.

mdgbayly (approved):
LGTM

AlexBedley (approved):
thanks for the explanation!

PR #1123, opened 2020-09-24
Integrate content editor with activity store to get visibility and availability features for free

imcdonald-d2l:
Integrating our content editor with `activity-store.js`. This file/store gives us boilerplate access to standard `activity-usage` properties in the store so that we can integrate with shared components like `d2l-activity-availability-dates-editor` and `2l-activity-availability-dates-summary`. These components integrate with `activity-store.js` to both access and set values on the activity usage entity. Gives us a lot of functionality with doing almost no work! 

**Note** had to fix an error being thrown when loading a new `activity-usage` into the store that naively assumed the `activity-usage` entity would always have score/grade information. This needed to be fixed for the `ObjectStore` to properly load the entity into state so that other components could access it. I know this is confusing to follow without having worked in these files so let me know if you have questions.

**EDIT** some work needed in `siren-sdk` repo as well. Similar issue with assuming `scoreOutOf` exists and `toString()` fails. Related [PR here](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/217).

Screenshot with new visibility and availability components:
visibility + availability integration

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
@@ -1,3 +1,5 @@
+import '../d2l-activity-availability-dates-summary.js';
+import '../d2l-activity-availability-dates-editor.js';
 import '@brightspace-ui-labs/accordion/accordion-collapse.js';
 import { bodySmallStyles, heading3Styles, heading4Styles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
@@ -81,8 +83,11 @@ class ContentAvailabilityEditor extends LocalizeActivityEditorMixin(RtlMixin(Lit
 	_renderAvailabilityDatesEditor() {
 
 		return html`
-			<div>
-				TODO - add availability dates editor
+			<div class="d2l-editor">
+				<d2l-activity-availability-dates-editor
+					href="${this.href}"
+					.token="${this.token}">
+				</d2l-activity-availability-dates-editor>
 			</div>
 		`;
 	}
@@ -90,9 +95,10 @@ class ContentAvailabilityEditor extends LocalizeActivityEditorMixin(RtlMixin(Lit
 	_renderAvailabilityDatesSummary() {
 
 		return html`
-			<div>
-				TODO - add availability dates summary
-			</div>
+			<d2l-activity-availability-dates-summary
+				href="${this.href}"
+				.token="${this.token}">
+			</d2l-activity-availability-dates-summary>
 		`;
 	}
 }
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-footer.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-footer.js
@@ -37,7 +37,8 @@ class ContentEditorFooter extends SaveStatusMixin(RtlMixin(LitElement)) {
 		return html`
 			<div class="d2l-activity-content-editor-footer-left">
 				<d2l-activity-editor-buttons></d2l-activity-editor-buttons>
-				<!-- TODO: more work/investigation is needed to get visibility editor working -->
+				<!-- TODO: figure out how to set activity usage entity "canEditDraft" property to true -->
+				<!-- so that we can properly toggle the visibility of the entity -->
 				<d2l-activity-visibility-editor
 					.href="${this.href}"
 					.token="${this.token}">
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
@@ -7,6 +7,7 @@ import '@brightspace-ui/core/components/colors/colors.js';
 import { css, html } from 'lit-element/lit-element.js';
 import { ActivityEditorContainerMixin } from '../mixins/d2l-activity-editor-container-mixin.js';
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
+import { shared as activityStore } from '../state/activity-store.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { shared as store } from './state/content-store.js';
@@ -68,6 +69,7 @@ class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditor
 		if ((changedProperties.has('href') || changedProperties.has('token')) &&
 			this.href && this.token) {
 			super._fetch(() => store.fetchContentActivity(this.href, this.token));
+			super._fetch(() => activityStore.fetch(this.href, this.token));
 		}
 	}
 
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -6,7 +6,7 @@ configureMobx({ enforceActions: 'observed' });
 export class ActivityScoreGrade {
 
 	constructor(entity, token) {
-		this.scoreOutOf = entity.scoreOutOf()?.toString();
+		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : undefined;
 		this.scoreOutOfError = null;
 		this.token = token;
 		this.inGrades = entity.inGrades();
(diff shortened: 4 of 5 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
```diff
@@ -81,18 +83,22 @@ class ContentAvailabilityEditor extends LocalizeActivityEditorMixin(RtlMixin(Lit
...
 	_renderAvailabilityDatesSummary() {
 
 		return html`
-			<div>
-				TODO - add availability dates summary
-			</div>
+			<d2l-activity-availability-dates-summary
+				href="${this.href}"
+				.token="${this.token}">
+			</d2l-activity-availability-dates-summary>
```
imcdonald-d2l, line 101: Functionality of these components comes free when properly hooking into the `activity-store` 😄

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-footer.js
```diff
@@ -37,7 +37,8 @@ class ContentEditorFooter extends SaveStatusMixin(RtlMixin(LitElement)) {
 		return html`
 			<div class="d2l-activity-content-editor-footer-left">
 				<d2l-activity-editor-buttons></d2l-activity-editor-buttons>
-				<!-- TODO: more work/investigation is needed to get visibility editor working -->
+				<!-- TODO: figure out how to set activity usage entity "canEditDraft" property to true -->
+				<!-- so that we can properly toggle the visibility of the entity -->
```
imcdonald-d2l, line 41: Visibility editor is not currently allowing you to toggle visibility state because the `canEditDraft` property on the activity usage entity is set to `false`... I'm assuming this is something we need to address in the back-end @CodeBaboon ?

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -68,6 +69,7 @@ class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditor
 		if ((changedProperties.has('href') || changedProperties.has('token')) &&
 			this.href && this.token) {
 			super._fetch(() => store.fetchContentActivity(this.href, this.token));
+			super._fetch(() => activityStore.fetch(this.href, this.token));
```
imcdonald-d2l, line 72: This loads the activity usage entity for our content item into the `activityStore`

components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -6,7 +6,7 @@ configureMobx({ enforceActions: 'observed' });
 export class ActivityScoreGrade {
 
 	constructor(entity, token) {
-		this.scoreOutOf = entity.scoreOutOf().toString();
+		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : null;
```
imcdonald-d2l, line 9: @mdgbayly added you to this PR since you were on the latest commit for this piece of code. Do you mind confirming that this doesn't break any of your functionality? Was getting the following error when loading our content activity usage entity into the store:
toString_error

Had to do the following above so an error wouldn't be [caught here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/state/object-store.js#L26) and prevent the store from loading the entity into state. Error is a result of `scoreOutOf()` returning `undefined` on our content activity usage entity.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
ianlaidlaw: so to be clear, this will go hit the href, get the entity and then put it in the store. and since that entity has availability dates stuff on it for free, we can just call those other components with this href and we will have what we need?

ianlaidlaw: also what is the difference between the above line:
```
super._fetch(() => store.fetchContentActivity(this.href, this.token));
```
store.fetchContentActivity and activityStore.fetch just sound so similar lol

imcdonald-d2l: To your first question, yes, the `activityStore` will utilize the `object-store` and get the `href` [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/object-store.js#L24) to load it into state. This allows you to call `activityStore.get(href)` to grab the entity from state (`activity-store` is wrapped with `object-store` so when you call `activityStore.get` it is actually calling the object-store defined function [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/object-store.js#L32)). 

To populate all the necessary `activity-usage` details the `activity-store` takes care of that [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/activity-usage.js#L32). 

So the components would grab the activity-usage entity using `activityStore.get(href)` where the `href` is the key in the object map. They then can access or update all `activity-usage` properties through the `activity-usage` object defined [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/activity-usage.js) utilized in the `activity-store` [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/activity-store.js#L6).

imcdonald-d2l: To your second question, `activityStore` implements the `activity-store.js` https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/state/activity-store.js. Where the `store` defined above implements the [custom store](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/d2l-activity-content-editor/state/content-store.js) created for `content-activity-usage` specific details like `title` defined [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js). More content specific details will be added to state as we come across them.

imcdonald-d2l: For further context `super._fetch()` is referencing [this function](https://github.com/BrightspaceHypermediaComponents/activities/blob/d4b8a3f95d08e867bdea5fbbe499f45866e63de4/components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js#L91) defined in `ActivityEditorMixin` . which takes care of tracking async state stuff.

imcdonald-d2l: I know this is a lot to digest so let me know if you need further clarification.

components/d2l-activity-editor/state/activity-score-grade.js
mdgbayly: I think this is fine, but I'm wondering if we would be better off handling things like this a little more pro-actively. I guess this is happening because we always construct an scoreOutOf object within the activity usage mobx state.

https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/state/activity-usage.js#L39

But possibly it would be better to only construct that object if the activity actually has a scoreOutOf entity. I guess it's a toss up because then you'd have to make other places in the mobx state that reference that scoreOutOf object handle it being null if you went that way.

I think you might run into quite a few things like this as you do initial integration of some of this stuff, because up to now it's only been used for FACE assignments, and it's hard to generalize/abstract when you only have a single implementation.

We're also starting to use this for FACE quiz this iteration, so I'm sure we'll hit many of the same issues.

imcdonald-d2l: Yes, it is a result of constructing `scoreAndGrade` object in the activity usage state. And I agree, it looks like it would be difficult to exclude it completely, especially if FACE quiz needs this information from their activity usage entity. Will leave implementation as is for now and keep iterating on this `activity-usage` entity definition as we run into more roadblocks/issues. Totally get that it is hard to generalize things right off the bat so will make adjustments as we go 👍

ianlaidlaw (approved):

pbrown-d2l (approved):

PR #1127, opened 2020-09-28
US120836 use SkeletonMixin from BrightspaceUI/core

mdulat:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F437879817656&fdp=true


mostlyfabulous (commented):
I've checked the https://github.com/BrightspaceUI/core/blob/master/components/skeleton/skeleton-mixin.js we're using instead of `SkeletizeMixin`
and the only thing to note is the 
```
constructor() {
	super();
	this.skeleton = false;
}
```

mostlyfabulous (dismissed):

mostlyfabulous:
It seems like the build can't find the new mixin: 
image

mdgbayly (approved):

PR #1129, opened 2020-09-30
Rfegan/us119463 refactor footer

ronan-f:
Summary

Start of work on FACE quizzing. The footer is going to be identical with assignments footer as per JC so this PR is to refactor the existing footer into one that can be reused between quizzing/assignments.


mdgbayly (approved):
Looks great

PR #1131, opened 2020-09-30
Rfegan/us119463 add quiz editor component

ronan-f:
New component based on [this story](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/413201069796).

The styling has been duplicated from `d2l-activity-assignment-editor` and there might be a better way to share styles between these 2 files.

It currently results in a view like this:

face-quiz

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js
@@ -13,6 +13,7 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityAssignmentEditorMixin } from './mixins/d2l-activity-assignment-lang-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { secondarySlotStyles } from '../shared-styles';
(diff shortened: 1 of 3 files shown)


mdgbayly:
Is your issue with the visibility toggle that it doesn't work? :-)
If so, it's probably not expected too yet. As the draft status API requires the quiz activity usage entity provider to plugin an API that can be used to update the draft status. I think we have an upcoming story for that. But when we get to it, I would like us to consider moving the implementation of that API to the activity usage API and then use a provider to actually update the status. Currently we expose the API through the activity usage but it is implemented on the assignments hypermedia domain as a completely separate API endpoint I think. If we could move it to the activity domain as a PATCH on the activity usage API it would cut down the number of API calls we have to make when saving.

ronan-f:
> Is your issue with the visibility toggle that it doesn't work? :-)
> If so, it's probably not expected to yet. As the draft status API requires the quiz activity usage entity provider to plugin an API that can be used to update the draft status. I think we have an upcoming story for that. But when we get to it, I would like us to consider moving the implementation of that API to the activity usage API and then use a provider to actually update the status. Currently we expose the API through the activity usage but it is implemented on the assignments hypermedia domain as a completely separate API endpoint I think. If we could move it to the activity domain as a PATCH on the activity usage API it would cut down the number of API calls we have to make when saving.

Thanks for explaining that. Makes sense. I was very confused as to the default canEditDraft status.

mdgbayly (approved):

PR #1133, opened 2020-10-01
Rfegan/us119463 fix back to quizzes btn

ronan-f:
As part of the AC for [US119463](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/413201069796) the "Back to manage quizzes" button should be working. 

Currently, the "back to manage <activity>" buttons rely on there being a dialog component available to trigger the confirmation flow. In this case, the confirmation dialog will be handled in a separate story and so I have added a check for when we have a non-existant dialog.

In this case if the dialog does not exist then we consider the "cancel" event to be completed and automatically redirect back to the manage page.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
@@ -1,9 +1,8 @@
 import { html, LitElement } from 'lit-element/lit-element.js';
 import { ActivityEditorContainerMixin } from '../mixins/d2l-activity-editor-container-mixin';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
-import { LocalizeActivityAssignmentEditorMixin } from '../d2l-activity-assignment-editor/mixins/d2l-activity-assignment-lang-mixin';
 
-class QuizEditor extends ActivityEditorContainerMixin(EntityMixinLit(LocalizeActivityAssignmentEditorMixin(LitElement))) {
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
```diff
@@ -1,9 +1,9 @@
 import { html, LitElement } from 'lit-element/lit-element.js';
 import { ActivityEditorContainerMixin } from '../mixins/d2l-activity-editor-container-mixin';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
-import { LocalizeActivityEditorMixin } from '../mixins/d2l-activity-editor-lang-mixin';
+import { LocalizeActivityAssignmentEditorMixin } from '../d2l-activity-assignment-editor/mixins/d2l-activity-assignment-lang-mixin';
```
ronan-f, line 4: This mixin is needed to have the correct langterms in `d2l-dialog-confirm` but it's stored within /d2l-activity-assignment-editor/mixins which implies it's assignment specific. At a quick glance it seems like quite a few of the lang terms can be used between different activites so possibly a shared mixin is suitable for this use case?

mdulat: yeah seems strange to have an "Assignment" specific localize mixin here, I think better to refactor the common terms into the general "Activity" editor localize mixin.

stevenao (approved):
LGTM

PR #1136, opened 2020-10-06
- Adding read only property to d2l-activity-attachments-editor

jmurray-d2l:
Hello! I am looking for a way to make this component readonly, basically we need to see the list of current attachments, but not allow the user to upload anymore.  I tried to use the `_canAddAttachments` but it looks private to me as well is looks to be reset in the renderer based upon the collection - let me know if you think there is a better way to do this!

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
@@ -10,10 +10,10 @@ class ActivityAttachmentsEditor extends ActivityEditorMixin(SkeletonMixin(MobxLi
 	static get properties() {
 		return {
 			readOnly : {
-				attribute: "read-only", 
-				type: Boolean 
+				attribute: 'read-only',
+				type: Boolean
 			},
-			_canAddAttachments: { 
+			_canAddAttachments: {
 				type: Boolean
 			}
 		};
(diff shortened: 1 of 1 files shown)


emil-furniss:
I think that the collection entity that this component is using should be setting the `canAddAttachments` property correctly. Is that not the case in your use case? Maybe @mdgbayly could chime in here?

Also, the private `_canAddAttachments` looks to be unused? Likely a vestige of the switch to mobX.

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js
```diff
@@ -9,7 +9,13 @@ import { shared as store } from './state/attachment-collections-store.js';
 class ActivityAttachmentsEditor extends ActivityEditorMixin(SkeletonMixin(MobxLitElement)) {
 	static get properties() {
 		return {
-			_canAddAttachments: { type: Boolean },
+			readOnly : {
```
mdgbayly, line 12: Out of interest how are you populating data for this control?
`canAddAttachments` is set based on the state of the actions in the associated siren entity here?
https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/AttachmentCollectionEntity.js#L34

So I would assume that if you don't want the user to add attachments, your attachments collection entity would not include any of those actions and so it would already be read only? Is that not working for you for some reason?

jmurray-d2l: using the existence of actions as well, it did not work initially, but I was unaware of this workflow - I will dig more into this, thanks!

jmurray-d2l:
I was using the wrong actions! Thank you all for the help, this is not needed

PR #1137, opened 2020-10-06
fix score initialized to undefined, and fix intermittent failing CI t…

mdulat:
…ests
Fixes https://trello.com/c/QpLFVmie/256-add-to-grades-default-value-showing-as-undefined
as well as intermittent failing CI tests due to store getting cleared between tests.

### DIFF
--- a/components/d2l-activity-editor/state/activity-score-grade.js
+++ b/components/d2l-activity-editor/state/activity-score-grade.js
@@ -6,7 +6,7 @@ configureMobx({ enforceActions: 'observed' });
 export class ActivityScoreGrade {
 
 	constructor(entity, token) {
-		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : undefined;
+		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : '';
 		this.scoreOutOfError = null;
 		this.token = token;
 		this.inGrades = entity.inGrades();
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -6,7 +6,7 @@ configureMobx({ enforceActions: 'observed' });
 export class ActivityScoreGrade {
 
 	constructor(entity, token) {
-		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : undefined;
+		this.scoreOutOf = entity.scoreOutOf() ? entity.scoreOutOf().toString() : '';
```
wongvincent, line 9: Would this change cause this line to always not match https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/217/files#diff-a1a6463aac8fe17f27e514f398c520fcR642
Because when `entity.scoreOutOf()` is `undefined`, then `scoreAndGrade.scoreOutOf` will be `''` and `scoreOutOf` will be `undefined`.

mdulat: yeah probably, can we set it to `''` there as well instead of `undefined`?

mdulat: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/219

imcdonald-d2l (approved):

PR #1139, opened 2020-10-06
US121067 use styled divs instead of external labels for score and ins…

mdulat:
…tructions; use internal label and skeleton for date-time
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F439155249712


mdgbayly (approved):
LGTM

PR #1141, opened 2020-10-08
Refactor the relationship between root editors, the d2l-activity-edit…

mdgbayly:
…or and the ActivityEditorContainerMixin to enable more shared functionality.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Ftask%2F440737131360

This PR is to work around some duplication that was occurring in how activity editors were being setup, particularly for things like using the page template, and having access to the confirmation dialog for handling save/cancel.

This does require a change in which component is now in control of all the editors (implements the `ActivityEditorContainerMixin`). It used to be the top level domain editor, e.g. `d2l-activity-assigment-editor`, but that made it difficult to share functionality like the cancel confirmation dialog, because the dialog was being duplicated in the local dom of each of those editors. To move it into the render of the `d2l-activity-editor` requires shifting the `ActivityEditorContainerMixin` implementation to the `d2l-activity-editor` instead. 

By moving the mixin to `d2l-activity-editor` component, it means that technically the root component e.g. `d2l-activity-assignment-editor` is no longer tracked as an editor (because it is a parent of the `d2l-activity-editor` which is now the `ActivityEditorContainerMixin`, and so it doesn't get registered using the `d2l-activity-editor-connected` event. 

So this does subtly change the role of the top level editor like `d2l-activity-assignment-editor`, `d2l-activity-quiz-editor` and `d2l-activity-content-editor` as previously they were implementing features like `save` and `hasPendingChanges` which was convenient, but also kind of meant that some of the lower level components are not really reusable in their own right. Going forward we should just consider that root component as the thing that knits together all the domain components. 

@imcdonald-d2l This shouldn't affect the content component too much yet as you don't have any save implementation (at least not committed) etc, but I did update it to remove the duplication of the template and some styling stuff. Tested it locally and looked OK to me but let me know if you see any issues.

The impact of changing which component implements the `ActivityEditorContainerMixin` on assignments is that for now I've just moved the save/cancel related implementations to the `d2l-activity-editor-assignment-detail` but I plan on putting up some follow on PRs that allow us to support these callbacks on multiple nested components. eg. we should be able to implement the `save` callback on both the `d2l-activity-assignment-editor-detail` and the `d2l-activity-assignment-editor-secondary` (or lower level components). The problem with that at the moment is that it would lead to multiple save calls. So I plan to enhance the save implementation to support that using promises to enable multiple save calls to be made against a single store, but where only one call will end up calling the API. We also needed to add some ordering support for assignments, that makes this more tricky too, but I have a PR in the works that will support the ordering for the multiple save calls too.

There are some other things I'd like to fix up with the way the base editor is implemented. e.g. currently we instantiate an ActivityUsage and an AssignmentActivityUsage (in order to get at the assignment href), but this is kind of pointless now because we can get the assignment href from the activity usage itself using the specialization rel href. So I plan to include that in a follow up PR too. That will simplify the assignment store.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
@@ -5,35 +5,28 @@ import './d2l-activity-content-editor-secondary.js';
 import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import { css, html } from 'lit-element/lit-element.js';
-import { ActivityEditorContainerMixin } from '../mixins/d2l-activity-editor-container-mixin.js';
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { shared as activityStore } from '../state/activity-store.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { shared as store } from './state/content-store.js';
 
-class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditorMixin(MobxLitElement))) {
+class ContentEditor extends RtlMixin(ActivityEditorMixin(MobxLitElement)) {
 
 	static get properties() {
 		return {
-			widthType: { type: String, attribute: 'width-type' }
+			widthType: { type: String, attribute: 'width-type' },
+			/**
+			 * Is Creating New
+			 */
+			isNew: { type: Boolean }
 		};
 	}
 
 	static get styles() {
 		return css`
 			:host {
-				--d2l-primary-padding: 20px;
-				--d2l-secondary-padding: 10px;
 				display: block;
-			}
-			div[slot="primary"] {
-				padding: var(--d2l-primary-padding);
-			}
-			div[slot="secondary"] {
-				background: var(--d2l-color-gypsum);
-				height: calc(100% - 2 * var(--d2l-secondary-padding));
-				padding: var(--d2l-secondary-padding);
 			}
 			d2l-icon {
 				padding-right: 1rem;
@@ -58,6 +51,9 @@ class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditor
 				telemetryId="content"
 				.href=${this.href}
 				.token=${this.token}
+				width-type="${this.widthType}"
+				error-term=""
+				?isnew="${this.isNew}"
 			>
 				${this._editorTemplate}
 			</d2l-activity-editor>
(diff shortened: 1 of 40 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -84,29 +80,21 @@ class ContentEditor extends ActivityEditorContainerMixin(RtlMixin(ActivityEditor
```
imcdonald-d2l, line 80: Everything looks good from my perspective and is functioning as expected but just to confirm my understanding, the `hasPendingChanges()` and `save()` functions above will have to move to a different child component of `d2l-activity-editor` (`content-editor-detail` for example) since the `ActivityEditorContainerMixin` has been moved to `d2l-activity-editor`?

imcdonald-d2l (approved):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
mdgbayly: @imcdonald-d2l Yes, that's correct because the root components won't be added to the list of editors in the `ActivityEditorContainerMixin` as they are the parent of the `d2l-activity-editor` which now implements that mixin, so the connected events they are currently sending won't be handled. Conceptually this means that these root components are no longer editors. I have not removed the usage of the `ActivityEditorMixin` from the root components yet, because that is where they are also getting their href/token attributes from.

Regarding the `hasPendingChanges` and `save` methods, yes, they would move to say `d2l-activity-content-editor-detail` in your case.

Technically if you wanted these components to be reusable standalone, you might want to implement those methods on multiple child components. e.g. if you had secondary components that could be used independently of the detail component, then they should also implement the pending/save methods. The implementation would be identical in each case, but you'd need to make the store component handle that fact that it might be called multiple times as part of the same save operation. I have a PR in the works for assignments that handles that by tracking whether a save is in progress in the assignment store (using the promise from the first API call) and returning that for subsequent save calls. It makes things slightly more complex and I guess is the price of trying to make these components decoupled and be able to "live alone". Always seems to be more complicated to do that than have everything just managed in a single place. So if there are other ideas...

Also there is the `hmc-foundation` stuff in the works that might have other ideas on how to handle all of this which I plan to look at.

imcdonald-d2l: Perfect, thanks for clarifying that! In regards to the save functionality, that hasn't been implemented quite yet but I assume we will start with the action in one location and add it to the other components to make them more stand-alone as we wrap up the need-to-have features to get this UI functioning properly. Unless it isn't a big lift to implement at first. Either way I'll be keeping eyes on the work you're doing along with any `hmc-foundation` work to make sure we implement best practices and provide feedback if I have any. 👍

stevenao (approved):

PR #1142, opened 2020-10-08
Add component/state tests for new content FACE related components

imcdonald-d2l:
- Adds basic accessibility testing to `d2l-activity-content-editor` and `d2l-activity-content-editor-detail` components. This is not needed for secondary or footer components since they are only using shared components right now.
- Adds a test to ensure fetching content works as expected. 

`content.spec` should be expanded on as we add more properties and actions to content state.


test/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.html
```diff
@@ -0,0 +1,42 @@
...
+	<body>
+
+		<test-fixture id="basic">
+			<template>
+				<d2l-activity-content-editor-detail></d2l-activity-content-editor-detail>
+			</template>
+		</test-fixture>
+
+		<script type="module">
+			import { runAxe } from '@brightspace-ui/core/tools/a11y-test-helper.js';
```
imcdonald-d2l, line 25: Link to the [shared accessibility file](https://github.com/BrightspaceUI/core/blob/master/tools/a11y-test-helper.js#L15) if you're curios.

CodeBaboon (dismissed):
Been awhile since I wrote/reviewed a11y or Jest tests so it's probably good to get more eyes on this, but fwiw I think this looks good (and adding tests is always good!)

sarahboyd (dismissed):

test/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+		<title>d2l-activity-content-editor-detail tests</title>
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script src="/node_modules/mocha/mocha.js"></script>
+		<script src="/node_modules/chai/chai.js"></script>
+		<script src="/node_modules/@polymer/test-fixture/test-fixture.js"></script>
```
pureooze, line 10: Would it be possible to use `karma` and `open-wc` for these instead of `polymer test`? We added support for it during dragon slayer so if its possible to not use `polymer test` I think that would be awesome 😃 

Here is an example of a similar test as the axe ones below that runs in Karma (docs [here](https://open-wc.org/testing/testing-chai-a11y-axe.html)):
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/test/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.test.js#L132-L135

You can run them locally like this: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/package.json#L17-L18

Just a thought, I know there is existing stuff already using the polymer stuff but it would be [nice to be more consistent with core, etc.](https://github.com/BrightspaceUI/core/pull/344) 😄

imcdonald-d2l: Thanks, will look into this!

test/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.test.js
```diff
@@ -0,0 +1,51 @@
...
+		contentStore.clear();
+	});
+
+	it('should construct', async() => {
+		await loadComponent();
+		runConstructor('d2l-activity-content-editor-detail');
+	});
+
+	it('passes accessibility test', async() => {
+		const el = await loadComponent();
```
pureooze, line 40: One issue I found with testing MobX components for accessibility is that if the store is not populated with certain values (like say a value for the accordion summaries) the tests may not catch those issues. [I found when I set values they caught issues with the components that were not caught before](https://github.com/BrightspaceHypermediaComponents/activities/pull/949). Just something to keep in mind as things get added to the components 👍

pureooze (approved):
LGTM, thanks for making the time for this 😃

test/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.test.js
imcdonald-d2l: Thanks for the heads up! Won't need any additional information in these tests just yet since we haven't implemented a lot of features but good to keep in the mind for the future 👍

sarahboyd (approved):

PR #1143, opened 2020-10-08
WorkToDo Widget Init

KearseTrevor:
- Introduces work-to-do components to the light of day
- Components are still a work in progress, list-items and header are mostly complete but may see small tweaks in the future dependent on design shifts
- Tests for WorkToDo and its components will come at a later point once more legwork has been done on the bones
- Note that these components will be exploring a switch over to the HMC Foundation POC Framework - so code specifically related to siren entity handling should be viewed as stop gap


components/d2l-work-to-do/state/fetch-entity.js
```diff
@@ -0,0 +1,14 @@
+export async function fetchEntity(href, token, bypassCache) {
```
m6jones, line 1: Read only probably doesn't need this.

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+			this.requestUpdate();
+		}
+	}
+
+	_onActivityUsageChange(usage) {
+		this._date = this.getActivityDate(usage);
+	}
+
+	render() {
+		const str = this.setActivityDateString(this._date, this.format, this.includeTime);
```
mbardakovD2L, line 76: nitpick: "set" kind of implies it sets a value as a side effect, rather than computing and returning one

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+				date.type,
+				'date', formatDate(date.date, { format: format }),
+				'time', formatTime(date.date)
+			);
+		} else {
+			dateString = this.localize(
+				date.type,
+				'date', formatDate(date.date, { format: format }),
+			);
+		}
```
bearfriend, line 116:
```javascript
const type = includeTime ? `${date.type}Timed` : date.type;
const dateString = this.localize(
  type,
  'date', formatDate(date.date, { format }),
  'time', includeTime && formatTime(date.date)
);
```
Maybe?

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+					break;
+			}
+
+			if (translations && translations.val) {
+				return {
+					language: lang,
+					resources: translations.val
+				};
+			}
+		}
```
mbardakovD2L, line 50: questions about how this works: do we support multiple languages simultaneously? or does it just look for the first one that succeeds? if `langs` is empty, do we want to fall back on a default?

KearseTrevor: It would eventually grow to look somewhat similar to this https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js

Can be pulled out into a mixin like this example or just pasted into the component itself.

mbardakovD2L: makes sense - I think that example covers my concerns so I'm happy with that

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+		this.includeTime = false;
+		this._dateString = '';
+		this._setEntityType(ActivityUsageEntity);
+	}
+
+	set _entity(entity) {
+		if (this._entityHasChanged(entity)) {
+			this._onActivityUsageChange(entity);
+			super._entity = entity;
+			this.requestUpdate();
```
m6jones, line 67: entity should not trigger a request update. You should be setting the things you need from the response individually so only if those values change the update process will happen. 
Stick to types that don't get passed as references.

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+		const str = this.setActivityDateString(this._date, this.format, this.includeTime);
+
+		const activityDateTemplate = str
+			? html `<div class="d2l-activity-date-text">${str}</div>`
+			: nothing;
+
+		return activityDateTemplate;
+	}
+
+	getActivityDate(usage) {
```
m6jones, line 85: Private method?

components/d2l-activity-date/d2l-activity-date.js
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		const eventDetails = {
+			bubbles: true,
+			composed: false,
+			detail: {
+				date: dateString.length > 0 && dateString
+			}
+		};
+		this.dispatchEvent(new CustomEvent('d2l-activity-date-changed', eventDetails));
```
m6jones, line 125: Why do we need this event? It is being called whenever render happens. Seems like a lot of events flying up.

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
```diff
@@ -0,0 +1,104 @@
...
+		});
+		this._count = collection._items().length || 0;
+		const eventDetails = {
+			bubbles: true,
+			composed: false,
+			detail: {
+				count: this._count
+			}
+		};
+		this.dispatchEvent(new CustomEvent('d2l-collection-changed', eventDetails));
```
m6jones, line 78: Why the event here? This could lead to some weird race conditions if the href is changed quickly.

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * @description Update widget's activity usage collection (_items) to current target collection
+	 * @param {ActivityUsageCollectionEntity} collection Current target collection entity
+	 */
+	_onActivityUsageCollectionChanged(collection) {
+		collection.onItemsChange(item => {
+			item.onActivityUsageChange(usage => {
+				usage.onOrganizationChange(organization => {
+					const newItem = { usage, organization };
+					this._items = [...this._items, newItem];
```
m6jones, line 66: If an individual organization changed it will add duplication.

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/#two-way-binding
Was following the recommendation of event firing to inform the parent component of the activity count found by the collection

components/d2l-activity-date/d2l-activity-date.js
KearseTrevor: Yup, good catch

components/d2l-activity-date/d2l-activity-date.js
KearseTrevor: Think this was an artifact of porting from org-date, will look a bit further but don't think it's needed

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		const items = this._items
+			? this._items.slice(0, this.maxDisplay).map(item =>
+				html`
+					<d2l-work-to-do-activity-list-pane
+						href=${ifDefined(item.usage._entity.getLinkByRel(Rels.Activities.activityUsage).href)}
+						token=${ifDefined(this.token)}></d2l-work-to-do-activity-list-pane>
+				`)
+			: nothing;
```
m6jones, line 93: this should probably use lit-html directive reorder.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
```diff
@@ -0,0 +1,206 @@
...
+		this._icon = ActivityAllowList.userAssignmentActivity.icon;
+		this._type = '';
+		this._setEntityType(ActivityUsageEntity);
+	}
+
+	set _entity(entity) {
+		if (this._entityHasChanged(entity)) {
+			this._onActivityUsageChange(entity);
+			super._entity = entity;
+			this.requestUpdate();
```
m6jones, line 117: entity should never requestUpdate

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
```diff
@@ -0,0 +1,206 @@
...
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ActivityUsageEntity } from 'siren-sdk/src/activities/ActivityUsageEntity';
+import { ActivityAllowList } from './env';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
+import { ifDefined } from 'lit-html/directives/if-defined';
+import { ListItemMixin } from '@brightspace-ui/core/components/list/list-item-mixin';
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { nothing } from 'lit-html';
+
+class ActivityListCard extends ListItemMixin(EntityMixinLit(LocalizeMixin(LitElement))) {
```
m6jones, line 19: @zinabat I didn't think people could change the style of the list item into a card by extending this... Does this give too much power?

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
```diff
@@ -0,0 +1,206 @@
...
+
+		return this._renderListItem({
+			illustration: activityIconTemplate,
+			content: html`
+				<d2l-list-item-content id="content">
+					<div id="content-top-container">
+						<d2l-organization-name
+							class="d2l-heading-3"
+							href="${ifDefined(this._organizationHref)}"
+							token="${ifDefined(this.token)}">
```
m6jones, line 144: `.token`

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
KearseTrevor: tbh - I think I'm using the term card loosely - renaming the component may be more appropriate

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
```diff
@@ -0,0 +1,206 @@
...
+		}
+		this._hasOrgCode = e.detail.organization.code && e.detail.organization.code.length > 0 ;
+
+		// Override Org Code's text-transform
+		if (this._hasOrgCode) {
+			const orgInfo = this.shadowRoot.querySelector('d2l-organization-info');
+			if (orgInfo) {
+				const orgCodeClass = orgInfo.shadowRoot.querySelector('.d2l-organization-code');
+				if (orgCodeClass) {
+					orgCodeClass.style.textTransform = 'none';
```
m6jones, line 200: Shouldn't edit things outside render.
Also, waiting on event to change things could end badly for race conditions. If you need to go deeper in hypermedia tree just use siren-sdk to do it.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
m6jones: Card was a miss named component.

m6jones: Looking really good!
image

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
```diff
@@ -0,0 +1,206 @@
...
+						</d2l-organization-info>
+						${separatorTemplate}
+						${this._type}
+					</div>
+					<div id="content-supporting-info-container" slot="supporting-info" class="d2l-body-compact">
+						Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.
+					</div>
+				</d2l-list-item-content>
+			`,
+			actions: nothing,
```
m6jones, line 162: Is this the solution for that space thing we found a while ago? If so I'll put that in as a default value.

KearseTrevor: Nope, I can specifically look into that while doing the rewrite from this PR and will get back to you on it. This was just setting "actions: nothing" since I wanted transparency on the component level that it was deliberate

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
m6jones: If the parent component needs this information it is better to get it from the API. Keeps thing consistent.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
m6jones: My hope the omitting it tells the devs it is deliberate. Made it an object param so we can omit things that we aren't using to keep the user code clean.

KearseTrevor: Got it

components/d2l-activity-date/d2l-activity-date.js
KearseTrevor: Praise be

components/d2l-activity-date/d2l-activity-date.js
KearseTrevor: Roger roger

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 * @description Update widget's activity usage collection (_items) to current target collection
+	 * @param {ActivityUsageCollectionEntity} collection Current target collection entity
+	 */
+	_onActivityUsageCollectionChanged(collection) {
+		collection.onItemsChange(item => {
+			item.onActivityUsageChange(usage => {
+				usage.onOrganizationChange(organization => {
+					const newItem = { usage, organization };
```
m6jones, line 65: Do you use more than the usage.href?

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
```diff
@@ -0,0 +1,104 @@
...
+	render() {
+		if (!this._items || this._count === 0 || this.maxDisplay === 0) {
+			return nothing;
+		}
+
+		const items = this._items
+			? this._items.slice(0, this.maxDisplay).map(item =>
+				html`
+					<d2l-work-to-do-activity-list-pane
+						href=${ifDefined(item.usage._entity.getLinkByRel(Rels.Activities.activityUsage).href)}
```
m6jones, line 90: You shouldn't have to go through the entity like this. usage should have a way to get href.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
m6jones: This is actually pretty bad and shows there an error in the code.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
m6jones: This also needs to be `.token="${ifDefined(this.token)}" since `.token` makes token a property which is the only way to pass functions.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
m6jones: This needs to be removed.

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: Correct - usage.userActivityUsageHref - it is. Will swap

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -0,0 +1,126 @@
...
+				}
+				:host([hidden]) {
+					display: none;
+				}
+				#status {
+					background-color: var(--d2l-color-carnelian-minus-1);
+					border-color: var(--d2l-color-carnelian-minus-1);
+					border-radius: 0.6rem;
+					border-style: solid;
+					border-width: 1px;
```
m6jones, line 42: Nit: combine width, style and color to one line.

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: Wait that's for the user usage, not the usage itself

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -0,0 +1,126 @@
...
+				};
+			}
+		}
+
+		return null;
+	}
+
+	constructor() {
+		super();
+		/** Represents if corresponding entities are overdue or upcoming */
```
m6jones, line 82: I rather comments live in `static get properties`

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: Nope
Just feeding that href straight to the list-item component

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -0,0 +1,126 @@
...
+		if (this.isOverdue) {
+			return this.localize('overdue');
+		}
+
+		const now = new Date();
+		const endDate = new Date(now.getFullYear(), now.getMonth(), now.getDate() + Config.FUTURE_DAY_LIMIT, 23, 59, 59, 999);
+		const startDate = new Date(now.getFullYear(), now.getMonth(), now.getDate(), 0, 0, 0, 0);
+
+		const start = formatDate(startDate, { format: 'MMM d' });
+		const end = formatDate(endDate, { format: 'MMM d' });
```
m6jones, line 121: Dates can be localized differently per language. I would split this into startMonth, startDay and explain the the context comment in the language file.

components/d2l-work-to-do/d2l-work-to-do-activity-list-pane.js
```diff
@@ -0,0 +1,183 @@
...
+		this._hasOrgCode = false;
+		this._icon = ActivityAllowList.userAssignmentActivity.icon;
+		this._setEntityType(ActivityUsageEntity);
+	}
+
+	set _entity(entity) {
+		if (this._entityHasChanged(entity)) {
+			this._onActivityUsageChange(entity);
+			super._entity = entity;
+			this.requestUpdate();
```
m6jones, line 92: This is bad.

components/d2l-work-to-do/d2l-work-to-do-activity-list-card.js
KearseTrevor: This has been updated - just deleted the line and it continued working

components/d2l-work-to-do/d2l-work-to-do-activity-list-pane.js
```diff
@@ -0,0 +1,183 @@
...
+		}
+		this._hasOrgCode = e.detail.organization.code && e.detail.organization.code.length > 0 ;
+
+		// Override Org Code's text-transform
+		if (this._hasOrgCode) {
+			const orgInfo = this.shadowRoot.querySelector('d2l-organization-info');
+			if (orgInfo) {
+				const orgCodeClass = orgInfo.shadowRoot.querySelector('.d2l-organization-code');
+				if (orgCodeClass) {
+					orgCodeClass.style.textTransform = 'none';
```
m6jones, line 177: Why do you need this?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -0,0 +1,252 @@
...
+		this._emptyEntity = {};
+		this._items = [];
+		this._setEntityType(UserEntity);
+	}
+
+	set _entity(entity) {
+		if (this._entityHasChanged(entity)) {
+			this._onEntityChanged(entity);
+			super._entity = entity;
+			this.requestUpdate();
```
m6jones, line 94: This is bad

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: 🤦  usage.self() ...

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -0,0 +1,252 @@
...
+	_onEntityChanged(user) {
+		if (!user || !user._entity) {
+			return;
+		}
+		this._setCollectionSources(user._entity);
+	}
+
+	render() {
+		const overdueCollection = html`
+				<d2l-activity-collection
```
m6jones, line 111: You might need to fix this tag. Is it `d2l-activity-work-to-do-collection`?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -0,0 +1,252 @@
...
+				query[field.name] = decodeURI(field.value);
+			}
+		});
+
+		const queryString = Object.keys(query).map(key => {
+			return `${key}=${query[key]}`;
+		}).join('&');
+
+		return queryString ? `${action.href}?${queryString}` : null;
+	}
```
m6jones, line 195: This code is done in `sirenAction` file in the siren-sdk

components/d2l-work-to-do/lang/en.js
```diff
@@ -0,0 +1,13 @@
...
+	dateHeader : "{start} - {end}",
+	goToDiscover : "Go To Discover",
+	myWorkToDo: "My Work To Do",
+	noActivitiesDiscoverActive : "You have no activities with due dates available. Check out Discover for more learning.",
+	noActivitiesDiscoverInactive : "You have no activities with due dates available.",
+	nothingHere : "There\'s nothing here...",
+	overdue : "Overdue",
+	viewAll : "View all work to do",
+	viewLearning : "View All Learning",
+};
```
m6jones, line 13: You need to add context comments for each language term.

components/d2l-work-to-do/d2l-work-to-do.js
KearseTrevor: I'm considering pull d2l-work-to-do out of this PR all together since it's a total mess that I hadn't fully addressed yet

components/d2l-work-to-do/d2l-work-to-do-activity-list-pane.js
KearseTrevor: Native style incoming from d2l-organization-info is text-transform: uppercase;
Wanted none (Could potentially do Capitalize as well but don't want uppercase) since it doesn't go with our design

components/d2l-work-to-do/d2l-work-to-do-activity-collection.js
KearseTrevor: Ah, understood. I'll make sure to tackle this soon (in the next PR for this component set).

components/d2l-work-to-do/d2l-work-to-do.js
KearseTrevor: Understood, I'm popping this component out of the PR since there's so much rework that is required - it was essentially all placeholder code but since it's going to a more visible place I'm just chopping off the rotten limb

mbardakovD2L (approved):

PR #1146, opened 2020-10-08
Part 2: Use activity usage specialization rel to get assignment href from activity and simplify mobx assignment store

mdgbayly:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Ftask%2F440737138828

By using the specialization rel to get the assignment href from the activity usage we no longer need to use the `AssignmentActivityUsage` siren-sdk entity which means we can simplify the loading of the `d2l-activity-assignment-editor` and the `assignment-store`. 

Most of the changes are just updates to dependent classes to change calls from `store.getAssignment(href)` to just `store.get(href)` and to update imports of the store object.

**Just pay attention to the second commit on this PR - This was branched from a previous PR which has now been merged, so I'm confused why the commits from that original PR are still showing up on this one. I thought Github was smart enough to handle this. I'm sure I've seen it handle this better in the past.**


ronan-f (approved):
As per description just looked at second commit. Looks good to me.

PR #1150, opened 2020-10-13
Refactor save to support grouping editors by save order including enh…

mdgbayly:
…anced promise handling to allow multiple editors to call save on the same store.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Ftask%2F440737132964

This refactoring is not essential but it makes components more reusable by allowing any component to call `save` on it's associated store. Previously we were hard wiring things such that a single 'dominant' component was responsible for calling save. To support multiple editor components being able to call `save`, the store's `save` implementation needs to be written in such a way that it generates a promise when save is called and subsequent calls to save before the initial save has completed just return that promise. This prevents the same save API being called multiple times for a single save interaction.

To support this approach, the `ActivityEditorAsyncContainerMixin` `save` implementation had to be refactored to execute multiple `save` calls for the same `saveOrder` grouping concurrently otherwise, it still led to the same `save` API being called multiple times. So now before calling `validate/save`, the mixin groups editors by their `saveOrder` attribute and then iteratively calls `validate/save` on each group in turn, but calls within the same group are processed as part of the same `Promise.all` batch.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
@@ -77,6 +77,11 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 		];
 	}
 
+	constructor() {
+		super(store);
+		this.saveOrder = 2000;
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
 		this._m4EmailNotificationEnabled = this._isMilestoneEnabled(Milestones.M4EmailSubmission);
@@ -108,6 +113,15 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 				${this._renderAssignmentSubmissionNotificationEmail(assignment)}
 			</d2l-labs-accordion-collapse>
 		`;
+	}
+
+	async save() {
+		const assignment = store.get(this.href);
+		if (!assignment) {
+			return;
+		}
+
+		await assignment.save();
 	}
 
 	_checkNotificationEmail(e) {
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -107,42 +120,62 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
 		this.isSaving = true;
 		this.markSaveStart(this.type, this.telemetryId);
 
-		const orderedEditors = Array.from(this._editors).sort((a, b) => a.saveOrder - b.saveOrder);
+		const groupedEditors = this._groupBy(Array.from(this._editors), 'saveOrder');
```
mdgbayly, line 123: This code groups editors using their `saveOrder` property. The grouping creates an object with the saveOrder as the key and the array of editors with that saveOrder as the value.

The next line then sorts that grouped object into a list of editors again using the key of the object to determine the ordering.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -110,6 +115,15 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 		`;
 	}
 
+	async save() {
```
emil-furniss, line 118: Why does this editor need special treatment?

mdgbayly: it was mostly as an example... we should really be doing this on all our components to make then actually reusable outside the context of one big activity editor. e.g. if you just wanted to drop the submissions component on a page and not use any of the other components.

mdgbayly: e.g. currently if someone just wanted to just the availability dates editors - it would not work currently without all the other plumbing.

emil-furniss (approved):

wongvincent (approved):

PR #1151, opened 2020-10-14
US121271 prep - move summarizer-styles into activity-editor base folder

mdulat:
As preparation for my story, I'm moving the `summarizer-styles` into the activity-editor folder instead of assignment specific, as it is used by other activities (including now quizzing).


wongvincent (approved):

mdgbayly:
I've kind of wondered ever since we started using this component, whether there is any value in trying to encapsulate this into an actual component. It still feels like it's almost not worth it, as we're basically just using the accordion, but there is quite a lot of duplicated setup with importing the styles and setting the classes, attributes, etc on the `d2l-labs-accordion-collapse` and the header and the list.

We could possibly remove that duplication by making this a component with slots for the header text, list items and contained components.

Again, still feels like it almost not worth it, but if we wanted to change them all from using an accordion to something else at some point, it would make that future refactoring simpler. Or if we wanted to change specific attributes of the accordion - dunno.

Thoughts?

mdulat:
I see the duplication in the styles, and format of header, list items, contained components. What about the state-changed event and opened attributes? There are differences in these, so a wrapper component would have to pass those along. Maybe it's worth it, I am not feeling strongly either way right now.

mdgbayly:
> I see the duplication in the styles, and format of header, list items, contained components. What about the state-changed event and opened attributes? There are differences in these, so a wrapper component would have to pass those along. Maybe it's worth it, I am not feeling strongly either way right now.

Yeh - that seems to be the main bit that is different right now, but if we had to handle errors in other accordions, we'd need to duplicate that too. I guess currently we only have date errors.

But that event handling could probably be generalized so that the event handler could be internalized into the shared component, and the shared component could expose a `has-errors` attribute or something that could be set  from the host.

mdulat:
> But that event handling could probably be generalized so that the event handler could be internalized into the shared component, and the shared component could expose a `has-errors` attribute or something that could be set from the host.

That would work. I guess the assumption is that we will always want the accordion to load in the opened state if it contains errors. Ok, I'll create this wrapper, it'll simplify the accordion work going forward.

mdulat:
This `summarizer-styles` file has been removed in preference of an accordion wrapper component.

PR #1152, opened 2020-10-14
US121271 prep - create accordion-collapse wrapper

mdulat:
Create a wrapper for the accordion-collapse component, since there is a lot of repetition of styles, code templates, properties, classes, error handling. And refactor existing code to use the wrapper.


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -75,25 +68,21 @@ class ActivityAssignmentAvailabilityEditor extends SkeletonMixin(ActivityEditorF
...
-				?no-icons="${this.skeleton}"
-				@d2l-labs-accordion-collapse-state-changed=${this._onAccordionStateChange}>
-				<h3 class="d2l-heading-3 d2l-activity-summarizer-header d2l-skeletize" slot="header">
+			<d2l-activity-accordion-collapse ?has-errors=${this._errorInAccordion()}>
+				<span slot="header">
 					${this.localize('hdrAvailability')}
-				</h3>
-				<ul class="d2l-body-small d2l-activity-summarizer-summary d2l-skeletize" slot="summary">
+				</span>
+				<span slot="summary-items">
```
mdgbayly, line 75: Hmm, had not considered it to this level of detail but was vaguely aware that the `<li>` might cause some issues here. I wonder if slotting the list items in the span into the ul in the component kind of breaks the semantics of the `<ul>`. (`<ul><span><li>data</li></span></ul>`)

I guess it's a little bit weird that we are assuming the slot takes `<li>` here, but maybe that is OK. An alternative would be to just allow the whole list to be slotted, but then that kind of puts us back at square one with the styling of the list items, as the component won't be able to style the child list items I don't think as it can only style immediately slotted children.

Another alternative if we are OK with assuming the slotted elements have to be `<li>` would be to slot the `<li>` s directly. e..g I think if you give a bunch of light DOM elements the same slot name they will just iteratively fill the slot

e..g in the component

```
<ul>
  <slot name="items"></slot>
</ul>
```

In the host

```
   <li slot="items">abc</li>
   <li slot="items">def</li>
```

dbatiste: > I wonder if slotting the list items in the span into the ul in the component kind of breaks the semantics...

Yes, it would be invalid HTML. [Spec for ul](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element); [Spec for li](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element)  Even with aria roles `list` and `listitem`, the list must be a direct parent of the items.

> ...as I think it can only style immediately slotted children.

Correct.  The `::slotted` selector can only target immediate children of the slot.  There is a lot of discussion happening currently to expand what can be selected, but there is no consensus that a) it is needed; and b) how it would be done if there were consensus.

> I think if you give a bunch of light DOM elements the same slot name they will just iteratively fill the slot

Correct. Any children with the slot name will be distributed to that slot.

mdgbayly (approved):
LGTM

PR #1156, opened 2020-10-19
Us119416/implement quizzing dirty check

ronan-f:
Part of the logic needed to implement dirty check for quizzing based on [US119416](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/413005808740).

### DIFF
--- a/components/d2l-activity-editor/state/activity-usage.js
+++ b/components/d2l-activity-editor/state/activity-usage.js
@@ -16,12 +16,7 @@ export class ActivityUsage {
 		this.token = token;
 	}
 
-	async dirty(activityType = '') {
-		// alignments are not supported at the quiz level
-		if (activityType === 'quiz') {
-			return !this._entity.equals(this._makeUsageData());
-		}
-
+	async dirty() {
 		return !this._entity.equals(this._makeUsageData()) || await this._alignmentsDirty();
 	}
 	async fetch() {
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -16,7 +16,12 @@ export class ActivityUsage {
 		this.token = token;
 	}
 
-	async dirty() {
+	async dirty(activityType = '') {
+		// alignments are not supported at the quiz level
+		if (activityType === 'quiz') {
```
ronan-f, line 21: This check is needed as alignments are not supported at the quiz level so this check would otherwise always return `true` indicating a change in content. 

The activity type is already set up as a property in `d2l-activity-editor.js` so I just passed it through for this check. Other activities eg assignments will not need to do this as `activityType` defaults to an empty string

mdgbayly: I don't think we should be including conditionals like this in the activity code. It will get unwieldy very quickly if we have to keep adding tool specific 'type' checks.

I think that type property that is being setup in the activity editor is just for telemetry logging, and could probably have been given a better name than just type, to make it clearer what it's purpose was because we don't want to start adding typed conditionals.

Where else are we referencing the string 'quiz'?

But yeh - we probably want to come up with a way to make `_alignmentsDirty()` always return false if alignments are not being used. Will touch base in slack about that.

mdgbayly (approved):

PR #1157, opened 2020-10-19
Stopped explicitly disabling cache as it was only needed to pick up s…

ronan-f:
…iren-sdk changes

This was removed as tests were failing due to travis using a cached version of node_modules that didn't have updated changes for siren-sdk. I want to re-add it as it slows down all future test runs. 

Edit: This is failing with an error about importing from siren-sdk so obviously it's still falling back to an older version of cached node_modules.


stevenao (dismissed):

stevenao:
https://travis-ci.com/github/BrightspaceHypermediaComponents/activities/caches

I think you could clear cache from the Travis web console too.

PR #1159, opened 2020-10-20
US119469 add start / end dates to availability dates accordion

mdulat:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F413481592160&fdp=true


ronan-f (approved):
LGTM

PR #1162, opened 2020-10-21
Updated default quiz date/ times to 12am

ronan-f:
As per Joseph, for quizzing (and eventually for assignments) the default start time should be 12AM.

I'm not a big fan of this approach as it's quite specific to just start date. What happens if we want to make the end date dynamic for example? Do we keep adding new properties? 

I believe this is just a temporary solution as eventually assignments will also adopt this default start time and it will no longer need to be dynamic.

Also wondering if it's better to have a fallback time e.g. `${this.startDateDefaultTime || 'endOfDay'}` or if it would be preferred to avoid the fallback and explicitly set the start time for each component?


mdulat (approved):
Looks good. I like having the endOfDay fallback so only quizzing has to pass in a value. Would be good to add a comment in the code to indicate it's a temporary solution, as in future we want all activities to behave the same way.

mdulat (approved):

mdulat (approved):

PR #1164, opened 2020-10-22
fix summarizer converting dates to browser settings instead of bright…

mdulat:
…space locale settings

Using javascript's `Date` constructor causes the date to be converted to the browser timezone instead of the user's locale setting in brightspace. The `BrightspaceUI/core/helper/dateTime` functions use the correct locale settings.


ronan-f (commented):
I'm just digging through this helper to see where it actually gets the users brightspace time settings from. If I'm following it correctly it seems to be from the call to `getDocumentLocaleSettings`?

mdulat:
> I'm just digging through this helper to see where it actually gets the users brightspace time settings from. If I'm following it correctly it seems to be in the call to `getDocumentLocaleSettings`?

Yeah

mdgbayly (approved):

PR #1165, opened 2020-10-23
Us121749/add timing display accordion

ronan-f:
Based on [US121749](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/448916982360) which is adding a new Timing & Display accordion to quizzing. This story just covers the initial setup so the accordion appears in the UI with correct lang term but no functionality or summary.

I used the same pattern as before to share styles between the quizzing accordions.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
@@ -9,7 +9,6 @@ import { html } from 'lit-element/lit-element.js';
 import { LocalizeActivityQuizEditorMixin } from './mixins/d2l-activity-quiz-lang-mixin';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
-import { shared as store } from '../state/activity-store.js';
 
 class ActivityQuizTimingAndDisplayEditor extends AsyncContainerMixin(LocalizeActivityQuizEditorMixin(SkeletonMixin(ActivityEditorFeaturesMixin(ActivityEditorMixin(MobxLitElement))))) {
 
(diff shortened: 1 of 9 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
```diff
@@ -0,0 +1,63 @@
...
+					${this.localize('hdrTimingAndDisplay')}
+				</span>
+
+			</d2l-activity-accordion-collapse>
+		`;
+	}
+	// Returns true if any error states relevant to this accordion are set
+	_errorInAccordion() {
+		const activity = store.get(this.href);
+		if (!activity || !activity.dates) {
```
mdulat, line 52: Should probably remove this whole conditional and return false always until we know what error handling this accordion will need.

ronan-f: Good point! I'll change that

components/d2l-activity-editor/styles/accordion-styles.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const accordionStyles = css`
+	:host {
+		display: block;
+	}
+
+	:host([hidden]) {
+		display: none;
+	}
+`;
```
mdulat, line 11: Just wondering if these styles can go into the `d2l-activity-accordion-collapse` component instead of needing a separate styles file, or do they need to be in the parent?

mdulat: Actually I think they do need to be in the parent based on how we show/hide these things. In this case can we update the other components that use the accordion as well?
`d2l-activity-assignment-availability-editor`
`d2l-activity-assignment-editor-submission-and-completion-editor`
`d2l-activity-assignment-evaluation-editor`
`d2l-activity-content-availability-editor`
`d2l-activity-quiz-availability-editor`

ronan-f: Yip tested locally adding the styles to `d2l-activity-accordion-collapse` but no joy. I've updated everywhere that implements accordions to use this shared style module now.

mdulat: thank-you!

mdulat (approved):

PR #1167, opened 2020-10-27
Add content module get/save actions + some cancel and re-direct work

imcdonald-d2l:
Integrate with newly defined `ContentModuleEntity` get/save functions defined in `siren-sdk` repo [here](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/222). This allows us to get/save title and description values on content modules.

save-actions

Note that a non-empty title is required for the save action to execute:
save-error

Some initial work for the cancel workflow was also added. Actually deleting of the newly created item and re-directing has yet to be implemented.


components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -131,8 +131,8 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
 		await this._saveEditors(orderedEditors);
 
 		this.isError = false;
+		this.isSaving = false;
```
imcdonald-d2l, line 134: @mdgbayly I believe you guys were tweaking this behavior recently. Noticed that `this.isSaving` was not properly set to `false` when successfully completing the save actions. Was only set to `false` when saving failed, not on success so i fixed that here.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -113,6 +140,10 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		}
 	}
 
+	_saveDescription(richText) {
+		store.getContentActivity(this.href).setDescription(richText);
```
CodeBaboon, line 144: Should we guard against null on `store.getContentActivity(this.href)`? We do in similar uses above (hasPendingChanges() and save() methods).

CodeBaboon (commented):
LGTM. One small comment, not sure if a legit concern or not.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: Good catch, we should be guarding against that in `saveDescription` and `saveTitle` to be safe. In a perfect world the content activity exists in the store but could be possible that it is not, so better to protect against it.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor.js
```diff
@@ -97,5 +98,17 @@ class ContentEditor extends RtlMixin(ActivityEditorMixin(MobxLitElement)) {
...
+		if (this.cancelHref) {
+			window.location.href = this.cancelHref;
+		}
+	}
+
+	_redirectOnSaveComplete() {
+		if (this.saveHref) {
+			window.location.href = this.saveHref;
+		}
+	}
```
imcdonald-d2l, line 112: If the `cancelHref` or `saveHref` are added to `d2l-activity-content-editor` component, then redirect once the corresponding event fires.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -21,21 +38,80 @@ export class Content {
...
+		if (!this._contentModule) {
+			return;
+		}
+		try {
+			await this._contentModule.setModuleTitle(this.moduleTitle);
+			await this._contentModule.setModuleDescription(this.moduleDescriptionRichText);
+			await this.fetch();
+		} catch (err) {
+			// TODO - what do we want to do with the error?
+		}
```
imcdonald-d2l, line 82: I know swallowing errors is not recommended but wasn't sure how we want to handle it. Do we want to log the error in telemetry? Do we just want to remove the try/catch and let it blow up? Thoughts?

CodeBaboon: Ideally we would have some kind of error label / toast we could display. I guess nothing like that is currently  in place?

imcdonald-d2l: A brief investigation leads me to believe nothing like that is currently in place. Although, I don't know how hard it would be to display a toast message if we already have pre-built components for that. 🤔

CodeBaboon: I'm sort of surprised this hasn't come up for assignments. Do they just swallow I guess? Seems bad. Maybe we don't have UX designs for the error cases yet? We might need to reach out to face-dev or Tara to see what should be done ideally.

imcdonald-d2l: After looking into it a bit more it would look as if by design I'm not supposed to have a try/catch and then let the container handle the thrown error in the container [mixin here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L152). That would prevent us from displaying a toast though. Are we okay with that?

imcdonald-d2l: The even have a comment saying they're not handling errors right now 😓 
```
// TODO - Once we decide how we want to handle errors we may want to add error handling logic to the save
```

CodeBaboon: Yeah, let's do that. It's good to have the error handling centralized like that. Of course, it should do something.... but that's for future work.

imcdonald-d2l: Sounds good, will remove the try/catch. Agreed that it should handle errors better though but at least it's a known issue.

test/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.test.js
```diff
@@ -40,12 +40,4 @@ describe('d2l-activity-content-editor-detail', function() {
...
 		await expect(el).to.be.accessible();
 	});
-
-	it('should set content title', async() => {
-		const testTitle = 'test title';
-		contentItem.setTitle(testTitle);
-		const el = await loadComponent();
-		const title = el.shadowRoot.querySelector('#content-title');
-		await expect(title.value).to.equal(testTitle);
-	});
```
imcdonald-d2l, line 50: Removing this test as it has become more complex to test with the additional complexity in `setTitle`  in `content.js`. Some additional setup would be required to get the content store set up to make this work. Arguably it is unnecessary as well since at all times we should just be grabbing the state variable which is already tested in `content.spec.js`. If anyone feels strongly that this test should be included I can look into it.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -9,33 +10,104 @@ export class Content {
...
+	}
+
+	async saveContentModule() {
+		if (!this._contentModule) {
+			return;
+		}
+		await this._contentModule.setModuleTitle(this.moduleTitle);
+		await this._contentModule.setModuleDescription(this.moduleDescriptionRichText);
+		await this.fetch();
```
ianlaidlaw, line 78: how come you need to refetch after saving?

ianlaidlaw: i guess it makes sense to get the "master version" but would the local copy not be the same?

CodeBaboon: I'll have to look into how the save is working down lower but the API will actually respond with the "master version" on those update requests also.

imcdonald-d2l: These are good points, I also don't know if it's worth re-fetching since we re-direct after a successful save anyways. So kinda pointless updating UI state if the user is kept on the new create page.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -9,33 +10,104 @@ export class Content {
...
+
+	loadContentModule(moduleEntity) {
+		this._contentModule = moduleEntity;
+		this.moduleTitle = moduleEntity.title();
+		this.moduleDescriptionRichText = moduleEntity.descriptionRichText();
+	}
+
+	async save() {
+		// TODO add more save functions as we add more content types
+		if (this.entityType === CONTENT_TYPES.module) {
```
ianlaidlaw, line 67: i did a quick search for `this.entityType === CONTENT_TYPES.module` and found 7 occurrences. i dont have any suggestions atm but i want to start a discussion about this. im worried this file will be trying to handle too many different things, especially as we add more types. should we start thinking less about reusability here and more about readability and better concerns?

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -9,33 +10,104 @@ export class Content {
...
 
 	setTitle(value) {
-		this.title = value;
+		if (this.entityType === CONTENT_TYPES.module) {
+			this.moduleTitle = value;
+		}
 	}
 
+	_makeModuleData() {
+		/* NOTE: if you add fields here, please make sure you update the corresponding equals method in siren-sdk.
```
ianlaidlaw, line 94: just curious: i checked the equals fn and it looks like its doing a `!==` check in all values. fine for now but will our properties ever be something like an array or object? in which case the equals compare wont work.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
imcdonald-d2l: I agree this file in particular should be re-worked as we add more types and our content-store state takes on an increased level of complexity. I was thinking this could be done when we start looking into web-links so we have a better idea of what that could look like with two different types being used across the various `content-editor` components. Was also thinking about addressing this later for the sake of getting this working in a shorter amount of time so we can release this faster. Open to improving this now if we feel that is necessary though.

CodeBaboon: I think Ian is right.

ianlaidlaw: thats fair, i dont think you need to change anything now in that case, as long as its something we're aware of and thinking about. we'll have a better idea once those other types exist.

CodeBaboon: Jokes aside, I am on board with refactoring as we bring the web links online.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
ianlaidlaw: this is also something that can be revisited

imcdonald-d2l: That is a good point, I don't know if that is a possibility. I think we would just cross that bridge later if we got there and customize the equals function like assignments [has done here](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L880) to meet the specific needs of assignment properties. So after the `left !== right` loop we just add custom checks to the equals function as we see fit.

ianlaidlaw (dismissed):
lgtm. my only "concerns" are things that will be easier to address in the future since they are kind of nebulous or marked todo right now.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -48,11 +48,13 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		this._debounceJobs = {};
 		this._setEntityType(ContentEntity);
 		this.skeleton = true;
+		this.saveOrder = 2000;
```
ianlaidlaw, line 51: what significance does 2000 have? and what does this do?

imcdonald-d2l: [This PR](https://github.com/BrightspaceHypermediaComponents/activities/pull/1150) should help clarify (or potentially confuse you more) on what the purpose of the `saveOrder` is used for. Essentially, how I understand it at a high level is a way to group editors for the `d2l-activity-editor-container-mixin` to then execute the save actions on.  So for all content related editors we create, we can put `this.saveOrder = 2000` so they are all grouped together in the container mixin.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
imcdonald-d2l: Removed `this.fetch()` after saving since we re-direct back to lessons after saving anyways. Since we aren't leaving them in the new create view there is no reason to update the UI with the most up-to-date version. Let me know if you have concerns with this.

ianlaidlaw: makes sense to me

ianlaidlaw (approved):

CodeBaboon (approved):

CodeBaboon (commented):
LGTM!

PR #1178, opened 2020-11-04
Add due date to face create view for content modules + edit description and loading skeleton fix

imcdonald-d2l:
### Notes
- When creating a new content item, the user should see the subtle button saying "Add Due Date". Once clicked, the due date field will appear.
- When editing, the user should see the "Add Due Date" button if a due date has not yet been added. Otherwise, the due date field should appear and show the selected due date. Note that this has not properly been tested yet since we haven't implemented loading in the edit view. Will be confirmed and fixed if necessary once this is implemented.
- The due date field does indeed appear to save on the activity-usage entity and should be visible in areas where implemented. 

UPDATE: refactored some logic so we see loading skeletons until content entity is loaded into state so you don't see a weird delay before the field value appears. Also had to set initial value of description value to `undefined` so the html editor would properly load in the value when grabbing the value from state.

### GIFS
Note: no animations/transitions are included in this implementation. 
add-due-date

### Screenshots
Screenshot below shows that a successful PATCH request was sent with the entered due date value.
save-date-request

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -45,9 +45,9 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 					padding-bottom: 20px;
 				}
 				.duedate-field {
-					transition: opacity 650ms ease-in;
-					opacity: 1;
 					height: auto;
+					opacity: 1;
+					transition: opacity 650ms ease-in;
 					width: auto;
 				}
 				:host([showAddDueDateBtn]) .duedate-field {
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -168,6 +168,8 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 					</d2l-activity-due-date-editor>
 				</div>
 			`;
+		} else {
+			return html ``;
 		}
 	}
 
@@ -182,6 +184,8 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 					${this._titleError}
 				</d2l-tooltip>
 			`;
+		} else {
+			return html ``;
 		}
 	}
 
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -27,7 +27,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		return {
 			_titleError: { type: String },
 			_hasDatePermissions: { type: Boolean },
-			showAddDueDateBtn: { type: Boolean, reflect: true }
+			_showAddDueDateBtn: { type: Boolean }
 		};
 	}
 
@@ -44,17 +44,6 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				:host > div {
 					padding-bottom: 20px;
 				}
-				.d2l-duedate-field {
-					height: auto;
-					opacity: 1;
-					transition: opacity 650ms ease-in;
-					width: auto;
-				}
-				:host([showAddDueDateBtn]) .d2l-duedate-field {
-					height: 0;
-					opacity: 0;
-					width: 0;
-				}
 			`
 		];
 	}
@@ -66,7 +55,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		this.skeleton = true;
 		this.saveOrder = 2000;
 		this._hasDatePermissions = false;
-		this.showAddDueDateBtn = true;
+		this._showAddDueDateBtn = true;
 	}
 
 	render() {
@@ -147,7 +136,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		const dueDate =  dates.dueDate ? dates.dueDate : null;
 		// if due date exists on the activity, show the field
 		if (dueDate) {
-			this.showAddDueDateBtn = false;
+			this._showAddDueDateBtn = false;
 		}
 	}
 
@@ -161,23 +150,23 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 	}
 
 	_renderDueDate() {
+		// TODO - replace with shared component when one is created
 		if (this._hasDatePermissions) {
 			return html `
 				<div id="duedate-container">
 					<d2l-button-subtle
 						text="${this.localize('content.addDueDate')}"
 						@click=${this._showDueDate}
-						?hidden=${!this.showAddDueDateBtn}
+						?hidden=${!this._showAddDueDateBtn}
 					>
 					</d2l-button-subtle>
-					<div class="d2l-duedate-field">
-						<d2l-activity-due-date-editor
-							.href="${this.href}"
-							.token="${this.token}"
-							?skeleton="${this.skeleton}"
-						>
-						</d2l-activity-due-date-editor>
-					</div>
+					<d2l-activity-due-date-editor
+						.href="${this.href}"
+						.token="${this.token}"
+						?skeleton="${this.skeleton}"
+						?hidden=${this._showAddDueDateBtn}
+					>
+					</d2l-activity-due-date-editor>
 				</div>
 			`;
 		}
@@ -234,7 +223,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 	}
 
 	_showDueDate() {
-		this.showAddDueDateBtn = false;
+		this._showAddDueDateBtn = false;
 	}
 }
 customElements.define('d2l-activity-content-editor-detail', ContentEditorDetail);
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -44,15 +44,15 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				:host > div {
 					padding-bottom: 20px;
 				}
-				.duedate-field {
+				.d2l-duedate-field {
 					height: auto;
 					opacity: 1;
 					transition: opacity 650ms ease-in;
 					width: auto;
 				}
-				:host([showAddDueDateBtn]) .duedate-field {
-					opacity: 0;
+				:host([showAddDueDateBtn]) .d2l-duedate-field {
 					height: 0;
+					opacity: 0;
 					width: 0;
 				}
 			`
@@ -170,7 +170,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						?hidden=${!this.showAddDueDateBtn}
 					>
 					</d2l-button-subtle>
-					<div class="duedate-field">
+					<div class="d2l-duedate-field">
 						<d2l-activity-due-date-editor
 							.href="${this.href}"
 							.token="${this.token}"
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -27,7 +27,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		return {
 			_titleError: { type: String },
 			_hasDatePermissions: { type: Boolean },
-			showAddDueDateBtn: { type: Boolean }
+			showAddDueDateBtn: { type: Boolean, reflect: true }
 		};
 	}
 
@@ -43,6 +43,17 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				}
 				:host > div {
 					padding-bottom: 20px;
+				}
+				.duedate-field {
+					transition: opacity 650ms ease-in;
+					opacity: 1;
+					height: auto;
+					width: auto;
+				}
+				:host([showAddDueDateBtn]) .duedate-field {
+					opacity: 0;
+					height: 0;
+					width: 0;
 				}
 			`
 		];
@@ -159,13 +170,14 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						?hidden=${!this.showAddDueDateBtn}
 					>
 					</d2l-button-subtle>
-					<d2l-activity-due-date-editor
-						.href="${this.href}"
-						.token="${this.token}"
-						?skeleton="${this.skeleton}"
-						?hidden=${this.showAddDueDateBtn}
-					>
-					</d2l-activity-due-date-editor>
+					<div class="duedate-field">
+						<d2l-activity-due-date-editor
+							.href="${this.href}"
+							.token="${this.token}"
+							?skeleton="${this.skeleton}"
+						>
+						</d2l-activity-due-date-editor>
+					</div>
 				</div>
 			`;
 		}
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -39,6 +44,17 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+					height: auto;
+					opacity: 1;
+					transition: opacity 650ms ease-in;
+					width: auto;
+				}
+				:host([showAddDueDateBtn]) .duedate-field {
+					opacity: 0;
+					height: 0;
+					width: 0;
+				}
```
imcdonald-d2l, line 57: The exact styles and transitions here will potentially change if Tara requests them. However, the idea is that since we can't transition on `display: none` you have to get hacky and set `opacity`, `height` and `width` properties to `0` to transition a new component coming into view. If anyone has cleaner solutions or recommendations, I'm all ears!

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -126,6 +160,29 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		);
 	}
 
+	_renderDueDate() {
+		if (this._hasDatePermissions) {
```
imcdonald-d2l, line 164: Since we are displaying a due date button we have to double check permissions for viewing the field here as well. Would be silly to display the button if they don't indeed have permissions to add the field. Normally this wouldn't be needed because all use cases so far just display the field automatically and `d2l-activity-due-date-editor` has its own permission checks.

If others began adopting this approach/style it might be worth making it's own component.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -39,6 +44,17 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+					height: auto;
+					opacity: 1;
+					transition: opacity 650ms ease-in;
+					width: auto;
+				}
+				:host([showAddDueDateBtn]) .d2l-duedate-field {
+					height: 0;
+					opacity: 0;
+					width: 0;
+				}
```
imcdonald-d2l, line 57: The exact styles and transitions here will potentially change if Tara requests them. However, the idea is that since we can't transition on `display: none` you have to get hacky and set `opacity`, `height` and `width` properties to `0` to transition a new component coming into view. If anyone has cleaner solutions or recommendations, I'm all ears!

dbatiste: With the current styles, I think we will end up with an "invisible" element that can be accessed via the keyboard - it will seem like the focus is mysteriously lost, but the focus is in fact applied to the element (browser behavior may differ).

Usually if an element should not be displayed then we use `display: none;`.  To get around the inability to animate/transition `display`, or another property while also changing from `display: none`, the `display` property is changed, and then we begin the animation in the subsequent frame.

showing
```javascript
elem.setAttribute('data-animate-state', 'pre-show-hide');

requestAnimationFrame(() => {
    elem.setAttribute('data-animate-state', 'show');
});
```

```css
.d2l-elem {
    display: none;
    opacity: 0;
    transition: opacity 650ms ease-in;
}
[data-animate-state="pre-show-hide"] {
    display: block;
}
[data-animate-state="show"] {
    display: block;
    opacity: 1;
}
@media (prefers-reduced-motion: reduce) {
    .d2l-elem {
        transition: none;
    }
}
```

We need similar special treatment for hiding too, except we need to wait until the animation is complete before setting `display: none;`.

hiding
```javascript
const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches; // usually done once

if (reducMotion) {
    elem.removeAttribute('data-animate-state');
} else {
    elem.addEventListener('transitionend', () => {
        elem.removeAttribute('data-animate-state');
    }, {once: true})
    elem.setAttribute('data-animate-state', 'pre-show-hide');
}
```

There are a few ways to code this up, and it tends to look different for Lit elements because we do not modify the DOM outside of render.  I thought it was probably easiest to convey with vanilla.  The approach is pretty standard, but it's not the sort of thing we really want littered throughout app code, so typically it gets baked into the `core` components/mixins. If this is going to be a common thing, we really ought to be looking at wait to share this.

More design related - the animation seems pretty slow (650ms) - I think considerably longer than anything we've used in core for showing/hiding things.  VisibleOnAncestor (used to reveal icon buttons) uses 200ms, Tabs (when added or removed) uses 200ms, and Dropdowns (showing/hiding) uses 300ms.  Worth double checking.

imcdonald-d2l: Thanks for your input @dbatiste, would have never thought of the focusing issue! Will investigate the suggested fix and reach out to you with any questions 🙂 

In regards to the 650ms, yes it did feel slow, it was just a number I picked and was waiting for feedback from our designer. Can adjust it to be 200ms to be more consistent until I get the official styles from our designer.

dbatiste: No problem Ian.  Yeah feel free to reach out.  As you can see it's a bit of code, and although not complex, I think if we see this added in app code more than once we've sorta failed.  I wonder if the animation is really required, and if it is, will we do this consistently in our designs, in such case we should look at the best way to deliver this in a reusable way.

imcdonald-d2l: Sounds good, I'll confirm with our designer if an animation is required and if this will be a common pattern. Will go from there once we have confirmation. 👍

imcdonald-d2l:
**Update**: animations have been removed until design pattern is finalized and a shared component is created.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -117,6 +126,20 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+	_getDueDateAndPermission() {
+		const entity = activityStore.get(this.href);
+		if (!entity || !entity.dates) {
+			return;
+		}
+		const dates = entity.dates;
+		this._hasDatePermissions = dates.canEditDates;
+		const dueDate =  dates.dueDate ? dates.dueDate : null;
+		// if due date exists on the activity, show the field
+		if (dueDate) {
```
CodeBaboon, line 138: I think the `dueDate` variable is basically unnecessary. This could just be written as:

```
if (dates.dueDate) {
   this._showAddDueDateBtn = false;
}
```

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
CodeBaboon: In the `false` case here we return nothing. Is that ok? Or should it be something like

```
if (!this._hasDatePermissions) {
  return html ``;
}
```

even if it's fine to just not have a return (I honestly just don't know, it probably is), it might be cleaner to exit early with just:

```
if (!this._hasDatePermissions) {
  return;
}
```

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -126,6 +149,29 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+						text="${this.localize('content.addDueDate')}"
+						@click=${this._showDueDate}
+						?hidden=${!this._showAddDueDateBtn}
+					>
+					</d2l-button-subtle>
+					<d2l-activity-due-date-editor
+						.href="${this.href}"
+						.token="${this.token}"
+						?skeleton="${this.skeleton}"
+						?hidden=${this._showAddDueDateBtn}
```
CodeBaboon, line 167: why do some of the right-side values get wrapped in double-quotes while others do not? Is it a case of "string values should be in quotes"? If so, I think maybe `?skeleton="${this.skeleton}"` should be `?skeleton=${this.skeleton}`?

imcdonald-d2l: I have never noticed the inconsistency before, good catch! I think it is because people like myself forget to add it and lit element doesn't seem to care whether it is there or not. According to their [docs here](https://lit-element.polymer-project.org/guide/templates#bind-properties-to-templated-elements), double quotes is the desired syntax.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -117,6 +126,20 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
 	}
 
+	_getDueDateAndPermission() {
+		const entity = activityStore.get(this.href);
+		if (!entity || !entity.dates) {
+			return;
+		}
+		const dates = entity.dates;
+		this._hasDatePermissions = dates.canEditDates;
+		const dueDate =  dates.dueDate ? dates.dueDate : null;
```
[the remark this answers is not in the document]
imcdonald-d2l: Good to know, thanks!

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: I tested by hardcoding `_hasDatePermissions` to `false` and everything works as expected with no rendering issues when we don't return anything. So don't think we need to add additional return statements here.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
dbatiste: That's the practice that we have been following in `core` (to use the double quotes).  Too bad there doesn't appear to be a lit linting rule for this.  😞

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ianlaidlaw: Reading the docs, and I'm not positive what the correct approach is, but I will mention that they don't even talk about simply not returning. To me, it feels like bad practice as a function should have 1 return type, but here we could return html or "void". I know that js allows this kind of stuff but it feels slightly wrong to me. From the docs:

```
Sometimes, you may want to render nothing at all. The values undefined, null and the empty string ('') in a text binding all render an empty text node. In most cases, that's exactly what you want:

import {html} from 'lit-html';
${user.isAdmin
      ? html`<button>DELETE</button>`
      : ''
  }
In some cases, you want to render nothing at all. In these cases, you can use the nothing value provided by lit-html.

import {html, nothing} from 'lit-html';
${user.isAdmin
      ? html`<button>DELETE</button>`
      : nothing
  }
In this case, when user.isAdmin is false, no text node is rendered.
```

ianlaidlaw: https://lit-html.polymer-project.org/guide/writing-templates

dbatiste: The `nothing` from `lit-html` is handy for avoiding unwanted text nodes that would suppress default slot content, but is rarely actually needed.  That being said, we usually don't use `nothing` unless it's needed.

imcdonald-d2l: Returning empty `html`  when condition isn't met. Let me know if you'd prefer it set to `nothing`.

ianlaidlaw: i'm inclined to think that empty html is the best approach here. we don't need to take advantage of `nothing` here, and empty html is the other example the docs use.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -126,6 +148,31 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+					<d2l-activity-due-date-editor
+						.href="${this.href}"
+						.token="${this.token}"
+						?skeleton="${this.skeleton}"
+						?hidden="${this._showAddDueDateBtn}"
+					>
+					</d2l-activity-due-date-editor>
+				</div>
+			`;
+		} else {
```
ianlaidlaw, line 171: i like the fact that we're returning something in this path now, but small nit: you can remove the `else` statement if you like. since you're returning something in the `if`, it already is an `if/else`

ianlaidlaw: again this is all up to you, but imo the absolute cleanest way to write this would be to invert it. it keeps what you want not nested.

```
if (!hasDatePermissions) {
  return html``
}

return html`
 <main render stuff />
`
```

ianlaidlaw (approved):

ianlaidlaw (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -49,12 +54,22 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+		this._hasDatePermissions = false;
+		this._showAddDueDateBtn = true;
 	}
 
 	render() {
 		const contentEntity = store.getContentActivity(this.href);
-		const title = contentEntity && contentEntity.entityType === CONTENT_TYPES.module ? contentEntity.moduleTitle : '';
-		const descriptionRichText = contentEntity && contentEntity.entityType === CONTENT_TYPES.module ? contentEntity.moduleDescriptionRichText : '';
+		let title = '';
+		let descriptionRichText = undefined;
```
imcdonald-d2l, line 64: we have to initialize `descriptionRichText` value to `undefined` if we want the value updating properly when we grab it from state. This is because the line of code [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-html-editor.js#L102).  Specifically this piece: 
```
&& typeof changedProperties.get('value') === 'undefined'
```
The reason for this condition is better explained in this [PR comment](https://github.com/BrightspaceHypermediaComponents/activities/pull/1121#issuecomment-697950477). 

Thanks to @mdgbayly for helping me solve this issue!

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -49,12 +54,22 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
+		this.skeleton = true;
+		if (contentEntity) {
+			// Show loading skeleton until we have the content entity loaded into state
+			this.skeleton = false;
```
imcdonald-d2l, line 68: ### Before
edit-content-item-bad-skeleton

### After
edit-content-item

.. and yes, edit view is working! 🎉

ianlaidlaw (approved):

ChrisKraljevicD2L (approved):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -73,12 +88,13 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
 			</div>
+			${this._renderDueDate()}
 			<div id="content-description-container">
 				<label class="d2l-label-text d2l-skeletize" for="content-description">${this.localize('content.description')}</label>
 				<div class="d2l-skeletize">
 					<d2l-activity-html-editor
-						id='content-description'
-						ariaLabel="content-description"
+						.ariaLabel="content-description"
+						.key="content-description"
```
CodeBaboon, line 97: I think the `.` prefix isn't used for native html attributes? Not sure (though I do see that `.value` has it below...).

CodeBaboon (approved):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: Yeah, it might not be needed. I can look into that next time I'm working in activities.

dbatiste: In Lit, the `.` directive is for assigning to the **property**, which in this case is `ariaLabel`.  However most times I think we specify the **attribute**, which would be `aria-label`.  For some things, it would make a difference (attribute vs property) but in this case I _think_ it's six or one half dozen of the other.

PR #1181, opened 2020-11-13
Added 12am as start date for assignment date picker

ronan-f:
Based on [US121708](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/448686592196).

I had thought of removing the `startDateDefaultTime` property completely and having the default start time set at the activity level to ensure consistency across all current/future activities however the content tool has their default set to 11:59pm and I did not want to introduce any changes that affect other teams.


mostlyfabulous (commented):
I'm assuming the string `"00:00:00"` gets parsed without issue but LGTM

ronan-f:
> I'm assuming the string `"00:00:00"` gets parsed without issue but LGTM

Yip if you have a look at the [d2l-input-date-time component in core](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-date-time.js#L67) it takes a `timeDefaultValue` property which accepts the following: `Valid values are times in ISO 8601 time format ("hh:mm:ss"), "startOfDay", "endOfDay".`

stevenao (approved):

PR #1183, opened 2020-11-16
Use ListItemLinkMixin for custom work-to-do list item.

dbatiste:
Once this is merged, we can remove the `actionHref` and link stuff that's currently baked directly into `ListItemMixin`.

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
@@ -13,11 +13,10 @@ import { ActivityAllowList } from './env';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 import { ifDefined } from 'lit-html/directives/if-defined';
 import { ListItemLinkMixin } from '@brightspace-ui/core/components/list/list-item-link-mixin';
-import { ListItemMixin } from '@brightspace-ui/core/components/list/list-item-mixin';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { nothing } from 'lit-html';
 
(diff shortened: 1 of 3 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -12,11 +12,12 @@ import { ActivityUsageEntity } from 'siren-sdk/src/activities/ActivityUsageEntit
 import { ActivityAllowList } from './env';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 import { ifDefined } from 'lit-html/directives/if-defined';
+import { ListItemLinkMixin } from '@brightspace-ui/core/components/list/list-item-link-mixin';
 import { ListItemMixin } from '@brightspace-ui/core/components/list/list-item-mixin';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { nothing } from 'lit-html';
 
-class ActivityListItemDetailed extends ListItemMixin(EntityMixinLit(LocalizeMixin(LitElement))) {
+class ActivityListItemDetailed extends ListItemLinkMixin(ListItemMixin(EntityMixinLit(LocalizeMixin(LitElement)))) {
```
m6jones, line 20: Do we need to add both `ListItemMixin` and `ListItemLinkMixin`?

dbatiste: Currently yes.  But hmmm, I could have `ListItemLinkMixin` and `ListItemButtonMixin` extend `ListItemMixin`.  I think that would be reasonable... since I don't think they would ever extend anything else.

m6jones: That would be amazing. Seems redundant having to import both.

m6jones (dismissed):
Depending on when you are able to do the combining `ListItemLinkMixin` and `ListItemMixin` This looks great.

m6jones (approved):
Nice!

PR #1184, opened 2020-11-16
Delete module on cancel

imcdonald-d2l:
Deletes the module when canceling the creation of a new item. 

**Related PRs**:
- https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/226
- https://github.com/Brightspace/lms/pull/3414

create-cancel-delete

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -111,7 +111,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		}
 	}
 
-	cancelCreate() {
+	async cancelCreate() {
 		const contentEntity = store.getContentActivity(this.href);
 		return contentEntity && contentEntity.cancelCreate();
 	}
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -113,7 +113,11 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 
 	async cancelCreate() {
 		const contentEntity = store.getContentActivity(this.href);
-		return contentEntity && contentEntity.cancelCreate();
+		if (!contentEntity) {
+			return;
+		}
+
+		await contentEntity.cancelCreate();
 	}
 
 	hasPendingChanges() {
--- a/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
@@ -16,8 +16,9 @@ export class Content {
 	}
 
 	cancelCreate() {
-		// This is the function that is called when cancelling the creation of a NEW content item
-		// TODO - add functionality to delete created activity
+		if (this.entityType === CONTENT_TYPES.module) {
+			return this._contentModule.deleteModule();
+		}
 		return;
 	}
 
--- a/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
@@ -15,11 +15,10 @@ export class Content {
 		this.moduleDescriptionRichText = '';
 	}
 
-	cancelCreate() {
+	async cancelCreate() {
 		if (this.entityType === CONTENT_TYPES.module) {
-			return this._contentModule.deleteModule();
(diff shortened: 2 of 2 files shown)


CodeBaboon (approved):

brissons (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -111,7 +111,7 @@ class ContentEditorDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 		}
 	}
 
-	cancelCreate() {
+	async cancelCreate() {
```
ianlaidlaw, line 114: what in this function is async?

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
```diff
@@ -16,8 +16,9 @@ export class Content {
 	}
 
 	cancelCreate() {
-		// This is the function that is called when cancelling the creation of a NEW content item
-		// TODO - add functionality to delete created activity
+		if (this.entityType === CONTENT_TYPES.module) {
+			return this._contentModule.deleteModule();
```
ianlaidlaw, line 20: Is this the function you were mentioning that will return the 204 status code and is needed?

ianlaidlaw (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
imcdonald-d2l: No, this is necessary in the `siren-sdk` repo:

This needs to be done so the siren-sdk responds properly to delete requests. We want it to hit this [line in SirenAction.js](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/SirenAction.js#L85) instead of [this one](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/SirenAction.js#L91) or the `resp.json()` will blow up.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: `contentEntity.cancelCreate()` returns the async function defined in siren-sdk but I guess it might not be necessary since we are just returning it.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
ianlaidlaw: Sorry, not sure I understand the answer. What I'm trying to get at is, what is the response of `this._contentModule.deleteModule();` ? and do you need to return it? Cuz I'm wondering how the empty `return;` below would come in to play

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ianlaidlaw: shouldnt be a problem, I think. making a function automatically changes the return type by wrapping it in a `Promise.resolve()` but if that is okay then this should be fine

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
imcdonald-d2l: I was just following the pattern assignments had used for cancel workflow. I don't know if returning is absolutely necessary since the function defined in siren sdk doesn't really return anything.

Not sure if we need it because of the container mixin here when it executes the cancelCreate functions:
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L81

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
imcdonald-d2l: I refactored this function @ianlaidlaw so it behaves the same way as saving. Turns out we don't need to return anything and can just await the delete request.

components/d2l-activity-editor/d2l-activity-content-editor/state/content.js
imcdonald-d2l: Turns out we don't need to return anything. Was playing around with it and it works fine if we set it up the same way we have saving working.

ianlaidlaw: cool, thats what i was thinking by the way it looked

imcdonald-d2l: Thanks for pointing that out though, it was a good catch! 🙂

ianlaidlaw (approved):

PR #1190, opened 2020-11-20
US122429 Quiz password ui

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F456674904728

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-availability-editor.js
@@ -1,6 +1,7 @@
 import '../d2l-activity-accordion-collapse.js';
 import '../d2l-activity-availability-dates-editor.js';
 import '../d2l-activity-availability-dates-summary.js';
+import './d2l-activity-quiz-password-editor.js';
 import { accordionStyles } from '../styles/accordion-styles';
 import { ActivityEditorFeaturesMixin } from '../mixins/d2l-activity-editor-features-mixin.js';
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
@@ -16,6 +17,7 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 	static get properties() {
 
 		return {
+			activityUsageHref: { type: String },
 			href: { type: String },
 			token: { type: Object }
 		};
@@ -45,13 +47,14 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 				<li slot="summary-items">${this._renderAvailabilityDatesSummary()}</li>
 				<span slot="components">
 					${this._renderAvailabilityDatesEditor()}
+					${this._renderPasswordEditor()}
 				</span>
 			</d2l-activity-accordion-collapse>
 		`;
 	}
 	// Returns true if any error states relevant to this accordion are set
 	_errorInAccordion() {
-		const activity = store.get(this.href);
+		const activity = store.get(this.activityUsageHref);
 		if (!activity || !activity.dates) {
 			return false;
 		}
@@ -64,7 +67,7 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 		return html`
 			<div class="d2l-editor">
 				<d2l-activity-availability-dates-editor
-					href="${this.href}"
+					href="${this.activityUsageHref}"
 					.token="${this.token}"
 					startDateDefaultTime="00:00:00">
 				</d2l-activity-availability-dates-editor>
@@ -76,11 +79,22 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 
 		return html`
 			<d2l-activity-availability-dates-summary
-				href="${this.href}"
+				href="${this.activityUsageHref}"
 				.token="${this.token}">
 			</d2l-activity-availability-dates-summary>
 		`;
 	}
+
+	_renderPasswordEditor() {
+
+		return html`
+			<d2l-activity-quiz-password-editor
+				.href="${this.href}"
+				.token="${this.token}">
+			</d2l-activity-quiz-password-editor>
+		`;
+	}
+
 }
 
 customElements.define(
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-password-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-password-editor.js
@@ -1,0 +1,66 @@
+import { css, html } from 'lit-element/lit-element.js';
+import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
+import { LocalizeActivityQuizEditorMixin } from './mixins/d2l-activity-quiz-lang-mixin';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { shared as store } from './state/quiz-store.js';
+
+class ActivityQuizPasswordEditor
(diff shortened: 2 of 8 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/lang/en.js
```diff
@@ -7,8 +7,9 @@ export default {
 	"hdrTimingAndDisplay": "Timing & Display", // timing/display accordion header
 	"hdrAttemptsAndCompletion": "Attempts & Completion", //attempts/completion accordion header
 	"displayTools": "Display", // Title for display tools
-	"hintsToolDescription": "Allow hints", // description for allow hints
-	"hintsAllowedSummary": "Hints allowed", // summary text when hints are enabled
+	"hintsToolDescription": "Allow hints", // decription for hints tool
+	"hintsAllowedSummary": "Hints allowed", // summary text when hints are enabled,
+	"passwordLabel": "Password", // Lable for the quiz password text input field
```
mostlyfabulous, line 12: Typo nit on code comment `// Lable`

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
```diff
@@ -1,6 +1,4 @@
 import '../d2l-activity-accordion-collapse.js';
-import '../d2l-activity-availability-dates-editor.js';
-import '../d2l-activity-availability-dates-summary.js';
```
mostlyfabulous, line 3: Where did the summary go? I may have missed it.

mostlyfabulous:
Not sure how many tests we should add for this but I would suggest:

1. Expect failure when setting password and `canEditPassword` is `false`
2. Expect failure when setting password to a password `> 20` chars and `canEditPassword` is `true`

Probably don't need more as I'm hoping the `<d2l-input-text>` component has checks on empty string, weird characters and other edge cases.

ronan-f:
> Not sure how many tests we should add for this but I would suggest:
> 
> 1. Expect failure when setting password and `canEditPassword` is `false`
> 2. Expect failure when setting password to a password `> 20` chars and `canEditPassword` is `true`
> 
> Probably don't need more as I'm hoping the `<d2l-input-text>` component has checks on empty string, weird characters and other edge cases.

Yip good suggestions. To address them:

1. So the saving of a password happens in the `siren-sdk` repo and there's a check to ensure that a user has edit permissions at that point. Updating the mobx store does not perform such a check so I don't see that we can write a test for this. Also, the input component _should_ be disabled and prevent input if the user does not have the correct permissions.
2. There's no logic in any of the functions I've added that will restrict input to 20 chars. This is implemented in the `<d2l-input-text>` component which seems to have pretty [comprehensive test coverage](https://github.com/BrightspaceUI/core/blob/master/components/inputs/test/input-text.test.js#L215).

Would be curious to hear other thoughts around maybe adding some additional checks for these things in this repo.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
ronan-f: Yeah neither the dates editor or summary are rendered in this component so no need for these imports. Probably got left over from a previous change.

components/d2l-activity-editor/d2l-activity-quiz-editor/lang/en.js
ronan-f: Good catch! Updated.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
mostlyfabulous: Good to know

test/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.spec.js
```diff
@@ -74,4 +76,12 @@ describe('Quiz', function() {
 		expect(quiz.name).to.equal('No Homework');
 	});
 
+	it('setPassword', async() => {
+		const quiz = new Quiz('http://quiz/1', 'token');
+		await quiz.fetch();
+		quiz.setPassword('super-secret-passord');
```
wongvincent, line 82: nit. typo on `password`

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -87,11 +94,13 @@ decorate(Quiz, {
 	canEditHints: observable,
 	canEditDisableRightClick: observable,
 	hintsToolEnabled: observable,
+	password: observable,
```
wongvincent, line 97:
```suggestion
	password: observable,
	canEditPassword: observable,
```

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-availability-editor.js
```diff
@@ -64,7 +67,7 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 		return html`
 			<div class="d2l-editor">
 				<d2l-activity-availability-dates-editor
-					href="${this.href}"
+					href="${this.activityUsageHref}"
```
wongvincent, line 70: Are these `activityUsageHref` changes supposed to be part of this PR?

ronan-f: Yeah I believe so. `<d2l-activity-availability-dates-summary>` needs the `activityUsageHref` and `<d2l-activity-quiz-password-editor>` needs the specific quizzing href from what I could tell. Would love to know if there's a better way to do that?

ronan-f: The dates-editor accesses `activity-store.js` as seen [here ](https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js?r=bd2c687d#8) which to my understanding means that it would need an `activityUsageHref` whereas the password editor uses `quiz-store.js` which is expecting the specific quizzing  href.

wongvincent: yeah, I see now that 
`d2l-activity-quiz-editor-secondary.js` is passing in the quiz href as `this.href` so I understand why this is changed now to use `this.activityUsageHref`.

wongvincent (approved):

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-password-editor.js
```diff
@@ -0,0 +1,66 @@
...
+			<d2l-input-text
+				?skeleton="${this.skeleton}"
+				id="quiz-password"
+				maxlength="20"
+				value="${entity.password}"
+				@input="${this._setPassword}"
+				label="${this.localize('passwordLabel')}"
+				?disabled="${!entity.canEditPassword}"
+				prevent-submit>
+			</d2l-input-text>
```
mdulat, line 48: Sorry I'm late to the game here, but should this input have `type="password"`? I think that just hides the input while the user types it in. Maybe a question for Patrick or Joseph.

ronan-f: Yeah when I first started this story I spoke with Joseph about password security (do we need to hash it/hide it etc) and my understanding was that it's not treated as a "secure" password but more like how a Zoom password would work. Looking at classic quizzing the password is not hidden so I suspect we should have the same behaviour. 

i will confirm this with Joseph though.

PR #1191, opened 2020-11-20
Add qe widget for inspiration 2020

AlexBedley:
Made the widget.
image

It expects 3 properties:

- `token` the same tokenGetter as QE and CE
- `href` an href that returns a collection of Activity Usages
- `quick-eval-href` an href that points to Quick Eval

I tried to re-use as much as I could. One thing I noticed right away is we tied all our controller methods to Polymer so I couldn't use them in Lit. Oops. Luckily this app is small so not a ton of duplication and the existing code provided a good example to follow.

I only had the screenshot to go on so I guesstimated a lot of the finer details of the UI. Hopefully not too many changes are needed.

Lastly, I didn't get around to adding langterms or checking accessibility. We got responsiveness and RTL for free by using core components. I also added a quick and dirty loading skeleton.

**NOTES:**
- Just a reminder to add as a web-component in brightspace-integration
- US123014


components/d2l-quick-eval-widget/controller.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export async function fetchSubmissionCount(activityUsageEntity, token) {
+
+	const evalStatusRel = 'https://activities.api.brightspace.com/rels/evaluation-status';
+	const evalStatusLink = activityUsageEntity.getLink(evalStatusRel);
+
+	const evaluationStatusEntity = await fetch(evalStatusLink, token);
+
+	return evaluationStatusEntity.properties.completed || 0;
```
amauryperez, line 47: I think here we probably want `newsubmissions` + `resubmissions` which is the subset of the `completed` ones that have not being evaluated yet or evaluated with resubmissions

amauryperez (dismissed):

amauryperez:
remember to add it bsi

components/d2l-quick-eval-widget/activity-card.js
```diff
@@ -0,0 +1,40 @@
...
+		` ];
+	}
+
+	render() {
+		return html`
+			<d2l-list-item>
+				<d2l-icon slot="illustration" icon="${this.icon}"></d2l-icon>
+				<d2l-list-item-content>
+					<d2l-link href="${this.evaluateAllHref}">${this.activityName}</d2l-link>
+					<p class="d2l-body-compact d2l-body-small">${this.courseName + '\t\u2022\t' + this.dueDate}</p>
```
AlexBedley, line 33: not really sure if we want `/t` as a separator, I don't really understand the implications

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -0,0 +1,130 @@
+import { css, html, LitElement } from 'lit-element';
+import { linkStyles } from '@brightspace-ui/core/components/link/link.js';
+import '@brightspace-ui/core/components/list/list.js';
+import { heading2Styles, bodyCompactStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import { fetchActivities, fetchActivityName, fetchCourseName, getDueDate, fetchSubmissionCount, determineIcon, fetchEvaluateAllHref, setToggleState } from './controller.js';
+import './activity-card';
```
AlexBedley, line 6: hmm I should probably give it a more specific name to differentiate it from the many other activity cards

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -0,0 +1,130 @@
...
+		this.count = 3;
+	}
+
+	async updated(changedProperties) {
+		super.updated();
+
+		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token) {
+			this.skeleton = true;
+			this.activities = await this.getActivities(this.href, this.token);
+			this.skeleton = false;
```
AlexBedley, line 60: hmm this should probably be in a `finally`

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+
+	async getActivities(href, token) {
+		const unassessedActivityCollection = await fetchActivities(href, token);
+		const activityUsages = unassessedActivityCollection.entities.slice(0, this.count);
```
AlexBedley, line 66: I'm not a huge fan of the way we're dealing with `count` here. I know the API was changed to only return 3 items but that feels weird. Ideally the API just accepts a `count` parameter and the front-end tells it how many it wants. I'm loath to get rid of count here because 1. it makes the loading animation better and 2. it makes the front-end more reusable

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -0,0 +1,130 @@
...
+		// delete when d2l-list supports SkeletonMixin
+		const loading = [];
+		for (let i = 0; i < this.count ; i++) {
+			loading.push(html`<ol class="d2l-skeletize"><li></li><li></li></ol>`);
+		}
+
+		return html`
+			<div class="d2l-heading-2 d2l-quick-eval-widget-heading">Evaluations To Do</div>
+			<hr>
+			${ this.skeleton ? loading : loaded }
```
AlexBedley, line 124: I wonder if doing it this way is expensive (since I construct both html strings regardless of `this.skeleton`)

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -0,0 +1,130 @@
...
+			<div @click="${this.handleViewAll}" class="d2l-link d2l-body-compact" href="${this.quickEvalHref}">View All</div>
```
AlexBedley, line 125: this probably shouldn't be a `div` but I was worried about all the `button` semantics

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
amauryperez: The API takes a count but the actual widget renderer in the Lms calls it with 3 always so I'm fine with the front-end not worrying about the count

amauryperez (approved):

amauryperez:
remember to add it to bsi

PR #1196, opened 2020-11-23
Changed css selectors to target slotted elements

ronan-f:
Defect fix based on this trello item https://trello.com/c/pXTHX0jK/266-fix-summarizer-spacing.

I believe this defect was introduced during [a refactor](https://github.com/BrightspaceHypermediaComponents/activities/pull/1152/files) to create an accordion wrapper which meant that the css selectors were no longer targeting the slotted `li` elements. From a quick [bit of reading](https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted)/experimenting with the `::slotted()` selector I tried this approach which seems to correctly target the `li` elements. 

Not sure if there's a preferred way to target slotted elements. 

Before 

image

After

image


mdulat (approved):
Nice!

PR #1198, opened 2020-11-24
US121861 Quizzes: Add summarizer for disable pager and alerts

mostlyfabulous:
[US121861](https://rally1.rallydev.com/#/detail/userstory/449846810872?fdp=true): [ui] Summarizer: block alerts

```
Summarizer text displays, "Alerts and communication blocked" when block alerts and communications is checked.

Rules below:

Accordion:  Timing & Display
Not active / default text:  Show nothing

Active state:
"Alerts and communication blocked"

Other:
Ensure summarizer text is skeletonized on load
"Alerts and communication blocked" will be the third summary text item on the accordion.
Summary text order should match field ordering i.e. should appear beneath "Hints allowed" and/or "Right clicks disabled" if they are shown.
```


ronan-f (approved):
LGTM

PR #1200, opened 2020-11-25
US120978 - Notification Email Summary

wongvincent:


mdulat:
Just wondering is the email notification stuff different for assignments and quizzes? Or can we have one version of it be moved into the parent activity folder and used by both? I guess both the input and the summary?

mdulat:
(sorry clicked the wrong button to comment, didn't mean to close it!)

wongvincent:
@mdulat No, they're not different. I'm working on the error handling right now, and noticed that the implementation is the exact same. Even the crazy regex to determine if the email will be the same.

So yeah, I'm working on refactoring it.

wongvincent:
Oh wait, if it's an activity and not quiz or assignment specific, how will it know which store to use...

ronan-f:
> Oh wait, if it's an activity and not quiz or assignment specific, how will it know which store to use...

Even though the component is defined at the activity level won't it still be rendered at the quiz/assignment level? So when you render `<d2l-activity-notification-email-summary href="${this.href}">`, `this.href` would still be the specific quiz/assignments href, no?

wongvincent:
Yes, but to reference the correct MobX object, I think we would need which one to import
https://github.com/BrightspaceHypermediaComponents/activities/pull/1200/files#diff-f6d98737c85f44d121b9bb02d2b95f51add9cf4a36c2156eba48242cc1734deeR5

wongvincent:
The 2 summary strings are also different for Assignments and Quizzing. I'm thinking maybe we can keep the 2 summaries as separate components.

There is an improvement that can be made for the error validation. I'm thinking I could do this: https://github.com/BrightspaceHypermediaComponents/activities/commit/bbda0ec78eb3408c42edeea27faa6a281499ce3d
which could be used by both Assignments and Quizzing so that I don't duplicate the notification email checking and error tooltip logic.

mdulat (approved):

mostlyfabulous:
> > Oh wait, if it's an activity and not quiz or assignment specific, how will it know which store to use...
> 
> Even though the component is defined at the activity level won't it still be rendered at the quiz/assignment level? So when you render `<d2l-activity-notification-email-summary href="${this.href}">`, `this.href` would still be the specific quiz/assignments href, no?

> Yes, but to reference the correct MobX object, I think we would need which one to import

If our components have the need to import the correct MobX store for activity we're developing, is there a way we could make this component more extendable/abstract. Ideally we would have generic/abstract summarizer component that would have a child component that implements the relevant activity MobX store? But is that too complicated for a simple component?

PR #1211, opened 2020-11-27
US119477 add release conditions compoment to quiz

mdulat:
First part of https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F413484054592&fdp=true
There is no dirty check (also not in Assignments, need to confirm whether one is needed)
Subsequent PR will show read-only version if permission lacking, for both quiz and assignment.


ronan-f (approved):
Looks good to me!

PR #1226, opened 2020-12-02
adding empty stage

dlockhart:
For the visual-diff "Regenerate Goldens" button to work, it searched for "Stage 2: Visual-difference-tests" -- which means it can't be the first and only stage. 🤦 

This is just a temporary measure until the visual-diff GitHub Action is ready.


svanherk (approved):
LOL

PR #1228, opened 2020-12-02
Fix: FACE Assignments - no help/info text below 'Individual Assignment'

mostlyfabulous:
https://trello.com/c/0kvcVuBm/3-course-with-no-groups-missing-no-groups-exist-create-new-groups-in-the-groups-tool-under-assignment-type

Should always show `No groups exist. Create new groups in the Groups tool.`, unless Groups exist.

Also nudged the `Create new groups in the Groups tool` to align with the Group Category dropdown selector.

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
@@ -70,11 +70,11 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 			return html``;
 		}
 
-		const isIndividualType = assignment.assignmentTypeProps.isIndividualAssignmentType;
 		const infoText = this._getInformationText(assignment);
+		const isIndividualType = assignment.assignmentTypeProps.isIndividualAssignmentType;
 		const canEditAssignmentType = assignment.assignmentTypeProps.canEditAssignmentType;
 		const groupTypeDisabled = assignment.assignmentTypeProps.isGroupAssignmentTypeDisabled;
-		const folderTypeText =	isIndividualType ? this.localize('txtIndividual') : this.localize('txtGroup');
+		const folderTypeText = isIndividualType ? this.localize('txtIndividual') : this.localize('txtGroup');
 		const groupTypeText = !isIndividualType && assignment.assignmentTypeProps.selectedGroupCategoryName
 			? this.localize('txtGroupCategoryWithName', 'groupCategory', assignment.assignmentTypeProps.selectedGroupCategoryName)
 			: '';
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
@@ -82,7 +82,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 		return html`
 			<div ?hidden=${canEditAssignmentType} id="read-only-assignment-type-container">
 				<div class="d2l-body-compact">${folderTypeText}</div>
-				<p slot="description" class="d2l-info-text-flush-left d2l-body-small">${infoText}</p>
+				<p class="d2l-info-text-flush-left d2l-body-small">${infoText}</p>
 				<div class="d2l-body-compact">${groupTypeText}</div>
 			</div>
 
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
@@ -45,7 +45,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 
 				.d2l-info-text {
 					margin: 0.1rem 0 0 0;
-					padding-left: 1.7rem;
+					padding-left: 1.8rem;
 				}
 
 				.d2l-info-text-flush-left {
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
@@ -145,14 +145,18 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 			return;
 		}
 
+		const canEditAssignmentType = assignment.assignmentTypeProps.canEditAssignmentType;
 		const isIndividualAssignmentType = assignment.assignmentTypeProps.isIndividualAssignmentType;
 		const hasSubmissions = assignment.submissionAndCompletionProps.assignmentHasSubmissions;
+		const groupTypeDisabled = assignment.assignmentTypeProps.isGroupAssignmentTypeDisabled;
 
-		if (!hasSubmissions &&  isIndividualAssignmentType && assignment.assignmentTypeProps.groupCategories.length === 0) {
+		if (hasSubmissions) return; // don't display either of the information texts about groups
+
+		if (isIndividualAssignmentType && groupTypeDisabled && assignment.assignmentTypeProps.groupCategories.length === 0) {
 			return this.localize('folderTypeNoGroups'); // this only displays below the 'Individual Assignment' text
 		}
 
-		if (!hasSubmissions && !isIndividualAssignmentType) {
+		if (!isIndividualAssignmentType && canEditAssignmentType) {
 			return this.localize('folderTypeCreateGroups'); // this only displays below the 'Group Assignment' text
 		}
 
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -78,6 +82,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 		return html`
 			<div ?hidden=${canEditAssignmentType} id="read-only-assignment-type-container">
 				<div class="d2l-body-compact">${folderTypeText}</div>
+				<p slot="description" class="d2l-info-text-flush-left d2l-body-small">${infoText}</p>
```
mdulat, line 85: Why is this a `slot`?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
```diff
@@ -78,6 +82,7 @@ class AssignmentTypeEditor extends ActivityEditorMixin(RtlMixin(LocalizeActivity
 		return html`
 			<div ?hidden=${canEditAssignmentType} id="read-only-assignment-type-container">
```
mdulat, line 83: Looks like this block is rendered when the user does not have permissions to edit assignment type. Do we want to display this text even when the user is not able to change assignment type?

mdulat: I notice that when I create a new assignment, then by default it does not have these actions checked for here: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L227 and therefore returns assignment is read-only. So maybe the issue is the setting of the assignment as read-only (i.e. canEditAssignmentType is false), and that's why we don't see the text?

mostlyfabulous: >Looks like this block is rendered when the user does not have permissions to edit assignment type. Do we want to display this text even when the user is not able to change assignment type?

I'll double check with Joseph as I may not have phrased it clearly that it would render when they can't change assignment type.

mostlyfabulous: >So maybe the issue is the setting of the assignment as read-only (i.e. canEditAssignmentType is false), and that's why we don't see the text?

Yeah I noticed that it would not display at all as `canEditAssignmentType` would keep it's containing element `<div ?hidden=${!canEditAssignmentType} id="editable-assignment-type-container">` from being displayed. Perhaps moving the info text to display outside that div would help but there would still be a styling issue to fix as one info text needs to be indented to line up (`Create new groups in the Groups tool`).

mostlyfabulous: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L227

Are those actions being set in the LMS when a new assignment is created?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
mostlyfabulous: That was copied over from when I was trying to get some styles to apply but that should go since it doesn't do anything here.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js
mdulat: That's what I'm wondering, maybe the problem isn't with the UI code here, maybe it's that those actions aren't being set correctly?

wongvincent: sorta. When there are no groups. The user shouldn't be able to edit between individual and groups. So that's why the action is missing.

wongvincent: but yeah, I think the action should be present even in there are no groups. Because right now, we can't tell when something can't be edited of missing permissions (in which case we shouldn't be showing the infoText) or if they have permission but there are no groups.

If we changed it, so the action is always included if they have edit permissions (disregarding whether there are groups or not), then we can handle it in the UI like this...

So there are 3 scenarios:
1. Have edit action, has groups -> show dropdown, show infoText
2. Have edit action, individual assn type, no groups -> don't show dropdown (show read-only type), show infoText
3. Lack edit action -> don't show dropdown, don't show infoText

mdulat: Just looking at the `folder-type` entity in different scenarios. Probably those actions shouldn't be present when there are no groups because a `set-to-group` action doesn't make sense when there are no groups. But there are other ways to determine whether permissions are present (and therefore info text for groups should be displayed). Like if permissions exists, but there are no groups in the course, then the `folder-type` entity has a `no-group-type` class. 

It also has a `informationText` property set to "No groups exist. Create new groups in the Groups tool." and I'm wondering why we're not using this property rather than duplicating that lang term within activities? Does anyone know?

wongvincent: Yes, a class would work too. not sure why we're not using that string.

mostlyfabulous: >Yes, a class would work too. not sure why we're not using that string.

Just to leave an answer here, it's due to moving away from autosave behaviour and now using Save and Close instead. Apparently now that we don't have autosave, the MobX store behaves differently and doesn't update with the langterms/informationText we would want to show the user when they choose different options when editing an Assignment.

> the terms to exist on the client side in order to allow switching things without calling save

mostlyfabulous: So after much digging, I've traced back what in the LMS decides an Assignment is read-only/they have _edit permissions_.

Permissions for `CanEditFolder` (which maps to the security constant `AddEditFolder` which pairs with _Add/Edit Assignment Submission Folders_ in **Roles and Permissions**) are checked here:
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=0ccdf8d6#570

No permissions doesn't have an explicit singular symbol/property/thing to state it (which confounds things IMO):
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=0ccdf8d6#931

Permissions to edit will append classes, actions & properties to the response depending on what is valid:
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=0ccdf8d6#788 

So there are two actions which will indicate a user has permissions to edit the assignment (read-only might be a bit different as I'm guessing we should say that is the case when there's been a submission): 
`update-submissions-rule` and `update-submissions-rule`, could we use those or should we do something like `builder.AddClass( "read-only" );`?

wongvincent: Yep, I would add a class for when they cannot edit the assignment, to differentiate between missing the action because they're missing the permission and missing the actions because they can edit but there are no groups.

mdulat: Couldn't we just use the information we currently have for `folder-type` ? I.e.

- No permissions, no groups: actions within folder-type entity don't exist, `no-group-type` class is NOT there. In this case, don't display help text.
- No permissions, with groups: same as above.
- User has permissions, no groups: actions within folder-type entity don't exist, `no-group-type` class IS there. In this case, display the "no groups exist" help text.
- User has permission, with groups: actions within folder-type entity exist. In this case, display the "create new groups" help text.

(You should double-check that the actions/classes I listed for each case is correct, I'm just going off memory).

wongvincent: `_getInformationText` function needs to be updated to be something like this I think...
```
if (has has-submissions class) {
  return;
}
if (is individual type && has no-groups-type class) {
  return that string about no groups exist;
}
if (is group && hasAction) { //we have to check the action here
  return that other string about creating new groups
}
```

mostlyfabulous:
### Course with no Groups
**Left: No perms | Right: Perms**

_Individual Assignment Type_
image

### Course with Groups
**Left: No perms | Right: Perms**

_Group Assignment Type_
image
_Individual Assignment Type_
image

emil-furniss (approved):

PR #1229, opened 2020-12-03
Mdulat/us119477 no perms read only

mdulat:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F413484054592&fdp=true

This task shows release conditions as read-only if the user doesn't have edit permissions. Previously they were hidden completely without permissions.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
@@ -108,8 +108,7 @@ class ActivityAssignmentAvailabilityEditor extends SkeletonMixin(ActivityEditorF
 	}
 
 	_renderReleaseConditionEditor() {
-		const activity = store.get(this.href);
-		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.canEditReleaseConditions) {
+		if (!this._m3ReleaseConditionsEnabled) {
 			return html``;
 		}
 
--- a/components/d2l-activity-editor/d2l-activity-conditions-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-conditions-editor.js
@@ -20,7 +20,8 @@ class ActivityConditionsEditor
 	static get properties() {
 
 		return {
-			description: { type: String }
+			description: { type: String },
+			canEdit: { type: Boolean, attribute: 'can-edit' }
 		};
 	}
 
@@ -101,6 +102,7 @@ class ActivityConditionsEditor
 
 	constructor() {
 		super(store);
+		this.canEdit = false;
 	}
 
 	render() {
@@ -267,6 +269,9 @@ class ActivityConditionsEditor
 		}
 	}
 	_renderAddCondition(entity) {
+		if(!this.canEdit) {
+			return html``;
+		}
 
 		const { canAttachExisting, canCreateNew } = entity;
 
@@ -322,7 +327,7 @@ class ActivityConditionsEditor
 						.innerHTML="${title}">
 					</span>
 				</span>
-				<span class="d2l-list-item-deleter">
+				<span class="d2l-list-item-deleter" ?hidden="${!this.canEdit}">
 					<d2l-button-icon
 						text="${this.localize('editor.btnRemoveCondition')}"
 						icon="tier1:close-default"
@@ -388,6 +393,7 @@ class ActivityConditionsEditor
 				<select
 					class="d2l-input-select"
 					id="operator"
+					?disabled="${!this.canEdit}"
 					@change="${this._setOperator}">
 					${operators.map(this._renderOperator, this)}
 				</select>
--- a/components/d2l-activity-editor/d2l-activity-conditions-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-conditions-editor.js
@@ -269,7 +269,7 @@ class ActivityConditionsEditor
 		}
 	}
 	_renderAddCondition(entity) {
-		if(!this.canEdit) {
+		if (!this.canEdit) {
 			return html``;
 		}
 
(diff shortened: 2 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -108,8 +108,7 @@ class ActivityAssignmentAvailabilityEditor extends SkeletonMixin(ActivityEditorF
 	}
 
 	_renderReleaseConditionEditor() {
-		const activity = store.get(this.href);
-		if (!this._m3ReleaseConditionsEnabled || !activity || !activity.canEditReleaseConditions) {
+		if (!this._m3ReleaseConditionsEnabled) {
```
mostlyfabulous, line 111: Removing those other bools because the parent component will do this check anyway?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-availability-editor.js
```diff
@@ -126,10 +126,6 @@ class ActivityQuizAvailabilityEditor extends AsyncContainerMixin(LocalizeActivit
 	}
 
 	_renderReleaseConditionEditor() {
-		const activity = store.get(this.activityUsageHref);
-		if (!activity || !activity.canEditReleaseConditions) {
```
mostlyfabulous, line 130: Same question as in the `assignment-editor`: Removing those other bools because the parent component will do this check anyway?

components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -322,7 +327,7 @@ class ActivityConditionsEditor
 						.innerHTML="${title}">
 					</span>
 				</span>
-				<span class="d2l-list-item-deleter">
+				<span class="d2l-list-item-deleter" ?hidden="${!this.canEdit}">
```
mostlyfabulous, line 330: So this allows the user to 'read' the release conditions (if any) and remove the UI elements that would 'let' them edit it (since the back-end wouldn't let them edit it anyway)?

components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -20,7 +20,8 @@ class ActivityConditionsEditor
 	static get properties() {
 
 		return {
-			description: { type: String }
+			description: { type: String },
+			canEdit: { type: Boolean, attribute: 'can-edit' }
```
emil-furniss, line 24: Do we need to pass this as an attribute rather than just getting the value from the entity from within this component?

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
emil-furniss: I think this can be removed since we don't want to hide the component any more, just render it as read-only.

wongvincent: off the top of my head, I think this is supposed to gate the FACE assignments release conditions. 
And I don't think FACE Quizzing is including this as an attribute.
So if it's okay for FACE assignments to also no longer be controlled by the Flag, then we can remove this check.

components/d2l-activity-editor/d2l-activity-conditions-editor.js
```diff
@@ -388,6 +393,7 @@ class ActivityConditionsEditor
 				<select
 					class="d2l-input-select"
 					id="operator"
+					?disabled="${!this.canEdit}"
```
wongvincent, line 396: just wanna double check, should it be disabled or hidden if they don't have the permission?

mostlyfabulous: Yeah the US isn't clear on what UI elements to disable or hide when read-only. Any design docs for this other than the image from the US?
image

components/d2l-activity-editor/d2l-activity-conditions-editor.js
mdulat: yes

components/d2l-activity-editor/d2l-activity-conditions-editor.js
mdulat: I wasn't sure the best way to do it. I can get the value from the store from within this component, but I'll need to add the activityUsage store, as right now it only has the conditions store. Is that the better way?

components/d2l-activity-editor/d2l-activity-conditions-editor.js
mdulat: Disabled, as they should be able to see the current value, but not be able to change it.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-availability-editor.js
mdulat: We don't need to know the value of  `activity.canEditReleaseConditions` here any more because we no longer render anything differently in this component based on that value (the distinction between read-only and editable is made further down).

components/d2l-activity-editor/d2l-activity-conditions-editor.js
mostlyfabulous: @emil-furniss Would like to know the pros and cons of passing the `attribute` vs accessing the `activityUsage` store 😃

emil-furniss: @mdulat Nah you're right, we'd have to add the extra store which might muddy the waters unnecessarily.
@mostlyfabulous I'd say that using an attribute complicates the contract between the component and its consumers, and manipulating the DOM is likely to be less performant.

emil-furniss (approved):

PR #1230, opened 2020-12-03
US119278 - Work To Do Loading State

KearseTrevor:
- Creating loading states per [US119278](https://rally1.rallydev.com/#/detail/userstory/411540292156?fdp=true)
- Make resuable subcomponents skeleton aware where applicable
- Load More Spinner (or skeleton) is outside the scope of this PR


components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -44,6 +46,23 @@ class ActivityListHeader extends LocalizeMixin(LitElement) {
...
+				:host([skeleton]) .d2l-activity-list-header-content.d2l-heading-4 {
+					width: 50%;
+				}
+				:host([skeleton]) .d2l-activity-list-counter {
+					min-height: 0.8rem;
+					width: 5%;
+				}
+				:host([skeleton]) .d2l-activity-list-counter.d2l-heading-3 {
+					max-width: 2rem;
+				}
```
mbardakovD2L, line 65: are the skeleton styles still necessary, or were they just a holdover from the mixins-needing-to-be-in-a-specific-order debacle?

demo/d2l-work-to-do/data/base/activity-collection.json
```diff
@@ -0,0 +1,240 @@
+{
```
mbardakovD2L, line 1: do we want to commit the `json` files in the demo page, or keep leaving them out?

KearseTrevor: I purposely added in a couple during this work because waiting on server responses was taking too long

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
KearseTrevor: I've updated this - added a container around the _status component_ to help alleviate this. It's still necessary in a couple locations like that are now the only ones remaining

mbardakovD2L (approved):

PR #1233, opened 2020-12-04
Cleanup unused store

wongvincent:
The host passes a value to this component as notification value comes from the quiz or assignment store and not from the activities store. And so this component doesn't actually retrieve any values from stores. The import was left over from when I was moving it around from quizzing to the more generic activity folder.


mdulat (approved):

wongvincent:
I forgot to merge with "fix".. but it's minor enough that I'll just leave it for the next release to pick this up.

PR #1236, opened 2020-12-04
feat: US119477: make add button readonly when can't edit

mdulat:
I initially made the "Add Release Condition" button hidden when user doesn't have permissions to add conditions, but it's supposed to be visible but disabled, so this PR fixes that.


mostlyfabulous (approved):
LGTM

PR #1237, opened 2020-12-04
Dispatching async event when loading alignments

ronan-f:
Based on [this trello card](https://trello.com/c/Yx8JT17F/7-face-standards-link-has-an-extra-delay-in-showing-up-exceeding-testcafe-timeout).

This PR is designed to address the skeleton loading of standards and ensure we the assignment detail component stays in "skeleton mode" until alignments have finished loading.

This is accomplished by dispatching a 'pending' event which is handled by the `asyncContainerMixin` so that the detail component knows to continue displaying the skeleton view.

After a discussion with @emil-furniss we also thought that it made sense for `d2l-activity-assignment-editor-detail` to control the hidden state of `d2l-activity-outcomes` and have that based on the skeleton state rather than `d2l-activity-outcomes` trying to manage it's own hidden state internally. I've added those changes here as well.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-outcomes.js
+++ b/components/d2l-activity-editor/d2l-activity-outcomes.js
@@ -4,6 +4,7 @@ import 'd2l-activity-alignments/d2l-select-outcomes-hierarchical.js';
 import { ActivityEditorFeaturesMixin, Milestones } from './mixins/d2l-activity-editor-features-mixin.js';
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
+import { AsyncStateEvent } from '@brightspace-ui/core/helpers/asyncStateEvent';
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-outcomes.js
```diff
@@ -86,12 +88,28 @@ class ActivityOutcomes extends ActivityEditorFeaturesMixin(ActivityEditorMixin(R
...
 	}
 	_closeDialog() {
 		this._opened = false;
 	}
+
+	_dispatchLoadingAlignments() {
+		let res;
+
+		const promise = new Promise(resolve => res = resolve);
+		promise.resolve = res;
```
ronan-f, line 103: This all feels a little bit strange but seemed appropriate as we want the promise to resolve separately from where it's defined.

mostlyfabulous (approved):
Promise that this resolves the automation tests?

wongvincent (approved):

PR #1240, opened 2020-12-07
Update WorkToDo with Config from LE

surags:
https://rally1.rallydev.com/#/detail/userstory/422999259560?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -148,7 +148,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
 	}
 
 	attributeChangedCallback(name, oldval, newval) {
-		if(!window.D2L.workToDoOptions) {
+		if (!window.D2L.workToDoOptions) {
 			window.D2L.workToDoOptions = {};
 		}
 
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -146,7 +146,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
 	updated(changedProperties) {
 		changedProperties.forEach((oldValue, propName) => {
 			switch (propName) {
-        		case 'options': this._optionsChanged(this.options);
+				case 'options': this._optionsChanged(this.options);
 					break;
 			}
 		});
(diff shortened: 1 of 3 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -130,6 +134,23 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
...
+					break;
+			}
+		});
+	}
+
+	_optionsChanged(newOptions) {
+		const options = JSON.parse(newOptions);
+		this.fullscreen = options.fullscreen;
+		this._overdueWeekLimit = options.overdueWeekLimit;
+		this._upcomingWeekLimit = options.upcomingWeekLimit;
```
KearseTrevor, line 150: In the case that options doesn't carry these properties, do the values for the variables stay at what is set in the constructor? Or do they get reassigned to undefined? If the later, we'll require || assignments

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -130,6 +134,23 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
...
+		window.D2L.options = options;
```
KearseTrevor, line 151: Are we positive we aren't stepping on anyone's toes by using this window property? Potentially may want something more specific to WorkToDo (ex: window.D2L.WorkToDoOptions)

surags: Renamed

components/d2l-work-to-do/d2l-work-to-do.js
surags: Now only updates if exists in the options object

KearseTrevor:
In holding pattern until conversation in https://github.com/Brightspace/lms/pull/4295 (sibling PR on LE side) is put further along.  Suggest potentially moving this PR back to draft form if it looks like things are going to get shook up a reasonable amount over there.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -139,6 +147,29 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
...
 	}
 
+	attributeChangedCallback(name, oldval, newval) {
+		if (!window.D2L.workToDoOptions) {
+			window.D2L.workToDoOptions = {};
+		}
+
+		switch (name) {
+			case 'data-fullscreen':
+				this.fullscreen = Boolean(newval.toLowerCase());
```
KearseTrevor, line 157: Per conversation, needs to be changed to a different solve - current always evaluates to truthy

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -139,6 +147,29 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeMixin(LitElement)) {
...
+				break;
+			case 'data-upcoming-week-limit':
+				this._upcomingWeekLimit = parseInt(newval);
+				window.D2L.workToDoOptions.upcomingWeekLimit = this._upcomingWeekLimit;
+				break;
+			case 'data-overdue-week-limit':
+				this._overdueWeekLimit = parseInt(newval);
+				window.D2L.workToDoOptions.overdueWeekLimit = this._overdueWeekLimit;
+				break;
+		}
```
KearseTrevor, line 168: We may want to add protection against negative numbers

KearseTrevor (approved):

PR #1246, opened 2020-12-09
Us123060/quizzing password description

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/461232159644

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
@@ -59,6 +59,10 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 				.d2l-input-radio-label {
 					margin-bottom: 10px;
 				}
+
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -59,6 +59,10 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 				.d2l-input-radio-label {
 					margin-bottom: 10px;
 				}
+
+				#notification-email-editor-description {
```
ronan-f, line 63: This will affect assignments but ran it by Joseph/Patrick first. The idea is that if an input has a description then there should be margin between the description and the subheader. 

**With description/margin**
image

**Without description/no margin**
image

ronan-f: The change was added in this PR as quizzes password now has a description field and should be styled consistently with similar components in assignments.

wongvincent (approved):

PR #1205, opened 2020-11-27
US122816 - Refactor Activities attachment file picker to support entity and link

amauryperez:
This change will allow us to use `d2l-activity-attachments-picker-presentational` from the consistent evaluation web component without needing to provide the attachments siren link

PR #1247, opened 2020-12-09
Fix audio and video attachments not attaching

wongvincent:
See: https://trello.com/c/9r5lB48J/274-face-assignments-audio-video-fails-to-attach
> I think it's because this is storing the file in an array in `files`
https://github.com/BrightspaceHypermediaComponents/activities/pull/1205/files#diff-bfdd45cba209e9d6b9a6c4035f88d0893b72255673353a34b4fd39f33d3cd951R131
and this is looking for it in
`e.detail.file` https://github.com/BrightspaceHypermediaComponents/activities/pull/1205/files#diff-8fb51195a645ae0ccaf34162ad052073e0ab93377c35b6b42001687672f2228aR82
It needs to be `const file = e.detail.files[0]`
Same for video
The change was merged in December, so the defect will appear in 20.21.1 builds, but not 20.20.12

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -79,7 +79,7 @@ class ActivityAttachmentsPicker extends ActivityEditorMixin(SkeletonMixin(Locali
 	}
 
 	async _handlerAudioUploaded(e) {
-		const file = e.detail.file;
+		const file = e.detail.files[0];
 
 		const fileId = file.GetId();
 		const fileName = file.GetName();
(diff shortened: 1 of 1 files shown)


ryantmer:
Strange - the old https://github.com/BrightspaceHypermediaComponents/activities/pull/1205/files#diff-8fb51195a645ae0ccaf34162ad052073e0ab93377c35b6b42001687672f2228aL116 looks the same as the new https://github.com/BrightspaceHypermediaComponents/activities/pull/1205/files#diff-bfdd45cba209e9d6b9a6c4035f88d0893b72255673353a34b4fd39f33d3cd951R127 to me - which isn't to say this change doesn't make sense, but more how was this working before?!

wongvincent:
The dispatch is fine.
It's here where it gets the id from the file previously https://github.com/BrightspaceHypermediaComponents/activities/pull/1205/files#diff-8fb51195a645ae0ccaf34162ad052073e0ab93377c35b6b42001687672f2228aL109

But now that is handled from the event listener, so it receives it as an array in `e.detail.files`

ryantmer (approved):
Ah yes, I see now. Reading comprehension is hard 😜

PR #1255, opened 2020-12-10
Fix logged error with mobx when running karma tests

pureooze:


dlockhart:

PR #1257, opened 2020-12-11
Updated accessible text to match story

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F461235145752

Updated accessibility text to match user story.


mostlyfabulous (approved):
👍

PR #1262, opened 2020-12-11
Added poc mixin to handle dialog state

ronan-f:
An idea for a mixin which can clean up some of the logic regarding opening/closing dialogs. The idea is to prevent us from having to implement logic to manage the open/closed state of the dialog in each component that uses it and instead inherit that behavior from a shared mixin.


dustxd (approved):
💯

PR #1264, opened 2020-12-11
Us123063/refactor dialog open state

ronan-f:
Added a mixin to handle opening/closing dialogs as part of a refactor for this story https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F461234358860.

The idea is that any component that wants to implement a dialog can extend this mixin and get the open/closing logic for free. I had a discussion with Martin on this and as he pointed out the dialog natively supports managing it's own open state but would require us to do stuff like this to open the dialog:

```javascript
document.querySelector('#open').addEventListener('click', async() => {
  const action = await document.querySelector('d2l-dialog').open();
  console.log('dialog action:', action);
});
```

which in my opinion is nowhere near as nice/clean as the declarative approach:

```javascript
<d2l-dialog ?opened="${this.opened}"></d2l-dialog>
```

This mixin lets us use the declarative approach while also not needing to repeat ourselves with open/close logic in multiple components.


mdulat (approved):
LGTM

PR #1266, opened 2020-12-14
Returning entity.dirty() instead of entity.dirty

ronan-f:
Fix defect which always renders dirty check dialog


pureooze (approved):
Thanks for catching this!

PR #1267, opened 2020-12-14
Us123063/Added help dialog for disabling alerts

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F461234358860

Added a button which renders a help dialog for disabling alerts

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-disable-pager-and-alerts-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-disable-pager-and-alerts-editor.js
@@ -1,13 +1,16 @@
+import '@brightspace-ui/core/components/button/button-icon.js';
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import { ActivityEditorDialogMixin } from '../mixins/d2l-activity-editor-dialog-mixin.js';
 import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { checkboxStyles } from '../styles/checkbox-styles.js';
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-disable-pager-and-alerts-editor.js
```diff
@@ -33,8 +36,35 @@ class ActivityQuizDisablePagerAndAlertsEditor
 				@change="${this._setDisablePagerAndAlertsTool}"
 				ariaLabel="${this.localize('disablePagerAndAlertsDescription')}"
 				?disabled="${!entity.canEditDisablePagerAndAlerts}">
-				${this.localize('disablePagerAndAlertsDescription')}
+
+				<label class="d2l-input-checkbox-text">${this.localize('disablePagerAndAlertsDescription')}</label>
```
mdulat, line 40: I'm trying to remember why this wasn't a label initially. Is it because the screen reader reads it out twice, once for the ariaLabel, and once for the Label? I think it was something like that, would be good to test this with a screen reader to be sure it's what we want.

mdulat (approved):

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-disable-pager-and-alerts-editor.js
ronan-f: Interesting. I'm not sure on the procedure for getting a test like that done but I'll add it to the test notes and we can figure something out.

PR #1271, opened 2020-12-15
US123107: Support handling Audio/Video attachments as links which store URNs

ktonon:
Current audio/video attachment experience is:

- user clicks on a video and must download an html file
- file contains a link, which they must click
- link takes them to a page in the lms to view the media

This PR adds support for handling links which store a URN which points to an audio or video resource. The new experience is:

- user clicks on a video and is taken to the media player in a new tab

Future work will add audio/video embedded player. Storing the media as URNs decouples the storage from the mode of playback.

Related: https://github.com/Brightspace/lms/pull/4695

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
+++ b/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
@@ -79,15 +79,25 @@ class ActivityAttachmentsPicker extends ActivityEditorMixin(SkeletonMixin(Locali
 	}
 
 	async _handlerAudioUploaded(e) {
-		const file = e.detail.files[0];
+		this._handlerAudioVideoUploaded(e, { storeCreateFileMethod: 'createAudio' });
+	}
 
-		const fileId = file.GetId();
-		const fileName = file.GetName();
-		const fileSystemType = file.GetFileSystemType();
+	async _handlerAudioVideoUploaded(e, { storeCreateFileMethod }) {
+		const attachment = e.detail.files[0];
 
-		const collection = store.get(this.href);
-		const previewUrl = await collection.getPreviewUrl(fileSystemType, fileId);
-		this._addToCollection(attachmentStore.createAudio(fileName, fileSystemType, fileId, previewUrl));
+		if (attachment.GetFileSystemType) {
+			const file = attachment;
+			const fileId = file.GetId();
+			const fileName = file.GetName();
+			const fileSystemType = file.GetFileSystemType();
+
+			const collection = store.get(this.href);
+			const previewUrl = await collection.getPreviewUrl(fileSystemType, fileId);
+			this._addToCollection(attachmentStore[storeCreateFileMethod](fileName, fileSystemType, fileId, previewUrl));
+		} else if (attachment.GetLocation) {
+			const link = attachment;
+			this._addToCollection(attachmentStore.createLink(link.GetName(), link.GetLocation(), link.GetUrn()));
+		}
 	}
 
 	async _handlerFilesUploaded(e) {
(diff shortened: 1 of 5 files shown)


components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -79,15 +79,25 @@ class ActivityAttachmentsPicker extends ActivityEditorMixin(SkeletonMixin(Locali
...
+		this._handlerAudioVideoUploaded(e, { storeCreateFileMethod: 'createAudio' });
+	}
 
-		const collection = store.get(this.href);
-		const previewUrl = await collection.getPreviewUrl(fileSystemType, fileId);
-		this._addToCollection(attachmentStore.createAudio(fileName, fileSystemType, fileId, previewUrl));
+	async _handlerAudioVideoUploaded(e, { storeCreateFileMethod }) {
+		const attachment = e.detail.files[0];
+
+		if (attachment.GetFileSystemType) {
```
ktonon, line 88: Keeping the event detail as `files` so as not to break consistent evaluation

dlockhart:

amauryperez (approved):

wongvincent (approved):
lgtm

components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-picker.js
```diff
@@ -79,15 +79,25 @@ class ActivityAttachmentsPicker extends ActivityEditorMixin(SkeletonMixin(Locali
...
+			const fileId = file.GetId();
+			const fileName = file.GetName();
+			const fileSystemType = file.GetFileSystemType();
+
+			const collection = store.get(this.href);
+			const previewUrl = await collection.getPreviewUrl(fileSystemType, fileId);
+			this._addToCollection(attachmentStore[storeCreateFileMethod](fileName, fileSystemType, fileId, previewUrl));
+		} else if (attachment.GetLocation) {
+			const link = attachment;
+			this._addToCollection(attachmentStore.createLink(link.GetName(), link.GetLocation(), link.GetUrn()));
```
mdgbayly, line 99: @ktonon Mostly out of interest. What is the difference here between `GetLocation()` and `GetUrn()`. I'm assuming `GetLocation()` is giving us a URL we can use immediately for previewing the attachment before we save to the backend, whereas `GetUrn()` is the only thing we save on the assignment attachment to the back end? I took a very quick look at the LMS PR you linked from here, and it looks like that is likely where all the URN clever handling is taking place? And presumably that gets rendered back into a usable URL when we render the assignment for viewing? Just wanting to check that will also be working via the assignments hypermedia APIs etc so that when we refetch the assignment for editing, those links will still work.

mdgbayly: @ktonon Also if we want to regression test this feature in FACE once it's merged, is it behind a feature flag that we should be aware of to enable testing? Plus any other setup steps we would need?

ktonon: Your explanation is correct. The `GetLocation()` returns a url which can be used immediately in the attachment preview. The urn is what gets persisted.

[LinkAttachment.PreparedUrl](https://github.com/Brightspace/lms/blob/master/lp/framework/core/D2L.LP/LP/Attachments/Domain/Links/Default/LinkAttachment.cs#L55) resolves the urn (if present) into a canonical url for viewing the resource

ktonon: It is behind this feature flag: https://github.com/Brightspace/lms-launch-darkly-flags/blob/master/features/media/vn-link-attach-assignment-face.feature.json

You also need to have

- tool 699000 enabled (Content Service)
- tool 633000 enabled (Media Platform)
- `d2l.Tools.ContentService.Endpoint` setup. We use `https://api-cert.us-east-1.content-service.brightspace.com` with our qa sites.
- `d2l.Tools.AudioCapture.UseMediaPlatformCapture` set to `on`
- `d2l.Tools.VideoCapture.DialogSrc` set to  `/d2l/wcs/mp/dialogs/VideoCaptureDialog.d2l`

PR #1273, opened 2020-12-15
Show immersive navbar in fullscreen mode

mbardakovD2L:
See https://rally1.rallydev.com/#/357251704080ud/search?detail=%2Fuserstory%2Fd9f49007-9857-4795-b2e7-28761bba75ee&keywords=immersive%20nav


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -117,6 +122,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		this._overdueWeekLimit = Config.OverdueWeekLimit;
 		this._upcomingWeekLimit = Config.UpcomingWeekLimit;
 		this._viewAllSource = 'http://www.d2l.com';  // TODO: Update to actual tool location
+		this._backLinkHref = window.D2L.workToDoOptions ? window.D2L.workToDoOptions.href : '';
```
mbardakovD2L, line 125: Question: is the constructor the correct place to put this, or is this more appropriate in `attributeChangedCallback`?

surags: This would probably be better in the attributeChangedCallback method. I don't think href is the right link here. It contains the whoami URL.

mbardakovD2L: that makes sense. where should we be getting the right link from? `window.D2L.workToDoOptions.href` was the closest thing there I could find

surags: It looks like the homepage for a user can be found with this API [REST API](https://docs.dev.d2l/index.php/Rest_API)
image. I think there's an existing PR for the LMS to send the homepage link or an endpoint to get the link. For now the navbar PR could look for a new attribute and maybe default to /d2l/home in the constructor as temporary solution

KearseTrevor: Agreed that constructor is not the place for this logic to occur. As window's properties are getting set by this specific component in the attributeCallback and therefore shouldn't rely on window being prepopulated.

dlockhart:

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -295,7 +301,19 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				`;
 			});
 
+			const immersiveNav = (isFullscreen) => {
+				return isFullscreen
+					? html`
+						<d2l-navigation-immersive back-link-href="${this._backLinkHref}" back-link-text="Back to D2L">
```
KearseTrevor, line 307: back-link-text should also get localized

mbardakovD2L:
Discussed with Rory - We can have "back to home" default to `<organization>/d2l/home` until we can find a better way to find the current user's homepage from their profile.

components/d2l-work-to-do/lang/en.js
```diff
@@ -1,6 +1,7 @@
 export const val = {
 	activitiesAvailable: "Activities that are due or ending in two weeks are complete! Check View All Work to see what's coming later.", // 'Empty View' - When widget has no activities to display within the next two weeks, but there are more activities further into the future that can be shown on the full screen view
 	Assignment : "Assignment",  // Meta-data descriptor that informs which type of activity is being displayed on a line item
+	backToD2L : "Back to D2L", // Displayed in the immersive navbar to escape out of fullescreen view
```
surags, line 4: nit: spelling of fullscreen

mbardakovD2L: i am a fulle

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -502,5 +520,11 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 			});
 	}
+
+	async _getHomeHref() {
```
surags, line 524: Should probably not be async until we actually make a call

surags (dismissed):

surags (approved):

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -502,5 +520,11 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 			});
 	}
+
+	_getHomeHref() {
+		// TODO: this is a default (and kind of a hacky way to get to it),
+		// ideally we want to get the user's homepage from their profile
+		this._homeLinkHref = window.location.href.substring(0, window.location.href.indexOf('/d2l/') + 5) + 'home';
```
KearseTrevor, line 527: Is this ripped from somewhere / what is the confidence levels that this approach is prod safe

mbardakovD2L: this particular snippet I tested with Surag's local LE, and that link does go back to home. I fully admit that long-term it's not a good solution, but I have full confidence that it works _right now_

PR #1281, opened 2020-12-16
Scy/quick eval widget refactor

Scy-D2L:
Refactored the `d2l-quick-eval-widget` to use `d2l-link` and `d2l-work-to-do-activity-item-basic` instead of unique components.

Added demo data for testing `d2l-quick-eval-widget` as connected json files - demo data removes most of the unused sections of HM responses to simplify future troubleshooting and updates.

Updated/refactored `d2l-work-to-do-activity-item-basic` to allow handling `evaluateAllHref` and a `submissionCount` from the `d2l-quick-eval` widget - `submission-count` is not currently used but will be required for counting badges in an upcoming story.

Appearance of `d2l-quick-eval-widget` on demo page:
image


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -171,9 +182,7 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 	}
 
 	set actionHref(href) {  // This is a hack - Garbage setter function since list-mixin initializes value
-		const oldVal = this._actionHref;
```
Scy-D2L, line 174: @mbardakovD2L This _seems_ to work in the demos, but I'm not entirely sure why this was set up this way so I'm wondering if there might be a problem I've missed?

KearseTrevor: The reason why this was done - if a custom getter is defined on a property, lit automatically drops its generated getter/setter functionality. Which would usually be fine, but since the list-item-mixin (or somewhere near this code) was setting actionHref, we were getting a call to a non-existent setter method. 

Though looks like this may be outdated code at this point - it doesn't seem like there is any incoming calls for setting this property. Should be alright to remove it now without issue (quick check would let you know for sure)

Scy-D2L: ```list-item-link-mixin.js:40 Uncaught TypeError: Cannot set property actionHref of #<ActivityListItemBasic> which has only a getter``` - looks like it stays, but it seems to work fine without the extra oldVal/overwrite thing.

Scy-D2L:
> Visual diff tests failed - pull request #1283 has been opened with the updated goldens.

This confuses me - I haven't touched any files in `quick-eval` itself as far as I can see, my changes are only in the widget. 🤔 @AlexBedley do you think this could be related or is it more likely some kind of flake?

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -242,13 +257,14 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 		}
 
 		const entity = this._usage._entity;
+		const allowList =  this.evaluateAllHref ? QuickEvalActivityAllowList : ActivityAllowList;
```
jeroach, line 260: What is the difference between `QuickEvalActivityAllowList` and `ActivityAllowList`

Scy-D2L: [ActivityAllowList](https://github.com/BrightspaceHypermediaComponents/activities/blob/scy/quick-eval-widget-refactor/components/d2l-work-to-do/env/index.js) is from the work-to-do widget and includes a lot of different `user-activity-usage`s and [QuizEvalActivityAllowList](https://github.com/BrightspaceHypermediaComponents/activities/blob/scy/quick-eval-widget-refactor/components/d2l-quick-eval-widget/env/index.js) is a list of regular `activity-usage` types applicable to quick-eval. I'm not sure the quick-eval-activity list is exhaustive of the activity types used in quick-eval since I translated it from the list in the existing version of the widget, but there doesn't seem to be any overlap between the two lists because of the user-activity/activity differences.

jeroach: Sounds good, thanks :)

jeroach (approved):
LGTM

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -47,70 +56,62 @@ export class QuickEvalWidget extends SkeletonMixin(LitElement) {
...
 	async updated(changedProperties) {
 		super.updated();
 
-		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token) {
-			this.skeleton = true;
-			try {
-				this.activities = await this.getActivities(this.href, this.token);
-			} finally {
-				this.skeleton = false;
+		if ((changedProperties.has('href') || changedProperties.has('token')) && this.activitiesHref && this.token) {
```
AlexBedley, line 66: Sorry I hadn't gotten around to this review yet. This line should be `if ((changedProperties.has('activitiesHref ')...`

PR #1290, opened 2020-12-18
fix: Semver keyword in PR title doesn't release

mostlyfabulous:
Not sure if there's a better way to word the reminder as this didn't release: https://github.com/BrightspaceHypermediaComponents/activities/pull/1289

### DIFF
--- a/.github/workflows/prchecklist.yml
+++ b/.github/workflows/prchecklist.yml
@@ -20,7 +20,7 @@ jobs:
               body: `
             ## PR Checklist:
             
-            **Did you use a semver keyword in the PR commit(s) when creating it?**
+            **Did you use a semver keyword in the PR's commit(s) when creating it?**
             - [ ] Yes
             - [ ] No but I will use a semver keyword in the merge commit message when I merge and squash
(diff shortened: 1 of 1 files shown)


pureooze:
Hmm so I actually used that wording because I found [this change](https://github.com/BrightspaceHypermediaComponents/activities/pull/1254) didnt release and I had to do another change [here](https://github.com/BrightspaceHypermediaComponents/activities/pull/1258) to trigger a release. I figured that if its in the title it adds the prefix to the merge commit but I guess it could get removed.

Not sure if I did something wrong in that original PR too 🤷. I'm fine with the wording change though if I just did something wrong in my PR or the action/github had issues at that time.

wongvincent:
If `fix: ` appears in one of the commits, it should release. If that's not happening, we should mention this to d.lockhart.

It should kinda be...
```
Did you use a semver keyword in a commit?

 Yes (and this PR will not be squashed)
 No, but I will use a semver keyword in the merge commit message
```

wongvincent (approved):

PR #1299, opened 2020-12-22
View All now redirects to full page view

surags:
https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?detail=%2Fuserstory%2F459381433584&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -208,7 +208,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				<div class="d2l-upcoming-list">
 					${collectionTemplate(this._upcomingCollection, this._upcomingDisplayLimit, false)}
 				</div>
-				<d2l-link aria-label="${this.localize('fullViewLink')}" href="${this._viewAllSource}" small>${this.localize('fullViewLink')}</d2l-link>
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -116,7 +118,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		this._upcomingCollection = undefined;
 		this._overdueWeekLimit = Config.OverdueWeekLimit;
 		this._upcomingWeekLimit = Config.UpcomingWeekLimit;
-		this._viewAllSource = 'http://www.d2l.com';  // TODO: Update to actual tool location
+		this._viewAllSource = undefined;
```
mbardakovD2L, line 121: might be outside the scope of this PR, but have we given any thought to what we should do in case we can't get this URL? like if `_viewAllSource` is undefined, should we hide the button?

mbardakovD2L (dismissed):

JasperAtD2L (approved):

components/d2l-work-to-do/d2l-work-to-do.js
KearseTrevor: I think hiding the button is probably a fairly safe play - should be fine as a "Shoot the barrel of fish first, ask for forgiveness from UX once they discover it" strategy

KearseTrevor (approved):
Think that adding an auto hiding function on 'undefined' case as Max mentioned makes sense on this PR but it's fine to go as is.

mbardakovD2L (approved):

PR #1304, opened 2021-01-04
Switch lang files to use double quotes so we don't have CI issues.

dbatiste:
When we get translations back from Serge that contain single embedded apostrophes, it breaks the JSON and CI fails.  This PR converts single quotes to double quotes for the language terms to avoid that.


dlockhart (approved):
Changes look good -- looks like some flake in the visual-diff and Sauce tests though.

dlockhart (approved):

PR #1307, opened 2021-01-04
US123218 Activity editor save-in-place

emil-furniss:
This adds a third button to the footer which only does `Save` without closing the editor and redirecting to the previous page - aka "Save-in-place".

Attachments are a special case since they have their own add-delete-undo lifecycle. On save-in-place, any attachments that are marked for deletion should be removed from the UI. There should be no option to "undo" the deletion of an attachment where the deletion has already been saved via save-in-place.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F462977517920


components/d2l-activity-editor/d2l-activity-attachments/state/attachment-collection.js
```diff
@@ -66,24 +66,35 @@ export class AttachmentCollection {
...
 			}
+
 			if (attachment.creating && !attachment.deleted) {
 				await attachment.save(this._entity);
 			}
 		}
+
+		runInAction(() => {
+			if (saveInPlace) {
+				this.attachments = remainingAttachments;
```
emil-furniss, line 95: All these shenanigans with the `remainingAttachments` is to remove the hrefs of deleted attachments from the array so that the UI updates and no longer shows deleted attachments after a save-in-place.

components/d2l-activity-editor/d2l-activity-attachments/state/attachment.js
```diff
@@ -81,6 +85,7 @@ export class LinkAttachment extends Attachment {
 
 	async save(entity) {
 		await entity.addLinkAttachment(this.attachment.name, this.attachment.urn || this.attachment.url);
+		runInAction(() => this.creating = false);
```
emil-furniss, line 88: After each save-in-place, we should no longer consider an attachment to be _new_ so `creating` is set to false.

components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -132,9 +154,19 @@ class ActivityEditor extends ActivityEditorContainerMixin(ActivityEditorTelemetr
 		}
 	}
 
+	_onSaveComplete(e) {
+		if (e.detail.saveInPlace) {
+			this._saveToastVisible = true;
+			e.stopPropagation();
```
emil-furniss, line 160: If it's a save-in-place, show the "Saved sucessfully" alert toast and stop propagation which will disable the redirect.

components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -128,21 +121,38 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
...
-		await this._saveEditors(orderedEditors);
+		const saveInPlace = e && e.detail && e.detail.saveInPlace;
+		await this._saveEditors(orderedEditors, saveInPlace);
 
 		this.isError = false;
-		this.dispatchEvent(this.saveCompleteEvent);
+
+		this.dispatchEvent(this._saveCompleteEvent(saveInPlace));
+		if (saveInPlace) {
+			this.isSaving = false;
```
emil-furniss, line 131: Setting `isSaving` to false will disable the opaque saving overlay when the save-in-place is complete. For save-and-close we don't want to ever hide it since we're just waiting for the redirect in that case.

mdgbayly (approved):
LGTM

PR #1309, opened 2021-01-06
Notification Email Description Text Size Fix

dustxd:
### Motivation/Context
The notification email description text should have `d2l-body-small` styling applied to it but the slot styling isn't passed down correctly for browsers that use Shady DOM, i.e. Legacy Edge and Mac Safari. I tried applying the styling directly to the `<p>` element but it's not ideal because 

1. We'd have to apply it to both assignment and quiz editors. 
2. The css modules seem to be applied in different orders across browsers so the styling would be off. 

Therefore, the best place to do this is using the `::slotted(p[slot=description])` selector because it solves both issues mentioned above.

### Summary of Changes
Copied `d2l-body-small` styling over to the slot style for `description`

#### Current Behaviour
image

#### Expected Behaviour
image

### Trello Card
https://trello.com/c/nGil5iTq/4-face-mac-safari-notification-email-description-text-is-huge


components/d2l-activity-editor/d2l-activity-notification-email-editor.js
```diff
@@ -41,6 +41,10 @@ class ActivityNotificationEmailEditor
 				}
 
 				::slotted(p[slot=description]) {
+					color: var(--d2l-color-tungsten);
```
dustxd, line 44: This file imports `bodySmallStyles` which contains `d2l-body-small`. Is there a way I can directly spread the css props in that style module into this `::slotted(p[slot=description])` style module instead of copying over the props like what I'm doing now?

wongvincent:
It LGTM, but I will see if others know of a better way!

components/d2l-activity-editor/d2l-activity-notification-email-editor.js
mdulat: I don't think so without using sass.

mdulat:
I don't really like copying over the styles this way. If they decide to change the styling of `d2l-body-small` then we would want the entire lms that uses this class to change consistently without us having to remember that there is this copy somewhere that will need to be changed as well. 

I would prefer to apply to the style twice (in assignments and quizzes) than to have a copy in one place. 

I just tried wrapping the slot in a div with the class, that seemed to work in Chrome at least, but I don't have a legacy Edge working at the moment so can't test it there. Does this work for you?
```
<div class="d2l-body-small">
    <slot name="description"></slot>
</div>
```

dustxd:
> I don't really like copying over the styles this way. If they decide to change the styling of `d2l-body-small` then we would want the entire lms that uses this class to change consistently without us having to remember that there is this copy somewhere that will need to be changed as well.
> 
> I would prefer to apply to the style twice (in assignments and quizzes) than to have a copy in one place.
> 
> I just tried wrapping the slot in a div with the class, that seemed to work in Chrome at least, but I don't have a legacy Edge working at the moment so can't test it there. Does this work for you?
> 
> ```
> <div class="d2l-body-small">
>     <slot name="description"></slot>
> </div>
> ```

I just tested this and it works on Legacy Edge! There was a problem with the order of applied styles but this overrides all the children styling. Thanks Margaret 😄

wongvincent (approved):

mdulat (approved):

PR #1318, opened 2021-01-11
Immersive navbar top padding

mbardakovD2L:
Immersive navbar was overlapping the title of the fullscreen view instead of pushing the content down, so I added padding to compensate. I'm sure there's a cleaner way to do this, but I couldn't think of one and wanted to get this out ASAP.

See rally story: https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?detail=%2Fuserstory%2F460369362992

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -94,7 +94,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 				.d2l-work-to-do-fullscreen-container {
 					background-image: linear-gradient(to bottom, #f9fbff, #ffffff);
-					padding: 3.35rem 2rem 0;
+					padding: 3.35rem 2rem 0 2rem;
 				}
 				d2l-button {
 					padding: 1.8rem 0;
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -94,7 +94,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 				.d2l-work-to-do-fullscreen-container {
 					background-image: linear-gradient(to bottom, #f9fbff, #ffffff);
-					padding: 0 2rem;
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -94,7 +94,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 				.d2l-work-to-do-fullscreen-container {
 					background-image: linear-gradient(to bottom, #f9fbff, #ffffff);
-					padding: 0 2rem;
+					padding: 67px 2rem 0;
```
JasperAtD2L, line 97: How'd you come up with 67?

mbardakovD2L: Inspected the height of the navbar itself in chrome

surags: Would that vary based on the window size?

JasperAtD2L: No, it'll vary based on the DPI of the person's display. It should probably be a rem.

mbardakovD2L: rem makes more sense - I'll figure out what the conversion is and then update that

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -94,7 +94,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				}
 				.d2l-work-to-do-fullscreen-container {
 					background-image: linear-gradient(to bottom, #f9fbff, #ffffff);
-					padding: 67px 2rem 0;
+					padding: 3.35rem 2rem 0;
```
JasperAtD2L, line 97: nit: not sure how it's done elsewhere, but expanding to the full 4 piece padding, IMHO is easier to look at than a 3 piece padding. With 2 it's easy to see top/bottom vs left/right.

mbardakovD2L: that's totally fair - I have to google it every time so I'm all in favour of whatever's easier

JasperAtD2L (approved):

PR #1319, opened 2021-01-11
fix: FACE Assignments activity-outcomes rendering

mostlyfabulous:
https://trello.com/c/F5wKvedY/287-face-assignments-outcomes-editor-taking-up-page-space-when-not-displayed

When there are no alignments and user lacks permissions to update alignments, hide the activity-outcomes component to prevent styles (`padding-bottom: 20px`) being applied to an editor component with no content. This fixes the spacing problem it otherwise creates.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-outcomes.js
+++ b/components/d2l-activity-editor/d2l-activity-outcomes.js
@@ -57,14 +57,14 @@ class ActivityOutcomes extends ActivityEditorFeaturesMixin(ActivityEditorMixin(R
 			return html``;
 		}
 
-		if (!this._hasAlignments) {
-			this.hidden = true;
-		}
-
 		const {
 			canUpdateAlignments,
 			alignmentsHref
 		} = activity;
+
+		if (!this._hasAlignments && !canUpdateAlignments) {
+			this.hidden = true;
+		}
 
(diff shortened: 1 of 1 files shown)


mdulat (approved):
I'm wondering if it would be cleaner to check the conditions [in the parent](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js#L126), and not render it at all if there are no alignments or permissions. 

But I'm not sure, curious what others think.

mostlyfabulous:
I chatted with @emil-furniss and he mentioned that flickering (while loading) from `activity-outcomes` is the reason why it is rendered as an empty component with styles applied to avoid this behaviour. The flickering causes the components below `activity-outcomes` to jump up and down as the `activity-outcomes` completes loading.

mdulat:
> I chatted with @emil-furniss and he mentioned that flickering (while loading) from `activity-outcomes` is the reason why it is rendered as an empty component with styles applied to avoid this behaviour. The flickering causes the components below `activity-outcomes` to jump up and down as the `activity-outcomes` completes loading.

Interesting! So many nuances.

PR #1320, opened 2021-01-11
Filter upcoming and overdue collections based on WTD config

surags:
https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?detail=%2Fuserstory%2F464048560492&fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -186,23 +186,9 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				return nothing;
 			}
 
-			let activities = collection.getSubEntitiesByRel(Rels.Activities.userActivityUsage);
+			const activities = this._getFilteredActivities(collection, isOverdue);
 			if (activities.length === 0) {
 				return nothing;
-			}
-
-			if (isOverdue) {
-				// Filter overdue activities based on config
-				const cutOffDate = new Date();
-				cutOffDate.setDate(cutOffDate.getDate() - (getOverdueWeekLimit() * 7));
-
-				activities = activities.filter((activity) => {
-					const activityDate = activity.hasSubEntityByClass('due-date')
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -180,11 +180,25 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				return nothing;
 			}
 
-			const activities = collection.getSubEntitiesByRel(Rels.Activities.userActivityUsage);
+			let activities = collection.getSubEntitiesByRel(Rels.Activities.userActivityUsage);
 			if (activities.length === 0) {
 				return nothing;
 			}
 
+			if (isOverdue) {
```
JasperAtD2L, line 188: Pull this out into a helper function `getFilteredActivities(collection, isOverdue)`, and combine this logic and line 183 together. This will also fix if no overdue items are within the last x weeks, it won't display `nothing`.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -180,25 +180,11 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				return nothing;
 			}
 
-			let activities = collection.getSubEntitiesByRel(Rels.Activities.userActivityUsage);
+			const activities = this._getFilteredActivities(collection, isOverdue);
```
JasperAtD2L, line 183: So I commented on Max's PR here: https://github.com/BrightspaceHypermediaComponents/activities/pull/1323, that we shouldn't be filtering in the templates/render. The collection should be passed in filtered. Because don't we need to apply this filter to the fullscreen view too?

surags: This filter should only apply to the widget view, not the fullscreen view. That being said, I like the idea of not having it in the template

JasperAtD2L (dismissed):

JasperAtD2L (approved):

PR #1323, opened 2021-01-11
W2D: Load More button behaviour

mbardakovD2L:
Rally: https://rally1.rallydev.com/#/357251704080ud/dashboard?detail=%2Fuserstory%2F424735308936

To facilitate this change, I had to update the fullscreenCollectionTemplate to take in a list of activities, rather than a sirenEntity, because now we get multiple entities (for multiple pages), and stitch them together.

In addition, once Load More has been clicked, we want to now ignore the display limit. (We may also want to ignore the display limit altogether in fullscreen mode - if we decide that we do, that's an easy change to make).


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -117,7 +119,12 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		this._overdueWeekLimit = Config.OverdueWeekLimit;
 		this._upcomingWeekLimit = Config.UpcomingWeekLimit;
 		this._viewAllSource = 'http://www.d2l.com';  // TODO: Update to actual tool location
+		this._upcomingActivities = [];
+		this._overdueActivities = [];
+		this._ignoreViewLimit = false;
 		this._setEntityType(UserEntity);
+
+		this.nextPageRel = 'https://activities.api.brightspace.com/rels/next-page'; // this should be somewhere in Hypermedia Constants
```
mbardakovD2L, line 127: Not sure how this is supposed to work - is there supposed to be a new `Rel` in `siren-sdk` where we get our other Rels? Am I allowed to go there and add it myself?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -314,17 +320,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			: nothing;
 
 		const fullscreenTemplate = () => {
-			if (!this._overdueCollection || !this._upcomingCollection || !this._maxCollection) {
+			if (!this._overdueCollection && !this._upcomingCollection && !this._maxCollection) {
```
mbardakovD2L, line 323: This seemed like a bug to me - my thinking is - if you have _any_ items overdue or upcoming, then we should be able to display them in fullscreen mode - I don't see why having all three (overdue, upcoming, and upcoming but far in the future) would be a necessity.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -314,17 +320,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
 				return nothing;
 			}
 			return html`
 				<div class="d2l-work-to-do-fullscreen-container">
 					<div class="d2l-heading-1 d2l-work-to-do-fullscreen-title">${this.localize('myWorkToDo')}</div>
 					<div class="d2l-overdue-collection-fullscreen">
-						${fullscreenCollectionTemplate(this._overdueCollection, this._overdueDisplayLimit, true)}
+						${fullscreenCollectionTemplate(this._overdueActivities, this._overdueDisplayLimit, true, true)}
```
mbardakovD2L, line 330: here I set `ignoreLimit` to `true` because if there is no pagination for overdue items, we might as well show them all - if we have reason to do otherwise I'm willing to change that.

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: I think you can, there is a couple things that need to get added for activities. After the is started change is in, I can add the changes.

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: we pre load the next page to minimize deduping weirdness. Eg. If the user clicks "load more" but the next page is fully deduped, nothing new would be displayed. Pre loading the page allows us to detect if the next page actually has information.

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: do something like: `Rels.Activities.nextPage = 'https://activities.api.brightspace.com/rels/next-page';` that way when the siren-sdk is updated we can remove this line and it'll just work.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -467,10 +473,14 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
-		// eslint-disable-next-line no-console
-		console.log('I am an empty block of code, please complete me.');
+	async _handleLoadMoreClicked() {
+		if (!this._upcomingCollection.hasLinkByRel(this.nextPageRel)) return;
+
+		const upcomingSource = this._upcomingCollection.getLinkByRel(this.nextPageRel).href;
+		const upcomingNextPage = await fetchEntity(upcomingSource, this.token, true);
+		this._upcomingActivities = this._upcomingActivities.concat(upcomingNextPage.getSubEntitiesByRel(Rels.Activities.userActivityUsage));
+		this._ignoreViewLimit = true;
+		this._upcomingCollection = upcomingNextPage; // moves "next page" forward every time this succeeds
```
JasperAtD2L, line 483: need to update: https://github.com/BrightspaceHypermediaComponents/activities/pull/1323/files#diff-d300a684213152d82078d4d614f2f7ede435053c312f283cc93533dbae942e94R409 to return false when the next-page link is missing.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -276,19 +283,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		const errorTemplate = nothing;
 
 		/** Fullscreen state templates */
-		const fullscreenCollectionTemplate = (collection, displayLimit, isOverdue) => {
-			if (!collection || displayLimit === 0) {
+		const fullscreenCollectionTemplate = (activities, displayLimit, ignoreLimit, isOverdue) => {
```
JasperAtD2L, line 286: not sure I care for all the activity pairing down done in the rendering functions. Adding in the dedup logic will make this even more convoluted. We should filter/cap the activities before rendering.

JasperAtD2L: There shouldn't be a need for either displayLimit or ignoreLimit

mbardakovD2L: right - that makes sense now that we have a pageSize (i.e. the "limit", before you request more, is just however many you asked to be on the first page)

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: that makes sense - I'll revert this

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: did you mean to link to a specific line? That link just goes to the top of the `files` page for this PR

JasperAtD2L: yes: https://github.com/BrightspaceHypermediaComponents/activities/blob/7a86836124884f6b9971751e56c79faa83cc58cf/components/d2l-work-to-do/d2l-work-to-do.js#L409

mbardakovD2L: right - totally glossed over that

mbardakovD2L:
I kept the display limit in place for the widget view, but removed it for fullscreen view - so now when you go into fullscreen view it'll show all the overdue items (since those aren't paginated) and one page's worth of upcoming items.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -467,10 +470,13 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
 	 */
-	_handleLoadMoreClicked() {
-		// Nothing here right now - blocked on US120246
-		// eslint-disable-next-line no-console
-		console.log('I am an empty block of code, please complete me.');
+	async _handleLoadMoreClicked() {
+		if (!this._upcomingCollection.hasLinkByRel(Rels.Activities.nextPage)) return;
+
+		const upcomingSource = this._upcomingCollection.getLinkByRel(Rels.Activities.nextPage).href;
+		const upcomingNextPage = await fetchEntity(upcomingSource, this.token, true);
```
JasperAtD2L, line 477: this could be undefined if the call fails

JasperAtD2L (approved):
This PR is good WRT the load more functionality. However I don't think the future loading (calling _loadUpcoming() twice) is functioning properly. If there isn't a story to look into it, we should add one.

PR #1325, opened 2021-01-11
localize quick eval widget

vix633:


activities.serge.json
```diff
@@ -145,19 +145,22 @@
       "zh-tw zh-tw"
     ]
   },{
-    "name": "assignmentActivityEditor",
```
jeroach, line 148: Is this just a rename?

jeroach (approved):

activities.serge.json
vix633: Something changed in master since this i think, let me double check

vix633: Yeah this is weird, i dont know whats with this change.

jeroach: Should the new code be a copy into a new block? Instead of editing an existing block?

kieranderson: Seems like this still got removed?
It's gone in master.
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/activities.serge.json

vix633:
FYI: the diff for this that's showing now, is not what the diff showed at merge time. There's some weird git diff caching weirdness that went down here.

PR #1330, opened 2021-01-12
W2D: fix/remove extra padding for fullscreen view

mbardakovD2L:
See rally: https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?Name=Full%20page%20padding&Priority=Normal&Project=%2Fproject%2F357251704080&Requirement=%2Fhierarchicalrequirement%2F463435916636&detail=%2Fdefect%2F482551412748&iteration=u&parentRef=%2Fhierarchicalrequirement%2F463435916636&rankTo=BOTTOM

I moved the immersive navbar to be at the top level of fullscreen view instead of having a duplicate for each collection component, which was causing it to set aside extra space for the navbar for each of these components. This also undoes #1318 , because now that we're putting the immersive navbar in the correct spot, it allocates its own space at the top of the page and we no longer need to add extra padding there ourselves.

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -324,19 +324,19 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 				</d2l-button>`
 			: nothing;
 
+		const immersiveNav = () => {
+			return html`
+				<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
+					<div class="d2l-typography d2l-body-standard" slot="middle">
+						<p>${this.localize('myWorkToDo')}</p>
+					</div>
+				</d2l-navigation-immersive>`;
+		};
+
 		const fullscreenTemplate = () => {
 			if (!this._overdueCollection || !this._upcomingCollection || !this._maxCollection) {
 				return nothing;
 			}
-
-			const immersiveNav = () => {
-				return html`
-						<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
-							<div class="d2l-typography d2l-body-standard" slot="middle">
-								<p>${this.localize('myWorkToDo')}</p>
-							</div>
-						</d2l-navigation-immersive>`
-			};
 
 			return html`
 				${immersiveNav()}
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -340,7 +328,20 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
+						<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
+							<div class="d2l-typography d2l-body-standard" slot="middle">
+								<p>${this.localize('myWorkToDo')}</p>
+							</div>
+						</d2l-navigation-immersive>`
+					: nothing;
+			};
+
 			return html`
+				${immersiveNav(this.fullscreen)}
```
JasperAtD2L, line 344: won't this always be true?

JasperAtD2L: Also shouldn't we show the immersive nav when there are no activities to show.

mbardakovD2L: oops - that's totally right. I think I needed the `isFullscreen` back when I had the navbar in the wrong place

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -340,7 +328,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
+				return html`
+						<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
+							<div class="d2l-typography d2l-body-standard" slot="middle">
+								<p>${this.localize('myWorkToDo')}</p>
+							</div>
+						</d2l-navigation-immersive>`
+			};
+
 			return html`
+				${immersiveNav()}
```
JasperAtD2L, line 342: We may already take care of this. But during loading is the nav missing? I don't see the nav added in the detailed skeleton view.

mbardakovD2L: That's a good point and one I initially wanted to bring up - the "is loading" state takes precedence over the "is fullscreen" state, but adding the immersive nav to the skeleton view should fix that problem. I can add that

JasperAtD2L (approved):

PR #1337, opened 2021-01-13
Scy/quick eval widget/empty states

Scy-D2L:
Add empty/error states to Quick Eval Widget
image
image

- Includes localization
- Added new states to demo page
- Added quick-eval-widget demo page to demo index


components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -1,133 +1,214 @@
...
+
+	get noSubmissionTemplate() {
+		return html`
+			<div class="d2l-quick-eval-widget-no-submissions">
+				<d2l-quick-eval-widget-no-submissions-image></d2l-quick-eval-widget-no-submissions-image>
+				<div class="d2l-quick-eval-widget-no-submissions-text-container">
+					<h4 class="d2l-heading-4">${this.localize('CaughtUp')}</h4>
+					<p class="d2l-body-compact">${this.localize('NoSubmissions')}</p>
+				</div>
+				<d2l-button primary @click="${this.handleViewAll}" href="${this.quickEvalHref}">
```
Scy-D2L, line 163: I hadn't thought about it until just now - @dbatiste is there an accessibility issue with using a button as a navigable link?

jmurray-d2l (approved):

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
dbatiste: Normally I would say yes, because it changes the semantics, which affects how they would be announced and exposed by ATs.  With ATs, sometimes the user can get a list of links (and other elements), and there are keyboard short-cuts for navigating between them.  For actual links, a user would also be able to right-click on the link and "Open Link in New Tab".  However in this case, it looks like this button performs some action, then it navigates so it seems reasonable in this case.

Aside, any properties that are not defined by the component (`href` in this case) should be data attributes (`data-href`).  The idea being that it avoids any possibility of future collision, on our components or native elements.  The browser allows it (something I think implementors might regret), but it's not recommended.  In this case it's extra iffy because `href` (in the native context) is only applicable to a couple of elements (such as `a` and `link`) and it also comes into play for HM components, which `d2l-button` is not.

Another thing I noticed - is `handleViewAll` intended to be private?  If so, all such methods should be prefixed with a leading `_` to indicate this.  It really helps for future developers understanding the code and the scope of risk when making changes.  Some documentation tooling will also regard such methods accordingly.

PR #1342, opened 2021-01-13
fix: put page size into a constant, separate from MaxWidgetDisplay

mbardakovD2L:
See https://rally1.rallydev.com/#/357251704080ud/dashboard?detail=%2Fuserstory%2F424735308936 for original rally ticket, and https://d2l.slack.com/archives/G0150CJ3SB1/p1610572621089200 for the thread on us deciding to do 20 for the page number.

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -521,7 +521,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	 * @param {Number} [forwardLimit] - [Default: Config.UpcomingWeekLimit * 7] Number of days into future to look for activities
 	 */
 	async _loadUpcoming(entity, forwardLimit, pageSize) {
-		if (!pageSize) pageSize = Constants.MaxWidgetDisplay;
+		if (!pageSize) pageSize = Constants.PageSize;
 
 		if (!entity || !entity.hasActionByName(Actions.activities.selectCustomDateRange)) {
 			return;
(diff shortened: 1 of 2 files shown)


JasperAtD2L (approved):

surags (approved):

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -521,7 +521,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	 * @param {Number} [forwardLimit] - [Default: Config.UpcomingWeekLimit * 7] Number of days into future to look for activities
 	 */
 	async _loadUpcoming(entity, forwardLimit, pageSize) {
-		if (!pageSize) pageSize = Constants.MaxWidgetDisplay;
+		if (!pageSize) pageSize = Constants.PageSize;
```
m6jones, line 524: Why do you have constants in a separate file? These are suppose to be components?

mbardakovD2L: I'm following what was already there before I took over the project, but my thinking was because there are multiple components all related to WorkToDo they might share some constants (not necessarily this constant, but in general) - @KearseTrevor might be able to comment more

KearseTrevor: The original intent of the constants file was an abstraction of any general customizable settings per tenant/user (before we actually had a workflow to provide this information to the components from the LE).
- It's true that multiple components will want the same information, however if we want to remove the smell associated with library components using a const file - we can have the parent component pass down the retrieved variables to the child (with the parent itself having some set for the default state in place of this file)

PR #1349, opened 2021-01-14
US122757: Use work to do filter action

JasperAtD2L:
https://rally1.rallydev.com/#/357251704080d/custom/367300408400?detail=%2Fuserstory%2F459454738300&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -523,7 +523,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	async _loadUpcoming(entity, forwardLimit, pageSize) {
 		if (!pageSize) pageSize = Constants.PageSize;
 
-		if (!entity || !entity.hasActionByName(Actions.activities.selectCustomDateRange)) {
+		if (!entity || (!entity.hasActionByName(Actions.activities.filterWorkToDo) && !entity.hasActionByName(Actions.activities.selectCustomDateRange))) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -534,12 +534,21 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
-		const action = entity.getActionByName(Actions.activities.selectCustomDateRange);
-		const fields = [
-			{ name: 'start', value: start },
-			{ name: 'end', value: end },
-			{ name: 'pageSize', value: pageSize }
-		];
+		const action = entity.hasActionByName(Actions.activities.filterWorkToDo)
+			? entity.getActionByName(Actions.activities.filterWorkToDo)
+			: entity.getActionByName(Actions.activities.selectCustomDateRange);
+		const fields = [].concat(action.fields).reduce((acc, field) => {
```
JasperAtD2L, line 540: I was hoping the siren-sdk would add default values if they weren't explicitly added. But alas it does not 👎

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -523,7 +523,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	async _loadUpcoming(entity, forwardLimit, pageSize) {
 		if (!pageSize) pageSize = Constants.MaxWidgetDisplay;
 
-		if (!entity || !entity.hasActionByName(Actions.activities.selectCustomDateRange)) {
+		if (!entity || (!entity.hasActionByName(Actions.activities.filterWorkToDo) && !entity.hasActionByName(Actions.activities.selectCustomDateRange))) {
```
m6jones, line 526: You shouldn't need to do this? Why are you accessing the siren entity?

components/d2l-work-to-do/d2l-work-to-do.js
m6jones: It actually does? Not sure why you are experiencing issues. But perform siren action does add the defaults.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -534,12 +534,21 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		const end = new Date(now.getFullYear(), now.getMonth(), now.getDate() + forwardLimit, 23, 59, 59, 999).toISOString();
 		const start = new Date(now.getTime()).toISOString();
 
-		const action = entity.getActionByName(Actions.activities.selectCustomDateRange);
-		const fields = [
-			{ name: 'start', value: start },
-			{ name: 'end', value: end },
-			{ name: 'pageSize', value: pageSize }
-		];
+		const action = entity.hasActionByName(Actions.activities.filterWorkToDo)
```
m6jones, line 537: You shouldn't be accessing the entity.

m6jones (changes requested):
You should never access the siren entity. You are skipping the entire workflow of the siren sdk. This is pretty bad. Fix that :)

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: Only if you don't specify a fields object: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/SirenAction.js#L110

JasperAtD2L:
> The siren sdk is designed to create a way to work with the entities without directly accessing them in the components. If you need to add new ways of working with the entities, I recommend a separate PR to siren-sdk to add that.

While the team agrees this should be done. The whole widget's data access is done via the underlying entities/collection. Changing this now would be a fairly large undertaking which we don't have the resources/time to take on. We will create a ticket to analyze and migrate to domain entities, for a dragon slayer, M2, or if we have time at the end.

jcrnac:
Given the time constraints for meeting the GA release date and the amount of work involved to refactor the code so that it is not accessing the entity directly, Luna should proceed with their current PR. 
Post GA, Maya, Rory, Luna dev, and myself will get together to discuss the work involved using using hmc-foundations rather than siren-sdk.

surags (dismissed):

surags (approved):

mbardakovD2L (approved):

PR #1362, opened 2021-01-15
Us123716 new html editor update on blur

mdulat:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F479012707364&fdp=true

Conditionally shows new html editor, if new attribute exists.


components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -0,0 +1,43 @@
...
+
+	static get properties() {
+		return {
+			value: { type: String },
+			ariaLabel: { type: String },
+			disabled: { type: Boolean }
+		};
+	}
+
+
```
ian-wallace, line 16: Super-nit:  I suspect this extra whitespace won't pass LINT (on my machine).  I'd recommend running this through a different linter to clean this up - it happened to me when you reviewed my code.  I guess there must be a couple different linting standards and it only picks one?

ian-wallace (dismissed):

ian-wallace (dismissed):

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
```diff
@@ -86,6 +90,12 @@ class QuizEditor extends AsyncContainerMixin(RtlMixin(LocalizeActivityQuizEditor
 			e.stopPropagation();
 			return;
 		}
+
+		if (e.detail.key === 'd2l-provider-html-new-editor-enabled') {
```
mdgbayly, line 94: minor nit: this could possibly have been moved to the `d2l-activity-editor` and have the top components pass the enabled attributes down.

components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -0,0 +1,40 @@
...
+	render() {
+		return html`
+			<d2l-htmleditor
+				html="${this.value}"
+				title="${this.ariaLabel}"
+				?disabled="${this.disabled}"
+				type="inline"
+				height="7rem"
+				@d2l-htmleditor-blur="${this._onContentChange}">
```
mdgbayly, line 23: Is this the only change handling the new HTML editor provides? I always feel wary about using blur handling to handle changes. Has this tripped us up before? Maybe that's just FUD and it is 100% reliable. 

It possibly has the advantage, that we don't have to worry about the change event, updating, the HTML editors value and then resetting the actual value of the HTML editor wiping out any internal tinymce selection handling which was an issue we had with our change based implementation.

mdulat: Currently yes, `d2l-htmleditor-blur` is the only event. Another option would be to never update the Mobx state and just grab the content when the user clicks one of the Save buttons. I have a stashed solution that does this. This solution that updates on blur seemed cleaner in a way since it's similar to how everything else works. It would have to be tested on other devices, which is easier once it's deployed to a quad site.

mdulat:
I don't know why visual tests keep failing, the generated PRs keep showing me a diff that doesn't look any different.

mdgbayly (approved):
OK to get this merged once you sort out the CI. The sooner we can start testing the new editor (and ditch the old one :-) ) the better.

components/d2l-activity-editor/d2l-activity-html-new-editor.js
dbatiste: Margaret and I talked about this a bit - we may eventually add a `d2l-htmleditor-change` event since I suspect it will be needed by the Quizzing FRA in order for it to update its side-by-side preview.  My recommendation was to avoid unnecessary work for every keystroke, but that the change event could be exposed if needed.

dbatiste:
> I don't know why visual tests keep failing, the generated PRs keep showing me a diff that doesn't look any different.

We've seen this sort of thing before - the diff is showing some subtle differences in the text within the meter component.  It seems flakey/inconsistent.  I would try rerunning.  You can update the goldens but which version is the most common is hard to know.  It would be nice to get to the bottom of this - I sorta wonder if it has something to do with rendering text in svgs (which is what the meter is doing).

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
mdulat: I will refactor this in a separate PR.

ian-wallace (approved):

PR #1403, opened 2021-01-19
DE42132: Clicking Back/Cancel on unmodified new weblinks launches warning popup

brissons:
This PR [address this defect](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F486166661432&fdp=true).

When creating a new weblink then immediately hitting 'Back' or 'Cancel', a popup is launched saying "Are you sure you want to discard your changes?".

This is because when new weblinks are created, we initially set them to a default 'garbage' url. We stopped this url from being displayed to the user by clearing the link when the page is loaded (which shows an empty textbox). However, this causes a difference between the entity and what is on the page. When the `dirty()` method is called when you press 'Cancel' or 'Back', it returns true because of the difference between the default url and the empty link, which triggers the popup.

This change modifies the entity to also have an empty url for new weblinks. So when a new weblink is made, there are no changes found and the popup won't trigger.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
@@ -37,12 +37,17 @@ export class ContentWebLink {
 		this.title = webLinkEntity.title();
 
 		const entityUrlValue = webLinkEntity.url();
-		// in order to create a new weblink entity, we need to assign it a
-		// default 'garbage' url, however we want to display an empty url on first load.
-		// This does not change the actual entity.
-		this.link = entityUrlValue === defaultPlaceholderLink
-			? ''
-			: entityUrlValue;
+		if(entityUrlValue === defaultPlaceholderLink) {
+			// in order to create a new weblink entity, we need to assign it a
+			// default 'garbage' url, however we want to display an empty url on first load.
+			this.link = '';
+			// we also need to modify the entity to have no url which is needed for
+			// the dirty() method. this does not save the entity.
+			this._contentWebLink._entity.properties.url = '';
+		}
+		else {
+			this.link = entityUrlValue;
+		}
 
 		this.isExternalResource = webLinkEntity.isExternalResource();
 	}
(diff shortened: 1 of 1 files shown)


ianlaidlaw (dismissed):
I think this all seems safe! ideally we wouldn't have that garbage url, but this looks like this would work

components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
```diff
@@ -37,12 +37,17 @@ export class ContentWebLink {
...
-		this.link = entityUrlValue === defaultPlaceholderLink
-			? ''
-			: entityUrlValue;
+		if(entityUrlValue === defaultPlaceholderLink) {
+			// in order to create a new weblink entity, we need to assign it a
+			// default 'garbage' url, however we want to display an empty url on first load.
+			this.link = '';
+			// we also need to modify the entity to have no url which is needed for
+			// the dirty() method. this does not save the entity.
+			this._contentWebLink._entity.properties.url = '';
```
CodeBaboon, line 46: I'm not super keen about reaching into the private properties like this. I wonder if there is a way for the `equal` method on a siren-sdk weblink entity to treat blank and the default placeholder as equal, would that maybe be less leaky? I guess it depends if the siren-sdk entity knows about this default url concept or not (probably not I guess?).

ianlaidlaw: One other idea I had, what if we didn't serialize the garbage url? then we don't need to do any comparison on the front end, we will simply have an empty string for the url. the user won't be able to overwrite this, as empty string isn't a valid url.

ianlaidlaw: this will keep the front end dumber. so just in the serializer, if it's the default, insteand send `''`

CodeBaboon: I like that. Then when we fix the back-end with something like Working Copy we potentially don't even have to touch the front end (well, I'm sure we will, but still I like it).

prajapac (approved):
LGTM, except if you can address Dave's comment 🤔 (Edit: I ++ Ian's Idea)

brissons:
I'm going to close this PR [for Ian's idea](https://github.com/Brightspace/lms/pull/5624). I'm going to open another PR shortly to get rid of references to the `defaultPlaceholderLink`.

PR #1406, opened 2021-01-20
DE42174 Fix score-out-of throwing error on save after save-in-place

emil-furniss:
On save-in-place, the activity-usage entity is reloaded which re-instantiates the ActivityScoreGrade (from [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/4af04474a3b577b096853529283f81446763b24f/components/d2l-activity-editor/state/activity-usage.js#L39)). This clears the state, so that the `ActivityScoreGrade.newGradeName` is `undefined` if the score-out-of is [saved](https://github.com/BrightspaceHypermediaComponents/activities/blob/4af04474a3b577b096853529283f81446763b24f/components/d2l-activity-editor/state/activity-usage.js#L187) after that.

Calling save with an undefined `newGradeName` throws a 400 error.

Since no props on the components are actually changing, this resets the `newGradeName` when the d2l-activity-score-editor component is updated without any changed properties.

Would love to hear if there is a more elegant way to handle this?

Ultimately this is a side-effect of the activity-usage needing to know about an assignment-activity property - the name. So that when the ActivityScoreGrade state is instantiated with the activity-usage entity, it can't know the assignment name. We're relying on the UI to [pass that value around](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js#L148) and set it when the `activityName` [prop value changes](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-score-editor.js#L258).

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -270,6 +270,10 @@ class ActivityScoreEditor extends SkeletonMixin(ActivityEditorMixin(LocalizeActi
 				this._setNewGradeName(this.activityName);
 			}
 		});
+
+		if (changedProperties.size === 0) {
+			this._setNewGradeName(this.activityName);
+		}
 	}
 
 	_addOrRemoveMenuItem() {


mdulat:
As another option, would it make sense to set the `newGradeName` within the constructor of `ActivityScoreGrade` by following the `specialization` link of the activity usage, and getting the `name` property of that result? Then we wouldn't have to rely on the UI to pass that value around.

emil-furniss:
> As another option, would it make sense to set the `newGradeName` within the constructor of `ActivityScoreGrade` by following the `specialization` link of the activity usage, and getting the `name` property of that result? Then we wouldn't have to rely on the UI to pass that value around.

Interesting idea, that would probably work. It's a little bit of extra network overhead, but I think the entity store should de-dupe the request and entity so it would end up only being a single request to the assignment.

mdulat:
Maybe we can apply this simple fix as the cert fix, for the sake of keeping cert fixes as small as possible. Then we can decide if we want to do a bigger refactor to handle activity-usage things needing access to non-activity-usage things? @mdgbayly ?

mdgbayly:
Yeh, definitely a weird workflow...state duality nastiness with having the same data in two places, I guess even if we did initialize it in the ActivityScoreGrade state on initialization, we'd still need the UI hooks to keep that in sync if the name changes in the assignment name field.

Not sure what the best way to approach it is. Ideally, we'd only have the name in one part of the state and everything would access it from there, but that is also tricky with our setup where our activity state doesn't really have any connection to our assignment state. 

I'd be OK with doing as Margaret suggested and just going with your fix for cert if it works and think more on this. Maybe stick a big FIXME or something in there explaining the problem it is solving and that we'd like to find a better, more elegant solution.

mdgbayly:
just some more thoughts for post cert. Kind of wishing `name` was an activity property. I know this was discussed at one point, and rejected because not all activities have names. But not all activities have dates either...so why couldn't we have an activity with an empty name.

But even if we don't go that far possibly we could manage the name in the activity mobx state. I feel less bad about assignment state knowing about activity state than vice versa. So the assignment could potentially just delegate all name changes to the the activity state. Possibly raises concerns about order of operations. e..g when the assignment loads can we guarantee that the activity usage will have already loaded so we can store the name from the assignment in the activity state. Probably we can at the moment but feels a bit icky. 

Anyway, something to consider.

emil-furniss:
@mdulat @mdgbayly Thanks for the input guys. I'll move forward with this as a small fix for cert.

mdulat (approved):

PR #1410, opened 2021-01-20
fix: US123716 new html editor was not rendering correctly in FACE pages.

mdulat:
After the addition of a `disabled` property, the new html editor was not rendered correctly on FACE pages. This happened because the value of the passed-in disabled property was based on a property that is initially `null`, which caused disabled to be initially true, which meant tinymce didn't have the proper component rendered on `init`. Ensuring disabled is initially set to false allows tinymce to initialize correctly.

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
@@ -45,6 +45,9 @@ class AssignmentEditorDetail extends AsyncContainerMixin(SkeletonMixin(SaveStatu
 				:host([hidden]) {
 					display: none;
 				}
+				:host([skeleton]) .d2l-skeletize::before {
+					z-index: 2;
+				}
 				#score-and-duedate-container {
 					display: flex;
 					flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -45,6 +45,9 @@ class AssignmentEditorDetail extends AsyncContainerMixin(SkeletonMixin(SaveStatu
 				:host([hidden]) {
 					display: none;
 				}
+				:host([skeleton]) .d2l-skeletize::before {
+					z-index: 2;
+				}
```
mdulat, line 50: Note that we are setting this here because skeletonization is done on a top-level div, allowing it to be applied to any of the 3 components that may go into that div (old html editor, new html editor, or plain text area). Even though the new html editor now supports skeleton, we would have to pass the value to it and all the other components, so instead we are keeping it on the top-level div here.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-detail.js
```diff
@@ -218,7 +221,7 @@ class AssignmentEditorDetail extends AsyncContainerMixin(SkeletonMixin(SaveStatu
 		if (!assignment) {
 			return;
 		}
-
+		this._saveOnChange('instructions');
```
mdulat, line 224: This was needed to ensure the `debounceJobs` are flushed before the component is saved, to ensure changes to instructions are processed if the Save button is pressed immediately after making changes.

dbatiste (approved):

PR #1414, opened 2021-01-21
W2D: Activity Descriptions

mbardakovD2L:
https://rally1.rallydev.com/#/357251704080ud/workviews?detail=%2Fuserstory%2F483217975936&view=b064b584-6337-43c1-b9a9-28047568284d&fdp=true?fdp=true


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -270,9 +270,14 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
-			? this._activity.properties.description
-			: '';
+		if (this._activity && !this.skeleton) {
+			if (this._activity.hasSubEntityByClass('description')) {
+				return this._activity.getSubEntityByClass('description').properties.text;
+			} else if (this._activity.properties.instructionsText) {
+				return this._activity.properties.instructionsText;
+			}
+		}
+		return '';
```
mbardakovD2L, line 280: Depending on the type of activity, the description could be in a different place - e.g. quizzes have a `description` entity, whereas assignments have a `instructionsText` as a property.

I broke this ternary operator down into an if statement because I suspect this code will get more complex as we accommodate more kinds of activities

components/d2l-work-to-do/env/index.js
```diff
@@ -25,7 +25,7 @@ export const ActivityAllowList = {
 	userChecklistActivity: {
 		class: Classes.activities.userChecklistActivity,
 		icon: 'tier2:checklist',
-		rel: Rels.Checklists.checklist,
+		rel: Rels.Checklists.checklistItem,
```
mbardakovD2L, line 28: What we want to actually show up in W2D is individual items on a checklist, not the checklist itself, because multiple checklist items could have different due dates and we want to show them all

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
m6jones: This should return undefined if a description doesn't exist.

If you are using this in render (ideally not something you should do). Then it should return `nothing` directive.

mbardakovD2L: can I get more context around what I shouldn't be doing in `render` and why? Currently that is the only place this getter is used, but I could see it coming up elsewhere so I think `undefined` makes more sense than `nothing`

m6jones: The main reason is you are using an object to set the values. This only happens when the reference to the object changes. So you could change the object and render wont run. Or change only the object but the content stays the same and run render when you don't need too.

m6jones: Nothing directive is good. Incase the method gets used within slot. It basically says use the default value for the slot.

surags (approved):

PR #1420, opened 2021-01-21
Moved management of _alignmentsHref to outcomes component

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/defect/486581000692

Implemented Martin's solution for standards/alignments loading. Looking at the network tab I cannot see this adding any additional requests. 

The result is that the standards button renders in sync with the end of the skeleton loading but the tags are still slightly delayed:

standardsLoading

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-outcomes.js
+++ b/components/d2l-activity-editor/d2l-activity-outcomes.js
@@ -3,6 +3,7 @@ import '@brightspace-ui/core/components/dialog/dialog.js';
 import 'd2l-activity-alignments/d2l-select-outcomes-hierarchical.js';
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
+import { fetchEntity } from './state/fetch-entity.js';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
@@ -18,7 +19,8 @@ class ActivityOutcomes extends ActivityEditorMixin(RtlMixin(MobxLitElement)) {
 			alignButtonText: { type: String, attribute: 'align-button-text' },
 			_opened: { type: Boolean },
 			_outcomesTerm: { type: String },
-			_browseOutcomesText: { type: String }
+			_browseOutcomesText: { type: String },
+			_alignmentsHref: { type: String }
 		};
 	}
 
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-outcomes.js
```diff
@@ -57,12 +58,16 @@ class ActivityOutcomes extends ActivityEditorMixin(RtlMixin(MobxLitElement)) {
 			alignmentsHref
 		} = activity;
 
-		if (!canUpdateAlignments && !this._hasAlignments || this._hasAlignments === undefined) {
+		if (!canUpdateAlignments && !this._hasAlignments) {
```
ronan-f, line 61: Had to remove this check as leaving it in defeats the purpose of this change. `this._hasAlignments` will be `undefined ` and prevent rendering until `_alignmentTagsEmptyChanged` fires.

ronan-f: Also had to leave this check here as without it, if a course has no outcomes, the outcomes editor is still rendered (albeit without displaying any elements) with 20px of padding causing an offset in the layout. I saw @mdulat had suggested moving this check to the parent component previously which seemed like a good suggestion to me to avoid this component having to manage it's own hidden state.

ronan-f: Oh looks like that was avoided due to concern over some flickering behavior. In which case I guess we should just leave it here although could maybe refactor to just:

```js
this.hidden = !canUpdateAlignments && !this._hasAlignments;
```

mdulat:
I have 2 questions: 

1. Can we use d2l-select-outcomes-hierarchical's `d2l-alignment-list-loaded` event to manage our loading state?
2. Why doesn't the outcomes component skeletize? Should it? (probably a separate issue)

ronan-f:
> I have 2 questions:
> 
> 1. Can we use d2l-select-outcomes-hierarchical's `d2l-alignment-list-loaded` event to manage our loading state?
> 2. Why doesn't the outcomes component skeletize? Should it? (probably a separate issue)

1. Oh interesting. I'm going to look into that.

2. My understanding is that we don't know if an assignment has associated outcomes when the skeleton loading begins (we only render it when `canEditName` becomes truthy) and so from a UX perspective we decided to not show it and only add it in once the skeleton stops loading.

components/d2l-activity-editor/d2l-activity-outcomes.js
emil-furniss: I believe that removing this condition is what is resulting in the behavior of the gif, where the button renders at the correct time but the tags load in after. Not sure if that behavior is acceptable - there was work done to explicitly get rid of that behavior.

And it has to check for `undefined` explicitly, since it starts out being `undefined` and is then set to `true` or `false` depending on if there are alignments present or not.

mdulat (approved):

PR #1421, opened 2021-01-22
DE42177 - tooltip is out of position when saved and there is an error

quinnbudan:
[Rally defect link](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F486456166756&fdp=true?fdp=true)

The `ErrorHandlingMixin.setError()` function lets you provide a `tooltipId` that it will show even if the input no longer has focus and isn't being hovered. This means that the tooltip is rendered before the `d2l-activity-editor` re-renders itself and adds the alert, so the tooltip stays in the spot it was originally rendered.

Here I'm not passing in the `tooltipId` to the `ErrorHandlingMixin.setError()` function so that the tooltip is not rendered if the user isn't focused on the input or hovering over it. When the error happens, the `d2l-activity-editor` component (through the `ActivityEditorContainerMixin`) still focuses on the input on save when there's an error, so the tooltip still shows up when you save, but in the right position.

I'm also changing `d2l-activity-content-editor-title` because the same issue happens there as well.

DE42177_tooltip

### DIFF
--- a/components/d2l-activity-editor/error-handling-mixin.js
+++ b/components/d2l-activity-editor/error-handling-mixin.js
@@ -5,11 +5,11 @@ export const ErrorHandlingMixin = superclass => class extends LocalizeMixin(supe
 	clearError(errorProperty) {
 		this[errorProperty] = '';
 	}
-	async setError(errorProperty, langterm, tooltipId) {
+	async setError(errorProperty, langterm, tooltipIdToForceShow) {
 		this[errorProperty] = this.localize(langterm);
 
 		await this.updateComplete;
-		this._showTooltip(tooltipId);
+		this._showTooltip(tooltipIdToForceShow);
 	}
 
 	_showTooltip(tooltipId) {
--- a/components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-title.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-title.js
@@ -106,7 +106,7 @@ class ContentEditorTitle extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivi
 			timeOut.after(ContentEditorConstants.DEBOUNCE_TIMEOUT),
 			() => {
 				if (isTitleEmpty) {
-					this.setError('_titleError', 'content.emptyNameField', 'title-tooltip');
+					this.setError('_titleError', 'content.emptyNameField');
 				}
 				else {
 					this.clearError('_titleError');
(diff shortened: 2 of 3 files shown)


components/d2l-activity-editor/error-handling-mixin.js
```diff
@@ -1,21 +1,21 @@
...
-	}
-
-	_showTooltip(tooltipId) {
-		const tooltipElem = this.shadowRoot.getElementById(tooltipId);
-		if (tooltipElem) {
-			tooltipElem.show();
-		}
-	}
-};
+	async setError(errorProperty, langterm, tooltipIdToForceShow) {
```
quinnbudan, line 8: I'm not sure why the line endings seem to be changing, I think I have my editorconfig set up properly. The changes are just on line 8 and line 12.

CodeBaboon:
Line endings aside (I am also not sure what's up with that) I have no issues with the code itself. But we do probably need to vet this functional change with our UX designers. Additionally, are there any a11y changes or considerations with this change?

quinnbudan:
> Additionally, are there any a11y changes or considerations with this change?

I don't believe there should be any a11y considerations, because `aria-invalid` is still true, `aria-describedby` is still set, and the tooltip still has the error so I think a screen reader can still read it. I'll double check though! I just need to set up a screen reader first. From the visual side of things, the input is still highlighted red and the tooltip still shows up on focus/hover/save.

quinnbudan:
I have confirmed that the screen reader functionality hasn't changed with these changes, but I'm not sure if it's right either. Basically the only way the screen reader will read out that there was an error is if you refocus to the input (or if you save, because that refocuses the first invalid input). It also doesn't read the tool tip message, just "Invalid selection".

components/d2l-activity-editor/error-handling-mixin.js
ianlaidlaw: i think the line endings are supposed to be `LF` might be worth checking what you have them set as

ianlaidlaw: also i think i'm missing something obvious, how is this working? looks like you're removing the 3rd parameter from calls to this function, so where does `tooltipIdToForceShow` come from?

quinnbudan: I'll double check thanks! This `ErrorHandlingMixin` is being used by another component (`ActivityNotificationEmailEditor`) that does want the tooltip to show up even if the user no longer has focus on the input. I'm updating the name of the third parameter to be clearer about what it's going to do (i.e. it's going to force show that tooltip), but we don't want this behaviour on our weblink/name inputs.

ianlaidlaw: gotcha! ugh, I hate mixins.

ianlaidlaw (approved):

ChrisKraljevicD2L (approved):

CodeBaboon (approved):

PR #1425, opened 2021-01-22
W2D: show Start Dates in the future

mbardakovD2L:
For context, we want to show "Starts (Month) (Date)" for items that start in the future, because otherwise people who see them in W2D would be confused why they aren't clickable. 

See: https://rally1.rallydev.com/#/357251704080d/workviews?detail=%2Fuserstory%2F483215752572&view=b064b584-6337-43c1-b9a9-28047568284d&fdp=true?fdp=true

Note: the `d2l-status-indicator` component doesn't exactly match the mockups (note that the production version has the text in all caps). I'm inclined to use the component as-is because it's probably consistent with other, similar places in Brightspace

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
@@ -16,6 +16,7 @@ import { LocalizeWorkToDoMixin } from './localization';
 import { nothing } from 'lit-html';
 import { QuickEvalActivityAllowList } from '../d2l-quick-eval-widget/env';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
+import { formatDate } from '@brightspace-ui/intl/lib/dateTime';
 
 class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinLit(LocalizeWorkToDoMixin(LitElement)))) {
 
@@ -164,8 +165,15 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 
 		const dateTemplate = html `<d2l-activity-date href="${this.href}" .token="${this.token}" format="MMM d" ?hidden=${this.skeleton}></d2l-activity-date>`;
 
-		const separatorTemplate = !this.skeletize && this._date && (this._orgName || this._orgCode)
+		const separatorTemplate = !this.skeleton && this._date && (this._orgName || this._orgCode)
 			? html `<d2l-icon class="d2l-icon-bullet" icon="tier1:bullet"></d2l-icon>`
+			: nothing;
+
+		const startDateTemplate = !this.skeleton && !this._started
+			? html `
+			<div class="d2l-status-container">
+				<d2l-status-indicator state="none" text="${this._startDateFormatted}"></d2l-status-indicator>
+			</div>`
 			: nothing;
 
 		return html `${this._renderListItem({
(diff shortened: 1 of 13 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -157,10 +163,17 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 
 		const dateTemplate = html `<d2l-activity-date href="${this.href}" .token="${this.token}" format="MMM d" ?hidden=${this.skeleton}></d2l-activity-date>`;
 
-		const separatorTemplate = !this.skeletize && this._date && (this._orgName || this._orgCode)
+		const separatorTemplate = !this.skeleton && this._date && (this._orgName || this._orgCode)
```
mbardakovD2L, line 166: not related to this ticket specifically, just a typo I noticed

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -99,6 +100,14 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 					overflow: hidden;
 					text-overflow: ellipsis;
 				}
+				.d2l-supporting-info-show-start-date {
+					display: flex;
+					flex-direction: column;
+				}
```
mbardakovD2L, line 106: the `display: flex` here is because we want to be able to display the description with the "Starts ___ _" status/description underneath, as two separate items stacked one above the other.

mbardakovD2L: update: we went with a slightly different design here to this is no longer relevant to the detailed view, though it still applies to the widget view

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -157,10 +163,17 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
...
 			? html `<d2l-icon class="d2l-icon-bullet" icon="tier1:bullet"></d2l-icon>`
 			: nothing;
 
+		const startDateTemplate = !this.skeleton && !this._started
```
AntonBazhal, line 170: not totally related to this PR, but should we use `started` class to calculate `this._started` instead of comparing dates?

mbardakovD2L: we can do that, it should be pretty straightforward

JasperAtD2L: there is a ticket to address this.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -223,6 +237,13 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 			: true;
 	}
 
+	/** Start Date formatted like (Short month) (Day) e.g. "Aug 15" */
+	get _startDateFormatted() {
+		return `Starts ${this._usage && this._usage.startDate()
```
AntonBazhal, line 242: Doesn't this need to be localized?

mbardakovD2L: oops - you're totally right

JasperAtD2L (approved):

PR #1426, opened 2021-01-22
Changed date pickers to render as disabed if user lacks permissions

ronan-f:
Based on this trello card: https://trello.com/c/79bACFo0/301-when-missing-delivery-permissions-date-fields-are-hidden-instead-of-disabled

The idea is that if a user doesn't have permissions to edit dates then it will show as a read only picker instead of not rendering at all. 

Without content:

image

With content:

image

I noticed `d2l-activity-availability-dates-editor` is being used by Assignments, Quizzing and Content so should probably get someone from the Content team on this PR as well.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
@@ -37,17 +37,14 @@ class ActivityAvailabilityDatesEditor extends (ActivityEditorMixin(LocalizeActiv
 			endDateError
 		} = this._getDateValues();
 
-		if (!canEditDates) {
-			return html``;
-		}
-
 		return html`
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -100,7 +98,6 @@ class ActivityAvailabilityDatesEditor extends (ActivityEditorMixin(LocalizeActiv
 
 		if (!dates.canEditDates) {
 			datesEntity.canEditDates = false;
-			return datesEntity;
```
ronan-f, line 103: Not returning here so that dates that have already been set still get added to `datesEntity`

mdulat (approved):

PR #1427, opened 2021-01-25
W2D: Update empty states

surags:
https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?detail=%2Fuserstory%2F483243445644&fdp=true

image
image
image

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -279,7 +279,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			}
 
 			return html`
-				${immersiveNav()}
+				${this.fullscreen ? immersiveNav() : ''}
 				<div class="d2l-empty-template">
 					<div class="d2l-empty-icon-container">
 						<d2l-work-to-do-empty-state-image id="empty-icon"></d2l-work-to-do-empty-state-image>
@@ -343,16 +343,12 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			: nothing;
 
 		const immersiveNav = () => {
-			if (this.fullscreen) {
-				return html`
-					<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
-						<div class="d2l-typography d2l-body-standard" slot="middle">
-							<p>${this.localize('myWorkToDo')}</p>
-						</div>
-					</d2l-navigation-immersive>`;
-			}
-
-			return nothing;
+			return html`
+				<d2l-navigation-immersive back-link-href="${this._homeLinkHref}" back-link-text="${this.localize('backToD2L')}">
+					<div class="d2l-typography d2l-body-standard" slot="middle">
+						<p>${this.localize('myWorkToDo')}</p>
+					</div>
+				</d2l-navigation-immersive>`;
 		};
 
 		const fullscreenTemplate = () => {
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -4,6 +4,7 @@ import '@brightspace-ui/core/components/link/link';
 import './d2l-work-to-do-activity-list-header';
 import './d2l-work-to-do-activity-list-item-basic';
 import './d2l-work-to-do-activity-list-item-detailed';
+import './d2l-work-to-do-empty-state-image';
 
 import '@webcomponents/webcomponentsjs/webcomponents-loader';
 import 'd2l-navigation/d2l-navigation-immersive';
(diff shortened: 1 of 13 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -91,16 +92,13 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
-					width: 100px;
+					max-width: 18rem;
+					width: 100%;
 				}
 				.d2l-work-to-do-fullscreen-container {
 					background-image: linear-gradient(to bottom, #f9fbff, #ffffff);
 					padding: 0 2rem;
 				}
-				d2l-button {
-					padding: 1.8rem 0;
```
mbardakovD2L, line 102: is this just for the "load more" button? does this interfere with the non-empty state at all?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -220,34 +218,56 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
+					</div>
+					<div class="d2l-body-standard d2l-empty-body-text-container">
+						${this.localize('comeBackNoFutureActivities')}
+					</div>
+				`;
 			}
 
-			return discoverActive
-				? html`${this.localize('noActivitiesDiscoverActive')}`
-				: html`${this.localize('noActivitiesDiscoverInactive')}`;
```
mbardakovD2L, line 230: do we no longer have a concept of "you have no activities, use discover to find more"?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -256,19 +276,14 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			}
 
 			return html`
+				${this.fullscreen ? immersiveNav() : ''}
```
mbardakovD2L, line 279: I _think_ it'd be easier/cleaner to handle the `this.fullscreen?` logic inside `immersiveNav()` itself, i.e. if we're not in fullscreen view then `immersiveNav()` would return the `nothing` directive

components/d2l-work-to-do/lang/en.js
```diff
@@ -12,12 +14,16 @@ export const val = {
 	loadMore : "Load More", // Button text displayed in 'Fullscreen View' that allows the user to access the next page of activities which will append to the bottom of the list currently shown
 	loadMoreDescription : "Display more assigned activities", // Additional description text to accompany the load more button for additional clarity for the user
 	myWorkToDo: "My Work To Do", // Widget title
+	noActivities : "You have no incomplete activities with due or end dates available.", // 'Empty state' - When widget has no activities in full page view
 	noActivitiesDiscoverActive : "You have no activities due or ending in the next two weeks. Check out Discover to find more learning.", // 'Empty View' - When widget has no activities to display within the next two weeks, but discover is active and can be navigated to by interacting with the provided element
 	noActivitiesDiscoverInactive : "You have no activities with due or end dates available.", // 'Empty View' - When widget has no activities to display within the next two weeks & there are no more activities in the future & discover is not active for this user
```
mbardakovD2L, line 19: if we _are_ getting rid of the "go to discover" linking we could also clean up these two langterms

components/d2l-work-to-do/d2l-work-to-do.js
surags: This change was intended for the view all work button. The only other place d2l-button is used is the load more button and I have updated the styling of the load more class to reflect the change

components/d2l-work-to-do/d2l-work-to-do.js
surags: The code seems to make references to discover but none if it is ever used. This was my attempt to clean it up a little. We can add it back when we work on adding discover. I don't feel strongly about it so I can revert if needed

components/d2l-work-to-do/d2l-work-to-do.js
surags: Updated

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: right - that makes sense. I'm also good either way but we should ask product/design if that's still planned

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -256,19 +279,14 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			}
 
 			return html`
+				${immersiveNav()}
```
JasperAtD2L, line 282: It's weird that we have empty/fullscreen/and widget (activity) states. It should really be empty/activities with a modifier of fullscreen. That being said we can't make that change.

Where I am going is; the immersiveNav is a full screen only thing, so the choice to display the immersive nav should be made here and not in the immersiveNav function.

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: update: nevermind, see Jasper's comments below

JasperAtD2L (approved):

mbardakovD2L (approved):

PR #1429, opened 2021-01-26
feat: [US124145] W2D telemetry

AntonBazhal:
https://rally1.rallydev.com/#/357251704080/dashboard?detail=/userstory/484006453552
https://rally1.rallydev.com/#/357251704080/dashboard?detail=/userstory/484008675520

`LoadMore` event includes upcoming activities request duration only (it does not measure the time till new activities show up). Not sure how much that is needed, so decided not to invest time for now. All these events can be easily extended in the future.

Sample events:
```json
{
  "EventType": "PerformanceEvent",
  "SourceId": "worktodo",
  "Date": "2021-01-31T03:16:25.715Z",
  "EventBody": {
    "EventTypeGuid": "02a00ca0-e67d-4a6c-908b-ee80b8a61eec",
    "Action": "LoadView",
    "Object": {
      "Id": "View",
      "Type": "Widget",
      "Url": "View"
    },
    "Custom": [
      {
        "name": "OverdueCount",
        "value": "1"
      },
      {
        "name": "UpcomingCount",
        "value": "20"
      },
      {
        "name": "UpcomingCount",
        "value": "20"
      }
    ],
    "UserTiming": [
      {
        "name": "d2l-work-to-do.view.loaded",
        "entryType": "measure",
        "startTime": 0,
        "duration": 5780.180000001565
      },
      {
        "name": "d2l-work-to-do.api.overdue.loaded",
        "entryType": "measure",
        "startTime": 2893.400000000838,
        "duration": 1621.979999996256
      },
      {
        "name": "d2l-work-to-do.api.upcoming.loaded",
        "entryType": "measure",
        "startTime": 2893.7649999861605,
        "duration": 2345.6200000364333
      },
      {
        "name": "d2l-work-to-do.api.upcoming.loaded",
        "entryType": "measure",
        "startTime": 2893.945000017993,
        "duration": 2429.2049999930896
      }
    ]
  },
  "name": "LoadView",
  "ts": 1612062985
}
```

```json
{
  "EventType": "PerformanceEvent",
  "SourceId": "worktodo",
  "Date": "2021-01-31T03:32:21.256Z",
  "EventBody": {
    "EventTypeGuid": "02a00ca0-e67d-4a6c-908b-ee80b8a61eec",
    "Action": "LoadMore",
    "Object": {
      "Id": "https%3A%2F%2Factivities.api.brightspace.com%2Frels%2Fnext-page",
      "Type": "ActivitiesNextPage",
      "Url": "https://activities.api.brightspace.com/rels/next-page"
    },
    "Custom": [
      {
        "name": "UpcomingCount",
        "value": "4"
      }
    ],
    "UserTiming": [
      {
        "name": "d2l-work-to-do.api.upcoming.loaded",
        "entryType": "measure",
        "startTime": 15883.139999990817,
        "duration": 1634.9199999822304
      }
    ]
  },
  "name": "LoadMore",
  "ts": 1612063941
}
```

```json
{
  "EventType": "TelemetryEvent",
  "SourceId": "worktodo",
  "Date": "2021-01-31T03:16:27.498Z",
  "EventBody": {
    "EventTypeGuid": "d6b021b6-ffa2-4ee8-8206-8710e04396ce",
    "Action": "NavigatedTo",
    "Object": {
      "Id": "https%3A%2F%2Fbazhalwtd.devlms.brightspace.com%2Fd2l%2Fhome%2F121535",
      "Type": "Course",
      "Url": "https://bazhalwtd.devlms.brightspace.com/d2l/home/121535"
    }
  },
  "name": "NavigatedTo",
  "ts": 1612062987
}
```


components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,79 @@
+import Events from 'd2l-telemetry-browser-client';
+
+const W2D_BASE_MARK = 'd2l-work-to-do';
+const W2D_OVERDUE_MARK = `${W2D_BASE_MARK}.ovedue`;
+const W2D_OVERDUE_LOAD_START_MARK = `${W2D_OVERDUE_MARK}.start`;
+const W2D_OVERDUE_LOADED_MEASURE = `${W2D_OVERDUE_MARK}.loaded`;
```
AntonBazhal, line 6: Not sure this is the best way to define consts inside mixin

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -523,8 +524,12 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		}
 
 		const source = entity.getLinkByRel(Rels.Activities.overdue).href;
+		this.markLoadOverdueStart();
 		await fetchEntity(source, this.token)
 			.then((sirenEntity) => {
+				const overdueCount = sirenEntity.getSubEntitiesByRel(Rels.Activities.userActivityUsage).length;
+				this.markAndLogLoadOverdueEnd(source, Rels.Activities.overdue, overdueCount);
```
AntonBazhal, line 531: Not sure what type to use here. I could not find any examples of measuring URL load time. It's mostly how long it takes for the whole page/widget to load and the type there is something like `MainView` or something.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,79 @@
...
+	markAndLogLoadOverdueEnd(href, type, count) {
+		this._logPerformanceEvent('LoadOverdue', href, type, W2D_OVERDUE_LOAD_START_MARK, W2D_OVERDUE_LOADED_MEASURE, { OverdueCount: count });
+	}
+
+	_markEventStart(startMark) {
+		if (!startMark) {
+			return;
+		}
+
+		performance.clearMarks(startMark);
```
JasperAtD2L, line 48: where is performance defined?

AntonBazhal: https://developer.mozilla.org/en-US/docs/Web/API/Performance

JasperAtD2L: image

JasperAtD2L: I was looking everywhere *except* the browser api.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,79 @@
...
+			return;
+		}
+
+		performance.clearMeasures(measureName);
+		performance.measure(measureName, startMark);
+
+		const eventBody = new Events.PerformanceEventBody()
+			.setAction(action)
+			.setObject(encodeURIComponent(href), type, href)
+			.addUserTiming(performance.getEntriesByName(measureName));
```
JasperAtD2L, line 63: we might want to normalize this object. Not sure what entries from other areas of the BSI look like.

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: I would almost set the id of the object as the rel and the type as XHR, HyperMedia, or something to that effect.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
AntonBazhal: this is consistent with what other places do. [this official example](https://docs.dev.d2l/index.php/Product_Telemetry_Service_Example_Events#Performance_Event) is the same as well.

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: After looking through that telemetry example, the type seems like just a human readable object. So maybe "UserOverdueActivities"? And probably could be defined in the mixin itself.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,79 @@
...
+			});
+		}
+
+		const event = new Events.TelemetryEvent()
+			.setType('PerformanceEvent')
+			.setDate(new Date())
+			.setSourceId(W2D_TELEMETRY_ID)
+			.setBody(eventBody);
+
+		this._client.logUserEvent(event);
```
JasperAtD2L, line 77: Looking at the telemetry service, it seems like the performance event crosses the entirety of the loading sequence. Should we follow this guide and measure all the load api times and package them in 1 "worktodo" performance event?

components/d2l-work-to-do/mixins/d2l-work-to-do-localization-mixin.js
```diff
@@ -48,7 +48,7 @@ export const LocalizeWorkToDoMixin = superclass => class extends LocalizeMixin(s
 			}
 		}
 
-		translations = await import('./lang/en.js');
+		translations = await import('../lang/en.js');
 		return {
 			language: 'en',
 			resources: translations.val
```
mbardakovD2L, line 54: this is more for my own curiosity than actual feedback - did you just move this file because it makes more sense here, or was there a technical reason as well?

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
AntonBazhal: 🤷‍♂️ it can go either way. some requests happen during the widget load, but some don't. like, we get the first upcoming activities page on load, but we get more pages later. the former can be a part of the loading sequence, but the latter can't. I guess it's more about what questions we want answered. The current rough list is:

1. When users click on the “load more” button
2. When users open the view all view
3. Time to load overdue
4. Time to load upcoming
5. When users click on an entity from the widget and the full page view

components/d2l-work-to-do/mixins/d2l-work-to-do-localization-mixin.js
AntonBazhal: just have more sense to have all mixins in one place

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
JasperAtD2L: I think we can get all those with 1 load performance event, just separate user timing values. We could then have a "load more" performance event for when they click the button.
Would allow us to add later things like render time to the load more and initial load events.

Not sure how we would get "open the view all view" since this is a page redirect, unless the browser keeps those performance marks across pages.

For full page view, the fullscreen flag should be a context field on every event, so we can see the difference.

JasperAtD2L: Not sure what the UI is for the telemetry stuff and if 1 event vs many events would be more useful.

AntonBazhal: yeah, we can definitely try to put all the perf measures in one event. I don't think there is a big difference between one or many measures in one event aside from custom fields being relevant to one measure only.

"open the view all view" is a TelemetryEvent, not PerformanceEvent. Like, we just need to log the fact that the full page view was opened.

This is how "UI" looks like:
Screen Shot 2021-01-26 at 3 33 01 PM

JasperAtD2L: > "open the view all view" is a TelemetryEvent,

Gotcha that makes a lot more sense.

> This is how "UI" looks like:

I see 1 metric per event would probably be an easier query 😆

AntonBazhal: Just figured out a sample query:
```sql
WITH a AS (
   SELECT
   source,
   date as datetime,
   date_trunc('hour', from_iso8601_timestamp(date)) as date,
   tenantid,
   eventbody.tenantUrl,
   eventbody.context.id as courseId,
   eventbody.actor.id as userId,
   eventbody.browser.name as browser,
   eventbody.browser.device as device,
   eventbody.browser.os as os,
   eventbody.usertiming
   FROM "product_telemetry"."event_data_partition_by_source"
   WHERE  source = 'worktodo'
   AND (EventType = 'https://schema.brightspace.com/ns/common/ui-performance-event.json'
   OR EventType = 'PerformanceEvent')
 )

SELECT
source,
datetime,
date,
tenantid,
tenantUrl,
courseId,
userId,
browser,
device,
os,
dots.name,
dots.duration
FROM a
CROSS JOIN UNNEST(usertiming) as t(dots)
WHERE dots.name = 'd2l-work-to-do.ovedue.loaded'
AND dots.duration > 0
```
Looks like `UNNEST` is just flattening all measures, so it does not matter for the query whether we have one or many. But it's definitely more efficient to have more than one to have fewer events and less data for the query to scan.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	markAndLogLoadMoreEnd(count) {
+		this.markLoadUpcomingEnd(false, count);
+		this._logPerformanceEvent('LoadMore', Rels.Activities.nextPage, 'ActivitiesNextPage', [W2D_UPCOMING_LOADED_MEASURE]);
+	}
+
+	markAndLogWidgetLoaded(fullscreen) {
+		this._markEventEnd(W2D_VIEW_LOADED_MEASURE);
+		this._logPerformanceEvent('LoadView', 'View', fullscreen ? 'Fullscreen' : 'Widget', W2D_WIEW_LOAD_MEASURES);
```
AntonBazhal, line 59: I can't decide whether Fullscreen/Widget should be Id or Type

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -134,6 +138,21 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
...
 	}
 
+	/**
+	 * Logs activity navigated telemetry event
+	 */
+	_onItemTriggered(event) {
+		const activityType = this._activityProperties && this._activityProperties.type;
+		if (this.skeleton ||
+			!activityType ||
+			!this.actionHref ||
```
bearfriend, line 148: Can we check `actionHref` against the `target`'s `href`? Slightly uncomfortable with assuming the event has actually triggered navigation.

bearfriend: clickity-click!
We should fix this, but there's 15 telemetry events that didn't actually happen.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -163,6 +167,21 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
 
+	/**
+	 * Logs activity navigated telemetry event
+	 */
+	_onItemTriggered(event) {
+		const activityType = this._activityProperties && this._activityProperties.type;
+		if (this.skeleton ||
+			!activityType ||
+			!this.actionHref ||
+			(event.type === 'keydown' && event.keyCode !== 13 && event.keyCode !== 32)) {
```
JasperAtD2L, line 178: For those of us who needed to look up an ascii table to remember 13 == return, and 32 == space. Can you add a comment.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -619,15 +637,21 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			}
 			return acc;
 		}, []);
+
+		this.markLoadUpcomingStart(isMax);
```
JasperAtD2L, line 641: This is kind of being refactored a bit: https://github.com/BrightspaceHypermediaComponents/activities/pull/1451

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -573,11 +587,15 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
 		await fetchEntity(source, this.token)
 			.then((sirenEntity) => {
 				if (sirenEntity) {
+					this.markLoadOverdueEnd(sirenEntity.getSubEntitiesByRel(Rels.Activities.userActivityUsage).length);
 					this._overdueActivities = this._getFilteredOverdueActivities(sirenEntity);
 					this._overdueCollection = sirenEntity;
+				} else {
+					this.markLoadOverdueEnd(0);
 				}
 			});
```
JasperAtD2L, line 600: Do we need to unmark or log the errors?

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	logActivityNavigatedTo(href, type) {
+		this._logTelemetryEvent('NavigatedTo', href, type);
+	}
+
+	_markEventStart(startMark) {
```
JasperAtD2L, line 66: Do we think it's worth putting these in a sub class that other telemetry mixins could extend from and get this functionality?

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
JasperAtD2L: Wouldn't id (href) be fullscreen = '/d2l/worktodo' and widget = 'homepage'?

JasperAtD2L: What would also be helpful is the href for emptystate which will tell us if it's on the org homepage or a course homepage.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -105,6 +106,9 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 		this._activityProperties = undefined;
 		this._organization = undefined;
 		this._setEntityType(ActivityUsageEntity);
+
+		this.addEventListener('click', this._onItemTriggered.bind(this));
```
dbatiste, line 110: This seems kinda risky is practice - I think it is assuming that all item clicks will translate into navigation. It might be ok in this particular usage, but I don't think it would be safe for all usages.  Generally speaking, there could be actions such as selection, secondary actions, or even drag & drop. Eventually we may support nested lists, in which case clicks on those lists would bubble up and get caught by this.

image

What if we were to expose a custom event - `d2l-list-item-link-before-navigate`?  We'd wire up to the actual `<a>` element's click.

AntonBazhal: Yeah, I was debating about this myself. I can do something like:
```js
const link = this.shadowRoot.querySelector('a');
link.addEventListener('click', this._onItemTriggered.bind(this));
```
Which is gonna be better, but still quite fragile, I guess. A custom event should be better.

components/d2l-work-to-do/d2l-work-to-do.js
AntonBazhal: We do not need to unmark. Log errors - yes, potentially. There is another event designed for this purpose. The whole `if (sirenEntity)` check is strange to me as I can't think of a case when this happens.

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
AntonBazhal: Dunno, all the examples I've seen so far, do things differently and build helpers from scratch 🤷‍♂️

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
dbatiste: Yeah we definitely wouldn't want consumers depending on the internals of the component since they are always subject to change.  I've create the custom event `d2l-list-item-link-click` event - release in core v1.115.0.

bearfriend: `d2l-list-item-link-click` originates from a native click event so it'll handle the enter key, and we don't actually want the space key handled so we can drop the `keydown` listener entirely.

AntonBazhal: Perfect, thanks Dave

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: It's been merged in, you likely will want to rebase

components/d2l-work-to-do/mixins/d2l-work-to-do-telemetry-mixin.js
AntonBazhal: I changed Id to `window.location.pathname`. 👍 👎 ?

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
m6jones: Why don't you want the space key handled?

AntonBazhal:
I think I addressed all the comments. Should be good for another 👁️

JasperAtD2L (approved):

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
bearfriend: The space key does not activate links.

PR #1434, opened 2021-01-26
fix: Add retry logic in W2D

surags:
https://rally1.rallydev.com/#/357251704080ud/custom/367300408400?detail=%2Fdefect%2F486828986160&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -494,8 +494,8 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	}
 
 	// !TODO Further investigate defect DE42208 to remove need for retry
-	async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 1)  {
-		if (maxRetries) {
+	async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 0)  {
+		if (maxRetries > -1) {
 			return performSirenAction(token, action, fields, immediate)
 				.catch(() => {
 					return this._performSirenActionWithRetry(token, action, fields, immediate, maxRetries - 1);
@@ -563,7 +563,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			}
 			return acc;
 		}, []);
-		this._performSirenActionWithRetry(this.token, action, fields, true, 2)
+		this._performSirenActionWithRetry(this.token, action, fields, true, 1)
 			.then((sirenEntity) => {
 				if (sirenEntity) {
 					if (!isMax) {
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -493,6 +493,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		}
 	}
 
+	// !TODO Further investigate defect DE42208 to remove need for retry
+	async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 1)  {
+		if (maxRetries) {
+			return performSirenAction(token, action, fields, immediate)
+				.catch(() => {
+					return this._performSirenActionWithRetry(token, action, fields, immediate, maxRetries - 1);
+				});
+		}
+
+		return Promise.reject();
+	}
+
 	/**
 	 * Load collection of overdue activities.
 	 * Will set collection of overdue activities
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -493,6 +493,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		}
 	}
 
+	// !TODO Further investigate defect DE42208 to remove need for retry
+	async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 1)  {
+		if (maxRetries) {
```
JasperAtD2L, line 498: Max retries should be x + 1 to account for the original request. So here should probably be `maxRetries > -1` to allow for retries = 0, and then maxRetries can be default to 0, and below we can specify 1 instead of 2.

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -493,6 +493,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
+			return performSirenAction(token, action, fields, immediate)
+				.catch(() => {
+					return this._performSirenActionWithRetry(token, action, fields, immediate, maxRetries - 1);
```
mbardakovD2L, line 501: my gut feeling is that recursion is a weird way to solve this problem but I don't know how else it could be done - if you do this in a `while` loop and `await performSirenAction`, could that accomplish the same thing? if not, I'm okay with this too

surags: `async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 1)  {
	for(i = -1; i < maxRetries; i++) {
		try {
			return await performSirenAction(token, action, fields, immediate);
		} catch {
			continue;		
		}
	}
}`

An iterative solution will probably look something like this. Not sure which is better/more readable

JasperAtD2L: I don't think recursion is bad here. Since we control the exit condition, the call stack won't get terribly big. Maybe @AntonBazhal has other concerns either way.

AntonBazhal: Recursion is fine. My concern with the code above is that we swallow an actual error when the retry limit is reached.

surags: The new commit should reject the promise with last retry error. The failure reason for the retries will likely be the same so that should be sufficient. The other alternative being append the error message for each retry and reject with a list of errors

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -507,8 +519,10 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		const source = entity.getLinkByRel(Rels.Activities.overdue).href;
 		await fetchEntity(source, this.token)
 			.then((sirenEntity) => {
-				this._overdueActivities = this._getFilteredOverdueActivities(sirenEntity);
-				this._overdueCollection = sirenEntity;
+				if (sirenEntity) {
```
AntonBazhal, line 522: in which case we may get no siren entity back in case of successful call?

AntonBazhal (approved):
Do you need to spin up a new site every time to reproduce this error?

surags:
> Do you need to spin up a new site every time to reproduce this error?

From what I have seen yes. I was unable to reproduce the error any other way.

components/d2l-work-to-do/d2l-work-to-do.js
surags: Not quite sure what chain of events trigger this but I noticed a console error in  `this._getFilteredOverdueActivities()`  on first load where no error was thrown(_.catch() was not executed_), but the sirenEntity was undefined. Additionally, since we do a check everywhere else it also keeps the code consistent

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -493,6 +493,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		}
 	}
 
+	// !TODO Further investigate defect DE42208 to remove need for retry
+	async _performSirenActionWithRetry(token, action, fields, immediate, maxRetries = 0, error = null)  {
+		if (maxRetries > -1) {
+			return performSirenAction(token, action, fields, immediate)
```
JasperAtD2L, line 499: I just want to make sure, but does this actually throw an error? I thought these swallowed the error and spit out undefined?

surags: Verified that in this case an error is actually thrown and it shows in console in the unhandled case.

JasperAtD2L: Thanks Surag. That must've been a long one to verify.

PR #1436, opened 2021-01-27
fix: hover state of detailed w2d list items

bearfriend:
Oversight from https://github.com/BrightspaceHypermediaComponents/activities/pull/1335

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
@@ -66,11 +66,10 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 					text-overflow: ellipsis;
 					white-space: nowrap;
 				}
-				.d2l-activity-icon-container.d2l-hovering,
-				.d2l-activity-icon-container.d2l-focusing,
-				.d2l-activity-name-container.d2l-hovering,
-				.d2l-activity-name-container.d2l-focusing {
-					--d2l-list-item-content-text-decoration: underline;
+				.d2l-hovering .d2l-activity-icon-container.d2l-has-action,
+				.d2l-focusing .d2l-activity-icon-container.d2l-has-action,
+				.d2l-hovering .d2l-activity-name-container.d2l-has-action,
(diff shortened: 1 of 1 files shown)


mbardakovD2L (approved):
I've made this mistake enough times you'd think I'd start catching it in reviews by now 😅

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -66,11 +66,10 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 					text-overflow: ellipsis;
 					white-space: nowrap;
 				}
-				.d2l-activity-icon-container.d2l-hovering,
-				.d2l-activity-icon-container.d2l-focusing,
-				.d2l-activity-name-container.d2l-hovering,
-				.d2l-activity-name-container.d2l-focusing {
-					--d2l-list-item-content-text-decoration: underline;
+				.d2l-hovering .d2l-activity-icon-container.d2l-has-action,
```
dbatiste, line 69: Are some fixes to the list-item components or their mixins in order?  The reason I ask is that it seems odd that consumers would need to override the color for these states.  I think the `d2l-hovering` and `d2l-focusing` classes should be regarded as purely internal to the core mixns.

bearfriend: In this specific scenario, it looks like there's no issue with the list-item components. We appear to be using the link mixin which defaults to the blue text, but our design dictates it be grey until hover. Perhaps a discussion for the designers?

Separate but related, there does appear to be an issue with the link mixin that does not properly attach/update the `actionHref` property/attribute so we are forced to compute and check `has-action` here separately. This could, however, be an issue inside the w2d components. I haven't had time to dig into it yet.

dbatiste: Yeah sounds like that color is a question for the designers.  Usually we question variants a lot before they spiral out of control.  😏 

I quickly checked the link action demo in core and it appears to be reflecting `actionHref` to `action-href`.  If you get a chance to check w2d components more and conclude we need to investigate further in core, I'm happy to do that.

PR #1437, opened 2021-01-27
fix: DE41706: Sync item rendering.

JasperAtD2L:
https://rally1.rallydev.com/#/357251704080d/workviews?detail=%2Fdefect%2F464034359184&view=b064b584-6337-43c1-b9a9-28047568284d&fdp=true?fdp=true

https://rally1.rallydev.com/#/357251704080d/custom/367300408400?detail=%2Fdefect%2F483995040488&fdp=true?fdp=true

Render all items only once the item details have been fetched.

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -469,7 +469,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 
 	/**
 	 * Fired when all the data required to render the activity is present.
-	 * Event is guarunteed to fire for success and failures.
+	 * Event is guaranteed to fire for success and failures.
 	 * @param {CustomEvent} event Event data from the activity item
 	 */
 	_itemLoaded(event) {
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -48,7 +48,9 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 			/** Represents current session's UpcomingWeekLimit */
 			_upcomingWeekLimit: { type: Number, attribute: 'data-upcoming-week-limit' },
 			/** individual items within upcomingCollection + subsequent pages once the UI has them */
-			_upcomingActivities: { type: Array }
+			_upcomingActivities: { type: Array },
+			/** keeps track of the sub items being loaded, so we can show all data at once and not a partial activity */
+			_initialLoad: { type: Boolean }
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -381,13 +389,18 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
 		const detailedSkeleton = html`
 			${immersiveNav()}
-			<d2l-work-to-do-activity-list-header skeleton fullscreen></d2l-work-to-do-activity-list-header>
-			<d2l-list separators="none">
-				<d2l-work-to-do-activity-list-item-detailed skeleton href=' ' token=' '></d2l-work-to-do-activity-list-item-detailed>
-				<d2l-work-to-do-activity-list-item-detailed skeleton href=' ' token=' '></d2l-work-to-do-activity-list-item-detailed>
-				<d2l-work-to-do-activity-list-item-detailed skeleton href=' ' token=' '></d2l-work-to-do-activity-list-item-detailed>
-			</d2l-list>
-			<div class="d2l-load-more-button-skeleton d2l-skeletize"></div>
+			<div class="d2l-work-to-do-fullscreen-container">
```
JasperAtD2L, line 392: Not related but noticed the fullscreen skeleton layout was quite different than the final layout. And it was causing quite a bit of jumping between renders.

mbardakovD2L:
how does this work in fullscreen view, when we load more? does it see that we have X number of new activities, create X skeletons, and then flip all X skeletons to actual items once the data is all there?

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -454,6 +467,21 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		return 'loading';
 	}
 
+	/**
+	 * Fired when all the data required to render the activity is present.
+	 * Event is guarunteed to fire for success and failures.
```
mbardakovD2L, line 472: nit: guaranteed

bearfriend (dismissed):
I was personally confused by the name `initialLoad`. I thought it would start as `false` and switch to `true`, but not a big deal. It's obvious what it's doing.

JasperAtD2L:
> how does this work in fullscreen view, when we load more? does it see that we have X number of new activities, create X skeletons, and then flip all X skeletons to actual items once the data is all there?

On load more the _initalLoad is false and thus the skeleton attribute will never get re-applied. And the data loaded hander is effectively disabled.

bearfriend (approved):
🔤

mbardakovD2L:
> > how does this work in fullscreen view, when we load more? does it see that we have X number of new activities, create X skeletons, and then flip all X skeletons to actual items once the data is all there?
> 
> On load more the _initalLoad is false and thus the skeleton attribute will never get re-applied. And the data loaded hander is effectively disabled.

That makes sense - I'm happier with that than with what I described

mbardakovD2L (approved):

PR #1438, opened 2021-01-27
Fix d2l-icon border radius in w2d list items

bearfriend:
https://rally1.rallydev.com/#/357251704080/custom/367300408400?detail=%2Fdefect%2F482557191752&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
@@ -50,8 +50,8 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 					display: none;
 				}
 				.d2l-activity-icon-container {
+					border-radius: 0 !important;
 					padding-top: 0.2rem;
-					border-radius: 0px !important;
 				}
 				:host([skeleton]) .d2l-activity-icon-container {
 					height: 1.3rem;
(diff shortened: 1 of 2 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -51,6 +51,7 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 				}
 				.d2l-activity-icon-container {
 					padding-top: 0.2rem;
+					border-radius: 0px !important;
```
mbardakovD2L, line 54: I'm a little concerned about the `!important` here - I'm assuming there's a border-radius property somewhere else that's interfering? do we know where from?

bearfriend: Sorry should have included this:
https://d2l.slack.com/archives/C0PHG3QB0/p1611759514022800

Also discussed with Maya further separately and we'll add `!important` here for now but she's going to probably add specific CSS for icons to the list item.

mbardakovD2L: okay - as long as we note somewhere to revisit this when we can then I'm happy with it

mbardakovD2L (approved):

mbardakovD2L (approved):

PR #1439, opened 2021-01-27
fix: DE42230: Hide org info for courses in W2D

JasperAtD2L:
https://rally1.rallydev.com/#/357251704080d/custom/367300408400?detail=%2Fdefect%2F486936639864&fdp=true?fdp=true

Added this logic to the basic item, but forgot to add it to the detailed item.


bearfriend (approved):
©️

PR #1443, opened 2021-01-28
Clean up w2d styles and actionHref logic

bearfriend:
Somewhat dependent on https://github.com/BrightspaceUI/core/pull/1076

Refactors the way links and icons are styled, `actionHref` is set, and restores the link background color on hover.

PR #1444, opened 2021-01-28
fix: DE42230: Add follow rel path to list-item-detailed

surags:
https://rally1.rallydev.com/#/357251704080d/custom/367300408400?detail=%2Fdefect%2F486936639864&fdp=true&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
@@ -365,7 +365,7 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 						&& this._activity.getLinkByRel(this._activityProperties.linkRel)
 					) || this._activity.getLinkByRel('alternate');
 
-					this.actionHref = (this._started && (this.evaluateAllHref || (link && link.href))) || null;
+					this.actionHref = (this._started && (link && link.href)) || null;
 				}
 			}
 		}
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
@@ -352,24 +352,45 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 		for (const allowed in ActivityAllowList) {
 			if (entity.hasClass(ActivityAllowList[allowed].class)) {
 				this._activityProperties = ActivityAllowList[allowed];
-				const source = (
-					entity.hasLinkByRel(ActivityAllowList[allowed].rel)
-					&& entity.getLinkByRel(ActivityAllowList[allowed].rel)
-					|| {}).href;
-				if (source) {
-					await fetchEntity(source, this.token)
-						.then((sirenEntity) => {
-							if (sirenEntity) {
-								this._activity = sirenEntity;
-								const link = sirenEntity.getLinkByRel('alternate');
-								this.actionHref = (this._started && link && link.href) || null;
-							}
-						});
-				}
+				const relList = [].concat(this._activityProperties.rel);
 
-				break;
+				const foundEntity = await this._followRelPath(relList, entity);
+
+				if (foundEntity) {
+					this._activity = foundEntity;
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -374,24 +357,47 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
+					this._activity = foundEntity;
+
+					const link = (
+						this._activityProperties
+						&& this._activityProperties.linkRel
+						&& this._activity.getLinkByRel(this._activityProperties.linkRel)
+					) || this._activity.getLinkByRel('alternate');
 
-				break;
+					this.actionHref = (this._started && (this.evaluateAllHref || (link && link.href))) || null;
```
bearfriend, line 373: Only thing is there is no `evaluateAllHref` in the detailed view, so:
`(this._started && link && link.href) || null;` I think?

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -374,24 +357,47 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
-					|| {}).href;
-				if (source) {
-					await fetchEntity(source, this.token)
-						.then((sirenEntity) => {
-							if (sirenEntity) {
-								this._activity = sirenEntity;
-							}
-						});
-				}
+				const relList = [].concat(this._activityProperties.rel);
```
mbardakovD2L, line 360: probably a dumb question but I feel obligated to ask - do we know for sure that the rels we need to dig through are going to be in the right order?

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
surags: 🤦‍♂️

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
surags: The relList is defined [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/854bbe1ce5a77387633dd0ef4d9120d302212348/components/d2l-work-to-do/env/index.js#L47) so the order will always be known.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -374,24 +357,47 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
+
+				const foundEntity = await this._followRelPath(relList, entity);
+
+				if (foundEntity) {
+					this._activity = foundEntity;
+
+					const link = (
```
JasperAtD2L, line 367: Should leave this as a getter/setter, to be consistent with the basic component. These 2 components really should've extended from each other.

JasperAtD2L: Oh I see you are basing of Danny's PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/1443

bearfriend: Yeah working on getting that merged right now so we can clear this chain of dependent PRs.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -374,24 +357,47 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
...
+						this._activityProperties
+						&& this._activityProperties.linkRel
+						&& this._activity.getLinkByRel(this._activityProperties.linkRel)
+					) || this._activity.getLinkByRel('alternate');
 
-				break;
```
JasperAtD2L, line 390: Keep the break, as we don't have to go through the rest of the allowed list.

JasperAtD2L (approved):

PR #1445, opened 2021-01-28
Abbreviate months when header is too narrow

bearfriend:
I used `resizeObserver` to listen to size changes on the list header and show long or short month names accordingly. I did not use the [polyfill](https://www.npmjs.com/package/resize-observer-polyfill) as it was not already part of the repo and this is not critical functionality. This means it will not work in legacy Edge and Firefox on Android only. Tested in Chrome, Edge (Chromium), and Firefox and seems to work fine. Safari should be tested, though any failure should fall back to existing long month names.

It's possible that doing calculations on the width to convert to `rem` would be beneficial, though I did not see any evidence of this in my tests.

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
@@ -9,6 +9,17 @@ import { formatDate } from '@brightspace-ui/intl/lib/dateTime';
 import { LocalizeWorkToDoMixin } from './localization';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
+const ro = new ResizeObserver(entries => {
+	entries.forEach(entry => {
+		const observerSize = entry && ((entry.contentBoxSize && entry.contentBoxSize[0]) || entry.contentBoxSize || entry.contentRect);
+		const width = observerSize && observerSize.width || observerSize.inlineSize;
+
+		if (entry && entry.target && entry.target.resizedCallback && width) {
+			entry.target.resizedCallback(width);
+		}
+	});
+});
+
 /**
  * Provides Title and Count for associated activity usage list
  */
@@ -93,6 +104,16 @@ class ActivityListHeader extends SkeletonMixin(LocalizeWorkToDoMixin(LitElement)
 		this.isOverdue = false;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		ro.observe(this);
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		ro.unobserve(this);
+	}
+
 	render() {
 		const containerClasses = {
 			'd2l-activity-list-header-basic': !this.fullscreen,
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -9,6 +9,17 @@ import { formatDate } from '@brightspace-ui/intl/lib/dateTime';
 import { LocalizeWorkToDoMixin } from './localization';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
+const ro = new ResizeObserver(entries => {
+	entries.forEach(entry => {
+		const i = entry && ((entry.contentBoxSize && entry.contentBoxSize[0]) || entry.contentBoxSize || entry.contentRect);
```
JasperAtD2L, line 14: `contentSize` ? instead of `i`.

JasperAtD2L:
> It's possible that doing calculations on the width to convert to rem would be beneficial, though I did not see any evidence of this in my tests.

I'm guessing you tried this on your high DPI screen for both Chrome and Safari.

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
bearfriend: Yeah to be honest I kind of forgot what exactly that object was and I didn't feel like restarting the environment and setting breakpoints. 🤷 

I'll stop being lazy and check.

bearfriend:
> I'm guessing you tried this on your high DPI screen for both Chrome and Safari.
BSI was not working in Safari (I probably need to run the demo page), but I did test in Chrome and emulated different devices/dpi/resolutions and it all seemed fine.

JasperAtD2L:
> BSI was not working in Safari (I probably need to run the demo page), but I did test in Chrome and emulated different devices/dpi/resolutions and it all seemed fine.

👍 I'll double check on my windows high DPI display once merged too.

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -9,6 +9,17 @@ import { formatDate } from '@brightspace-ui/intl/lib/dateTime';
...
+		const width = i && i.width || i.inlineSize;
```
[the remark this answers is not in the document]
bearfriend: That's a good question, I think doing it this generic way leaves it open to observe other elements as well, but this is a really simple component so maybe not worth it. I'll take a look.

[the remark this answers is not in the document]
bearfriend: Ahhh, ok so because this single instance of `ResizeObserver` is created outside the class it handles all instances of the header component so the `entries` correspond to however many header elements are on the page.

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
bearfriend: It's a `ResizeObserverSize` object, so I went with `observerSize`.

JasperAtD2L (approved):

bearfriend:
Safari tested ✅

JasperAtD2L (approved):

JasperAtD2L:
New Edge tested on windows high DPI ✅

PR #1449, opened 2021-01-29
Fix: DE41706 fix introduced regression

AntonBazhal:


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -475,7 +475,12 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	_itemLoaded(event) {
 		if (this._initialLoad) {
 			this._loadedElements.push(event.target);
-			if (this._loadedElements.length === this._overdueActivities.length + this._upcomingActivities.length) {
+
+			const expectedLoadedActivities = this.fullscreen
+				? this._overdueActivities.length + this._upcomingActivities.length
+				: this._overdueDisplayLimit + this._upcomingDisplayLimit;
```
surags, line 481: If not fullscreen this should probably be `min(this._overdueDisplayLimit, this._overdueActivities.length) + min(this._upcomingDisplayLimit, this._upcomingActivities.length)` to account for the scenario where there aren't enough activities

AntonBazhal: Ah, I think it actually needs to be `min(Constants.MaxWidgetDisplay, this._overdueActivities.length + this._upcomingActivities.length)`

JasperAtD2L (approved):

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: I'm wondering if we should just get the value from the templates: https://github.com/BrightspaceHypermediaComponents/activities/pull/1449/files#diff-d300a684213152d82078d4d614f2f7ede435053c312f283cc93533dbae942e94R192 so we know exactly how many sub components are on screen.

JasperAtD2L: Actually nvm the render gets called a bunch of times, this should be good.

PR #1454, opened 2021-02-01
Preview should only be visible when user has entered a URL

ianlaidlaw:
If the user enters a url and then deletes it, the preview should disappear.

The way this is all configured, is that preview will update when the url is valid. However, an empty link: `""` is technically an "invalid" url. This is condition is weird, however in that we don't want to show this error to the user right away, and I assume they will also know if they delete the url, it is invalid. So I removed the error when the user deletes the link. The error _will_ still show when they try to submit.

This allows `""` to be considered valid as long as the user is not submitting the form, and the preview will disappear.

previewfix2

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
@@ -3,6 +3,7 @@ import 'd2l-tooltip/d2l-tooltip';
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import { css, html } from 'lit-element/lit-element.js';
 import { ActivityEditorMixin } from '../../mixins/d2l-activity-editor-mixin.js';
+import { AsyncContainerMixin } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { ContentEditorConstants } from '../constants';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { ErrorHandlingMixin } from '../../error-handling-mixin.js';
@@ -48,6 +49,7 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		super();
 		this._debounceJobs = {};
 		this.skeleton = true;
+		this.saveOrder = 3000;
 	}
 
 	render() {
@@ -103,8 +105,12 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 	}
 
 	validate() {
+		const link = this.shadowRoot.getElementById('content-link').value;
+		const isExternalResource = this.shadowRoot.getElementById('open-new-tab').checked;
+		const invalidWeblinkError = getWeblinkError(link, isExternalResource, true);
+
 		this.clearError('_linkError');
-		this._saveLink(null);
+		this.setError('_linkError', invalidWeblinkError, 'link-tooltip');
 	}
 
 	_renderLinkTooltip() {
@@ -125,11 +131,10 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		`;
 	}
 
-	_saveLink(inputEvent = null) {
-		const isSubmitting = inputEvent === null;
+	_saveLink() {
 		const link = this.shadowRoot.getElementById('content-link').value;
 		const isExternalResource = this.shadowRoot.getElementById('open-new-tab').checked;
-		const invalidWeblinkError = getWeblinkError(link, isExternalResource, isSubmitting);
+		const invalidWeblinkError = getWeblinkError(link, isExternalResource);
 
 		this._debounceJobs.link = Debouncer.debounce(
 			this._debounceJobs.link,
--- a/components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
@@ -3,7 +3,6 @@ import 'd2l-tooltip/d2l-tooltip';
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import { css, html } from 'lit-element/lit-element.js';
 import { ActivityEditorMixin } from '../../mixins/d2l-activity-editor-mixin.js';
-import { AsyncContainerMixin } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { ContentEditorConstants } from '../constants';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { ErrorHandlingMixin } from '../../error-handling-mixin.js';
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
```diff
@@ -125,10 +125,11 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		`;
 	}
 
-	_saveLink() {
+	_saveLink(inputEvent = null) {
```
ianlaidlaw, line 128: `inputEvent` will be present via: 
```
@change="${this._saveLink}"
```
sending as a param. we're not using it, but it's there.

brissons:
This isn't 100% related to this defect, but does the preview also show up when we have 'Open in a New Tab' selected/do we want the preview to show up? I always assumed that the preview was to replicate the embedded links.

ianlaidlaw:
> This isn't 100% related to this defect, but does the preview also show up when we have 'Open in a New Tab' selected/do we want the preview to show up? I always assumed that the preview was to replicate the embedded links.

that's a good question... i don't know. but i can ask. if that's the case, that should be an simple change

quinnbudan:
I'm not sure how your changes could have affected this, but the page used to show an error alert and focus the first invalid element when save was clicked and there were errors (see gif below from a quad site) and I noticed this isn't happening in your gif. This was coming from the `ActivityEditorContainerMixin` in the `_save` function, which calls `validate` on the editors. Is it possible that when `_save` calls `validate` the `inputEvent` is not null somehow and then it gets called again later with a null `inputEvent`?

weblink_error_popup

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
```diff
@@ -104,7 +104,7 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 
 	validate() {
 		this.clearError('_linkError');
-		this._saveLink();
+		this._saveLink(null);
```
ChrisKraljevicD2L, line 107: If `inputEvent` is by default `null` for `_saveLink`, why is it passed in explicitly here?

ianlaidlaw: no reason really except to be explicit. but i can remove it

ianlaidlaw:
> I'm not sure how your changes could have affected this, but the page used to show an error alert and focus the first invalid element when save was clicked and there were errors (see gif below from a quad site) and I noticed this isn't happening in your gif. This was coming from the `ActivityEditorContainerMixin` in the `_save` function, which calls `validate` on the editors. Is it possible that when `_save` calls `validate` the `inputEvent` is not null somehow and then it gets called again later with a null `inputEvent`?
> 
> weblink_error_popup

@quinnbudan so you were right and there was something very weird going on. Basically my changes helped expose a problem with how this editor was setup:

When you hit save, it will call the validate function on _all_ editors according to their sort order. This validate is supposed to happen asynchronously. However, because of how the `validate` on the weblink was originally setup, was that it would call into a function that used a setTimeout for the debounce time... and it was not async. So essentially the top level would think this editor is valid, and then 500ms later, the weblink validate function would essentially set it to false.

I reworked the `validate` function to not call into save, but instead I duplicated a small amount of code for checking and setting the error. This also let me not have to use the `inputEvent` as I can be confident that function is being called via the save button so I know that `isSubmitting` will be true _only_ there.

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
ianlaidlaw: no longer using this approach.

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
ianlaidlaw: i changed these functions slightly since i commented, but i removed an instance of passing a param when there is a default

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
```diff
@@ -48,6 +48,7 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		super();
 		this._debounceJobs = {};
 		this.skeleton = true;
+		this.saveOrder = 2000;
```
ianlaidlaw, line 51: this was missing so it got the default save order and was actually placed above `content-weblink-detail` `2000` puts them inline together

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
```diff
@@ -103,8 +104,12 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 	}
 
 	validate() {
+		const link = this.shadowRoot.getElementById('content-link').value;
+		const isExternalResource = this.shadowRoot.getElementById('open-new-tab').checked;
+		const invalidWeblinkError = getWeblinkError(link, isExternalResource, true);
+
 		this.clearError('_linkError');
-		this._saveLink();
```
ianlaidlaw, line 107: I explained in a comment below, but calling into this was causing some weird async behavior.

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
```diff
@@ -103,8 +104,12 @@ class ContentEditorLink extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
...
+		this.setError('_linkError', invalidWeblinkError, 'link-tooltip');
```
quinnbudan, line 112: So the third parameter here is for a tooltip ID that you want to force show, even if the user doesn't have focus or isn't hovering over the input. Are you sure we want that here? It may also re-introduce the tooltip positioning bug but I'm not sure, I'm checking out your branch to see now

ianlaidlaw: yeah I did experiment with the tooltip positioning bug, but I only ever encountered it when I deleted this line:
`this.clearError('_linkError');`
At first I thought it wasn't needed, but looks like that's what will cause it to be in the correct position, I guess after all the things get pushed down from the error showing up.

quinnbudan: Yeah that makes sense, I think it will work and make a little more sense if you don't pass in the third parameter since we don't need to force the tooltip to show due to that mixin refocusing the invalid input, but it also isn't an issue so this is a super nit for sure!

quinnbudan (dismissed):
LGTM!

components/d2l-activity-editor/d2l-activity-content-editor/web-link/d2l-activity-content-editor-link.js
ianlaidlaw: i'll give that a test!

quinnbudan (approved):

brissons (approved):
lgtm!

PR #1460, opened 2021-02-02
W2D: Back To Home context link

mbardakovD2L:
https://rally1.rallydev.com/#/357251704080ud/workviews?detail=%2Fuserstory%2F478938056424&view=b064b584-6337-43c1-b9a9-28047568284d

This is a really simple and naive implementation - we save the current URL to session storage just before we go into fullscreen view, and then return to the saved URL when the user clicks "back to home". If there isn't already a saved URL there, we default to `/d2l/home`


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -581,9 +582,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
-		// TODO: this is a default (and kind of a hacky way to get to it),
-		// ideally we want to get the user's homepage from their profile
-		this._homeLinkHref = window.location.href.substring(0, window.location.href.indexOf('/d2l/') + 5) + 'home';
+
+		if (this.fullscreen) {
+			const prevPage = sessionStorage.getItem('prevPage');
+			if (prevPage) {
+				this._homeLinkHref = prevPage;
+			} else {
+				this._homeLinkHref = window.location.href.substring(0, window.location.href.indexOf('/d2l/') + 5) + 'home';
```
mbardakovD2L, line 591: relative path should work here so we don't need to substring anything

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -581,9 +582,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	}
 
 	_getHomeHref() {
-		// TODO: this is a default (and kind of a hacky way to get to it),
-		// ideally we want to get the user's homepage from their profile
-		this._homeLinkHref = window.location.href.substring(0, window.location.href.indexOf('/d2l/') + 5) + 'home';
+
+		if (this.fullscreen) {
+			const prevPage = sessionStorage.getItem('prevPage');
+			if (prevPage) {
```
mbardakovD2L, line 588: this needs some validation - i.e. not only that `prevPage` exists, but also at least looks like a proper URL and leads to somewhere meaningful

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: update: "relative path" would be relative to `blah.brightspace.com/d2l/le/worktodo/` so it would be something like `../../home` which still makes assumptions I don't like making. I'm hoping there's a cleaner way to handle this but I can't think of one

JasperAtD2L: can you just set href to '/' and let the browser figure it out. And then brightspace should redirect them to /d2l/home

mbardakovD2L: that works - I'll update this

components/d2l-work-to-do/d2l-work-to-do.js
JasperAtD2L: yeah I would validate the domain of the prevPage against window.location.href if it's an absolute path.

mbardakovD2L:
@JasperAtD2L answered all my questions so this is ready to be reviewed for real

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -136,8 +136,14 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 		this._overdueActivities = [];
 		this._viewAllSource = undefined;
 		this._setEntityType(UserEntity);
+<<<<<<< HEAD
```
JasperAtD2L, line 139: need to fix your merge conflicts ;)

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -642,9 +648,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 	}
 
 	_getHomeHref() {
-		// TODO: this is a default (and kind of a hacky way to get to it),
-		// ideally we want to get the user's homepage from their profile
-		this._homeLinkHref = window.location.href.substring(0, window.location.href.indexOf('/d2l/') + 5) + 'home';
+
+		if (this.fullscreen) {
+			const prevPage = sessionStorage.getItem('prevPage');
```
JasperAtD2L, line 653: make prevPage a const and also add a namespace like `workToDo.prevPage` so it doesn't collide with other variables.

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: 😬 forgot to hit "save" so I guess I comitted... nothing

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -642,9 +642,17 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
...
+
+		if (this.fullscreen) {
+			const prevPage = sessionStorage.getItem(Constants.HomepageSessionStorageKey);
+			if (prevPage && (new URL(prevPage)).hostname === window.location.hostname) {
+				this._homeLinkHref = prevPage;
+			} else {
+				this._homeLinkHref = '/';
+			}
+		} else {
+			sessionStorage.setItem(Constants.HomepageSessionStorageKey, window.location.href);
```
JasperAtD2L, line 654: nit: this makes this function no longer a 'get' function.

JasperAtD2L (dismissed):

components/d2l-work-to-do/d2l-work-to-do.js
mbardakovD2L: can you elaborate? this was always `setItem`, with a key and a value - the only difference now is that the key is a constant?

mbardakovD2L: I'll update this to make it clearer - `_updateHomeHref` is more accurate

JasperAtD2L (approved):

PR #1461, opened 2021-02-02
fix: Update W2D upcomingCollectionCount in full page view

surags:
https://rally1.rallydev.com/#/357251704080ud/workviews?detail=%2Fuserstory%2F486933827748&view=b064b584-6337-43c1-b9a9-28047568284d&fdp=true?fdp=true

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do.js
+++ b/components/d2l-work-to-do/d2l-work-to-do.js
@@ -538,9 +538,9 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 						: [ this._loadLimitedUpcoming(emptyEntity), this._loadMaxUpcoming(emptyEntity) ]);
 
 				if (upcomingCollection) {
-					this._totalUpcomingActivities = upcomingCollection.hasProperty('pagingTotalResults') && upcomingCollection.properties.pagingTotalResults;
 					this._upcomingCollection = upcomingCollection;
 					this._upcomingActivities = upcomingCollection.getSubEntitiesByRel(Rels.Activities.userActivityUsage);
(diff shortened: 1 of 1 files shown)


components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -332,7 +333,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 
 			return html`
 				<div class="d2l-activity-collection-container-fullscreen">
-					<d2l-work-to-do-activity-list-header ?skeleton=${this._initialLoad} ?overdue=${isOverdue} count=${activities.length} fullscreen></d2l-work-to-do-activity-list-header>
+					<d2l-work-to-do-activity-list-header ?skeleton=${this._initialLoad} ?overdue=${isOverdue} count=${isOverdue ? activities.length : this._totalUpcomingActivities} fullscreen></d2l-work-to-do-activity-list-header>
```
mbardakovD2L, line 336: what happens if `pagingTotalResults` doesn't exist? is there a reasonable default we can fall back to?

JasperAtD2L: pass in a totalRecords here that we just pass to the count. Also this logic needs to be applied to the `collectionTemplate`

components/d2l-work-to-do/d2l-work-to-do.js
```diff
@@ -537,6 +538,7 @@ class WorkToDoWidget extends EntityMixinLit(LocalizeWorkToDoMixin(LitElement)) {
 						: [ this._loadLimitedUpcoming(emptyEntity), this._loadMaxUpcoming(emptyEntity) ]);
 
 				if (upcomingCollection) {
+					this._totalUpcomingActivities = upcomingCollection.hasProperty('pagingTotalResults') && upcomingCollection.properties.pagingTotalResults;
```
JasperAtD2L, line 541: default to length if pagingTotalResults is undefined or null.

JasperAtD2L (approved):

PR #1469, opened 2021-02-04
Jroach/fix activites translations

jeroach:
Updating all activities translations to use `js` files via the Serge `parser_plugin`

There is a lot going on here, so let me break it down into steps:
- Update the `activities.serge.json` file to use `js` instead of `json` for the affected components
- Remove the old `build/lang` folders
- Update the files in `/lang` folders to be `js` instead of `json`
- Update the specific localization files per component to point to the new files under `lang` instead of the old files under `build`


Scy-D2L (approved):
LGTM - nice bunch of cleanup, too. 😃

components/d2l-activity-card/build/lang/ar.js
```diff
@@ -1,5 +0,0 @@
-export const LangAr = {
-	'course': 'Course',
```
svanherk, line 2: Oh wow, these never even got updated with the translations.  🎉 for removing the build step that's easy to forget to do

components/d2l-activity-list-item/lang/fi.js
```diff
@@ -0,0 +1,5 @@
+/* eslint quotes: 0 */
+
+export default {
+  "enroll": "Enroll"
```
svanherk, line 4: Not sure why `fi` is missing `clickToViewActivity`... might be a good idea to add the English translation here for now.  I _think_ it falls back to English, but can't remember with all the different implementations of the translation stuff 😞

svanherk:
Took a quick look at OSLO... looks like everything you're touching _isn't_ in OSLO so we don't need to worry about how this affects the LMS langterms: https://search.d2l.dev/xref/lms/lp/framework/core/D2L/lang/Provider/BlockedCollections.cs?r=ed917294

svanherk:
Related to that, you can probably remove https://search.d2l.dev/xref/lms/lp/framework/core/D2L/lang/Provider/BlockedCollections.cs?r=ed917294#25, but I _doubt_ it'll break anything if you don't

svanherk (approved):
Looks good, thanks for doing this Josh!  Looks like this will fix a bunch of missing translations. 🎉   

This might re-trigger translations for all these files (I'm not sure if serge will recognize that the terms in the English file didn't change at all), but shouldn't be an issue even if it does.

svanherk:
Also, seems like you can close https://github.com/BrightspaceHypermediaComponents/activities/pull/611 and https://github.com/BrightspaceHypermediaComponents/activities/pull/612

svanherk (approved):

PR #1471, opened 2021-02-04
W2D: Change header divs to headers where appropriate

mbardakovD2L:
context: some of our headers were divs with `d2l-heading-X` classes applied to them, instead of actual `h1`/`h2`/`h3` elements, which meant that they looked like headers but didn't behave like them when being read out by a screen reader. This fix should remedy that without any changes to how they look visually.

https://rally1.rallydev.com/#/357251704080ud/workviews?Name=01.%20Update%20Widget%2FFull%20Page%20Headers%20to%20be%20accessible&Priority=Normal&Project=%2Fproject%2F357251704080&Requirement=%2Fhierarchicalrequirement%2F463435916636&detail=%2Fdefect%2F492428081372&iteration=u&parentRef=%2Fhierarchicalrequirement%2F463435916636&rankTo=BOTTOM&view=b064b584-6337-43c1-b9a9-28047568284d

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
@@ -139,10 +139,15 @@ class ActivityListHeader extends SkeletonMixin(LocalizeWorkToDoMixin(LitElement)
 			'd2l-heading-3': this.fullscreen,
 		};
 
-		const messageTemplate = html`
-			<div class=${classMap(messageClasses)}>
-				${this._message}
-			</div>`;
+		const messageTemplate = this.fullscreen
+			? html`
+				<h2 class=${classMap(messageClasses)}>
+					${this._message}
+				</h2>`
+			: html`
+				<h3 class=${classMap(messageClasses)}>
+					${this._message}
+				</h3>`;
 
 		const counterTemplate = html`
 			<div class=${classMap(counterContainerClasses)}>
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
@@ -217,7 +217,7 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 
 		const dateTemplate = this.includeDate
 			? html `
-				<div class=${classMap(dateClasses)}>
+				<h3 class=${classMap(dateClasses)}>
 					<d2l-activity-date
 						class="d2l-heading-4"
 						href=${this.href}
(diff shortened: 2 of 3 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js
```diff
@@ -217,7 +217,7 @@ class ActivityListItemDetailed extends ListItemLinkMixin(SkeletonMixin(EntityMix
 
 		const dateTemplate = this.includeDate
 			? html `
-				<div class=${classMap(dateClasses)}>
+				<h3 class=${classMap(dateClasses)}>
 					<d2l-activity-date
 						class="d2l-heading-4"
```
mbardakovD2L, line 222: note: there's a discrepancy here between the `h3` tag and the `d2l-heading-4` class - we want it _styled_ like the `heading-4` but we want the screen reader to still read off "heading level 3" because logically that's what's happening on the page (there's a level 1, 2, and 3) - if we skipped level 3 and went straight to 4, users might be confused when navigating because it would appear like they missed something

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
```diff
@@ -139,10 +139,15 @@ class ActivityListHeader extends SkeletonMixin(LocalizeWorkToDoMixin(LitElement)
...
-			</div>`;
+		const messageTemplate = this.fullscreen
+			? html`
+				<h2 class=${classMap(messageClasses)}>
+					${this._message}
+				</h2>`
+			: html`
+				<h3 class=${classMap(messageClasses)}>
+					${this._message}
+				</h3>`;
```
bearfriend, line 150: I'm not at all convinced this is actually better. I'll let you decide.

```javascript
const headingLevel = this.fullscreen ? 'h2' : 'h3';
const messageTemplate = html`
	<${headingLevel} class=${classMap(messageClasses)}>
		${this._message}
	</${headingLevel}>`;
```

bearfriend (approved):

bearfriend (approved):

bearfriend (approved):

components/d2l-work-to-do/d2l-work-to-do-activity-list-header.js
mbardakovD2L: I like that more, I'll update this

mbardakovD2L: something with `classMap` breaks when we do it this way: `Uncaught (in promise) Error: The `classMap` directive must be used in the `class` attribute and must be the only part in the attribute.` - I tried moving the `h` outside `headingLevel` and made it part of `messageTemplate` which came up with a different error. I'm sure there's a way to get this working but I don't want to spend too much more time on it so I'll just merge what's already approved

bearfriend: We probably could do `<div role="heading" aria-level="${headingLevel}"...>` but that just seems silly.

PR #1472, opened 2021-02-04
Us123398/add ip restriction component 2

ronan-f:
Added part of https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F465115111280

This is using the D2L Polymer table which maybe isn't ideal but the plan is to replace it at some point when the table gets ported over to lit. This is still pretty rough but wanted to get feedback on the logic. 

Right now there is no working copy support and I'm saving all the IP restrictions when the user presses "add". 

I think for delete we can fire off network requests as the user presses the "delete" button.

The table is rendered by mapping over all of the IP restrictions in the mobx state and generating a row for each. I'm keeping track of the index of each row within the table and updating the specific row in mobx state when it changes.

To determine whether or not we add a new range or update an existing one the API accepts an "oldStart" and "oldEnd" value. If these are provided then it deletes the old range and adds the new one so effectively an update.

Siren-sdk PR which enables this behaviour:  https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/265

Edit: I don't think CI will pass until siren-sdk PR has been merged.


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -0,0 +1,108 @@
...
+		return html`
+			<d2l-activity-quiz-ip-restrictions-container
+				href="${this.ipRestrictionsHref}"
+				.token="${this.token}"
+				@restrictions-resize-dialog="${this._resizeDialog}">
+			</d2l-activity-quiz-ip-restrictions-container>
+		`;
+	}
+
+	_resizeDialog() {
```
ronan-f, line 98: Patrick ideally wants the dialogs to resize as users add more IP restrictions

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -0,0 +1,142 @@
...
+		const entity = store.get(this.href);
+		if (!entity || !entity.ipRestrictions || !entity.ipRestrictions.length) {
+			return html``;
+		}
+
+		return entity.ipRestrictions.map((restriction, index) => {
+			const { start, end } = restriction;
+
+			return html`
+				<d2l-tr rowindex=${index}>
```
ronan-f, line 105: I don't like having to add the index manually but it wasn't available natively from my experimentation.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,105 @@
...
+
+		await Promise.all(promises);
+	}
+
+	setIpRestriction(index, key, val) {
+		const currentVal = this.ipRestrictions[index];
+		this.ipRestrictions[index] = { ...currentVal, [key]: val };
+	}
+
+	_filterOldRestrictions() {
```
ronan-f, line 64: This is to limit network requests and only update IP ranges which have changed.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,105 @@
...
+	async saveRestrictions() {
+		const restrictionsToSave = this._filterOldRestrictions();
+
+		if (!restrictionsToSave) {
+			return;
+		}
+
+		const promises = restrictionsToSave.map(restriction => {
+			this._entity.addIpRestriction(restriction);
+		});
```
ronan-f, line 54: Currently not handling any potential errors here. From my understanding of Promise.all it's an all or nothing approach so it would likely only catch a single error? Might be a better approach here.

ronan-f: And potentially when working copy support is added the saves will be fired off based on a different event...maybe when focus is removed from the input element?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -0,0 +1,142 @@
...
+					<d2l-th>
+						<d2l-input-text @input="${this._handleChange}" value="${start}" name="start"></d2l-input-text>
```
ronan-f, line 107: Not sure of the best way to distinguish between which values are the start of an IP range vs which are the end. In this approach I gave each input a name attribute and used that to decide what part of the range it represents.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
mdgbayly: Yeh, might become moot if we decide to make API calls immediately with working copy. Alternatively, could use Promise.allSettled.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,105 @@
...
+		this._saving = null;
+		this.ipRestrictions = [];
+	}
+
+	addRestriction() {
+		this.ipRestrictions.push({ start: '', end: '' });
+	}
+
+	deleteIpRestriction(start, end) {
+		this._entity.deleteIpRestriction(start, end);
```
mdgbayly, line 20: does deletion need to remove it from the MobX state too?

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
mdgbayly: Does this work effectively if restrictions have been deleted? As then the MobX state index might not match the original entity index?

Maybe using the idea of the Siren entities having an index could be used to compare the MobX state with the original siren entity state. That way you wouldn't need to be using the `old` properties in the comparison below. You could compare the MobX values against the original values in the siren entity to see if they have changed using the id stored in the MobX state to find the correct siren state.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,131 @@
...
+
+	deleteIpRestriction(index) {
+		if (this.ipRestrictions.length === 1) {
+			return;
+		}
+
+		const restriction = this.ipRestrictions.splice(index, 1);
+		const isNew = restriction && restriction[0].id === undefined;
+
+		if (!isNew) {
```
ronan-f, line 27: If an IP restriction hasn't already been persisted then we only need to delete it from MobX store

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
ronan-f: Made use of `Promise.allSettled` to handle any rejected promises although need to confirm with Patrick how these should be handled

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,131 @@
...
+	addRestriction() {
+		this.ipRestrictions.push({ start: '', end: '' });
+	}
+
+	deleteIpRestriction(index) {
+		const restriction = this.ipRestrictions.splice(index, 1);
+		const isNew = restriction && restriction[0].id === undefined;
+
+		if (this.ipRestrictions.length === 0) {
+			this.ipRestrictions.push({ start: '', end: '' });
```
ronan-f, line 24: As per Patrick if the user deletes the final restriction then we should leave them an empty input

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -0,0 +1,152 @@
...
+		`;
+	}
+
+	_renderRestrictionRows() {
+		const entity = store.get(this.href);
+		if (!entity || !entity.ipRestrictions || !entity.ipRestrictions.length) {
+			return html``;
+		}
+
+		const deleteIp = this._deleteIpRestriction.bind(this);
```
mdgbayly, line 109: nit: Could these be bound and stored as class variables. Otherwise I assume it's creating a new function on each render?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -0,0 +1,152 @@
...
+					:host {
+						display: block;
+					}
+				</style>
+			</custom-style>
+
+			<d2l-table id="ip-restrictions-table">
+			 	<d2l-thead>
+			 		<d2l-tr>
+			 			<d2l-th class="title">IP Range Start</d2l-th>
```
mdgbayly, line 91: Will need localization at some point obviously.

ronan-f: Yip I've got another PR which cleans stuff up and adds localization etc almost ready.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -0,0 +1,131 @@
...
+
+		for (const restriction of this.ipRestrictions) {
+
+			const { start, end, id } = restriction || {};
+
+			if (!start || !end) {
+				continue;
+			}
+
+			const expectedRestriction = expectedRestrictions[id];
```
mdgbayly, line 105: nit: I guess this is making an assumption that the id we are getting back is an index into the list of restrictions in the entity. Maybe we should be finding by id instead?

mdgbayly (approved):
looks good - just the minor nits:

me-louie (approved):

PR #1476, opened 2021-02-06
US123706 added Attempt Conditions to dialog

me-louie:
- Adding basic UI for the Attempts Conditions editor, some todos remain for info to be fetched from Attempts entity
- Edited `handleClose` method in `Dialog Mixin` to prevent parent dialog from being closed when child dialog is closed
- Renamed some methods to include 'Editor'
- Double checked that everything also looks good in RTL

[US123706](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F478977336380%2Ftasks&fdp=true?fdp=true): [ui] Attempts: Conditions entities

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F478977336380%2Ftasks&fdp=true?fdp=true


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
```diff
@@ -52,12 +52,15 @@ class ActivityQuizRetakeIncorrectOnlyEditor
 			${this._renderDialog()}
 		`;
 	}
-
+	_handleClose(e) {
+		this.opened = false;
+		e.stopPropagation();
```
emil-furniss, line 57: Is there something special about this dialog that we need to stopPropagation, when other dialogs are using the default mixin behavior? Should we consider adding this to the mixin?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin';
+import { LocalizeActivityQuizEditorMixin } from './mixins/d2l-activity-quiz-lang-mixin.js';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles';
+import { shared as store } from './state/quiz-store';
+
+class ActivityQuizAttemptConditionsEditor extends ActivityEditorMixin(LocalizeActivityQuizEditorMixin(MobxLitElement)) {
+	static get properties() {
+		return {
+			isAttemptConditionsOpen: { attribute: false }
```
emil-furniss, line 12: For completeness I'd consider adding `type: Boolean` here and then initializing the value in the constructor with `this.isAttemptConditionsOpen = false;`.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+				}
+				.d2l-attempts-conditions-range-editor {
+					align-items: center;
+					border-bottom: 1px solid var(--d2l-color-corundum);
+					display: flex;
+					padding-bottom: 0.8rem;
+					padding-top: 0.5rem;
+				}
+				.d2l-input-number-text {
+					margin-left: 0.4rem;
```
emil-furniss, line 40: rtl?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+		// TODO: replace const with data fetched from attempts entity when #attempts allowed >=2
+		const attemptsConditions = [1, 2, 3, 4, 5];
+		return html`
+			${attemptsConditions.map((index) => html`
+				<div class="d2l-attempts-conditions-range-editor">
+					<div>
+						<div class="d2l-label-text">${this.localize('attemptConditionsRangePrefixText1')} ${index}:</div>
+						<div class="d2l-body-small">${this.localize('attemptConditionsRangePrefixText2')}</div>
+					</div>
+					<d2l-input-number label=${this.localize('minLabel')} min=0 max=100 input-width=3.5rem>
```
emil-furniss, line 86: We might get the min/max from the entity as well? Not sure yet since it's probably never going to change, just something to keep in mind.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+						<div slot="after">
+							<span class="d2l-label-text d2l-input-number-text">%</span>
```
emil-furniss, line 88: Do we need to localize the % sign and its placement?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+					</d2l-input-number>
+					<div class="d2l-body-small d2l-input-range-text">${this.localize('andRangeText')}</div>
+					<d2l-input-number label=${this.localize('maxLabel')} min=0 max=100 input-width=3.5rem>
+						<div slot="after">
+							<span class="d2l-label-text d2l-input-number-text">%</span>
+						</div>
+					</d2l-input-number>
+					<div>
+						<div class="d2l-body-small">${this.localize('attemptConditionsRangeSuffixText')}</div>
+						<div class="d2l-label-text">Attempt ${index + 1}</div>
```
emil-furniss, line 99: This should be localized

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+		`;
+	}
+	_renderAttemptConditionsRangeEditor() {
+		// TODO: replace const with data fetched from attempts entity when #attempts allowed >=2
+		const attemptsConditions = [1, 2, 3, 4, 5];
+		return html`
+			${attemptsConditions.map((index) => html`
+				<div class="d2l-attempts-conditions-range-editor">
+					<div>
+						<div class="d2l-label-text">${this.localize('attemptConditionsRangePrefixText1')} ${index}:</div>
```
emil-furniss, line 83: Just to keep in mind for later: Rather than using `index`, it's probably safer to replace this with the attempt number that is provided by the entity (when we get it).

emil-furniss: Oh and we shouldn't just append the index to the string like that, it should be part of the langterm in case other langs have a different ordering or use a different colon char.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
```diff
@@ -0,0 +1,110 @@
...
+						<div class="d2l-body-small">${this.localize('attemptConditionsRangePrefixText2')}</div>
+					</div>
+					<d2l-input-number label=${this.localize('minLabel')} min=0 max=100 input-width=3.5rem>
+						<div slot="after">
+							<span class="d2l-label-text d2l-input-number-text">%</span>
+						</div>
+					</d2l-input-number>
+					<div class="d2l-body-small d2l-input-range-text">${this.localize('andRangeText')}</div>
```
emil-furniss, line 91: Not sure if this is an issue: we're basically forcing the sentence structure `[ input ]% and [ input ]%` through the UI, which might be an issue for some langs? @mdulat @wongvincent Thoughts?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
me-louie: image
I noticed that when I have nested dialogs like this, and I close the RIO dialog, the Manage Attempts dialog closes too unless I stop the event propagation. We might want to consider adding this to the mixin. I looked through other examples in the code and haven't yet see any instances of nested dialogs like the one we have here.. do you know of any?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
me-louie: Thanks, I wasn't sure if symbols like % were translated but I can localize it and test out rtl

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
emil-furniss: Ah ok, that makes sense. I think adding it to the mixin should be fine then.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
me-louie: I can confirm that the `after` slot handles rtl :)

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
me-louie: Yup, the `index` is just temporary until we fetch the attempt number from the entity. I can add a comment so it is explicit. Thanks for the info on lang term parametrization.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
mdulat: Yeah that's a tricky one. Ideally the entire sentence would be a lang term, but then the lang term would include the input controls. I've never tried putting more html into a lang other than simple <bold> tags or the like. Maybe it would work?

mdulat: Might be worth asking Patrick and Joseph their thoughts on this.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
dbatiste: I think this is a bug in the dialog mixin.  The dialog mixin handles this scenario for non-Chrome browsers, but not for Chrome.  I think it also handles it for the Promise-based usage.  This is due to `DialogMixin` supporting both native `dialog` element as well as our own custom implementation for browsers that do not support - basically this was missed in the Chrome case during early testing.

It's worth mentioning that nested dialogs are sorta boss fight 2 for dialogs.  Lol.  The component was developed with nested dialogs in mind, however it's not "advertised" as supported because a) we typically want to avoid nested dialogs from a design perspective; and b) there were some issues that we encountered, either for legacy Edge or IE11 during development.  You may run into some issues since this usage has not been extensively tested.

emil-furniss: Thanks for clarifying. We'll just keep this in here for now and can clean it up if/when the `DialogMixin` issue is resolved.

We'll also clarify with design whether we _really_ want a nested dialog before worrying too much about it ^^.

test/d2l-activity-editor/mixins/d2l-activity-editor-dialog-mixin.js
```diff
@@ -43,9 +43,10 @@ describe('d2l-activity-editor-dialog-mixin', function() {
 
 	it('handles closing a dialog', async() => {
 		const el = await fixture(`<${editor}></${editor}>`);
+		const event = { preventDefault: () => {} };
```
ronan-f, line 46: Will need to be:

```js 
const event = { stopPropagation: () => {} };
```

as that's what `handleClose` will be looking for

me-louie: Thanks 😅

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
me-louie: Design confirmed that we're doing the nested dialog for now.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-attempt-conditions-editor.js
me-louie: Thanks for bringing this up Margaret and Emil! Patrick and Joseph are both unsure whether this will be an issue at this point but said we can go forward and adjust later if it proves to be a problem.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-retake-incorrect-only-editor.js
dbatiste: Sounds good.

emil-furniss (approved):

ronan-f (dismissed):

ronan-f (approved):

PR #1478, opened 2021-02-08
US123692 - Implement Working Copy for Quiz Timing

wongvincent:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478606416460

- Call Quiz Checkout and Quiz Checkin.
- Use `fixedCheckoutHref` to provide a base working copy for checkouts off the base quiz entity. Further checkouts by components will always be checked out from this working copy href.
- Check out again whenever dialog is re-opened.
- Load timing component based off the timing endpoint of the working copy.
- Timing container component uses working copy independent of other components. It does not depend on other components checking out the initial working copy.

- Includes a hard-coded dialog height.


wongvincent:
Don't merge until siren-sdk pr is merged

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-timing-and-display-editor.js
```diff
@@ -50,12 +50,10 @@ class ActivityQuizTimingAndDisplayEditor extends AsyncContainerMixin(LocalizeAct
...
 	render() {
 		const entity = store.get(this.href);
-		if (!entity) {
-			return html``;
-		}
+		if (!entity) return html``;
 
 		const {
-			timingHref
+			checkoutHref
```
wongvincent, line 56: This will be changed when we update the `checkout` action from being a GET to a POST.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -63,6 +96,7 @@ export class Quiz {
 		this.headerIsDisplayed = entity.headerIsDisplayed();
 		this.headerRichTextEditorConfig = entity.headerRichTextEditorConfig();
 		this.ipRestrictionsHref = entity.ipRestrictionsHref();
+		this.checkoutHref = entity.getCheckoutHref() || this.href;
```
wongvincent, line 99: This will be removed when we update the `checkout` action from being a GET to a POST.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-dialog.js
```diff
@@ -0,0 +1,80 @@
...
+import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin';
+import { LocalizeActivityQuizEditorMixin } from './mixins/d2l-activity-quiz-lang-mixin.js';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { shared as store } from './state/quiz-store';
+
+class ActivityQuizManageTimingDialog extends ActivityEditorMixin(ActivityEditorDialogMixin(LocalizeActivityQuizEditorMixin(MobxLitElement))) {
+	static get styles() {
+		return [
+			css`
+				#manage-timing-dialog-timing-editor {
```
wongvincent, line 14: will need to look into this further later

wongvincent:
do not merge

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
```diff
@@ -26,35 +35,40 @@ class ActivityQuizManageTimingContainer extends ActivityEditorMixin(ActivityEdit
...
+		this.checkedOutHref = '';
+	}
+
+	async _openDialog(e) {
+		const entity = store.get(this.href);
+		if (!entity) return;
+
+		this.checkedOutHref = await entity.checkout(store);
+
+		this.shadowRoot.querySelector('d2l-activity-quiz-manage-timing-dialog').openDialog(e);
```
emil-furniss, line 54: Maybe nicer to do this with a prop that is toggled here and then set on `d2l-activity-quiz-manage-timing-dialog`?

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -11,6 +11,39 @@ export class Quiz {
 		this._saving = null;
 	}
 
+	async checkin(quizStore) {
```
emil-furniss, line 14: :tada:

emil-furniss (dismissed):

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
```diff
@@ -26,35 +35,40 @@ class ActivityQuizManageTimingContainer extends ActivityEditorMixin(ActivityEdit
...
+
+	async _closeDialog() {
+		this.checkedOutHref = '';
+	}
+
+	async _openDialog(e) {
+		const entity = store.get(this.href);
+		if (!entity) return;
+
+		this.checkedOutHref = await entity.checkout(store);
```
mdgbayly, line 52: So this checkout is effectively the fork? Where does the original checkout happen? Or is that not included yet?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-dialog.js
```diff
@@ -0,0 +1,80 @@
...
+
+	async openDialog(e) {
+		this.open(e);
+	}
+
+	async _closeDialog(e) {
+		const entity = store.get(this.href);
+		if (!entity) return;
+		if (e.detail.action === 'ok') {
+			await entity.checkin(store);
```
mdgbayly, line 39: This checkin equates to the 'merge'? I wonder if it would be clearer if the checkout/checkin operations were in the same component rather than being split between the container and the dialog?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
```diff
@@ -26,35 +35,40 @@ class ActivityQuizManageTimingContainer extends ActivityEditorMixin(ActivityEdit
     	`;
 	}
 
+	async save() {
+		const entity = store.get(this.href);
+		if (!entity) return;
+		await entity.checkin(store);
```
mdgbayly, line 41: This checkin equates to the final checkin and is called when the main editor Save button is clicked? So the idea would be to add this to every component that was participating in working copy, so that that any of them can be used independently? And the `checkin` implementation will de-dupe multilple calls, the same as we do for our other final saves?

That seems to make sense. 
I wonder how we can achieve something similar for the initial checkout. That was what was nice about having checkout be a GET originally. We could allow multiple components to do that inital checkout and they would all end up with the same checkout. Maybe we have to encapsulate that in the main MobX quiz entity? e.g. allow any UI component to call checkout on the main quiz href, and the quiz MobX entity will de-dupe those checkout calls?

It will mean that each component will have multiple checkout/checkin calls. Once on the main href and once on the fork?

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-dialog.js
wongvincent: Yes, I agree, I will combine them back into one file.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
wongvincent: Combining into one file, so I'll just use the dialog mixin.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
wongvincent: Yes, called when main editor button Save is clicked.
Yes, will need to add to attempts, IP, other components using working copy.

Yes, I am still unsure of how to do the initial checkout. Part of the reason why it's not in this PR.

Yes, it would be correct to say each checkout of the main quiz entity will result in different working copy hrefs. Unsure how to get multiple components to checkout to use the same href, will chat with you on slack about this.

wongvincent:
Convert back to draft. I'm going to have this PR use the POST action for the initial checkout.
Don't merge until this is merged: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/266

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -33,13 +33,26 @@ export class Quiz {
...
+			quizStore.put(href, entity);
+		} else {
+			href = this.href;
+		}
 
-		const href = sirenEntity.self();
-		const entity = new Quiz(href, this.token);
-		entity.load(sirenEntity);
-		quizStore.put(href, entity);
+		if (this._entity.isBaseQuiz()) {
```
wongvincent, line 51:
```
isBaseQuiz() {
	if (this._canCheckin()) {
		return false;
	}
	return true;
}
```

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
```diff
@@ -42,14 +41,23 @@ class ActivityQuizManageTimingContainer extends ActivityEditorMixin(ActivityEdit
 	}
 
 	async save() {
-		const entity = store.get(this.href);
-		if (!entity) return;
-		await entity.checkin(store);
+		const quiz = store.get(this.href);
+		if (!quiz) return html``;
```
mdgbayly, line 45: nit: should just return rather than return a template.

components/d2l-activity-editor/mixins/d2l-activity-editor-mixin.js
```diff
@@ -65,7 +68,12 @@ export const ActivityEditorMixin = superclass => class extends superclass {
 
 		if ((changedProperties.has('href') || changedProperties.has('token')) &&
 			this.href && this.token) {
-			this.store && this._fetch(() => this.store.fetch(this.href, this.token));
+			this.store && this._fetch(async() => {
+				const entity = await this.store.fetch(this.href, this.token);
+				if (this.checkoutOnLoad) {
+					return entity && entity.checkout && entity.checkout(this.store);
```
mdgbayly, line 74: Possibly we could store the self link of the checked out entity as a class property here to avoid having to manage that in the mobx state as 'fixedCheckoutHref`. I wasn't really clear on the significance of the name `fixed` in that context?

mdgbayly (dismissed):
I think this works. Looking at it again now, I'm wondering if it would be clearer if your ` ActivityQuizTimingAndDisplayEditor` was the component that handled the checkout of the original quiz rather than the container component. The container component then just takes an href which it also checks out on fetch.

The `save` implementation would move to `ActivityQuizTimingAndDisplayEditor` and the container would then just be responsible for the forking.

This is just tweaking and doesn't change things much - but might help to keep the distinctions between what is checking out/in a bit clearer.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -11,6 +11,50 @@ export class Quiz {
...
+
+		if (this._saving) {
+			return this._saving;
+		}
+
+		this._saving = this._entity.checkin();
+		const sirenEntity = await this._saving;
+		if (!sirenEntity) return;
+
+		const href = sirenEntity.self();
```
mdgbayly, line 27: nit: I wonder if we should also delete the working copy reference that is being checked in from the store too, otherwise if we open dialogs multiple times, we will end up with a lot of orphaned working copies just sitting there?

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -11,6 +11,50 @@ export class Quiz {
...
+		const entity = new Quiz(href, this.token);
+		entity.load(sirenEntity);
+		quizStore.put(href, entity);
+
+		this._saving = null;
+	}
+
+	async checkout(quizStore, forceCheckout) {
+		if (!forceCheckout && this.checkedOutHref) {
```
mdgbayly, line 36: This is looking much better, but I still think there is the possibility of a race condition here?
E.g. if the timing container and the IP restrictions container both try to checkout the base quiz concurrently?
They will both enter the checkout function and as the `checkedOutHref` isn't set until after the call to `await this._entity.checkout()` they could both end up checking out the entity and adding it to the store.

I guess maybe with the way you had it before where the `fixedCheckedOutHref` was a MobX property it might have worked out in the end as the last checked out entity would have won, and as components were reacting to the props of the MobX state they would have both ended up with the last one winning.

But it still seems wrong to allow this. And I think now we've added the checkedOutHref to the mixin, it will cause the components that checkout the base concurrently to end up referencing different checkouts?

So I think it might need to use a pattern more like what we use on save where we use the promise from the API call as the flag.

https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js#L77

I think this is a bit more complex though because we still want to allow the non-forced checkouts to resolve with the href of the checkout from the base entity. So that probably means using some explicit promise handling in this method. Started trying to write out what that might look like but my brain isn't aligned to promise shenanigans right now, so will take a look first thing tomorrow.

mdgbayly: Brain wouldn't stop thinking about it....

Maybe something like this:

```
if (!forceCheckout && this._checkedOut) {
    return this._checkedOut;
}	

let href = this.href;

const p = new Promise((resolve, reject) => {
	const sirenEntity = await this._entity.checkout();
	if (sirenEntity) {
		href = sirenEntity.self();
		const entity = new Quiz(href, this.token);
		entity.load(sirenEntity);
		quizStore.put(href, entity);
	}
    resolve(href);	
})

runInAction(() => {
	if (!forcedCheckout) {
		this._checkedOut = p;
	}
});

return p;
```

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
```diff
@@ -32,27 +33,31 @@ export class Quiz {
...
-			href = sirenEntity.self();
-			const entity = new Quiz(href, this.token);
-			entity.load(sirenEntity);
-			quizStore.put(href, entity);
-		} else {
-			href = this.href;
-		}
+
+		const p = new Promise(async(resolve) => {
+			const sirenEntity = await this._entity.checkout();
```
mdgbayly, line 44: I assume it didn't like having the await in the executor function without async? Maybe we just have to use promises.

Also occurred to me that we probably don't need to use the `runInAction` below, as I don't think we need to expose the promise `this._checkedOut` as an observable property.

mdgbayly: or maybe there is just a plain out better way to write this. The main thing is we need to capture the promise that returns the href.

mdgbayly: Maybe something like this:

```
if (!forcedCheckout && this._checkedOut) {
	return this._checkedOut;
}	

let href = this.href;

const p = this._entity.checkout().then(function(sirenEntity) {
	if (sirenEntity) {
		href = sirenEntity.self();
		const entity = new Quiz(href, this.token);
		entity.load(sirenEntity);
		quizStore.put(href, entity);
	}
	return href;	
}, function() {
	return href;
});

if (!forcedCheckout) {
	this._checkingOut = p;
}

return p;
```

dbatiste: I kinda don't like this linting error.  I've run into it before as well and while I think it catches some mis-use I think there might still be legit cases for wanting an `async` function for a promise.  I think this case could be rewritten as:

```javascript
const getHrePromise = async() => {
   const sirenEntity = await this._entity.checkout();
   ...
   return href;
};

runInAction(() => {
   if (!forcedCheckout) {
      this._checkedOut = getHrefPromise;
   }
});
```

mdgbayly (approved):
LGTM

PR #1490, opened 2021-02-09
Attempt to Fix visual diff flake

svanherk:
Most of these are test changes, but there is also a production change - the `d2l-quick-eval-activity-card` will now support `prefers-reduced-motion`.  Let me know if there are issues with this, but all our core components support this and it has the benefit of respecting the setting, plus making visual-diff faster and less flakey (hopefully).


components/d2l-quick-eval/activity-card/d2l-quick-eval-activity-card.js
```diff
@@ -130,6 +130,11 @@ class D2LQuickEvalActivityCard extends QuickEvalLocalize(PolymerElement) {
 					d2l-quick-eval-activity-card-items {
 						transition: transform 200ms ease-out, opacity 200ms ease-out;
 					}
+					@media (prefers-reduced-motion: reduce) {
```
svanherk, line 133: If users have this option enabled for their OS, the transition will be turned off and the card flip will happen instantly.

test/d2l-quick-eval/perceptual/d2l-quick-eval-action-dismiss-dialog.visual-diff.js
```diff
@@ -9,8 +9,7 @@ describe.skip('d2l-quick-eval-action-dismiss-dialog', function() {
 
 	before(async() => {
 		browser = await puppeteer.launch();
-		page = await browser.newPage();
-		await page.setViewport({ width: 900, height: 800, deviceScaleFactor: 2 });
+		page = await visualDiff.createPage(browser, { viewport: { width: 900, height: 800 } });
```
svanherk, line 12: Using the `createPage` method means the tests will now run in a mode where `prefers-reduced-motion` is enabled.  I've updated every test to use this, since it will also disable dropdown/tooltip/other core component animations.

test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card.visual-diff.js
```diff
@@ -35,14 +34,12 @@ describe('d2l-quick-eval-activity-card', function() {
 	it ('dismiss', async function() {
 		// The differences between this and the "regular" card is only in the buttons which only appear on hover.
 		page.hover('#dismiss d2l-quick-eval-activity-card');
-		await page.waitFor(2000);
```
svanherk, line 38: And because we've disabled transitions/animations, we can remove all of these waits.

test/d2l-quick-eval/perceptual/d2l-quick-eval-activity-card-new-submissions.visual-diff.js
```diff
@@ -9,8 +9,7 @@ describe('d2l-quick-eval-activity-card-new-submissions', function() {
 
 	before(async() => {
 		browser = await puppeteer.launch();
-		page = await browser.newPage();
-		await page.setViewport({ width: 800, height: 800, deviceScaleFactor: 2 });
+		page = await visualDiff.createPage(browser);
```
svanherk, line 12: 800/800 is the default, so no viewport option specified here

jeroach (approved):
LGTM

PR #1498, opened 2021-02-09
Cleaned up ip restrictions

ronan-f:

Cleaned up a few things including adding lang terms/styling etc.


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -86,6 +108,29 @@ class ActivityQuizIpRestrictionEditor
...
+		if (!entity) {
+			return;
+		}
+
+		const { errors } = entity || {};
+
+		if (!errors || !errors.length) {
+			return;
+		}
+		// TODO: add some error handling
```
ronan-f, line 122: Waiting for feedback from design to decide how to handle potential errors

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -39,17 +54,19 @@ class ActivityQuizIpRestrictionsContainer extends ActivityEditorMixin(LocalizeAc
...
 
-	_deleteIpRestriction(e, index) {
+	_deleteIpRestriction(_, index) {
 		const entity = store.get(this.href);
 		if (!entity) {
 			return;
 		}
 
+		if (entity.ipRestrictions.length > 1) {
+			this._sendResizeEvent();
```
ronan-f, line 66: Patrick did opt for not resizing dialogs although if we don't add this on delete it leaves a bunch of whitespace at the bottom of the dialog.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -66,8 +75,16 @@ export class QuizIpRestrictions {
 		});
 
 		if (errors && errors.length) {
-			//TODO: handle errors
+			this.setErrors(errors);
```
ronan-f, line 78: Same as above. Awaiting design for error handling

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -66,8 +75,16 @@ export class QuizIpRestrictions {
...
+
+			return;
 		}
+
+		await this.fetch();
```
ronan-f, line 83: This is needed as we use IDs so that MobX state knows which IP address maps to siren SDK entities. The IDs are created by siren-sdk when loading the entity. In this case, when we add new entities we need siren sdk to update so that we can correctly assign IDs.

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-ipRestrictions.js
```diff
@@ -98,7 +115,7 @@ export class QuizIpRestrictions {
 
 			const { start, end, id } = restriction || {};
 
-			if (!start || !end) {
+			if (!start) {
```
ronan-f, line 118: I was previously continuing if there was no end however the correct behaviour is to set both start/end equal to the value of start which is handled server side

ronan-f, line 118: I was previously continuing if there was no end however the correct behaviour is to set both start/end equal to the value of start which is handled server side

mdulat (approved):

PR #1506, opened 2021-02-10
Set work-to-do-activity-list-item-basic to always use submission-icon in quick-eval widget

Scy-D2L:
Fixed a problem where items with no submissions were using the regular icon component instead of the submission icon component, causing misalignment in the list. 

Before:
image

After:
image

### DIFF
--- a/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
+++ b/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
@@ -11,6 +11,7 @@ import { ActivityAllowList, HideOrgInfoClasses } from './env';
 import { classMap } from 'lit-html/directives/class-map';
 import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit';
 import { fetchEntity } from './state/fetch-entity';
+import { ifDefined } from 'lit-html/directives/if-defined';
 import { ListItemLinkMixin } from '@brightspace-ui/core/components/list/list-item-link-mixin';
 import { LocalizeWorkToDoMixin } from './mixins/d2l-work-to-do-localization-mixin';
 import { nothing } from 'lit-html';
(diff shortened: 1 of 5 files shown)


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -185,11 +186,11 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 			: nothing;
 
 		return this._renderListItem({
-			illustration: this.submissionCount ? html`
+			illustration: this.evaluateAllHref ? html`
 					<d2l-quick-eval-widget-submission-icon style="overflow: visible;"
 						class="${classMap(iconClasses)}"
 						icon=${this._icon}
-						submission-count=${this.submissionCount > 99 ? '99+' : this.submissionCount}>
+						submission-count=${ifDefined(this.submissionCount > 0 ? (this.submissionCount > 99 ? '99+' : this.submissionCount) : undefined)} >
```
Scy-D2L, line 193: I recognize that this could be cleaner without ifDefined if we just checked for a 0 inside of the submission-icon component, but the submission-icon is a placeholder until we have a ui-core equivalent and it isn't expected to be making this specific check.

demo/d2l-quick-eval-widget/data/evaluation-status-quiz.json
```diff
@@ -4,8 +4,8 @@
     "evaluation-status"
   ],
   "properties": {
-    "newsubmissions": 8,
-    "resubmissions": 4
+    "newsubmissions": 0,
```
jeroach, line 7: do we still want a demo example with non-zero counts?

Scy-D2L: The screenshot above is from the demo, it has three entries and this is the only 0. Since the icon type logic is separate from the badge/no badge logic I think it's probably fine if the only quiz example has a zero-count.

jeroach (approved):

PR #1509, opened 2021-02-11
Us123398/cross browser styling

ronan-f:

As part of this story I made use of the Polymer [d2l-table component ](https://github.com/BrightspaceUI/table) by adding it to Lit Element components. Testing on Chrome based browsers the styling applied correctly however when I tested on firefox/safari styles were not being picked up at all.  

To get styling to work on Chrome I had added this style tag when rendering the table:

```
<custom-style>
    <style include="d2l-table-style">
    </style>
</custom-style>
```

The "fix" I added which works locally on firefox was to copy all of the styles from `d2l-table-style` and add them as part of LitElement CSS.

Is there a better solution for this? Am I missing a polyfill or something that's preventing other browsers from picking up this styling?

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
@@ -161,7 +161,7 @@ class ActivityQuizIpRestrictionEditor
 
 		await entity.saveRestrictions();
 
-		if (!entity.errors && !entity.errors.length) {
+		if (!entity.errors || !entity.errors.length) {
 			this.handleClose();
 		}
 	}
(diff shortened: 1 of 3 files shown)


dbatiste:
So `<custom-style>` is custom element that is built into Polymer.  It wouldn't be amazing to make this component depend on Polymer simply to use the table styles anyway.

I think what we want here, is for these styles to be implemented in core, either as shared CSS that can be imported and included with your element, or as ported and optimized components built on Lit.  From what I gather, transitioning to Lit will greatly simplify some of the CSS that's needed.  @dlockhart is far more familiar with the table issues in Polymer though, so he might have something to add.  This is something that is on our docket but I don't think anyone is currently assigned to it. I think we'd also welcome contribution for it, but there's probably some additional thought and conversation that's needed.

dlockhart:
@ronan-f asked specifically about Lit-based tables in `#web-platform` a couple weeks ago, so definitely not surprised to see this come up. It's very high on our list... to the point where I think based on the number of times these styles have been copy-pasted I think it should be at the very very top.

But yeah, it's not going to be a straight Polymer -> Lit conversion as a lot of things are now simpler. But there's also a lot of complexity in the old component around the scrolling behaviour and the sticky headers.

I'm going to make an argument for Gaudi (maybe me) to tackle this in the very short term.

dlockhart:
Having said that, @ronan-f can you use the technique to apply these styles to a native `<table>` element and not use the `<d2l-*>` elements? Those were only created for IE11 and will be going away in the conversion to Lit.

PR #1517, opened 2021-02-16
fix: refactor handling of d2l-request-provider for html-editor-enable…

mdulat:
…d and html-new-editor-enabled, used by all activity editor types.

@mdgbayly this is something you suggested in a previous PR, am just getting to it, as I was about to add another attribute for an LD flag. Figured I should refactor this first.


mdgbayly (approved):
LGTM

PR #1519, opened 2021-02-17
W2D: fix padding on list items to better match mockups

mbardakovD2L:
https://rally1.rallydev.com/#/357251704080ud/workviews?detail=%2Fdefect%2F492430695604&view=b064b584-6337-43c1-b9a9-28047568284d&fdp=true?fdp=true

Verified with Abbie over slack that this is how we want it to look


JasperAtD2L (dismissed):

JasperAtD2L (approved):

bearfriend:
This seems like something that should be done at the list-item level. Also, fiddling with a negative `margin` seems like it shouldn't be necessary. I would expect `line-height` and maybe a positive margin to be what we want but there's a weird gap between the divs I can't explain.

mbardakovD2L:
I noticed the same thing - I _think_ each list-item has a tertiary slot for more information we're not using (i.e. where the description goes in the fullscreen view) which may be inflating and/or messing with the spacing when it's empty.

dbatiste:
> This seems like something that should be done at the list-item level. Also, fiddling with a negative `margin` seems like it shouldn't be necessary. I would expect `line-height` and maybe a positive margin to be what we want but there's a weird gap between the divs I can't explain.

Agreed.  The `d2l-list-item-content` component provides three different slots for the purpose to provide consistent layout.  Ideally we're being consistent, and if a variant is needed we can look at options to provide that variant in a reusable/consistent way.

mbardakovD2L:
I think I found the issue - we were putting the bullet-separated list in the "supporting info" slot instead of the "secondary" slot - changing it to use the secondary slot instead (and removing the css changes I made earlier in this branch) looks like this:
image

bearfriend (approved):
↕️

PR #1520, opened 2021-02-18
Us123398/fix ip validation

ronan-f:

Addressing comments from @me-louie when testing. I refactored the validation logic to occur on save instead of on blur. Basically it prevents saving if there are validation errors and points out which boxes need corrected. When errors are corrected and the user presses save there is the same server side error handling as before. The reason for this change is that handling it on blur meant the error states became difficult to manage. It wasn't specified in the design when the validation should occur so hopefully this approach is ok. 

I added a few small fixes such as increasing the icon size to large and disabling the dialog opener if user lacks permissions as well.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
@@ -36,14 +36,6 @@ class ActivityQuizIpRestrictionEditor
 					display: none;
 				}
 
-				#ip-restriction-editor-label {
-					margin-bottom: 10px;
-				}
-
-				d2l-button {
-					margin: 1rem 0;
-				}
-
 				d2l-alert {
 					margin: 1rem 0;
 				}
@@ -69,13 +61,6 @@ class ActivityQuizIpRestrictionEditor
 		`;
 	}
 
-	_renderActionButtons() {
-		return html`
-			<d2l-button primary @click=${this._saveRestrictions}>${this.localize('btnIpRestrictionsDialogAdd')}</d2l-button>
-			<d2l-button data-dialog-action>${this.localize('btnIpRestrictionsDialogBtnCancel')}</d2l-button>
-		`;
-	}
-
 	_renderDialog() {
 		return html`
 			<d2l-dialog
@@ -86,18 +71,23 @@ class ActivityQuizIpRestrictionEditor
 				${this._renderErrors()}
 				${this._renderHelpDialog()}
 				${this._renderIpRestrictionsContainer()}
-				${this._renderActionButtons()}
 
 			</d2l-dialog>
 		`;
 	}
 
 	_renderDialogOpener() {
+		const entity = ipStore.get(this.ipRestrictionsHref);
+		if (!entity) {
+			return;
+		}
+
 		return html`
-			<div id="ip-restriction-editor-label" class="d2l-label-text">
+			<div class="d2l-label-text">
 				${this.localize('ipRestrictionLabel')}
 			</div>
 			<d2l-button-subtle
+				?disabled=${!entity.canEditIpRestrictions}
 				text="${this.localize('btnOpenIpRestrictionDialog')}"
 				h-align="text"
 				@click="${this.open}">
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -143,7 +133,8 @@ class ActivityQuizIpRestrictionEditor
 			<d2l-activity-quiz-ip-restrictions-container
 				href="${this.ipRestrictionsHref}"
 				.token="${this.token}"
-				@restrictions-resize-dialog="${this._resizeDialog}">
+				@restrictions-resize-dialog="${this._resizeDialog}"
+				@close-ip-dialog="${this.handleClose}">
```
ronan-f, line 137: I moved the action buttons/save logic into `d2l-activity-quiz-ip-restriction-container.js` so now it emits a close event if the dialog should be closed.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -71,6 +81,9 @@ class ActivityQuizIpRestrictionsContainer extends ActivityEditorMixin(ActivityEd
 		}
 
 		entity.deleteIpRestriction(index);
+
+		await this.updateComplete;
```
ronan-f, line 85: Need to wait for the table to update as we make assertions on the content of each of the input boxes.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -170,29 +192,68 @@ class ActivityQuizIpRestrictionsContainer extends ActivityEditorMixin(ActivityEd
 		});
 	}
 
+	async _saveRestrictions() {
+		const entity = store.get(this.href);
+		if (!entity) {
+			return;
+		}
+
+		const hasValidationError = this._validate();
```
ronan-f, line 201: Main difference in this method is handling validation here before save.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -170,29 +192,68 @@ class ActivityQuizIpRestrictionsContainer extends ActivityEditorMixin(ActivityEd
...
+
+		return hasValidationError;
+	}
+
+	_validateRestriction(restriction) {
+		if (!restriction) {
+			return true;
 		}
+
+		const isValid = !restriction.formValue || validateIp(restriction.formValue);
```
ronan-f, line 250: This part is kind of funky but basically empty IP ranges are technically valid as if the end range is empty it should still be sent to the server and the server will handle setting endVal = startVal i.e. a single restriction as opposed to a range.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js
```diff
@@ -60,7 +70,7 @@ class ActivityQuizIpRestrictionsContainer extends ActivityEditorMixin(ActivityEd
 		entity.addRestriction();
 	}
 
-	_deleteIpRestriction(_, index) {
+	async _deleteIpRestriction(_, index) {
```
me-louie, line 73: Just wondering but what is this ` _` argument for?

me-louie (approved):
lgtm

ronan-f:
> Just wondering but what is this _ argument for?

@me-louie So if you check out this [`_generateHandler`](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restrictions-container.js?r=e2507da4#76) method it returns a function that gets passed `e` (in this case the event that triggered the function call) and `rowIndex`. 

For deleting a restriction we only need the `rowIndex` so I marked the `e` param as unused by denoting it with an underscore. I don't think there's any semantic difference but it just lets the linter and other contributors know that I didn't expect that param to be used.

PR #1525, opened 2021-02-18
feat: Updated lang term as per design

ronan-f:
Updated lang term as per Patrick for consistency across tools. The copy on the subtle button will now read:

image


me-louie (approved):
v nice

PR #1526, opened 2021-02-18
W2D: Add "Course" subtitle to widget view

mbardakovD2L:
This also involved refactoring how the bullet-separated list of descriptors worked, since it was previously hard-coded to assume there would only be a date + orgCode/orgName and that's no longer the case.

https://rally1.rallydev.com/#/357251704080ud/workviews?detail=%2Fuserstory%2F496177602228&view=b064b584-6337-43c1-b9a9-28047568284d


components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -73,8 +73,6 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 				}
 				.d2l-icon-bullet {
 					color: var(--d2l-color-tungsten);
-					margin-left: -0.15rem;
-					margin-right: -0.15rem;
```
mbardakovD2L, line 77: for whatever reason, the spacing was really wonky after the refactor and getting rid of this seemed to fix it. I don't know why and I am very open to other ideas

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -190,6 +188,11 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 			</div>`
 			: nothing;
 
+		const supportingInfoTemplate = (items) => {
+			const filteredItems = items.filter(item => item);
+			return html `${filteredItems.map((item, idx) => [item, idx < filteredItems.length - 1 ? separatorTemplate : nothing]).flat()}`;
```
mbardakovD2L, line 193: There's a lot going on here so I'll break it down - we want to make sure there's a bullet after every item except the last one (hence the bit with the index) and since we're returning every pair of `[item, bullet]` as an array, I wanted to flatten them all at the end because that felt cleaner. An alternative to this would be to use a full loop instead of a map, and append every item to the list with bullets in between, which I can change it to if this is too hard to read.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -208,9 +211,7 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 						${this._name}
 					</div>
 					<div class=${classMap(secondaryClasses)} slot="supporting-info">
-						${dateTemplate}
-						${separatorTemplate}
-						${this._orgName || this._orgCode}
+						${supportingInfoTemplate([dateTemplate, (this._orgName || this._orgCode), this._type === this.localize('Course') ? this._type : null])}
```
mbardakovD2L, line 214: we only want to show the type if the type is `Course`, as seen in thread: https://d2l.slack.com/archives/G0150CJ3SB1/p1613669915074800

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
mbardakovD2L: update: there are visual-diff tests failing so this is definitely the wrong way to do this

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
bearfriend: Do we want to check the type against the allow list type instead of the translated string?

```javascript
this._type === ActivityAllowList.userCourseOfferingActivity.type ? this._type : null
```

mbardakovD2L: That feels better, I'll update this

bearfriend: Does that work though? Is `this._type` translated or is it always the english name?

bearfriend: Ahh, I see. It should be

```javascript 
this._activityProperties.type === ActivityAllowList.userCourseOfferingActivity.type ? this._type : null
```

mbardakovD2L: I don't know, _but_, `this._type` comes from `_activityProperties.type`, which is itself from the ActivityAllowList. So whether or not they're translated shouldn't affect this particular equality check (i.e. either it translates both sides of the `===` or it translates neither). Unfortunately I don't actually know the answer to your original question.

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -208,9 +211,7 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 						${this._name}
 					</div>
 					<div class=${classMap(secondaryClasses)} slot="supporting-info">
-						${dateTemplate}
-						${separatorTemplate}
-						${this._orgName || this._orgCode}
+						${supportingInfoTemplate([dateTemplate, (this._orgName || this._orgCode), this._type === ActivityAllowList.userCourseOfferingActivity.type ? this._type : null])}
```
bearfriend, line 214: Swap the other side for the pre-translated type, and give 'er some space.

```suggestion
						${supportingInfoTemplate([
						  dateTemplate,
						  this._orgName || this._orgCode,
						  this._activityProperties.type === ActivityAllowList.userCourseOfferingActivity.type && this._type
						])}
```

mbardakovD2L: oops - totally forgot that `this._type` has itself been translated

bearfriend (approved):
🥊

components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js
```diff
@@ -179,7 +177,7 @@ class ActivityListItemBasic extends ListItemLinkMixin(SkeletonMixin(EntityMixinL
 
 		const dateTemplate = html `<d2l-activity-date href="${this.href}" .token="${this.token}" format="MMM d" ?hidden=${this.skeleton}></d2l-activity-date>`;
 
-		const separatorTemplate = !this.skeleton && this._date && (this._orgName || this._orgCode)
+		const separatorTemplate = !this.skeleton
 			? html `<d2l-icon class="d2l-icon-bullet" icon="tier1:bullet"></d2l-icon>`
```
bearfriend, line 181: You might be able to just make this `'\n<d2l-icon class="d2l-icon-bullet" icon="tier1:bullet"></d2l-icon>\n'` or some other way of adding whitespace between all the tags.

bearfriend (approved):
⚫

PR #1531, opened 2021-02-19
QE Widget items with new submissions link to evaluate new submissions only

Scy-D2L:
Before: All items linked to 'evaluate all submissions'
Now: Items with > 0 submissions will link to 'evaluate new submissions'

### DIFF
--- a/components/d2l-quick-eval-widget/d2l-quick-eval-widget-controller.js
+++ b/components/d2l-quick-eval-widget/d2l-quick-eval-widget-controller.js
@@ -22,13 +22,14 @@ export async function fetchSubmissionCount(activityUsageEntity, token) {
 	return newSubmissions || 0;
 }
 
-export async function fetchEvaluateAllHref(activityUsageEntity, token) {
+export async function fetchEvaluationHref(activityUsageEntity, token, evaluateAll) {
 	const evalStatusRel = 'https://activities.api.brightspace.com/rels/evaluation-status';
 	const evalStatusLink = activityUsageEntity.getLink(evalStatusRel);
 
 	const evaluationStatusEntity = await fetch(evalStatusLink, token);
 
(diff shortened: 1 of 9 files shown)


components/d2l-quick-eval-widget/d2l-quick-eval-widget-controller.js
```diff
@@ -22,13 +22,14 @@ export async function fetchSubmissionCount(activityUsageEntity, token) {
 	return newSubmissions || 0;
 }
 
-export async function fetchEvaluateAllHref(activityUsageEntity, token) {
+export async function fetchEvaluationHref(activityUsageEntity, token, evaluateAll) {
```
jeroach, line 25: Are there any other usages of this? Are we ever expecting to need to evaluate all?

Just wondering if we can remove this parameter. Unless I'm mistaken, there doesn't appear to be a code path that sets it to `true`.

i.e. if there are 0 submissions, I wouldn't expect the item to show up in a "work do to" widget

Scy-D2L: When there are no new submissions something can show up in the Quick Eval widget the same as it can show up in Quick Eval - I believe this happens when there are unpublished evaluations or something to that effect. Because they don't have new submissions, we send them to evaluate all instead of evaluate new.

Scy-D2L: It may just be confusion that "submissionCount" here is counting new submissions and not all submissions. 🤔

jeroach: Gotcha, sounds reasonable :)

jeroach (approved):

PR #1532, opened 2021-02-19
US123360 - Support for quiz timing optimistic loading

wongvincent:
Support for Timing Type (Recommended/Enforced) radio buttons and SubmissionLateType radio buttons (show/hide extended deadline select box when uselatelimit radio button is selected.)

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F464625055264%2Ftasks

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
@@ -1,6 +1,7 @@
 import { action, configure as configureMobx, decorate, observable } from 'mobx';
 import { fetchEntity } from '../../state/fetch-entity.js';
 import { QuizTimingEntity } from 'siren-sdk/src/activities/quizzes/timing/QuizTimingEntity.js';
+import { Classes } from 'siren-sdk/src/hypermedia-constants';
 
 configureMobx({ enforceActions: 'observed' });
 
@@ -48,6 +49,7 @@ export class QuizTiming {
 	}
 
 	async setExceededTimeLimitBehaviour(data) {
+		this.isAutomaticZero = data === Classes.quizzes.timing.automaticZero;
 		await this._entity.setExceededTimeLimitBehaviour(data);
 		this.fetch();
 	}
@@ -68,6 +70,7 @@ export class QuizTiming {
 	}
 
 	async setTimingType(data) {
+		this.isTimingEnforced = data === Classes.quizzes.timing.enforced;
 		await this._entity.setTimingType(data);
 		this.fetch();
 	}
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
```diff
@@ -1,6 +1,7 @@
 import { action, configure as configureMobx, decorate, observable } from 'mobx';
 import { fetchEntity } from '../../state/fetch-entity.js';
 import { QuizTimingEntity } from 'siren-sdk/src/activities/quizzes/timing/QuizTimingEntity.js';
+import { Classes } from 'siren-sdk/src/hypermedia-constants';
```
wongvincent, line 4: Is it okay to reach into siren-sdk like this, or should I add it to `d2l-hypermedia-constants` instead? https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/index.js#L312

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
```diff
@@ -48,6 +49,7 @@ export class QuizTiming {
 	}
 
 	async setExceededTimeLimitBehaviour(data) {
+		this.isAutomaticZero = data === Classes.quizzes.timing.automaticZero;
 		await this._entity.setExceededTimeLimitBehaviour(data);
```
ronan-f, line 53: Do we need to await here? Won't this block the UI while we're waiting? I thought we could just update state, fire off the network request and let it process in the background?

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
```diff
@@ -48,6 +49,7 @@ export class QuizTiming {
 	}
 
 	async setExceededTimeLimitBehaviour(data) {
+		this.isAutomaticZero = data === Classes.quizzes.timing.automaticZero;
```
ronan-f, line 52: Do we need to `await` here? Won't this block the UI while we're waiting? I thought we could just update state, fire off the network request and let it process in the background?

ronan-f: I guess the point is that we need `this.fetch()` to be guaranteed to run after the network request has completed but should still be able to accomplish that

wongvincent: Nah, it won't block the UI as long as I put my optimistic loading state change before the action call. I need to update the MobX entity after calling the action with `this.fetch`, but it can only happen after the action request is complete, that's why I need the `await`

mdgbayly: I feel like we might have tenatively touched on this scenario but not sure if we really dove into it =>

Is there any danger with this approach of refetching, that the response from the API could end up causing jank in the UI if:
* UI is in state 1
* the user changes a radio button, 
* we update the UI immediately to state 2
* we make the API call
* before the API call returns the user changes the option back again
* we immediately update the UI again back to state 1
* the original API call returns  - does this put back state to 2 again?

wongvincent: Yes, it looks jank. Adding a task to the story so we can address this later.

ronan-f (approved):
As discussed on slack/comments likely not a perfect solution but jank to be addressed later so approving for now.

wongvincent:
Yes, would like others to see this jank on QA.

PR #1537, opened 2021-02-19
Mdulat/us124808 rename choose from grades

mdulat:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F496651489404&fdp=true

Note: only the last commit is unique to this PR, since it is branched off previous.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades-dialog.js
@@ -20,7 +20,8 @@ class ActivityGradesDialog extends ActivityEditorMixin(LocalizeActivityEditorMix
 		return {
 			_createNewRadioChecked: { type: Boolean },
 			_canLinkNewGrade: { type: Boolean },
-			_hasGradeCandidates: { type: Boolean }
+			_hasGradeCandidates: { type: Boolean },
+			_createSelectboxGradeItemEnabled: { type: Boolean }
 		};
 	}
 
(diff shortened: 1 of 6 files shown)


components/d2l-activity-editor/d2l-activity-grades-dialog.js
```diff
@@ -85,7 +100,11 @@ class ActivityGradesDialog extends ActivityEditorMixin(LocalizeActivityEditorMix
 		} = activity.scoreAndGrade;
 
 		return html`
-			<d2l-dialog title-text="${this.localize('editor.chooseFromGrades')}" width="460" @d2l-dialog-open="${this._onDialogOpen}">
+			<d2l-dialog
+				title-text="${this._createSelectboxGradeItemEnabled ? this.localize('editor.editLinkExisting') : this.localize('editor.chooseFromGrades')}"
+				width="460"
```
mostlyfabulous, line 105: I haven't touched dialogs before and I'm wondering if imported styles will have any effect on the `width` when resizing?

mostlyfabulous (approved):
LGTM

components/d2l-activity-editor/d2l-activity-grades-dialog.js
mdulat: I don't think so, the width is set by the [dialog mixin](https://github.com/BrightspaceUI/core/blob/cdfb626af5af09b105caf5c3370ede657ff62c14/components/dialog/dialog-mixin.js), but I haven't looked at this too closely either. (The change here was just a line break :) )

PR #1544, opened 2021-02-22
Added quick fixes for focus and footer

ronan-f:

Fixes requested by Joseph:

1. Moved action buttons to slot="footer" so that they would behave consistently with the attempts dialog i.e. the action buttons stick in place even when new elements force scroll bar. This meant lifting the actions buttons up to `d2l-activity-quiz-ip-restrictions-editor` so that I could set the slot attribute correctly. 
2. (Edit: this change has been removed as Joseph didn't like the janky behaviour but if anyone has any thoughts on this I'm still interested so leaving here) Added an event listener for `d2l-dialog-open` to focus on the first instance of `d2l-input-text`. Prior to this change it focused on the "help" button on open as the core dialog component [forces focus](https://github.com/BrightspaceUI/core/blob/master/components/dialog/dialog-mixin.js#L314) on the first focusable element when the dialog opens. Not sure if there's a better way of handling this. It's kinda janky as for a split second you can see the help button in focus and then it changes. I wanted to use a lifecycle method like `firstUpdated` or `connectedCallback` to handle setting this but as core overrides any value on dialog open this doesn't seem possible.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
@@ -10,6 +10,7 @@ import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityQuizEditorMixin } from './mixins/d2l-activity-quiz-lang-mixin';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { validateIp } from './helpers/ip-validation-helper.js';
 
 class ActivityQuizIpRestrictionEditor
 	extends ActivityEditorMixin(RtlMixin(ActivityEditorDialogMixin(LocalizeActivityQuizEditorMixin(MobxLitElement)))) {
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -61,6 +62,15 @@ class ActivityQuizIpRestrictionEditor
 		`;
 	}
 
+	_renderActionButtons() {
+		return html`
+			<div slot="footer" id="d2l-actions-container">
+				<d2l-button slot="footer" primary @click=${this._saveRestrictions}>${this.localize('btnIpRestrictionsDialogAdd')}</d2l-button>
```
wongvincent, line 68: doesn't only the top level div `d2l-actions-container` need the slot attribute?

ronan-f: True. Didn't mean to leave that in.

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -143,6 +153,66 @@ class ActivityQuizIpRestrictionEditor
...
+		const hasValidationError = this._validate();
+
+		if (hasValidationError) {
+			return;
+		}
+
+		await entity.saveRestrictions();
+
+		if (!entity.errors || !entity.errors.length) {
+			this.handleClose();
```
wongvincent, line 173: I haven't handled this either, but I think we'll want to show the errors and not close the dialog or something... still to be discussed.

wongvincent: Oh wait, this is a `!` oops. nevermind.

wongvincent (approved):

PR #1545, opened 2021-02-22
Quick Eval Widget Skeleton Fix

Scy-D2L:
Items in Quick-Eval widget will not render until fully loaded
Also changes the style of skeleton to be more detailed
image

### DIFF
--- a/components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
+++ b/components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
@@ -12,13 +12,15 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
 const errorState = 'error';
-const loadedState = 'loaded';
+const listState = 'list';
 const noSubmissionState = 'noSubmission';
 
 export class QuickEvalWidget extends LocalizeQuickEvalWidget(SkeletonMixin(LitElement)) {
 	static get properties() {
 		return {
 			_activities: { type: Array },
+			_loading: { type: Boolean },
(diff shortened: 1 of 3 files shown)


components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
```diff
@@ -162,13 +176,15 @@ export class QuickEvalWidget extends LocalizeQuickEvalWidget(SkeletonMixin(LitEl
 			</div>`;
 	}
 
-	get loadedTemplate() {
+	get listTemplate() {
 		const listItems = this._activities.map(activity => {
 			return html`<d2l-work-to-do-activity-list-item-basic
 					evaluate-href="${activity.evaluationHref}"
 					href="${activity.href}"
+					?skeleton=${this._loading || this.skeleton}
```
jeroach, line 184: It looks like `this.skeleton` no longer exists in this file

JamieTran (approved):
LGTM apart from Josh's comment

components/d2l-quick-eval-widget/d2l-quick-eval-widget.js
Scy-D2L: `this.skeleton` comes from the SkeletonMixin. We aren't actively using it at this level right now since the individual list items skeletize, but it would be a controlled way to force the entire component to skeletize regardless of the individual items loaded state. I can remove it, the only impact  is having no way to check the loading state in the demo without code changes or a severely throttled connection (right now the demo pages have a skeleton switch next to them).

jeroach: Ah right, that's fine. I forgot about that mixin

jeroach (approved):

PR #1553, opened 2021-02-23
US123360 - Change to improve quiz timing entity optimistic loading jank

wongvincent:
Only refetch entity if no entity is returned. If entity is returned, update the timing entity, but don't reload it to avoid janky UI updates.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F464625055264%2Ftasks

See: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/276

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
@@ -76,6 +76,10 @@ export class QuizTiming {
 
 	async updateProperty(updateFunc) {
 		const entity = await updateFunc();
+		// The siren-sdk function called to perform an action first checks that the entity has permission to do so.
+		// If the entity lacks permission, the function returns `undefined`, otherwise it returns a reconstructed siren-sdk timing entity.
+		// If `undefined` is returned, it likely means the UI is out of sync with the entity state, and disallowed actions can be performed.
+		// In this case, we should attempt to reload the MobX object, so that the UI state is in sync again.
 		if (!entity) {
 			this.fetch();
 		}
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
@@ -47,36 +47,39 @@ export class QuizTiming {
 		this.maxEnforcedGraceLimit = entity.maxEnforcedGraceLimit();
 	}
 
-	async setExceededTimeLimitBehaviour(data) {
+	setExceededTimeLimitBehaviour(data) {
 		this.isAutomaticZero = this._entity.isAutomaticZero(data);
-		await this._entity.setExceededTimeLimitBehaviour(data);
-		this.fetch();
+		this.updateProperty(() => this._entity.setExceededTimeLimitBehaviour(data));
+
 	}
 
-	async setExtendedDeadline(data) {
-		await this._entity.setExtendedDeadline(data);
-		this.fetch();
+	setExtendedDeadline(data) {
+		this.updateProperty(() => this._entity.setExtendedDeadline(data));
 	}
 
-	async setGracePeriod(data) {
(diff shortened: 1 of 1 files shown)


me-louie (dismissed):

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-timing.js
```diff
@@ -47,36 +47,39 @@ export class QuizTiming {
...
 		this.isTimingEnforced = this._entity.isTimingEnforced(data);
-		await this._entity.setTimingType(data);
-		this.fetch();
+		this.updateProperty(() => this._entity.setTimingType(data));
+	}
+
+	async updateProperty(updateFunc) {
+		const entity = await updateFunc();
+		if (!entity) {
+			this.fetch();
```
mdgbayly, line 80: I don't totally get this line but maybe it is not having any actual impact. If the API call to update the resource, doesn't return the updated entity in the response, then this call to `this.fetch()` won't actually cause a new GET request to the resource, because it will just return the existing entity in the store. 

So I'm not clear why we are gating the reloading behaviour of the component state based on whether the siren-sdk function returns an entity or not? That seems a bit obscure?

wongvincent: Oh, I think I should fetch with a bypass cache.

I was thinking of the case where if they don't actually have permission/action to edit it, but we change something in the UI which tries to perform the siren action, the request will not be made and return undefined. That probably means our UI is in some weird state where it thinks it has permission to make the request, but it doesn't. So I want to reload the timing entity in this case. Maybe I'm not completely understanding what you were trying to say yesterday.

mdgbayly: OK - yeh - that works - it's just a little bit unclear that's why it's doing it. Yesterday, I was suggesting we make it explicit in the MobX code by doing the canEdit check in there. But I guess if you assume that the siren-sdk call always returns an entity if the siren-sdk call succeeds, then what you have will work. Maybe if you have been working with the code it seems clearer :-).

In that case, if this is the way you were solving that problem, then I think it's fine that you don't bypass the cache in the fetch call, because you actually want to refresh the state from the existing entity.

mdgbayly: This is probably a weird enough scenario that it warrants a comment on what the function is doing and why. Normally I like code to be self-documenting unless there are subtle behaviours like what is going on here :-)

mdgbayly: Hmm, ok - becoming sort of clearer now - as I see in the other PR, you also modified the siren-sdk class to reconstruct a siren-sdk entity from the returned response. Is that nice or would it be better to recreate the siren-sdk class in the MobX state? hmm?

wongvincent: I did a similar thing in the siren-sdk QuizEntity checkin/checkout functions where it returns a reconstructred siren-sdk quiz entity, so I sorta followed that.

Let me add a comment here :)

mdgbayly: I guess it kind of makes sense to recreate the entity in the function itself - sort of consistent with the siren sdk entities being immutable, so a mutating function returns a new version of the entity. Just not sure we use that pattern elsewhere? but maybe we should.

wongvincent: I only see one example in siren-sdk of an update function that returns an entity, in the OrganizationCollection where an add function returns a new OrganizationEntity.
It can become a pattern if we have the attempts and ipaddress stuff similarly implemented. :)

mdgbayly (approved):

PR #1554, opened 2021-02-23
Validate IPs on delete

ronan-f:
I did some refactoring earlier and realized that `this._validate` had now been moved to parent component so it is no longer being triggered on delete. To handle this I added logic to emit a "deleted" event and run the validate logic from the parent component.


dustxd (approved):
Good catch - get those changes in ring leader!

PR #1562, opened 2021-02-24
Return

wongvincent:
if `entity` is undefined, and we call `fetch`, `this._entity` is updated with the correct entity. We shouldn't be re-setting it to `undefined`.


mostlyfabulous (approved):
Good catch!

PR #1576, opened 2021-03-02
US123692 - Add error handling to working copy dialog mixin

wongvincent:
Like the validation, the setting of `isError` is again inspired by the `activity-editor-container-mixin` https://github.com/BrightspaceHypermediaComponents/activities/blob/b7db0429ed7039258ed8d0e8270a229215c2ebdd/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js

The timing/workingcopy component needs to include a `d2l-alert` that uses the `isError` and `errorTerm` properties to determine what to display. The `isError` and `errorTerm` are set in the mixin based on whether the input validation and the `checkin` succeed or fail. The error term is set to the `validationErrorTerm` if the input validation check fails or `serverErrorTerm` if the `checkin` fails. The `validationErrorTerm` and `serverErrorTerm` properties are set in the `firstUpdated` function of the timing component. (I tried setting it in the constructor, but `this.localize(...)` returns nothing.

This PR also adds an `isSaving` property which is used to disable the 'OK' and 'Cancel' buttons. This is important because the dialog needs to wait for the `checkin` to complete before closing and thus there is a delay. If it fails, the dialog stays open and shows the `serverErrorTerm` in the alert as detailed above.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478606416460%2Ftasks

See this other PR that changes the `checkin` function to reject the promise if it fails: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/279

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
@@ -1,4 +1,5 @@
 import './d2l-activity-quiz-manage-timing-editor.js';
+import '@brightspace-ui/core/components/alert/alert.js';
 import '@brightspace-ui/core/components/dialog/dialog.js';
 import { css, html } from 'lit-element/lit-element';
 import { ActivityEditorWorkingCopyDialogMixin } from '../mixins/d2l-activity-editor-working-copy-dialog-mixin';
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
```diff
@@ -23,6 +27,12 @@ class ActivityQuizManageTimingContainer extends ActivityEditorWorkingCopyDialogM
 		super(store);
 	}
 
+	firstUpdated() {
+		super.firstUpdated();
+		this.serverErrorTerm = this.localize('quizTimingServerError');
+		this.validationErrorTerm = this.localize('quizTimingValidationError');
```
emil-furniss, line 33: Nit: use getters so the terms are only evaluated when they are needed

emil-furniss (approved):

PR #1579, opened 2021-03-02
US125185/ip restrictions fixes

ronan-f:
https://trello.com/c/Z6Pg1Vxe/321-face-quiz-ip-restrictions-no-start-defined-but-no-validation-error-surfaced
https://trello.com/c/6Sxx9w7q/322-face-quiz-ip-restrictions-multiple-duplicate-errors-surface

Fixes for above trello items. Both feel fairly janky. Any suggestions appreciated.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
@@ -131,7 +131,10 @@ class ActivityQuizIpRestrictionEditor
 			return;
 		}
 
-		return errors.map((error) => {
+		// using a set to filter out duplicate error messages
+		const uniqueErrors = Array.from(new Set(errors));
+
+		return uniqueErrors.map((error) => {
 
 			if (error === 500 || !error) {
 				error = this.localize('ipRestrictions500Error');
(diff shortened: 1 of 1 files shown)


mdulat:
2 questions:

1. If there are 2 different types of errors, should we see both? I see only one type at a time
2. When there are errors should the focus go to the top? Currently when there are multiple IPs that require a scrollbar, when I click Add I don't see the errors at the top unless I scroll manually.

ronan-f:
> 2 questions:
> 
> 1. If there are 2 different types of errors, should we see both? I see only one type at a time
> 2. When there are errors should the focus go to the top? Currently when there are multiple IPs that require a scrollbar, when I click Add I don't see the errors at the top unless I scroll manually.

So for 1 yes you should be able to see multiple different kinds of server side errors. There's only one for client side validation which is the "Incorrect format..." message. To see 2 different errors you can add 2 new ranges, one which duplicates the start value of a range that has already been saved and another that has start > end. 

image


For 2, yeah I was thinking about that and meant to look into it further as it seems like it would be better UX but I kept getting distracted with other things and haven't come back to it. I'm sure we're doing something similar elsewhere so I can look into adding it.

mdulat:
> So for 1 yes you should be able to see multiple different kinds of server side errors. There's only one for client side validation which is the "Incorrect format..." message. To see 2 different errors you can add 2 new ranges, one which duplicates the start value of a range that has already been saved and another that has start > end.

Ah ok, so the client error shows up by itself before the save is attempted, that makes sense.

mdulat (approved):

PR #1582, opened 2021-03-04
Shifted validation client side to prevent losing track of IDs

ronan-f:
Shifted IP restrictions validation client side to avoid losing track of IDs when sending network requests.


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-ip-restriction-editor.js
```diff
@@ -70,6 +70,68 @@ class ActivityQuizIpRestrictionEditor
 		`;
 	}
 
+	_handleValidationError(errorKey) {
```
mdgbayly, line 73: It feels weird to me that we're doing all this validation in the UI component rather than in the MobX state. Probably makes it harder to write unit tests for. But I expect we just want to move on.

mdgbayly (approved):

PR #1587, opened 2021-03-05
DE42624: Increasing editor size to take up all available vertical space

brissons:
[DE42624](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F502198563680&fdp=true): Lessons FACE > new HTML editor default size too small

**Actual behaviour**:
When editing/creating a module, the default size of the HTML editor size is really small.

**Expected behaviour**:
When editing/creating a module, the HTML editor should take up all available vertical space/dynamically resize with the window.

dynamic-height

**Note**: This requires `@brightspace-ui/htmleditor` version 1.2.6 or higher.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -24,7 +24,7 @@ class ContentEditorDetail extends MobxLitElement {
 				}
 				d2l-activity-content-module-detail {
 					display: flex;
-					flex-flow: column;
+					flex-direction: column;
 					height: inherit;
 				}
 			`
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -22,6 +22,11 @@ class ContentEditorDetail extends MobxLitElement {
 				d2l-loading-spinner {
 					width: 100%;
 				}
+				d2l-activity-content-module-detail {
+					display: flex;
+					flex-flow: column;
+					height: inherit;
+				}
 			`
 		];
 	}
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -34,7 +34,7 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				}
 				#content-description-container {
 					display: flex;
-					flex-flow: column;
+					flex-direction: column;
 					height: inherit;
 				}
 			`
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -27,6 +27,17 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 			css`
 				.d2l-activity-label-container {
 					margin-bottom: 7px;
+					display: inline-block;
+					vertical-align: bottom;
+				}
+				#content-description-container {
+					display: flex;
+					flex-flow: column;
+					height: inherit;
+				}
+				#html-editor-container {
+					flex-grow: 1;
+					min-height: 300px;
 				}
 			`
 		];
@@ -63,13 +74,15 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				<div class="d2l-activity-label-container d2l-label-text d2l-skeletize">
 					${this.localize('content.description')}
 				</div>
-				<div class="d2l-skeletize">
+				<div id="html-editor-container" class="d2l-skeletize">
 					<d2l-activity-text-editor
-						.ariaLabel="content-description"
+						id="d2l-activity-text-editor"
+						ariaLabel="${this.localize('content.description')}"
 						.key="content-description"
 						.value="${descriptionRichText}"
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
+						height="100%"
 					>
 					</d2l-activity-text-editor>
 				</div>
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -30,7 +30,7 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				}
 				.d2l-new-html-editor-container {
 					flex-grow: 1;
-					min-height: 10rem;
+					min-height: 300px;
 				}
 				#content-description-container {
 					display: flex;
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -82,7 +82,7 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						.value="${descriptionRichText}"
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
-						height="100%"
+						htmlEditorHeight="100%"
 					>
 					</d2l-activity-text-editor>
 				</div>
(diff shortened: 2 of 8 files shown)


components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -8,18 +8,34 @@ class ActivityHtmlNewEditor extends LocalizeActivityEditorMixin(LitElement) {
...
+				d2l-htmleditor {
+					height: 100%;
+				}
+			`
+		];
+	}
+
+	constructor() {
+		super();
+		this.htmlEditorHeight = '10rem';
```
brissons, line 28: This sets the default to `10rem` (the previous set value), since other FACE pages use this component.

components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
```diff
@@ -63,13 +83,14 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
...
-				<div class="d2l-skeletize">
+				<div class="d2l-skeletize ${htmlNewEditorEnabled ? 'd2l-new-html-editor-container' : ''}">
 					<d2l-activity-text-editor
-						.ariaLabel="content-description"
+						.ariaLabel="${this.localize('content.description')}"
 						.key="content-description"
 						.value="${descriptionRichText}"
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
+						htmlEditorHeight="100%"
```
brissons, line 93: Added this prop which sets the height of the new html editor. This gets passed down to the actual `d2l-htmleditor`.

components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
```diff
@@ -28,6 +28,15 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				.d2l-activity-label-container {
 					margin-bottom: 7px;
 				}
+				.d2l-new-html-editor-container {
+					flex-grow: 1;
+					min-height: 300px;
```
brissons, line 33: This is the [minimum height of the actual editor](https://github.com/BrightspaceUI/htmleditor/blob/a08403ed8e209371c70a3acb07011b0da2ecaa01/htmleditor-mixin.js#L268). Since the container is `flex-grow`, we should set the minimum here as well.

components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-styles.js
```diff
@@ -7,6 +7,9 @@ export const activityContentEditorStyles = css`
 	:host([hidden]) {
 		display: none;
 	}
+	:host([skeleton]) .d2l-skeletize::before {
+		z-index: 3;
+	}
```
brissons, line 12: Noticed that the loading skeleton is hidden behind the text-box when `d2l.Preferences.TurnOffWYSIWYG` is turned on.

components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
```diff
@@ -63,13 +83,14 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				<div class="d2l-activity-label-container d2l-label-text d2l-skeletize">
 					${this.localize('content.description')}
 				</div>
-				<div class="d2l-skeletize">
+				<div class="d2l-skeletize ${htmlNewEditorEnabled ? 'd2l-new-html-editor-container' : ''}">
 					<d2l-activity-text-editor
-						.ariaLabel="content-description"
+						.ariaLabel="${this.localize('content.description')}"
```
brissons, line 88: I missed this last time, but this allows the screenreader to read the description label properly.

ChrisKraljevicD2L (dismissed):

ChrisKraljevicD2L (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-styles.js
quinnbudan: I'm just curious because I don't really understand `z-index` (I just looked it up for the first time!) but why are we setting this to 3? Are there some other elements where it's set to 2 or something and we want this to be in front?

components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
```diff
@@ -63,13 +83,14 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 				<div class="d2l-activity-label-container d2l-label-text d2l-skeletize">
 					${this.localize('content.description')}
 				</div>
-				<div class="d2l-skeletize">
+				<div class="d2l-skeletize ${htmlNewEditorEnabled ? 'd2l-new-html-editor-container' : ''}">
```
quinnbudan, line 86: Does it cause issues to always have this class even if the new editor isn't enabled?

quinnbudan (approved):
LGTM, just have some clarification questions

components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-styles.js
brissons: Yeah! So if we turn on `d2l.Preferences.TurnOffWYSIWYG`, we get a `d2l-input-textarea`, which has a `z-index: 2;`. This is the same as the skeleton `z-index`. If we set the skeleton to have `z-index: 3;`, we guarantee that the skeleton will be in front of the textbox.

Without the z-index increase, the textbox is in front of the skeleton while loading:
image

components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
brissons: If the new editor isn't enabled, we end up with either the old html editor or the textbox. Both of which have a fixed height. If we were to leave this class in, the loading skeleton (which is mean to grow with the container) will not reflect the actual sizes of the old html editor or the textbox.

components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-editor-styles.js
dbatiste: Currently the skeleton uses `z-index: 1;`, and think there are a few places (a couple [here](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=lms&full=%22.d2l-skeletize%3A%3Abefore%22&defs=&refs=&path=&hist=&type=&xrd=&si=full&si=full)) that have overridden it to be `z-index: 2;` (including `d2l-input-textarea` as you found), resulting in this needing `z-index-3;`.  I am wondering if maybe we should update the skeleton mixin styles to be something like `z-index:4/5;` to provide a little breathing room for components when they need to use `z-index` for low level stacking.  @dlockhart , thoughts?

dlockhart: That seems reasonable yeah.

dbatiste: @brissons ... bumped z-index for skeletons in core@1.118.2.

brissons: Awesome thanks!

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -22,6 +22,11 @@ class ContentEditorDetail extends MobxLitElement {
 				d2l-loading-spinner {
 					width: 100%;
 				}
+				d2l-activity-content-module-detail {
+					display: flex;
+					flex-flow: column;
```
ianlaidlaw, line 27: I've never used `flex-flow` but my understanding is that it's a combo of `flex-direction` and setting the wrapping behavior. Since you didn't specify anything with `wrap` would simply using `flex-direction` be more appropriate?

ianlaidlaw (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
ianlaidlaw: though there's really no harm just keeping it like this.

brissons: It would be! `flex-direction` is better.

ianlaidlaw (approved):

CodeBaboon (approved):
LGTM, lots of good feedback and discussion on this PR, too! Nice.

PR #1590, opened 2021-03-05
US125529 Total Points UI

me-louie:
- Added a new component to render the Total Points of a quiz.
- I called it `d2l-activity-quiz-editor-action-bar` because the plan is to eventually put some buttons next to the Total Points 

- Made changes to `ActivityScoreGrade ` to accommodate new linked `score-out-of `entity. This involved splitting off an `async fetch(entity) ` which gets `siren-sdk `to fetch/return the `linked score-out-of entity` if it is present on the `ActivityUsage`. Following, `fetch ` does the `ActivityScoreGrade` initialization which was originally in the constructor.
- Updated the related tests

**Corresponding siren-sdk change:** https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/291

[US125529](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F505601651720&fdp=true?fdp=true): [ui] Display total points for a quiz

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F505601651720&fdp=true?fdp=true


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -159,6 +160,14 @@ export class ActivityUsage {
 		});
 	}
 
+	async _loadScoreOutOf(entity) {
+		const linkedScoreOutOfEntity = entity.scoreOutOfHref();
+		if (linkedScoreOutOfEntity) {
+			const scoreOutOfEntity = await fetchEntity(entity.scoreOutOfHref(), this.token);
+			runInAction(() => this.scoreOutOf = scoreOutOfEntity.properties.scoreOutOf);
```
emil-furniss, line 167: This is technically fine, but why not use the new `ActivityScoreOutOfEntity` with its `scoreOutOf()`? The `scoreOutOfEntity` could also not have `properties` on it.

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -159,6 +160,14 @@ export class ActivityUsage {
 		});
 	}
 
+	async _loadScoreOutOf(entity) {
+		const linkedScoreOutOfEntity = entity.scoreOutOfHref();
```
emil-furniss, line 164: Naming nit: Not really an entity, really just an href.

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -159,6 +160,14 @@ export class ActivityUsage {
...
+		if (linkedScoreOutOfEntity) {
+			const scoreOutOfEntity = await fetchEntity(entity.scoreOutOfHref(), this.token);
```
emil-furniss, line 166: Reuse the `linkedScoreOutOfEntity` value instead of getting it again?

me-louie: oops yes this is a refactoring typo

components/d2l-activity-editor/state/activity-usage.js
me-louie: Yes, totally that is better. Made the change.

mdgbayly: https://github.com/BrightspaceHypermediaComponents/activities/blob/f9612781b647de2755a87fa75fd77fb43c6e9c33/components/d2l-activity-editor/state/activity-score-grade.js#L8

mdgbayly: Ideally we want to be able to instantiate that `ActivityScoreGrade` MobX state from either the linked sub entity or the embedded sub entity, and all the UI components use is the `ActivityScoreGrade` state object.

me-louie: ok thanks for your comment Martin! I will look into utilizing `ActivityScoreGrade`. I was a bit unsure about having two scoreOutOf properties on the usage entity so doing it this way makes more sense.

test/d2l-activity-editor/state/activity-score-grade.spec.js
```diff
@@ -39,46 +39,58 @@ describe('Activity Score Grade', function() {
...
 
 		it('initializes with correct createNewGrade', async() => {
-			const activity = new ActivityScoreGrade(defaultEntityMock);
+			const activity = new ActivityScoreGrade('token');
+			await activity.fetch(defaultEntityMock);
 
 			expect(activity.createNewGrade).to.be.false;
 		});
+
+		it('initializes scoreOutOf with value provided by fetchLinkedScoreOutOfEntity', async() => {
```
me-louie, line 79: Not entirely sure if this is the correct way to test the linked `score-out-of `entity, hoping for some feedback @emil-furniss :)

emil-furniss: Looks good to me. Since ActivityScoreGrade should be agnostic to the underlying entity, there shouldn't really be much difference between the linked vs embedded entities - except for how `fetchLinkedScoreOutOfEntity` modifies the underlying entity which is exactly what you are covering here.

components/d2l-activity-editor/state/activity-score-grade.js
```diff
@@ -1,14 +1,23 @@
...
+	constructor(token) {
+		this.token = token;
+	}
+
+	addToGrades() {
+		this.inGrades = true;
+	}
+
+	async fetch(entity) {
+		await entity.fetchLinkedScoreOutOfEntity(fetchEntity);
```
emil-furniss, line 18: I can understand why we're doing it like this, but I'm honestly not keen on how `fetchLinkedScoreOutOfEntity` is modifying the underlying entity rather than returning a new one. It's just very opaque and side-effect-y.

mdgbayly: I kind of agree, but it was hard to see how to integrate this without doing a full-scale refactoring of all the existing score/grade stuff. With hindsight, it would have been better if the score/grade representation in siren-sdk had been split out from the activity usage from the beginning in the same way that we have split it out in our MobX state. In general, we don't seem to have a good story for how we handle whether something is including as either a linked or embedded sub entity, which for something that has it's own identity should always be an option in a siren response. In this case, we are making that even more complicated by trying to support a selfless sub-entity and a sub entity with a self link using the same code. But I think ultimately we should be looking to refactor to this being a separate siren-sdk entity.

There are other cases where the siren sdk entities allow mutation. They have a mechanism for keeping track of the loading of sub-entities which is primarily designed to allow them to signal changes using their `onchange` callback approach, but for various reasons when we did the mobx integration we chose to not use that `onchange` approach because it was harder to make fit with MobX reactions. So I didn't think it appropriate to use here either, but having the conditionally linked subentity is not too dissimilar, but I agree kind of opaque.

mdgbayly (approved):
Acknowledge that this requirement challenged us for how best to integrate the duality of the linked vs embedded sub-entity. OK to accept this as is. Created a hygiene item to revisit how we handle score information in the Activity Usage, ideally separating it out into it's own siren-sdk entity that we can manage indepedently.

PR #1594, opened 2021-03-06
Ensure check-in entity is refetched

wongvincent:
The checked out working copy of a base quiz needs to be re-fetched when it is checked-in as the working copy should theoretically no longer have a check-in action anymore (and to make sure everything is in sync).
https://trello.com/c/LbKqk61x/327-face-quiz-timing-saving-twice-causes-a-500-error

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js
@@ -12,7 +12,7 @@ export class Quiz {
 		this._checkedOut = null;
 	}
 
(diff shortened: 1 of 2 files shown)


ronan-f (commented):
Does this cause any jank? Like if a user makes changes before the refetch completes?

wongvincent:
Nope. I'm only refetching on the quiz `save`. So unless the fetch for some reason returns something out of sync with what was displayed in the UI before save, there shouldn't be jank.
For our case, our working copy is only in dialogs so there won't be any jank since they'll need to re-open the dialog.

ronan-f (approved):
LGTM

PR #1603, opened 2021-03-11
Refetch wc before checkin

wongvincent:
Problem: When a `PATCH` is made to the timing entity, the existing quiz working copy in the store is not updated, so we do not know if it has a `check-in` action or not.

Solution: Refetch the working copy prior to check in.
The code fix is a bit more complicated because we need to wait for the `PATCH` to the timing endpoint to complete before refetching the quiz working copy, and the timing entity and the quiz working copy are in different stores.

These are the steps of how things will be processed...
1. Say focus is currently on input, and we click OK to check in the dialog
2. The input focus out event causes a `PATCH` request to be sent. This promise is set to the `saving` property in the `quiz-timing` MobX object.
3. The OK click on the dialog causes a call to the `_checkin` function, which will now `await` the timing `saving` property before requesting `GET` on the quiz working copy. The `GET` on the working copy is required in order to know if there will be a check in action or not. (i.e. Did our latest timing `PATCH` cause the wc to be dirty or not)
4. Then after the check in, a working copy timing fetch occurs which updates the summary in the collapsed/expanded accordion.

Also refactored the summary update into a private function.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-timing-container.js
@@ -56,25 +56,17 @@ class ActivityQuizManageTimingContainer extends ActivityEditorWorkingCopyDialogM
 	}
 
 	async _checkinDialog(e) {
+		const { timingHref: dialogTimingHref } = (this.dialogHref && store.get(this.dialogHref)) || {};
+		const dialogTimingEntity = timingStore.get(dialogTimingHref);
+		// Wait for timing entity PATCH requests to complete before checking in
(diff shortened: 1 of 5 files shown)


ronan-f (approved):
I hope this works 🚀 🚀 🚀

wongvincent:
Please 🙏 that this fixes all things (other than the issue on the server side).

Just 🚢 it?

wongvincent:
@ronan-f
IP restrictions doesn't have this problem, because it is only PATCHing when the dialog "Add" button is clicked which also checks in and closes the dialog. So it is already `await`ing the PATCH before doing any sort of checking in.

@mostlyfabulous 
The `attempts` stuff might have the problem I've explained in the PR description. So you may have to do something similar to this fix.

ronan-f (approved):

mostlyfabulous:
Yes, `attempts` has this problem

PR #1610, opened 2021-03-12
Us125748/save editor images

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F508303488440

Added handler to move pasted images from /temp into /content. I couldn't work out how to clear /temp so not sure if the editor will be smart enough to pick up images directly from /content?

Also need to do something to handle any potential upload errors.

I was also concerned about duplicate uploads, e.g. saving multiple times might persist an image multiple times. This doesn't _seem_ to be the case, I assume because the file name is generated based on the temp guid ([here](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlEditor/Renderers/WysiwygEditor/Core/TinyMCEv4/TinyMceController.cs?r=4b476021#86)) but then I wonder is this doing extra work replacing images that don't need to be replaced?

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-html-new-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-new-editor.js
@@ -26,7 +26,7 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 	constructor() {
 		super();
 		this.htmlEditorHeight = '10rem';
-		this.context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
+		this._context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
 	}
 
 	render() {
@@ -61,7 +61,7 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 	}
 
 	_isPasteAllowed() {
-		return this.context.uploadFiles && this.context.viewFiles;
+		return this._context.uploadFiles && this._context.viewFiles;
 	}
 
 	_onContentChange() {
@@ -76,7 +76,7 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 	}
 
 	_uploadFile(file) {
-		const { orgUnitId } = this.context;
+		const { orgUnitId } = this._context;
 
 		return new Promise((resolve, reject) => {
 			D2L.LP.Web.UI.Rpc.Connect(D2L.LP.Web.UI.Rpc.Verbs.POST,
--- a/components/d2l-activity-editor/d2l-activity-html-new-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-new-editor.js
@@ -26,6 +26,7 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 	constructor() {
 		super();
 		this.htmlEditorHeight = '10rem';
+		this.context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
 	}
 
 	render() {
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -51,6 +74,20 @@ class ActivityHtmlNewEditor extends LocalizeActivityEditorMixin(LitElement) {
...
 		}));
 	}
+
+	_uploadFile(file) {
+		const { orgUnitId } = this._context;
+
+		return new Promise((resolve, reject) => {
+			D2L.LP.Web.UI.Rpc.Connect(D2L.LP.Web.UI.Rpc.Verbs.POST,
+				new D2L.LP.Web.Http.UrlLocation('/d2l/lp/htmleditor/tinymce/moveUploadedFile'),
+				{ orgUnitId: orgUnitId, fileId: file.Id },
```
ronan-f, line 84: I didn't need to do anything in terms of fetching the images etc from temp as all that's needed to move a given file is it's ID which we can get from the `files` array

dbatiste: @grant-cleary fyi.  I am concerned about such routes spreading throughout the app components.  I am wondering if perhaps we want to centralize this in a helper in the editor repo.  What are your thoughts on how apps should be doing this?

grant-cleary: Centralizing in a helper seems fine to me. It would certainly make it easier to control the use of some of these routes.

dbatiste: Maybe we put it in the Lms adapter.  🤔

dbatiste: Just thinking that technically, if a FRA consumer needed this, that's where it would need to go anyway.

grant-cleary: Yeah, good point. Right now FRAs call Valence APIs to do this, but if we want to centralize it around our helpers, it likely makes sense to include it there.

mdulat: Sounds like a good idea. Is this something you guys (Grant, Dave) would create, or should Ronan add this to d2l-htmleditor?

grant-cleary: I could take it on. We're trying to balance changes on top of a bigger toolbar rewrite, so it might be easier for us to coordinate changes ourselves.

dbatiste: Thanks Grant.  😄

components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -26,23 +27,41 @@ class ActivityHtmlNewEditor extends LocalizeActivityEditorMixin(LitElement) {
 	constructor() {
 		super();
 		this.htmlEditorHeight = '10rem';
+		this._context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
+		this._filesToReplace = {};
+		this.saveOrder = 500;
```
ronan-f, line 32: I had to set a saveOrder property here as I was facing a problem where the quiz's save method ([this one](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz.js?r=c56fb0b1#135)) was firing before this editor. 

This matters because I changed the dispatching of the `d2l-activity-html-editor-change` event so that it only gets triggered on save. 

If this event has not been fired then the quiz's state will not have been updated with the correct HTML content and so it will save with an outdated description. 

Changing the order here means that state gets set before the quiz's save event is called and so the network request which persists the HTML will contain the correct data.

Setting it to 500 was fairly arbitrary. Not sure if there's a "correct" `saveOrder` for this case. I saw some other editors using 1000/2000 etc.

mdulat (approved):

PR #1612, opened 2021-03-16
Add fr-on files, fix serge config rewrites, fix fallback file names, and remove unused files

bearfriend:
* Added `fr-on` files for all components
* Fixed all component serge rewrite configs
* Swapped `fr` to `fr-fr` and `fr-ca` to `fr` where necessary
* Renamed `cy-gb` to `cy`
* Renamed `da-dk` to `da`
* Renamed `zh` to `zh-cn`
* Removed unused `fi` and `cy-gb` files

`fr-on` is a [new langpack](https://rally1.rallydev.com/#/12024503567ud/custom/399523971908?detail=%2Fportfolioitem%2Ffeature%2F502780395048&fdp=true?fdp=true) that will not be managed through serge. We must copy `fr-ca` to `fr-on`.

Many components were missing rewrite configs for supported languages like `cy-gb` and `da-dk`, so any other region identifier (e.g. `da-*`) would fall back to `en` instead of their base language. I added these rewrites and renamed the files (e.g. `cy-gb` -> `cy`) accordingly. In additon, some rewrite were incorrect (e.g. `fr-fr fr`) and were fixed, along with renaming those files. Both `zh-cn` and `zh-tw` were removed from rewrites as we don't want either one falling back to the other for political reasons. Any other region identifier (`zh-*`) will fall back to `en`.

We do not officially support `fi`, but some specific repos have it for one particular client. It is not managed through serge so was removed as a rewrite, and none of these components actually support/require it so the files were removed as well.


components/d2l-work-to-do/lang/fr.js
```diff
@@ -1,28 +1,28 @@
 export const val = {
-  activitiesAvailable: "Les activités à remettre ou se terminant dans deux semaines sont terminées ! Cochez Afficher tout le travail pour voir les activités à venir.", // 'Empty View' - When widget has no activities to display within the next two weeks, but there are more activities further into the future that can be shown on the full screen view
```
mbardakovD2L, line 2: where are the changes in this file coming from? did this start as a serge PR or did you update the wording yourself?

bearfriend: This came from moving `fr` to `fr-fr` and `fr-ca` to `fr`, so this file changed from being France French to Canadian French. No manual edits were done to any files.

mbardakovD2L: so the new `fr` (formerly `fr-ca`) is now a fallback for the old `fr` (now `fr-fr`)?

mbardakovD2L:
The work-to-do side looks good to me. I'm not sure what the etiquette is for approving a PR that touches multiple teams' code but I can go back to this and approve it once the other teams have had a chance to look

components/d2l-work-to-do/lang/fr.js
bearfriend: Correct!

jeroach (commented):
No concerns for `d2l-quick-eval` and `d2l-quick-eval-widget`

Thanks very much!

JordanCorner:
Changes for `d2l-activity-editor` look good. I think we need an `fr-on` file for [`d2l-activity-assignment-editor`](https://github.com/BrightspaceHypermediaComponents/activities/tree/master/components/d2l-activity-editor/d2l-activity-assignment-editor/lang)

bearfriend:
@JordanCorner Thanks for pointing that out!

PR #1619, opened 2021-03-19
Fire update event on save to ensure correct state

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F508303488440

I need to add this event before moving files etc as otherwise if there are no files present on the HTML editor property then this line will evaluate to false and the changes won't be persisted.

```js
if (!editor || !editor.files || !editor.files.length || !editor.isDirty) return;
```

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-html-new-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-html-new-editor.js
@@ -42,13 +42,16 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 				label-hidden
 				?disabled="${this.disabled}"
 				height="${this.htmlEditorHeight}"
-				paste-local-images="${allowPaste}">
+				?paste-local-images="${allowPaste}">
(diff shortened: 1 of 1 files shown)


mdulat (approved):

components/d2l-activity-editor/d2l-activity-html-new-editor.js
```diff
@@ -42,13 +42,16 @@ class ActivityHtmlNewEditor extends ActivityEditorMixin(LocalizeActivityEditorMi
 				label-hidden
 				?disabled="${this.disabled}"
 				height="${this.htmlEditorHeight}"
-				paste-local-images="${allowPaste}">
+				?paste-local-images="${allowPaste}">
```
ronan-f, line 45: Not sure why it's saying this is a new change... https://search.d2l.dev/xref/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-html-new-editor.js?r=506e0325&mo=1152&fi=45#45

PR #1620, opened 2021-03-19
DE42834 Content > Default font size in new HTML editor should be 12pt

ianlaidlaw:
This pr makes the default font size `12pt` when using the Content Modules HTML editor.

Did some refactor for default values.

image


components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -10,7 +10,18 @@ class ActivityTextEditor extends LitElement {
 			disabled: { type: Boolean },
 			ariaLabel: { type: String },
 			key: { type: String },
-			htmlEditorHeight: { type: String }
+			htmlEditorHeight: {
+				type: String,
+				value: '10rem'
```
ianlaidlaw, line 15: I moved this here from `d2l-activity-html-new-editor.js`
```
this.htmlEditorHeight = '10rem';
```
but I'm unsure if these default values should live in this file or that one. Open to suggestions.

ianlaidlaw: I think on second glance, probably in the other file.

brissons: Either seems fine, but I like that the more `customizable` values are all together, so this looks good!

brissons (approved):

prajapac (approved):

PR #1623, opened 2021-03-23
DE42834 Content > Fix attributes and properties on new html editor

ianlaidlaw:
Previously, we were passing data down as attributes but treating them like properties. This had no visible real negative impact, but cluttered things up with several attributes showing up as `undefined`. Also, there was an instance of a boolean attribute that was not prepended with the `?` so it was not treated like a true boolean attribute.

This pr makes the attributes look more like attributes (lowercase kebab case) and uses `ifDefined` to prevent attributes from being rendered if they don't exist.

### Content Editor - Has values to pass down:

image

### Assignment Editor - Has no values to pass down and will not render undefined:

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -90,9 +90,9 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						.value="${descriptionRichText}"
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
-						htmlEditorHeight="100%"
-						fullPage="true"
-						fullPageFontSize="12pt"
+						html-editor-height="100%"
+						full-page="true"
+						full-page-font-size="12pt"
 					>
 					</d2l-activity-text-editor>
 				</div>
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -91,7 +91,7 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
 						html-editor-height="100%"
-						full-page="true"
+						full-page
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
```diff
@@ -90,9 +90,9 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						.value="${descriptionRichText}"
 						@d2l-activity-text-editor-change="${this._onRichtextChange}"
 						.richtextEditorConfig="${{}}"
-						htmlEditorHeight="100%"
-						fullPage="true"
-						fullPageFontSize="12pt"
+						html-editor-height="100%"
+						full-page="true"
```
dbatiste, line 94: Looks like this is a defined as a Boolean so you should be able to just have `full-page`, instead of `full-page="true"`.

components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -10,18 +11,9 @@ class ActivityTextEditor extends LitElement {
...
-			fullPage: {
-				type: Boolean,
-				value: false
-			},
-			fullPageFontSize: {
-				type: String,
-				value: '14pt'
-			}
+			htmlEditorHeight: { type: String, attribute: 'html-editor-height' },
+			fullPage: { type: Boolean, attribute: 'full-page' },
```
dbatiste, line 15: Should initialize `fullPage` in `constructor`.  I think there are certain scenarios where the analyzer will complain otherwise.

components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -10,21 +11,17 @@ class ActivityTextEditor extends LitElement {
...
-			},
-			fullPage: {
-				type: Boolean,
-				value: false
-			},
-			fullPageFontSize: {
-				type: String,
-				value: '14pt'
-			}
+			htmlEditorHeight: { type: String, attribute: 'html-editor-height' },
```
brissons, line 14: We may need this to default to '10rem' just so that this change doesn't affect assignments. Having said that, it looks like '10rem' is [smaller than the minimum size of the new editor](https://github.com/BrightspaceUI/htmleditor/blob/8a51dd25d0e3bc4404f0ad16795e6826151e194c/htmleditor-mixin.js#L272).

ianlaidlaw: good call, maybe i should bring that default back

quinnbudan (dismissed):

brissons (approved):
lgtm!

PR #1624, opened 2021-03-23
US126109 handle update-total-quiz-points event in quiz editor and ref…

me-louie:
…resh total points UI

- `d2l-activity-quiz-editor` handles the `d2l-question-updated` event triggered in `hmc` .
- `ActivityUsage` calls  `this.scoreAndGrade.fetchUpdatedScoreOutOf(entity, bypassCache = true)` which refetches the `linked score-out-of subentity`  so the Total Points UI gets updated
- refactored the setters in `activity-score-grade` fetch method into a new `load(..)` method

**Related PRs:**
`hmc`: https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/150
`siren-sdk`: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/301

[US126109](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F512322551332&fdp=true?fdp=true): Refresh total points on edit

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F512322551332&fdp=true?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
@@ -53,7 +53,7 @@ class QuizEditor extends AsyncContainerMixin(RtlMixin(LocalizeActivityQuizEditor
 				?isnew="${this.isNew}"
 				?html-editor-enabled="${this.htmlEditorEnabled}"
 				?html-new-editor-enabled="${this.htmlNewEditorEnabled}"
-				@update-total-quiz-points="${this._updateTotalQuizPoints}">
+				@d2l-question-updated="${this._updateTotalQuizPoints}">
 
 				${this._editorTemplate}
 
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
@@ -53,7 +53,7 @@ class QuizEditor extends AsyncContainerMixin(RtlMixin(LocalizeActivityQuizEditor
 				?isnew="${this.isNew}"
 				?html-editor-enabled="${this.htmlEditorEnabled}"
 				?html-new-editor-enabled="${this.htmlNewEditorEnabled}"
-				@d2l-question-updated="${this._updateTotalQuizPoints}">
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js
```diff
@@ -84,5 +85,11 @@ class QuizEditor extends AsyncContainerMixin(RtlMixin(LocalizeActivityQuizEditor
 			</div>
 		`;
 	}
+	_updateTotalQuizPoints() {
+		const activity = store.get(this.href);
+		if (activity) {
+			activity.fetchUpdatedScoreAndGrade(true);
```
me-louie, line 91: maybe I should rename this to `refreshTotalQuizPoints` and the methjod to `refreshScoreAndGrade`?

components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -27,6 +27,13 @@ export class ActivityUsage {
 		}
 		return this;
 	}
+	async fetchUpdatedScoreAndGrade(bypassCache) {
+		const sirenEntity = await fetchEntity(this.href, this.token);
```
mdgbayly, line 31: I think here we might need to be more selective because in this case we do only want to refresh the linked score out of entity. It seems kind of cludgy but I'm not sure what else we can do at this point. We don't want to refresh any other grading information.

So I was thinking this implementation would have to call a more fine-grained method on the `scoreAndGrade` to just force a refetch of `fetchLinkedScoreOutOfEntity` and then to update the `scoreAndGrade.scoreOutOf` with the new value. But I don't think we can change other values as that might overwrite other changes once we start adding grading capabilities to quiz.

me-louie: Thanks for your comment, that makes sense! I made the change

mdgbayly: I don't think we need to fetch a new activityusage sirenEntity here?
I think this could just be handled by passing `this._entity` into your `fetchUpdatedScoreOutOf` method?As all we really want to refetch is the linked siren sub entity representing the score out of?

me-louie: That's right, because that sirenEntity hasn't changed

mdgbayly (approved):

PR #1640, opened 2021-04-01
fix:  US123931 Fixed to work with HTML editor turned off.

ian-wallace:
https://rally1.rallydev.com/#/16349626795d/dashboard?detail=%2Fuserstory%2F482797593636&fdp=true?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-header-footer-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-header-footer-editor.js
@@ -60,17 +60,30 @@ class ActivityQuizManageHeaderFooterEditor extends ActivityEditorMixin(RtlMixin(
 
 		const el = this.shadowRoot.querySelector('d2l-activity-text-editor');
 		const newEd = el.shadowRoot.querySelector('d2l-activity-html-new-editor');
-		const htmlEd = newEd.shadowRoot.querySelector('d2l-htmleditor');
-		htmlEd.html = header;
-		return;
+
+		if (newEd) {
+			const htmlEd = newEd.shadowRoot.querySelector('d2l-htmleditor');
+			htmlEd.html = header;
+		} else {
+			const ed = el.shadowRoot.querySelector('d2l-input-textarea');
+			ed.value = header;
+		}
 	}
 
 	save() {
+		const entity = store.get(this.href);
+
 		const el = this.shadowRoot.querySelector('d2l-activity-text-editor');
 		const newEd = el.shadowRoot.querySelector('d2l-activity-html-new-editor');
-		const htmlEd = newEd.shadowRoot.querySelector('d2l-htmleditor');
-		const entity = store.get(this.href);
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-manage-header-footer-editor.js
```diff
@@ -60,17 +60,30 @@ class ActivityQuizManageHeaderFooterEditor extends ActivityEditorMixin(RtlMixin(
...
 		const newEd = el.shadowRoot.querySelector('d2l-activity-html-new-editor');
-		const htmlEd = newEd.shadowRoot.querySelector('d2l-htmleditor');
-		const entity = store.get(this.href);
-		entity.setHeader(htmlEd.html);
+
+		if (newEd) {
+			const htmlEd = newEd.shadowRoot.querySelector('d2l-htmleditor');
+			entity.setHeader(htmlEd.html);
+		} else {
+			// assuming htmlEd is turned off in user profile.  Deliberately won't handle old html editor.
```
mdulat, line 83: We're assuming quizzing will always use the new editor, so if the LD flag for new editor is off they will get just a text input?

mdulat: Actually, `d2l-activity-text-editor` checks whether the html editor is enabled (via user profile), and whether the new editor is enabled (via the LD flag), and if it's enabled via user profile but LD flag off then it will use `d2l-activity-html-editor` instead of `d2l-activity-html-new-editor`, not `d2l-input-textarea`.

mdulat: It might actually make sense to determine whether html editor is enabled, and whether new html editor is enabled using `d2l-request-provider` the same was it is done by [d2l-activity-text-editor](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-text-editor.js#L26), and then decide which component to access based on that. 

I'm also wondering, for the `reset()` case, could you not just set the `value` of `el` and have that propagate down to whichever editor is contained within `el`?

ian-wallace: 1)  I assumed quizzing was only using the new editor but now that has changed apparently - I will add you to a slack thread.
2)  To your second point on the editor, if there is a flag in play, I'll need to change things further.  I spoke with Joseph and work for handling the old editor would be done with the Footer story.
3)  For reset(), that is an astute comment.  It's bad practice to change values in other components directly.  I spoke with Dave B. about it and we tested the components expecting them to refresh and they did not behave as expected.  One answer is to add reset() methods on the sub components and request a refresh, but I did not put that as part of this PR to keep it simple.  

At this point, there is a lot going on here and we should probably discuss it directly.

mdulat (approved):

PR #1643, opened 2021-04-07
US125506 checkout WC on load and set up activity usage state for WC

mdulat:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F505562070580&fdp=true

### DIFF
--- a/components/d2l-activity-editor/state/activity-usage.js
+++ b/components/d2l-activity-editor/state/activity-usage.js
@@ -13,7 +13,7 @@ configureMobx({ enforceActions: 'observed' });
 export class ActivityUsage extends WorkingCopy {
 
 	constructor(href, token) {
-		super(ActivityUsage, ActivityUsageEntity)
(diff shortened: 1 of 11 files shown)


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -14,19 +14,86 @@ export class ActivityUsage {
...
+			sirenEntity = await this._saving;
+		} catch (e) {
+			return;
+		} finally {
+			this._saving = null;
+		}
+
+		if (!sirenEntity) return;
+		const href = sirenEntity.self();
+		const entity = new ActivityUsage(href, this.token);
```
emil-furniss, line 42: Apart from the class instantiated here, this (and checkout, dirty, and fetch) looks copy-pasted from the quiz state? Is there any way we can reuse that code rather than duplicating it?

emil-furniss (approved):

PR #1652, opened 2021-04-09
feat:  US126437 Added alert for when a header is made visible via FACE.

ian-wallace:
https://rally1.rallydev.com/#/29180338367d/dashboard?detail=%2Fuserstory%2F516914643580&fdp=true?fdp=true


mdulat:
If we are displaying this alert for more than one of the fields (description, header, footer) then are we displaying more than one alert? I thought we were going to create a "part 2" lang term that incorporates the multiple items that will now appear, within a single alert. Maybe confirm with Joseph and Patrick?

ian-wallace:
> If we are displaying this alert for more than one of the fields (description, header, footer) then are we displaying more than one alert? I thought we were going to create a "part 2" lang term that incorporates the multiple items that will now appear, within a single alert. Maybe confirm with Joseph and Patrick?

I was talking to Joseph about this - so right now he is ok with three separate alerts and then maybe we'd combine them all at the end, but the second thought was that these would be pretty rare, so they may be ok as separate...  We can talk again but the current thought is to keep them separate for now.

mdulat (approved):

PR #1655, opened 2021-04-09
WIP: Update .stylelintrc.json with experimental syntax processors

mostlyfabulous:
Would be nice if we could use `??` and `?.` syntax


mostlyfabulous:
Wrong linter, should be polymer instead of stylelint!

PR #1661, opened 2021-04-14
Handle redirecting to the lesson view page from the updated activity usage

CodeBaboon:
## Relevant Rally Links 

[US126672](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F519007142480&fdp=true?fdp=true): Modify activities saveHref to handle new working copy items

## Description

NOTE: Requires https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/313 and https://github.com/Brightspace/lms/pull/9275

Adds logic to notify and update the parent content activity usage when a web link is persisted and the activity usage href may have changed.

Additionally, adds logic to use the new `lesson-view-page` href when navigating back to the lessons page.

## Goal/Solution

The `lesson-view-page` link will allow the FACE UI to navigate back to the correct topic after saving (especially useful for creating a new topic). Currently when creating a new weblink and choosing "Save And Close" the user will receive a 404 Not Found page because the topic id did not exist at the time of loading the page and so it incorrectly attempts to navigate to `-1`.

The `activity-usage` link allows the web link to raise an event with the updated activity-usage href value when saving. This is important when committing a new unsaved weblink for the first time as the activity usage href will change and the parent control needs to be aware so it can fetch this updated entity.

## Video/Screenshots

fix-nav-on-weblink-create

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
@@ -12,6 +12,7 @@ export class ContentWebLink {
 		this.title = '';
 		this.link = '';
 		this.isExternalResource = false;
+		this.activityUsageHref = '';
 	}
 
 	async cancelCreate() {
@@ -34,9 +35,11 @@ export class ContentWebLink {
 
 	load(webLinkEntity) {
 		this._contentWebLink = webLinkEntity;
+		this.href = webLinkEntity.self();
 		this.title = webLinkEntity.title();
 		this.link = webLinkEntity.url();
 		this.isExternalResource = webLinkEntity.isExternalResource();
+		this.activityUsageHref = webLinkEntity.getActivityUsageHref();
 	}
 
 	async save() {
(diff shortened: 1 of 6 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/web-link/state/content-web-link.js
```diff
@@ -49,7 +52,8 @@ export class ContentWebLink {
 		await this._contentWebLink.setWebLinkExternalResource(this.isExternalResource);
 		const committedWebLinkEntity = await this._commit(this._contentWebLink);
 		const editableWebLinkEntity = await this._checkout(committedWebLinkEntity);
-		await this.load(editableWebLinkEntity);
+		this.load(editableWebLinkEntity);
```
quinnbudan, line 55: this used to be awaited, do we not need to wait for this to resolve anymore?

CodeBaboon: `load` is not an `async` function, so the `await` is unnecessary.

quinnbudan (approved):
LGTM with one question!

ianlaidlaw (approved):

brissons (approved):
👍

PR #1663, opened 2021-04-15
US126389 add categories

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F516602039000
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/516602044064

Apologies for a larger PR. This should be the last of the categories work including the dialog for adding a new category and the summary text on the accordion.

### DIFF
--- a/components/d2l-activity-editor/mixins/d2l-activity-editor-dialog-mixin.js
+++ b/components/d2l-activity-editor/mixins/d2l-activity-editor-dialog-mixin.js
@@ -13,11 +13,11 @@ export const ActivityEditorDialogMixin = superclass => class extends superclass 
 
 	handleClose(e) {
 		this.opened = false;
-		e.stopPropagation();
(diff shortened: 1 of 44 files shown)


components/d2l-activity-editor/mixins/d2l-activity-editor-dialog-mixin.js
```diff
@@ -13,11 +13,11 @@ export const ActivityEditorDialogMixin = superclass => class extends superclass
 
 	handleClose(e) {
 		this.opened = false;
-		e.stopPropagation();
+		e && e.stopPropagation();
```
ronan-f, line 16: This was necessary to call this method in a function in which case there will be no event object available.

mostlyfabulous (approved):

PR #1664, opened 2021-04-16
US125502 add WC implementation for creating new grade and changing gr…

mdulat:
…ade name and maxPoints

The plan is to add the WC implementation in small chunks of functionality per PR (while checkin of activity usage is disabled in siren-sdk), then at the end remove current functionality and enable checkin.

This PR adds the functionality to create a new grade, both on the main page, and in the grades dialog. Also to update the grade name and maxPoints when the WC state is in "new-grade".

https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F505558523792&fdp=true


components/d2l-activity-editor/d2l-activity-grades-dialog.js
```diff
@@ -196,10 +213,18 @@ class ActivityGradesDialog extends ActivityEditorMixin(LocalizeActivityEditorMix
...
+		associateGradeEntity.setGradebookStatus(gradebookStatus, scoreAndGrade.newGradeName, scoreAndGrade.scoreOutOf);
+	}
+
 	_dialogRadioChanged(e) {
 		const currentTarget = e.currentTarget;
 		if (currentTarget && currentTarget.value === 'createNew') {
 			this._createNewRadioChecked = true;
+			this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
 		} else if (currentTarget && currentTarget.value === 'linkExisting') {
 			this._createNewRadioChecked = false;
```
wongvincent, line 229: Do we need to call `this._associateGradeSetGradebookStatus(GradebookStatus.ExistingGrade);` here

components/d2l-activity-editor/state/associate-grade.js
```diff
@@ -31,14 +31,41 @@ export class AssociateGrade {
 		this._entity = entity;
```
wongvincent, line 31: I think we may also want to set the initial values of the properties from the entity in here?
Hard to wrap my head around everything.

wongvincent (approved):

components/d2l-activity-editor/d2l-activity-grades-dialog.js
```diff
@@ -213,5 +238,14 @@ class ActivityGradesDialog extends ActivityEditorMixin(LocalizeActivityEditorMix
 		e.target.resize();
 	}
 
+	async _saveAssociateGrade() {
```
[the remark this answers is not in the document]
mdulat: Yup this will be in a future PR.

components/d2l-activity-editor/d2l-activity-grades-dialog.js
mdulat: Yes, in a future PR. For now I'm only handling the "new-grade" and "not-in-gradebook" cases.

components/d2l-activity-editor/state/associate-grade.js
mdulat: Yes good point. I'll add in the next PR, should have at least the gradebook status here currently, but will add everything else as I add handling of the other states.

PR #1666, opened 2021-04-19
Adjusted check to ensure there is a selected category before renderin…

ronan-f:
…g summary

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F516602044064%2Ftasks

This was previously rendering `Category: ` even if there was no selected category.


mostlyfabulous (approved):
Does this default to `No Category` than?

ronan-f:
> Does this default to `No Category` than?

If there's no category selected (including 'No Category') then no summarizer will be rendered. Otherwise it should render: `Category: <category name>` as the summary.

PR #1667, opened 2021-04-20
Add title and due date to content file FACE page

ianlaidlaw:
This pr adds a new Content editor that will be used for Content File activities. Adds support for editing/saving title and due date.

Includes work for the following tickets:
- [US116825](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F392902164372)
- [US124484](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F490070201328)

### SCREENSHOTS / GIFS

image

image

filesfacepage

**NOTE: when making a NEW html file, the page will still default to the weblink view.** This is presumably because a new activity is getting serialized with both the weblink and file links, and weblinks just happen to be above files in this if statement:

image

Resolving this is dependent on [US126671](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F519002105712)

PR #1684, opened 2021-04-28
fix: trigger visual_diff tests manually only, not on pull_request.

mdulat:
Given the frequent visual diff test flakes that prevent unrelated code changes from being merged, I am proposing that these tests not be run on every pull request. Instead, they can be triggered manually whenever a related change needs to be tested. I realize this will not catch random changes to dependent components that may affect the components checked by these tests. Perhaps a cron job may be a better solution. 

Discussion is welcome, I am just proposing one option here. But would like to find a better solution than what we currently have in place.

### DIFF
--- a/.github/workflows/visual-diff.yml
+++ b/.github/workflows/visual-diff.yml
@@ -1,5 +1,5 @@
 name: Visual Diff
-on: pull_request
+on: workflow_dispatch
 jobs:
   build:
     name: Visual Diff


svanherk:
We had another Puppeteer bump on Monday so a one-time diff of very small quick-to-approve changes would be expected.  I assume it's not just that though, but a series of flakey tests as well?  Is it the same tests that fail every time that need to be reworked or skipped until the team that made them has time to investigate?

ianlaidlaw:
> We had another Puppeteer bump on Monday so a one-time diff of very small quick-to-approve changes would be expected. I assume it's not just that though, but a series of flakey tests as well? Is it the same tests that fail every time that need to be reworked or skipped until the team that made them has time to investigate?

It seems to be a few flaky tests. I've [had an older pr](https://github.com/BrightspaceHypermediaComponents/activities/pull/1628) where the goldens fail and I merge, then they fail again, and I merge and I'm back to no changes and all is well.

Also had a [more recent pr](https://github.com/BrightspaceHypermediaComponents/activities/pull/1667) where I attempted to do the same thing but it kept failing and eventually just closed the pr.

I think it might just be the `activity-card` suite that seems to be a problem and is inconsistent. Looking through the failures I've seen it appears to not be consistent which test fails.

In my opinion I haven't found visual diff testing or snapshot testing to be particularly helpful. In my past experience I find that people generally just ignore them and accept new goldens because it can be unclear if a failure is a false negative, especially as teams change or if they are unfamiliar with the components, etc. Looking at [these](https://visual-diff.d2l.dev/screenshots/BrightspaceHypermediaComponents/activities/activity-card/2021-03-24-19.39.27.579-510dc2e6f9274a950f550f3a8ebf8d5ed0e5b059-report.html) [failures](https://visual-diff.d2l.dev/screenshots/BrightspaceHypermediaComponents/activities/activity-card/2021-04-27-15.28.46.499-842279f6999af472864d4ce6a6aa33e90d3b1db0-report.html) I've encountered, I couldn't confidently tell you which one is supposed to be correct or not. Maybe that's a problem with how these tests were set up, or maybe a problem with the component itself and can be improved, which would be great.

I do think getting these tests to skip being run only manually would just be a temporary step between them never being run at all. But again, maybe it's just one flaky suite/component that can be improved and this won't be a problem anymore!

svanherk:
My concern with moving to manual is that it removes the main benefit of `visual-diff` - catching those _unexpected_ regressions you mentioned.  We could change something in `core` that seems fine there, but actually breaks something downstream in an `activities` component.  I agree with @ianlaidlaw - I think this will just lead to them never being run and the goldens falling out of date.  A cron job would be better than manual, but still make it a lot harder to triage what exactly broke things.

> In my opinion I haven't found visual diff testing or snapshot testing to be particularly helpful. In my past experience I find that people generally just ignore them and accept new goldens because it can be unclear if a failure is a false negative, especially as teams change or if they are unfamiliar with the components, etc. Looking at these failures I've encountered, I couldn't confidently tell you which one is supposed to be correct or not. Maybe that's a problem with how these tests were set up, or maybe a problem with the component itself and can be improved, which would be great.

This is good info.  I think what's really going on here is we have a mini monorepo setup.  It makes sense that you don't know which golden is correct if it's not a component you own/work on.  We have found the visual-diff tests extremely helpful with finding regressions in `core`, but we own every component and there is a cost in upkeep to fixing flakey tests when they appear.

Trying to think of ways to combat this - we could add the team that owns/cares about those components as codeowners for the goldens.  Because we now open draft PRs they wouldn't get added right away, but pressing "Ready for review" would get them added and notified about an actual issue/regression or a flakey test to look at.  Right now even if we did break something with a change to `core`, the right people might not even know if they're no longer working in this repo a lot.

@jmurray-d2l @jeroach @vix633 Thoughts here since I think this will mostly affect your team?

mdulat:
It sounds like ideally these tests would run for each PR so that it is easier to determine which change caused the visual-diff change. However the relevant teams should be automatically notified of the test failures, and also PRs should not be prevented from being merged if the tests fail.

So if I'm working on a component that is not related to the visual diffs, my work should not be delayed by these tests, but the interested team should be notified so that they can check the diffs and decide whether new goldens should be generated, or whether other changes need to be applied. Does that sound do-able?

svanherk:
That seems like a good thing to try to me.  I think the workflow would be:
1. You open a PR and visual-diff tests run.  If they fail, you go to that PR and see if the changes are related to you.  If not, you'd hit "Ready for review" which would notify the team based on CODEOWNERS file entries.
2. The interested team can then decide whether:
A. The change is fine and the goldens should be updated (PR merged)
B. The change is not fine and something has broken somewhere and needs to be fixed
C. The test is flakey and should be fixed
D. The test is flakey and not providing the intended value and should be deleted
3. Ideally, if it's option A. you get a response before you need to merge, so another PR doesn't need to be opened just to update the goldens.  But if not, you still merge.  The visual-diff PR will eventually be automatically closed if you merge the PR that opened it, but the report and info should all still be there to investigate.

This probably means more visual-diff PRs in the meantime (if nothing changes, the same PR will just be opened again against the next PR), but those PRs are no longer blocking.  This seems like a good compromise to me, but I'm not in this repo very often so it would still be good to hear from others.  I'm coming at this from the perspective of working on upcoming changes to filter, table etc. - it'd be great to have a quick way to verify these changes didn't affect anything in Quick Eval components.

jeroach:
Sorry for the delay - I'm okay with Stacey's proposed plan.

As a counter to the previous opinion around vdiff tests not being very useful, we've seen several times on Nimbus where they've caught unexpected changes. So I find them quite useful for visual components.

But of course, if vdiff tests that are completely unrelated to the component you're working on fail - that's a pain. So like I said, perhaps we can move forward with what Stacey and others have suggested :)

bearfriend:
I also have found them very useful, even with the flakiness. It's usually fairly obvious if a test flaked or if something you did could have caused the change, but I definitely like the idea of not just assuming something isn't related to your changes. That seems like almost the whole point of these tests; to catch changes to things you didn't anticipate or understand the relationship between.

There are two ways I've seen things fail:
1. Pure flake: I think the main issue is with timing/state. Things aren't exactly deterministic in UI and there are probably a decent number of assumptions in the tests that _usually_ work but sometimes just don't get into the right state quickly enough. Probably easy enough to start taking note of particular tests that fail and adding some wait-type functions.
2. Weird render changes: This usually looks like a random change to the way something is rendered by a browser, like antialiasing or something, which still seems odd that it would/could happen at random, but I've seen in multiple times, and even fix itself on a re-run. I suppose this could also be a timing/state thing at least some of the time.

mdulat:
@svanherk regarding workflow step 1:

> You open a PR and visual-diff tests run. If they fail, you go to that PR and see if the changes are related to you. If not, you'd hit "Ready for review" which would notify the team based on CODEOWNERS file entries.

Would it be possible for the PR to be set to "Ready for review" automatically to notify code owners, instead of PR owners having to click that explicitly? Because I think that if we are able to merge our PR without those tests passing people may ignore those diff PRs, especially those working on components that don't have Visual Diff tests (so failing tests will never be related to changes made to those components).

svanherk:
> Would it be possible for the PR to be set to "Ready for review" automatically to notify code owners, instead of PR owners having to click that explicitly?
So it _used_ to work that way, but we _just_ switched to draft PRs to cut down on the noise this was causing in `core` when someone was writing/changing visual-diff tests or if a flake did happen.  We figured draft PRs were a more accurate representation because it really is a "guess" until someone takes a look.

But maybe it depends on the repo context - if a team is codeowners over a whole repo they'll want less noise around it, versus a situation where different teams own different parts of a repo.  We could make it configurable - I'll raise that idea tomorrow

svanherk:
Added the feature discussed above to turn off draft PRs for this repo: https://github.com/BrightspaceHypermediaComponents/activities/pull/1694

svanherk:
@jeroach Who all makes sense to add to the CODEOWNERS file?  We've really liked [using a team](https://github.com/BrightspaceUI/core/blob/master/CODEOWNERS#L1) instead of listing people individually because once one person approves, the rest are taken off the review.  But if only a few members make sense it might be overkill to make a new team

jeroach:
> @jeroach Who all makes sense to add to the CODEOWNERS file? We've really liked [using a team](https://github.com/BrightspaceUI/core/blob/master/CODEOWNERS#L1) instead of listing people individually because once one person approves, the rest are taken off the review. But if only a few members make sense it might be overkill to make a new team

I'd say myself, @JamieTran , @Scy-D2L , and @vix633 are the Nimbus devs working primarily on front end these days.

PR #1687, opened 2021-04-28
Updated categories implementation to ensure summary reflects selected…

ronan-f:
… category

Fix addresses [this trello card](https://trello.com/c/6Mz8lmGe/366-categories-when-a-category-is-changed-and-accordion-is-collapsed-the-summarizer-text-shows-the-old-category-until-the-assignment).

Previously the summarizer was only updating once the 'save' network request competed. Now it should render based on the current UI state.


mdulat (approved):
LGTM

PR #1695, opened 2021-04-29
Updating Translations 2105 28/04/2021 (#1685)

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F600538373508

Certfix for translations


ronan-f:
> Looks good to me. Hopefully the change of `categories` to `category` will be translated soon.

The flag for categories will be off for 20.21.5 so we should be good 🤞 It's really only the Spanish translation that will be available

PR #1702, opened 2021-04-30
Add Nimbus to codeowners

jeroach:


Scy-D2L (dismissed):

JamieTran (dismissed):

CODEOWNERS
```diff
@@ -1,3 +1,5 @@
 *       @majones
+/components/d2l-quick-eval/ @BrightspaceHypermediaComponents/nimbus
```
svanherk, line 2: I _think_ if this is changed to `d2l-quick-eval/` and `d2l-quick-eval-widget/`, that would cover the component files and the test files (which I think covers almost all of the goldens)

jeroach: Makes sense - I'll update it. Thanks!

svanherk (approved):
Looks good to me

PR #1706, opened 2021-05-05
Enable fr-on

bearfriend:
In https://github.com/BrightspaceHypermediaComponents/activities/pull/1612 I added `fr-on` files for all components in this repo, but I left enabling it up to the teams responsible for each component. I don't think I was very clear about that, so for any components not using the dynamic localization mixin I've manually enabled `fr-on` here. Though I have done a cursory check to make sure the files exist and the correct patterns were followed, 🚨  _I have not tested these_.

Please test these components if your team is responsible for them, or tag someone you know is.

I have no reason to believe any of the affected components required `fr-on` in the May release, or even had any differences from `fr-ca` yet, so I don't see why a hotfix would be necessary, but if you believe it is please let me know.

### DIFF
--- a/components/d2l-activity-card/ActivityCardLocalize.js
+++ b/components/d2l-activity-card/ActivityCardLocalize.js
@@ -8,6 +8,7 @@ import es from './lang/es.js';
 import eses from './lang/es-es.js';
 import fr from './lang/fr.js';
 import frfr from './lang/fr-fr.js';
+import fron from './lang/fr-on.js';
 import ja from './lang/ja.js';
 import ko from './lang/ko.js';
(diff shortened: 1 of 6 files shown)


JordanCorner (approved):
The changes generally look fine, side note it would be nice to see everyone use the dynamic mixin and get rid of these ugly switch statements. 

I don't own any of these components specifically, and I'm not sure who to tag, but maybe a post in the #d2l-activities-repo channel in slack at least so people are aware.

jeroach (dismissed):

jeroach (approved):

JordanCorner (approved):

PR #1711, opened 2021-05-06
Content module editor: Description field text defaults to "undefined" for new modules

ianlaidlaw:
~~Because of the recent addition of the `live()` [directive on the new-html-editor component in this repo](https://lit-html.polymer-project.org/guide/release-notes/1.2.0.html#live-directive), the initial text was not getting updated from it's initial state of `undefined` to `""`. I'll be honest, I don't understand why not as `live` claims it does a strict equality check to update and so it should be updating. I suspect something is not setup quite right for how some other editors want to update (or not update) their content. This is also only happening for new content which makes this lifecycle all the more confusing.~~

~~This is a bit of a band-aid fix, as we should still be respecting `moduleEntity.descriptionRichText` (which is empty string) but for whatever reason the initial render is passing down that `undefined` and then when we get `""` and pass that down, `live` is rejecting the update.~~

Binding `value` as a property (as it is done for the other editors) seems to make `live` respect these changes.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -56,7 +56,7 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 
 	render() {
 		const moduleEntity = moduleStore.getContentModuleActivity(this.href);
-		let descriptionRichText = undefined;
+		let descriptionRichText = '';
 
 		if (moduleEntity) {
 			this.skeleton = false;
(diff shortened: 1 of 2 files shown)


brissons (dismissed):
lgtm! Yeah this is weird, but this will definitely work until we figure out why it's behaving this way.

quinnbudan (dismissed):

quinnbudan:
Based on the [live playground](https://lit.dev/playground/#sample=examples/directive-live) I feel like I'm not sure if this is an issue with `live`? It seems like `live` is for resetting the user-inputted value to the last rendered value, but I might be misunderstanding it

brissons (approved):
👏

ianlaidlaw:
Hi, @ian-wallace I noticed you were the one who added `live` and I was wondering if you could confirm if these changes are okay. To explain our problem, when using only the new editor our value was not getting overridden when we fetched our entity (it was staying at undefined) and it seemed to be because of the live directive. I could see the values were different but live refused to update the value. I noticed the new editor wasn't binding `value` as a property (like the other ones were), and changing that now seems to be triggering live and our value is getting set properly.

quinnbudan (approved):
LGTM! Nice catch!

CodeBaboon (approved):
LGTM

prajapac (approved):

ian-wallace (approved):
Nice catch.  This looks correct to me.

PR #1718, opened 2021-05-10
Mdulat/us127874 fix heading sizes

mdulat:
This makes all headings within accordions consistent.

https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F603233026972&fdp=true


mdulat:
> I'm wondering what elements should be used to express these headers from an accessibility point of view? Is there a reason to go with divs instead of `h2/h3`?

Headers may make more sense, but since the FACE page doesn't have an H1 Carin always gets confused seeing H3, H4, and asks where is the H1 and H2. We decided to use `div` for these instead of `label` since they are not labelling a specific input. The actual inputs below these "headings" do have their own appropriate labels.

ronan-f (approved):
Looks good to me. This _might_ have an impact on some TC selectors but we'll see during the next run. I'm monitoring it for this week anyway so can reference this PR in case anything breaks.

dustxd (approved):

PR #1725, opened 2021-05-12
Handled reset case for assignments without categories

ronan-f:
https://trello.com/c/Mbb5PkHC/377-assignment-aligned-to-new-category-even-on-cancel

Follow up from last PR to cover the case where we need to reset the category on an assignment which loads with "No Category".

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-categories.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-categories.js
@@ -2,6 +2,8 @@ import { action, configure as configureMobx, decorate, observable } from 'mobx';
 import { CategoriesEntity } from 'siren-sdk/src/activities/assignments/CategoriesEntity.js';
 import { fetchEntity } from '../../state/fetch-entity.js';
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-categories.js
```diff
@@ -41,10 +43,14 @@ export class AssignmentCategories {
...
 
+		if (this.initialCategory === UNSELECTED_ID) {
+			await this._entity.save({ categoryId: UNSELECTED_ID });
+		}
+
 		this.initialCategory && await this._entity.save({ categoryId: this.initialCategory.properties.categoryId });
 	}
 
-	async save() {
+	async save(shouldReset) {
```
dustxd, line 53: What's the difference between this `save` and the `save` on Line 47?

ronan-f: Oh line 47 is handling the reset case when a user cancels their changes and so it behaves slightly differently. Instead of accumulating any and all changes to the categories it just resets whatever category is currently chosen back to it's initial value.

dustxd (approved):

PR #1727, opened 2021-05-13
US128071 - don't load file entity if there's no link

quinnbudan:
## Relevant Rally Links 

- [US128071](https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fuserstory%2F604152072952&fdp=true?fdp=true): (FE) FACE HTML file - Don't load file usage link if not present, change where to check for type

## Description

To handle the fact that we don't want to have a file entity usage link when the file activity is new, we have removed the link from the entity response when it's a new entity. The goal is to update FACE to check if there is a file entity usage link before trying to load it, rather than checking the URL to determine if it's a new activity.

## Goal/Solution

The solution here is just to get the `fileEntityHref` and check if it has a value before proceeding to load its content.

## Video/Screenshots

N/A

## Related PRs

- https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/323 the SDK PR needs to be merged first

## Remaining Work

- In the future we will want to factor out this HTML specific stuff from the generic component. We have a story for that [here in US128197](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F604631601020)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
@@ -35,10 +35,10 @@ export class ContentFile {
 
 			entity = await this._checkout(entity);
 
-			// TODO: This will have to change once the BE changes (US128070)
-			if (!this._isNewFile(entity) && entity.getFileType() === FILE_TYPES.html) {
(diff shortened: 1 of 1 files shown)


CodeBaboon (approved):
image

ianlaidlaw (approved):

PR #1734, opened 2021-05-18
Default to font families in Content editors

ianlaidlaw:
This pr will tell the content html editors to use a certain font family. They were not being applied as we set our editor to "fullPage" so we can take advantage of customization.

One unfortunate thing is that manually setting the font family like this does not make it select it from the drop down list, so even though I am typing in Lato by default, I still see a different Lato in the dropdown menu.

image

Note: I had the font flipped when I made these screenshots, Modules should be Lato, Files should be Verdana.

MODULE EDITOR:

image

HTML EDITOR:

image


DiljotSG (approved):

quinnbudan (approved):
LGTM!

prajapac (approved):

brissons (approved):

PR #1752, opened 2021-05-27
DE43733: Clicking back or cancel with changes in editor does not trigger discard changes popup

brissons:
## Relevant Rally Links 

[DE43733:](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F600359414369&fdp=true) FACE HTML > clicking 'back' with changes in editor does not trigger discard changes popup

## Description

Clicking back or cancel with changes in editor was not triggering the 'Discard changes? Are you sure you want to discard your changes?'. This was happening for two reasons:

1. The current `dirty()` method was only comparing the `ContentFileEntity` values, which is the title and the description of the file activity, not the actual file content in the html editor.
2. The html editor triggers the save event on `blur` rather than on `change`, which is being debounced. If the user is working on the html editor then immediately presses 'Cancel' or 'Back', the changes in the editor won't be detected since it is being debounced.

## Goal/Solution

These two issues were addressed accordingly:

1. Added `persistedFileContent` to `content-file` which holds the most recent saved content on disk (or in working copy). This was then added to the` dirty() `method to compare to what's in the actual html editor. The comparison is done separately from the `contentFileEntity` since the `contentFileEntity` does not contain the actual file data.
2. A separate `onChange` method without the debouncer is used for the new editor, with the original `onChange` method added only to the old editor. This way, there is is no delay for saving the page content on blur (while maintaining a debouncer for the old editor).

## Remaining Work

There is one remaining issue where the discard changes popup window will be incorrectly triggered if a new html document is made. If the user makes a new html document, triggers the blur event on the editor, then hits cancel, the `dirty()` method will compare the persistedFileContent (which is "" since it is new) and what is in the editor (which will now have the head and body tags). Since this issue also affects modules, this will be addressed in a separate defect.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
@@ -15,6 +15,7 @@ export class ContentFile {
 		this.token = token;
 		this.title = '';
 		this.activityUsageHref = '';
+		this.persistedFileContent = '';
 		this.fileContent = '';
 		this.fileType = null;
 	}
@@ -24,7 +25,7 @@ export class ContentFile {
 	}
 
 	get dirty() {
-		return !this._contentFileEntity.equals(this._makeContentFileData());
+		return !(this._contentFileEntity.equals(this._makeContentFileData()) && this.persistedFileContent === this.fileContent);
 	}
 
 	async fetch() {
@@ -55,6 +56,7 @@ export class ContentFile {
 		this.href = contentFileEntity.self();
 		this.activityUsageHref = contentFileEntity.getActivityUsageHref();
 		this.title = contentFileEntity.title();
+		this.persistedFileContent = fileContent;
 		this.fileContent = fileContent;
 		this.fileType = contentFileEntity.getFileType();
 		this.fileHref = contentFileEntity.getFileHref();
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -66,15 +68,17 @@ export class ContentFile {
...
 		if (this._contentFileEntity.getFileType() === FILE_TYPES.html) {
 			const htmlEntity = new ContentHtmlFileEntity(this._contentFileEntity, this.token, { remove: () => { } });
 			await htmlEntity.setHtmlFileHtmlContent(this.fileContent);
+			fileContent = this.fileContent;
 		}
 
 		const committedContentFileEntity = await this._commit(this._contentFileEntity);
 		const editableContentFileEntity = await this._checkout(committedContentFileEntity);
-		this.load(editableContentFileEntity);
+		this.load(editableContentFileEntity, fileContent);
```
brissons, line 81: Something I noticed as well. We weren't updating the `fileContent` before, but this way needs to change as well. With our other face pages, I think after we save we refetch and repopulate to make sure what is saved reflects what is on the page. We'll eventually want to re-fetch the `fileContent` from the saved html entity, but this should probably go in a separate ticket.

ChrisKraljevicD2L (approved):

quinnbudan (approved):
Looks great!

PR #1753, opened 2021-05-27
feat: Translation Updates (#1748)

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F605195643028

Merging lang term changes to cert branch that were introduced here: https://github.com/BrightspaceHypermediaComponents/activities/pull/1748


wongvincent (approved):
i'm afraid to ask. but we don't need to update fr-on right

ronan-f:
> i'm afraid to ask. but we don't need to update fr-on right

Scary question. In theory we should let fr-ON terms fall back to base french unless the client manually overrides them. There is an open oslo bug that may affect fr-ON but it's known and I believe team USA are working on it https://rally1.rallydev.com/#/29180338367ud/dashboard?detail=%2Fdefect%2F3d089464-4b0c-48ea-856d-2b9e407b0d47

PR #1769, opened 2021-06-02
US118647 - Replaced simple-overlay with d2l-dialog-fullscreen component in rubrics list

annabelsegalld2l:
The d2l-simple-overlay component is being replaced with the d2l-dialog-fullscreen component. The functionality of the page should not be affected. 

Here's a gif of the new look: 
add_rubric_new

I noticed that the "Edit rubric" title is part of the body, rather than being in the header. We can move that to the header, but I decided to keep it as it was for this PR(the overlay header was also empty).

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
+++ b/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
@@ -78,6 +78,7 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 
 			<d2l-dialog-fullscreen
 				id="create-new-association-dialog"
+				title-text = ${this.localize('rubrics.hdrCreateRubric')}
 				@d2l-dialog-close="${this._clearNewRubricHref}">
 				${this._renderRubricEditor()}
 				<d2l-button slot="footer" primary  @click="${this._attachRubric}">
@@ -232,7 +233,8 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 				<d2l-rubric-editor
 					.href="${this._newlyCreatedPotentialAssociationHref}"
 					.token="${this.token}"
-					title-dropdown-hidden>
+					title-dropdown-hidden
+					title-hidden>
 				</d2l-rubric-editor>`;
 		} else {
 			return html``;
--- a/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
+++ b/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
@@ -3,7 +3,7 @@ import 'd2l-associations/add-associations.js';
 import 'd2l-rubric/d2l-rubric';
 import 'd2l-rubric/d2l-rubric-title';
 import 'd2l-rubric/editor/d2l-rubric-editor.js';
-import 'd2l-simple-overlay/d2l-simple-overlay.js';
+import '@brightspace-ui/core/components/dialog/dialog-fullscreen.js';
 import '@brightspace-ui/core/components/dropdown/dropdown.js';
 import '@brightspace-ui/core/components/dropdown/dropdown-content.js';
 import '@brightspace-ui/core/components/menu/menu.js';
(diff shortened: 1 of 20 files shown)


components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -76,23 +76,17 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 
 			${this._renderDefaultScoringRubric(entity)}
 
-			<d2l-simple-overlay
+			<d2l-dialog-fullscreen
 				id="create-new-association-dialog"
-				close-simple-overlay-alt-text="${this.localize('rubrics.btnClose')}"
```
svanherk, line 81: Looks like we can cleanup the `rubrics.btnClose` langterm

svanherk:
> I noticed that the "Edit rubric" title is part of the body, rather than being in the header. We can move that to the header, but I decided to keep it as it was for this PR(the overlay header was also empty).

Huh that's interesting.  Looks like the "Edit Rubric" is attached to the rubric editor so maybe it was done because having the two titles would look silly, but it also looks like there's a slot that you could maybe override to hide the editor title.  The dialog titles is an `h2` versus this title is just a `div` so not sure the screen reader experience difference.  We also say the `title-text` is required on the dialog, so maybe @dbatiste can weigh in there...

svanherk:
Is this used in FACE?  This looks good to me but getting someone from that project's eyes on this would be good as well

components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -76,23 +76,17 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 
 			${this._renderDefaultScoringRubric(entity)}
 
-			<d2l-simple-overlay
+			<d2l-dialog-fullscreen
```
annabelsegalld2l, line 79: One thing we can do to fix the empty title is consider adding that no-header option. I know that was in development by Dave B but the PR was never merged.

dbatiste:
> We also say the `title-text` is required on the dialog, so maybe @dbatiste can weigh in there...

Yeah, we should try to use the title - besides providing a more balanced layout, it labels the dialog element.  The a11y experience likely varies by browser + screen reader combination.

annabelsegalld2l:
I added @mdulat from FACE since she is last the last commit to this file. Feel free to add more reviewers if needed Margaret.

mdulat (approved):
LGTM.
I think the reason the title is in the body is because it's part of the `d2l-rubric-editor` component, though it should be easy enough to leave out the header as an option. I believe it was also left in there because Design wanted it to be inline with the status, and we couldn't put the status into the dialog header. So maybe check with Design before moving it into the dialog header, it does seem like the better a11y solution.

mdulat (approved):

components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -76,23 +76,18 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
...
 				id="create-new-association-dialog"
-				close-simple-overlay-alt-text="${this.localize('rubrics.btnClose')}"
-				no-cancel-on-outside-click
-				@d2l-simple-overlay-close-button-clicked="${this._clearNewRubricHref}"
-				@d2l-simple-overlay-canceled="${this._clearNewRubricHref}"
-			>
+				title-text = ${this.localize('rubrics.hdrCreateRubric')}
```
svanherk, line 81: In the slack thread Patrick mentioned changing this name just for FACE... is `d2l-activity-rubrics-list-container` only used in FACE?  Or whatever uses that only used in FACE?

mdulat: Currently it's only used in FACE, yes.

annabelsegalld2l: Yep, this is d2l-activity-editor, which is only used by the new FACE experience.

svanherk (approved):

PR #1770, opened 2021-06-02
Only show due date subtle button for modules, expand to show field for other content types

CodeBaboon:
## Relevant Rally Links 

[US128149](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F604435412380&fdp=true?fdp=true): FACE pages - due dates component - default to expanded state (but not on modules)

## Related PRs
Add skeletons for due dates under expanded or not scenarios: https://github.com/BrightspaceHypermediaComponents/activities/pull/1771

## Description
​
We don't want to hide the due date field by default when creating or editing non-module content items. But for modules we still want to default to the subtle button which needs to be clicked to display the date/time fields.
​
## Goal/Solution
​
I've replaced the internal `_showAddDueDateBtn` with a public `expanded` attribute. It defaults to false but then in the `_renderDueDate` function I'm passing in an override for the types that we want to default to expanded.

## Video/Screenshots
​
Before:
before-expanded-due-date

After:
after-expanded-due-date
​
## Remaining Work

US128149 calls for ensuring skeletonization of the due date controls. This change will be posted in a separate PR. UPDATE: #1771

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
@@ -63,7 +63,7 @@ class ContentEditorDetail extends MobxLitElement {
 					.href="${contentActivityHref}"
 					.token="${this.token}"
 				>
-					${this._renderDueDate()}
+					${this._renderDueDate(true)}
 				</d2l-activity-content-web-link-detail>
 			`;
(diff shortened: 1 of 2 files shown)


quinnbudan (approved):
LGTM!

ianlaidlaw (approved):

brissons (approved):
lgtm!

PR #1772, opened 2021-06-02
US125507 - Check in dialog

wongvincent:
This PR switches the grades dialog from `ActivityEditorMixin` to `ActivityEditorWorkingCopyDialogMixin` which has a bunch of helpers to help open/close the dialog and checkout/checkin.

The PR also simplifies the showing of the "+ Grade Category" button a little.

This PR also tries to separate some of the working copy logic and non-working copy through if/else.

There are 2 issues with this PR.
1: the `async` state on the dialog should be `true` to show the loading spinner.
2: when the selectbox LD flag is off, it is still checking out because the entity contains a checkout action, although it doesn't patch or checkin or do anything with the checked out entity. But I should prevent that somehow probably.

https://rally1.rallydev.com/#/29180338367ud/dashboard?detail=%2Fuserstory%2F505562087036&fdp=true?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades-dialog.js
@@ -6,7 +6,7 @@ import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/inputs/input-radio-spacer.js';
 import { sharedAssociateGrade as associateGradeStore, shared as store } from './state/activity-store.js';
 import { css, html } from 'lit-element/lit-element';
-import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
+import { ActivityEditorWorkingCopyDialogMixin } from './mixins/d2l-activity-editor-working-copy-dialog-mixin.js';
 import { bodySmallStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
 import { GradebookStatus } from 'siren-sdk/src/activities/associateGrade/AssociateGradeEntity.js';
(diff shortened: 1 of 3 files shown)


mdulat:
There was a reason I didn't use the `ActivityEditorWorkingCopyDialogMixin`, I don't recall the details now, but I think it had to do with the fact that the main face page is already checked out, then when you open the dialog you have to check out again (with parentWorkingCopyId). 

Also it needs both the main page's checked out href and the original activityUsageHref, because when it is in `new-grade` state then changes to the assignment name or max-points will only show up on the `activityUsageHref` in the store, and not in the checked out href.

If it works with `ActivityEditorWorkingCopyDialogMixin` that's great, just make sure it's actually making all the right network calls and using the correct `checkedOutHref`, etc. Test changing the assignment name or maxPoints while in new-grade state, then open the dialog and make sure it is using the correct name and max-points.

wongvincent:
> There was a reason I didn't use the ActivityEditorWorkingCopyDialogMixin, I don't recall the details now, but I think it had to do with the fact that the main face page is already checked out, then when you open the dialog you have to check out again (with parentWorkingCopyId).
Yes, it is expected that opening the dialog should `checkout` a 2nd working copy from the 1st working copy. When the dialog 'OK' is pressed, we can checkin our 2nd working copy changes into the 1st working copy.
Or if the dialog `Cancel` is pressed, then we do nothing with the 2nd working copy. When the dialog is re-opened, a fresh 2nd working copy is created by again doing a `checkout` on the 1st working copy.

wongvincent:
> Also it needs both the main page's checked out href and the original activityUsageHref, because when it is in new-grade state then changes to the assignment name or max-points will only show up on the activityUsageHref in the store, and not in the checked out href.
Yep, I noticed this, so I had to keep it the same so that both an `href` and a `baseActivityUsageHref` are passed into the `d2l-activity-grades-dialog`.

mdulat (approved):

PR #1777, opened 2021-06-04
US125507 - clean up all the hrefs!

wongvincent:
It's magic.

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F505562087036&expandApp=486232622048&fdp=true?fdp=true

We are only passing in the base activity usage as the `href` to the `grades-dialog`, and no other `href`s. The working copy will be checked out in the `score-editor` which is the parent of the `grades-dialog`. This 1st working copy `href` is stored as a property in the MobX object of the base activity usage.

With the selectbox flag on, aka using working copy, the `grades-dialog` component will mostly be making references to the `dialogHref`. The `dialogHref` which is a string set in the `workingCopyDialogMixin` via a checkout action off the first working copy. It knows the `href` of the first working copy because it can retrieve it from a property in the MobX object of the base activity usage.

With the selectbox flag off, aka not using working copy, the `grades-dialog` component will only be using the `href`. It will not use the `checkoutHref` or the `dialogHref`.

The `grades-dialog` component can still get the "Grade Out Of" and activity "Name" by using the `href` that is passed in and getting the values from the `scoreAndGrade` object of that activityUsage MobX object.

It's basically magic.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
@@ -19,7 +19,6 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 
 	static get properties() {
 		return {
-			baseActivityUsageHref: { type: String, attribute: 'base-activity-usage-href' },
 			_createNewRadioChecked: { type: Boolean },
 			_canLinkNewGrade: { type: Boolean },
 			_hasGradeCandidates: { type: Boolean },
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
```diff
@@ -162,12 +159,12 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
...
 
 		if (!this._createSelectboxGradeItemEnabled) {
-			this.openDialog(null, true);
+			this.openDialog();
 		}
 	}
 
 	async _associateGradeSetGradebookStatus(gradebookStatus) {
-		const scoreAndGrade = store.get(this.baseActivityUsageHref).scoreAndGrade;
+		const scoreAndGrade = store.get(this.dialogHref).scoreAndGrade;
```
mdulat, line 167: Is this going to get the updated name and points values? Or should this be using `this.href`?

wongvincent: Yes. Thanks

mdulat (approved):
Looks good. I like magic :)

PR #1782, opened 2021-06-08
DE43805 Fix rubrics behavior on save-in-place

emil-furniss:
Save-in-place was throwing off the existing rubric state management, so this just re-fetches the associations on save-in-place so the UI reflects the correct state (equivalent to loading the assignment from scratch).

There was also a small bug where the rubric 'X' removal button was relying on the event to provide the rubric id, but if the focus was elsewhere when clicking the target event was an unexpected element and the button wouldn't respond. Since the association is available at the time of the click, this switches it to just use the known association href directly.

https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fdefect%2F600844552471&fdp=true


ronan-f (approved):
Makes sense to me

PR #1803, opened 2021-06-11
Fix ending dialog tag typo

annabelsegalld2l:
The ending tag should be closing the d2l-dialog tag, not d2l-dialog-fullscreen. This was auto-fixed by the HTML, so it wasn't actually causing any errors or rendering problems.


mdulat (approved):
nice catch!

PR #1804, opened 2021-06-11
Fix loading skeleton on the HTML file type page

ianlaidlaw:
This pr creates a new component which just consists of a label and a full height div that will perpetually be in a loading state. On our file page, we use info from the entity to determine which elements to render as this page will eventually have multiple options based on the file type. As a result, before we got the entity the view would be blank so no loading skeletons would show. Now, before we receive and examine our entity we will default to this loading view.

Note: the loading state on the due date is not in sync with the other elements as we slot that in for some reason I can't remember. It's unrelated to this change, it happened before as well.

loading

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
@@ -1,4 +1,5 @@
 import '../shared-components/d2l-activity-content-editor-title.js';
+import './d2l-activity-content-file-loading.js';
 import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { ContentFileEntity, FILE_TYPES } from 'siren-sdk/src/activities/content/ContentFileEntity.js';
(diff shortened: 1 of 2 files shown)


ChrisKraljevicD2L (approved):

DiljotSG (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -174,6 +177,15 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 			</div>`;
 	}
 
+	_renderUnknownLoadingFileType() {
+		return html`
+			<d2l-activity-content-file-loading
```
ianlaidlaw, line 182: regular reminder i am bad at naming things. open to suggestions

CodeBaboon: I. along with my level 3 Cleric named Church Giver, have no qualms with your chosen name.

CodeBaboon (approved):

quinnbudan (approved):

PR #1822, opened 2021-06-18
fix: don't show category button if no categories, don't show type/sch…

mdulat:
…eme button if no schemes


wongvincent (approved):

wongvincent (commented):
Hmmm should the buttons be hidden or disabled

PR #1823, opened 2021-06-18
US129035 - fix grades/categories/schemes items being individually fetched

wongvincent:
https://rally1.rallydev.com/#/29180338367d/dashboard?detail=%2Fuserstory%2F601489857733&fdp=true?fdp=true

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
@@ -118,7 +118,7 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 			const associateGradeHref = this._associateGradeHref;
 			if (!associateGradeHref) return;
 
-			let associateGrade = await this._fetch(() => associateGradeStore.fetch(associateGradeHref, this.token));
+			const associateGrade = await associateGradeStore.fetch(associateGradeHref, this.token);
 			if (!associateGrade) return;
 
 			this._createNewRadioChecked = associateGrade.gradebookStatus === GradebookStatus.NewGrade || associateGrade.gradebookStatus === GradebookStatus.NotInGradebook;
@@ -188,9 +188,8 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 				await this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
 
 				//the entity resets the gradeType to numeric when switching from link existing to create new
-				if (gradebookStatus === GradebookStatus.NewGrade) {
-					await associateGrade.getGradeSchemes(true);
-				}
+				const associateGrade = associateGradeStore.get(this._associateGradeHref);
+				await associateGrade.getGradeSchemes(true);
 			}
 		} else if (currentTarget && currentTarget.value === 'linkExisting') {
 			this._createNewRadioChecked = false;
--- a/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
@@ -167,9 +167,6 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 		if (!scoreAndGradeBase || !associateGrade) return;
 
 		await associateGrade.setGradebookStatus(gradebookStatus, scoreAndGradeBase.newGradeName, scoreAndGradeBase.scoreOutOf);
-		if (gradebookStatus === GradebookStatus.NewGrade) {
-			await associateGrade.getGradeSchemes();
-		}
 	}
 
 	get _canCreateNewGrade() {
@@ -187,7 +184,14 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 		const currentTarget = e.currentTarget;
 		if (currentTarget && currentTarget.value === 'createNew') {
 			this._createNewRadioChecked = true;
-			this._createSelectboxGradeItemEnabled && this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
+			if (this._createSelectboxGradeItemEnabled) {
+				await this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
+
+				//the entity resets the gradeType to numeric when switching from link existing to create new
+				if (gradebookStatus === GradebookStatus.NewGrade) {
+					await associateGrade.getGradeSchemes(true);
+				}
+			}
 		} else if (currentTarget && currentTarget.value === 'linkExisting') {
 			this._createNewRadioChecked = false;
 			this._createSelectboxGradeItemEnabled && this._associateGradeSetGradebookStatus(GradebookStatus.ExistingGrade);
--- a/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
+++ b/components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
@@ -116,20 +116,20 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 			await this.openDialog();
 
 			const associateGradeHref = this._associateGradeHref;
-			if (associateGradeHref) {
-				await this._fetch(() => associateGradeStore.fetch(associateGradeHref, this.token));
+			if (!associateGradeHref) return;
 
-				const associateGrade = associateGradeStore.get(associateGradeHref);
-				if (!associateGrade) return;
+			let associateGrade = await this._fetch(() => associateGradeStore.fetch(associateGradeHref, this.token));
+			if (!associateGrade) return;
 
-				this._createNewRadioChecked = associateGrade.gradebookStatus === GradebookStatus.NewGrade && this._canCreateNewGrade;
+			this._createNewRadioChecked = associateGrade.gradebookStatus === GradebookStatus.NewGrade || associateGrade.gradebookStatus === GradebookStatus.NotInGradebook;
 
-				if (this._createNewRadioChecked) {
-					this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
-				} else {
-					this._associateGradeSetGradebookStatus(GradebookStatus.ExistingGrade);
-				}
+			if (associateGrade.gradebookStatus === GradebookStatus.NotInGradebook) {
+				await this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
 			}
+
+			associateGrade.getGradeSchemes();
+			associateGrade.getGradeCandidates();
+			associateGrade.getGradeCategories();
 		} else {
 			const entity = store.get(this.href);
 			if (!entity || !entity.scoreAndGrade) return;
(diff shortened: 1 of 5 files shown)


wongvincent:
I dont know what the problem is when doing an `await setGradebookStatus`, and then calling `getSchemes/Categories/Candidates()`. For whatever reason, it fetches each individual item.

If I don't call `setGradebookStatus` beforehand, like if the `gradebookStatus` is already `create-new` or `existing-grade`, and I open the dialog, then calling `getSchemes/Categories/Candidates()` is fine. It respects the cache primer, and doesn't try to fetch each individual item.

If I call `setGradebookStatus` and have a setTimeout to force it to wait 10 seconds or something before fetching the schemes/categories/candidates, then it is fine, and doesn't try to fetch each individual item.

So it seems like there may be some sort of issue with the timing and how the store is processing things.

But with my changes, this fetching individual items behavior will only happen if the user is "Not in Grades" and opening the dialog. It is fine in the other scenarios of opening the dialog when in grades (create-new or existing-grade).

components/d2l-activity-editor/state/associate-grade.js
```diff
@@ -25,15 +25,19 @@ export class AssociateGrade {
...
 	async getGradeCandidates() {
+		if (this.gradeCandidateCollection) return;
+
 		const gradeCandidateCollectionEntity = await this._entity.getGradeCandidates();
 		if (!gradeCandidateCollectionEntity) return;
 		runInAction(() => {
 			this.gradeCandidateCollection = new GradeCandidateCollection(gradeCandidateCollectionEntity, this.token);
 		});
-		await this.gradeCandidateCollection.fetch();
+		await this.gradeCandidateCollection.load(gradeCandidateCollectionEntity);
```
wongvincent, line 35: This makes it so that the `GradeCandidateCollection` entity `fetch` function doesn't need to process both href or entity. If we have the entity, like we do here, we can just call the `load` function.

components/d2l-activity-editor/state/associate-grade.js
```diff
@@ -60,10 +66,6 @@ export class AssociateGrade {
 		this.canCreateNewGrade = entity.canCreateNewGrade();
 		this.canEditNewGrade = entity.canEditNewGrade();
 		this.canGetSchemes = entity.canGetSchemesForType(this.gradeType);
-
-		this.gradeCandidateCollection || this.getGradeCandidates();
```
wongvincent, line 64: I'm moving this out of `load`. Otherwise, the schemes, candidates, and categories will be fetched even on the initial checkout and it's in grades. We don't actually need these entities unless the dialog is opened. So I've moved this to be fetched in the `d2l-activity-grades-dialog` component when the dialog is opened.

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
```diff
@@ -24,8 +20,6 @@ export class GradeCandidateCollection {
 		if (sirenEntity) {
 			const entity = new GradeCandidateCollectionEntity(sirenEntity, this.token, { remove: () => { } });
 			await this.load(entity);
-		} else {
-			this.gradeCandidateCollectionEntity && await this.load(this.gradeCandidateCollectionEntity);
```
wongvincent, line 28: If this class is constructed with an entity instead of an href, they can just call `this.load(entity)`, rather than `this.fetch()`.
I updated `components/d2l-activity-editor/state/associate-grade.js` to call `this.load(entity)`

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
```diff
@@ -116,20 +116,20 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
 			await this.openDialog();
 
 			const associateGradeHref = this._associateGradeHref;
-			if (associateGradeHref) {
-				await this._fetch(() => associateGradeStore.fetch(associateGradeHref, this.token));
```
wongvincent, line 120: wrapping this in `this._fetch` did not appear to be doing anything.

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
```diff
@@ -116,20 +116,20 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
...
 
-				const associateGrade = associateGradeStore.get(associateGradeHref);
-				if (!associateGrade) return;
+			const associateGrade = await associateGradeStore.fetch(associateGradeHref, this.token);
+			if (!associateGrade) return;
 
-				this._createNewRadioChecked = associateGrade.gradebookStatus === GradebookStatus.NewGrade && this._canCreateNewGrade;
+			this._createNewRadioChecked = associateGrade.gradebookStatus === GradebookStatus.NewGrade || associateGrade.gradebookStatus === GradebookStatus.NotInGradebook;
 
-				if (this._createNewRadioChecked) {
```
wongvincent, line 127: Simplifying this because I can't think of a reason why we would want to set it to NewGrade if it's already new-grade, same for existing grade. We would only want to call this if the item was `not-in-gradebook` and we open the dialog, as we will then need to set it to `new-grade`.

Removing this hides the problem where when we call `setGradebookStatus` before fetching schemes/candidates/categories, it would cause each individual scheme/candidate/category to be fetched.

components/d2l-activity-editor/state/associate-grade.js
```diff
@@ -60,10 +66,6 @@ export class AssociateGrade {
...
-		this.gradeCategoryCollection || this.getGradeCategories();
-		this.gradeSchemeCollection || this.getGradeSchemes();
```
wongvincent, line 66: I'm moving this out of load. Otherwise, the schemes, candidates, and categories will be fetched even on the initial checkout and it's in grades. We don't actually need these entities unless the dialog is opened. So I've moved this to be fetched in the d2l-activity-grades-dialog component when the dialog is opened.

components/d2l-activity-editor/state/associate-grade.js
```diff
@@ -60,10 +66,6 @@ export class AssociateGrade {
 		this.canCreateNewGrade = entity.canCreateNewGrade();
 		this.canEditNewGrade = entity.canEditNewGrade();
 		this.canGetSchemes = entity.canGetSchemesForType(this.gradeType);
```
wongvincent, line 68: this is a test, pls ignore

components/d2l-activity-editor/d2l-activity-grades/state/grade-scheme.js
```diff
@@ -39,7 +39,6 @@ export class GradeScheme {
 		const href = sirenEntity.self();
 		const entity = associateGradeStore.get(href);
 		await entity.load(sirenEntity);
-		await entity.getGradeSchemes();
```
mdulat, line 42: I think this was here to make sure the newly selected scheme has the "selected" class. Should we explicitly set `isSelected` for this selected scheme here?

components/d2l-activity-editor/state/associate-grade.js
wongvincent, line 66: I'm moving these out of load. Otherwise, the schemes, candidates, and categories will be fetched even on the initial checkout and it's in grades. We don't actually need these entities unless the dialog is opened. So I've moved this to be fetched in the d2l-activity-grades-dialog component when the dialog is opened.

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
```diff
@@ -116,20 +116,20 @@ class ActivityGradesDialog extends ActivityEditorWorkingCopyDialogMixin(Localize
...
-					this._associateGradeSetGradebookStatus(GradebookStatus.NewGrade);
-				} else {
-					this._associateGradeSetGradebookStatus(GradebookStatus.ExistingGrade);
-				}
```
wongvincent, line 131: Simplifying this because I can't think of a reason why we would want to set it to NewGrade if it's already new-grade, same for existing grade. We would only want to call this if the item was `not-in-gradebook` and we open the dialog, as we will then need to set it to `new-grade`.

Removing this hides the problem where when we call `setGradebookStatus` before fetching schemes/candidates/categories, it would cause each individual scheme/candidate/category to be fetched.

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
```diff
@@ -8,12 +8,8 @@ configureMobx({ enforceActions: 'observed' });
 
 export class GradeCandidateCollection {
 
-	constructor(hrefOrGradeCandidateCollectionEntity, token) {
-		if (hrefOrGradeCandidateCollectionEntity instanceof GradeCandidateCollectionEntity) {
-			this.gradeCandidateCollectionEntity = hrefOrGradeCandidateCollectionEntity;
-		} else {
-			this.href = hrefOrGradeCandidateCollectionEntity;
-		}
+	constructor(href, token) {
```
mdulat, line 11: Is this first param always going to be an href now? Currently when we `getGradeCandidates()` from `associate-grade` we pass in a `gradeCandidateCollectionEntity` instead, because we are performing a GET action rather than a fetch.

components/d2l-activity-editor/d2l-activity-grades/state/grade-scheme.js
wongvincent: Ah okay, yes, I think I should have it manually set `isSelected`. If we don't have re-fetch the `/grade-schemes`, we should probably try not to.

components/d2l-activity-editor/state/associate-grade.js
mdulat: Oh I see, smart! Ok ignore my comment about that in grade-candidate-collection.js. Maybe we should set the first param to `new GradeCandidateCollection` on line 33 to `null` then?

components/d2l-activity-editor/d2l-activity-grades/state/grade-candidate-collection.js
wongvincent: Ah, I should probably extract the `href` and pass that in

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
mdulat: > I can't think of a reason why we would want to set it to NewGrade if it's already new-grade

This may have been to make sure the name and maxPoints are set correctly. But if we're patching those before we open the dialog then it should be correct anyway. Just make sure that's the case.

components/d2l-activity-editor/state/associate-grade.js
wongvincent: I'm going to pass in the actual `self` link from the gradeCandidateCollectionEntity

components/d2l-activity-editor/d2l-activity-grades/d2l-activity-grades-dialog.js
wongvincent: Yep, good point.
I checked and we are!

components/d2l-activity-editor/d2l-activity-grades/state/grade-scheme.js
wongvincent: I'm going to put this line back and optimize it so we don't need to fetch `/grade-schemes` on every change in a different PR.

mdulat (approved):

PR #1825, opened 2021-06-18
Remove fullpage attributes for content modules

ianlaidlaw:
This pr removes the "full page" attributes when using the new html editor. This means that the font size and family that are selected by default will use the system settings when editing modules. The full page attributes are still present when editing html topics.

Removing these was necessary for modules as it would add extra template stuff which is not meant for modules. This was causing all those extra new lines to appear as they were added when the html content was filtered.

Rally: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F601628066803

### Modules

image

### HTML Files

image

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/module/d2l-activity-content-module-detail.js
@@ -83,9 +83,6 @@ class ContentModuleDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlin
 						@d2l-activity-text-editor-change="${activityTextEditorChange}"
 						.richtextEditorConfig="${{}}"
 						html-editor-height="100%"
-						full-page
-						full-page-font-size="12pt"
-						full-page-font-family="Lato"
 					>
 					</d2l-activity-text-editor>
 				</div>


brissons (approved):
Dev. tested and LGTM! A couple questions/comments:
- It looks like sequence viewer doesn't load the extra characters for students. Is it loading the content differently so that it doesn't include the new characters?
- Maybe include the rally link and defect number for this PR

quinnbudan (approved):
LGTM!

ianlaidlaw:
> Dev. tested and LGTM! A couple questions/comments:
> 
> * It looks like sequence viewer doesn't load the extra characters for students. Is it loading the content differently so that it doesn't include the new characters?
> * Maybe include the rally link and defect number for this PR

Done, included the rally ticket. And yeah sequences as well as the main content view do not access the html from the same place the "filtered" html is, and that's why this hasn't been noticed for as long as it has even though it's happened since the old html editor - it wasn't until someone was using the Visual TOC and decided to include empty module descriptions that it was very noticable.

brissons:
> > Dev. tested and LGTM! A couple questions/comments:
> > 
> > * It looks like sequence viewer doesn't load the extra characters for students. Is it loading the content differently so that it doesn't include the new characters?
> > * Maybe include the rally link and defect number for this PR
> 
> Done, included the rally ticket. And yeah sequences as well as the main content view in Lessons do not access the html from the same place the "filtered" html is, and that's why this hasn't been noticed for as long as it has even though it's happened since the old html editor - it wasn't until someone was using the Visual TOC and decided to include empty module descriptions that it was very noticable.

Oh cool ok! Ty!

DiljotSG (approved):

PR #1828, opened 2021-06-18
fix: don't patch after each character change in name

mdulat:


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -312,7 +315,11 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 				toFocus.focus();
 			} else if (propName === 'activityName') {
 				this._setNewGradeName(this.activityName);
-				this._associateGradeSetGradeName(this.activityName);
+				this._debounceJobs = Debouncer.debounce(
+					this._debounceJobs,
+					timeOut.after(1000),
+					() => this._associateGradeSetGradeName(this.activityName)
```
wongvincent, line 321: will this be run if the user clicks save before this request is complete (or might it not even be called within that 1 second)?

mdulat: I'm not clear on the internal workings of the Debouncer, but based on testing, it calls the function immediately after pressing save, even if the 1s did not elapse.

mdulat: Also the patch is always called before the checkin

PR #1832, opened 2021-06-22
fix: label grade units correctly

mdulat:
This fixes the issue where the units label of the grade becomes the grade type (i.e. "numeric" or "selectbox" instead of "points"). I'm sure there was a trello or defect logged somewhere but I could not locate it.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -183,15 +183,15 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 	render() {
 		const activity = store.get(this.href);
 
-		let gradeType;
+		let gradeUnits;
 		let inGrades;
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -183,15 +183,15 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 	render() {
 		const activity = store.get(this.href);
 
-		let gradeType;
```
mdulat, line 186: It turns out this variable name was misleading, this was not supposed to represent the grade type, it was supposed to represent the units label. In the old score-out-of sub-entity the grade type options each had a "title" which was "points" for numeric types and "letter" for selectbox types. So in the old implementation (flag off) selectbox grades will still show "letter" here instead of "points" but I figured leave the old stuff for now and just make sure with flag on it shows "points" always instead of the gradeType.

wongvincent (approved):

PR #1834, opened 2021-06-22
fix: only patch new grade's name and points on save

mdulat:


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -321,6 +320,17 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		}
 	}
 
+	async save() {
+		if (this._createSelectboxGradeItemEnabled) {
+			const associateGradeEntity = associateGradeStore.get(this._associateGradeHref);
+			if (associateGradeEntity && associateGradeEntity.gradebookStatus === GradebookStatus.NewGrade) {
+				const scoreOutOf = this.shadowRoot.querySelector('#score-out-of').value;
```
wongvincent, line 327: probably better to retrieve this from `activity.scoreAndGrade.scoreOutOf`.

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -321,6 +320,17 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		}
 	}
 
+	async save() {
+		if (this._createSelectboxGradeItemEnabled) {
+			const associateGradeEntity = associateGradeStore.get(this._associateGradeHref);
+			if (associateGradeEntity && associateGradeEntity.gradebookStatus === GradebookStatus.NewGrade) {
```
wongvincent, line 326: I'm always confused whether we should be checking the GradebookStatus here. Or in the MobX object. Or in siren-sdk to determine whether we are allowed to call the actions in the new-grade entity to update its properties.

For example, the function here always works even if the GradebookStatus isn't `new-grade` I think. https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/associateGrade/AssociateGradeEntity.js#L205

wongvincent (approved):

components/d2l-activity-editor/d2l-activity-score-editor.js
mdulat: This check wasn't about permissions, it's just that we don't need to set the name and max points if the gradebook status isn't new-grade.

wongvincent (approved):

PR #1836, opened 2021-06-22
US129210 Add existing questions

emil-furniss:
Receive activity usage hrefs of created questions and call the add-existing workflow in the hmc quiz editor with that information, then update the total points display. 

Related hmc PR: https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/250

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F601851661777

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor-detail.js
@@ -183,7 +183,8 @@ class QuizEditorDetail extends ActivityQuizEditorTelemetryMixin(ActivityEditorMi
 				?skeleton="${this.skeleton}"
 				href="${this.activityUsageHref}"
 				.token="${this.token}"
-				quiz-href="${this.href}">
+				quiz-href="${this.href}"
+				@d2l-question-activity-added="${this._onActivityAdded}">
 			</d2l-activity-quiz-editor-action-bar>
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor-detail.js
```diff
@@ -225,6 +226,15 @@ class QuizEditorDetail extends ActivityQuizEditorTelemetryMixin(ActivityEditorMi
 		await quiz.save();
 	}
 
+	async _onActivityAdded(e) {
+		if (e && e.detail && e.detail.activities && e.detail.activities.length) {
+			const editor = this.shadowRoot.querySelector('d2l-activity-quiz-question-editor');
```
mdgbayly, line 231: unfortunate we have to do all this `shadowRoot` hacking , but I guess there is no other option right now whilst we have these different models. The only alternative I can think of is to use some kind of dependency registration kind of like we do with the activity editors, but this is probably fine.

mdgbayly (approved):
Looks good - are we missing updating the question points? Is that coming in a separate story?

emil-furniss:
> are we missing updating the question points?

No, turns out that the action state includes an encoded `sketchpadQoId`, which means that the points are set as part of the `TryAdd`.
https://search.d2l.dev/xref/lms/le/qc/D2L.LE.QuestionCollection/QuestionCollection/Extensibility/Hypermedia/Plugins/Question/QcQuestionCollectionAction.cs?r=4f4547d7#52

PR #1858, opened 2021-06-29
fix: scoreOutOf error

mdulat:
Make sure main page is checked in BEFORE score-out-of api is called. This way, if we set to ungraded, then state will be not-in-gradebook (disassociating the grade) before setting score-out-of to empty. Fixes error "score out of must have a value when grade is associated".

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-editor.js
@@ -72,6 +72,7 @@ class ActivityEditor extends ActivityEditorContainerMixin(ActivityEditorTelemetr
 		this.resizable = false;
 		this._backdropShown = false;
 		this._saveToastVisible = null;
+		this.saveOrder = 1500;
 	}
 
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-editor.js
```diff
@@ -72,6 +72,7 @@ class ActivityEditor extends ActivityEditorContainerMixin(ActivityEditorTelemetr
 		this.resizable = false;
 		this._backdropShown = false;
 		this._saveToastVisible = null;
+		this.saveOrder = 1500;
```
wongvincent, line 75: It inherits this from `d2l-activity-editor-mixin` and is `1000`. Do we need to change it to `1500`?

mdulat: Yes it needs to be saved after the checkin

wongvincent (approved):

PR #1860, opened 2021-06-30
US127570 - sorting html templates by name

coffeexcode:
Co-authored with @quinnbudan 

## Relevant Rally Links

- [US127570](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F602050357608&fdp=true?fdp=true): FACE HTML file - Support Sorting HTML Templates By Name 

## Description

This PR enables a config variable to sort HTML templates by name.

## Goal/Solution

This PR enables a config variable to sort HTML templates by name. The config variable is passed into FACE by the LMS.

## Video/Screenshots

N/A

## Related PRs

- https://github.com/Brightspace/lms/pull/12034

## Remaining Work

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
@@ -158,10 +158,10 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 	async _getHtmlTemplates() {
 		const htmlTemplatesResponse = await fetchEntity(this.htmlTemplatesHref, this.token);
 		const htmlTemplatesEntity = new ContentHtmlFileTemplatesEntity(htmlTemplatesResponse, this.token, { remove: () => { } });
-		let templates = htmlTemplatesEntity.getHtmlFileTemplates();
+		const templates = htmlTemplatesEntity.getHtmlFileTemplates();
 
 		if (this.sortHTMLTemplatesByName) {
-			templates = templates.sort((a, b) => a?.properties?.title?.localeCompare(b?.properties?.title));
+			templates.sort((a, b) => a?.properties?.title?.localeCompare(b?.properties?.title));
 		}
 
 		this.htmlFileTemplates = templates;
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
@@ -41,6 +41,9 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 					justify-content: space-between;
 					margin-bottom: 6px;
 				}
+				.d2l-new-html-editor-container {
+					margin-top: 6px;
+				}
 			`
 		];
 	}
@@ -66,16 +69,23 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 		const contentFileEntity = contentFileStore.getContentFileActivity(this.href);
 		let pageContent = undefined;
 		let pageRenderer = undefined;
+		let htmlTemplatesHref = null;
 
 		if (contentFileEntity) {
 			this.skeleton = false;
 			pageContent = contentFileEntity.fileContent;
 
+<<<<<<< HEAD
 			this.htmlTemplatesHref = contentFileEntity.htmlTemplatesHref;
+=======
+			if (contentFileEntity.htmlTemplatesHref) {
+				htmlTemplatesHref = contentFileEntity.htmlTemplatesHref;
+			}
+>>>>>>> d361877c (Configure 'Select Template' subtle button to show dependent on wether 'contentFileEntity.htmlTemplatesHref' exists)
 
 			switch (contentFileEntity.fileType) {
 				case FILE_TYPES.html:
-					pageRenderer = this._renderHtmlEditor(pageContent);
+					pageRenderer = this._renderHtmlEditor(pageContent, htmlTemplatesHref);
 					break;
 			}
 		} else {
@@ -183,7 +193,7 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 		);
 	}
 
-	_renderHtmlEditor(pageContent) {
+	_renderHtmlEditor(pageContent, htmlTemplatesHref) {
 		const newEditorEvent = new CustomEvent('d2l-request-provider', {
 			detail: { key: 'd2l-provider-html-new-editor-enabled' },
 			bubbles: true,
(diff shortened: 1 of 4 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -203,21 +209,21 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 				<label class="d2l-label-text d2l-skeletize">
 					${this.localize('content.pageContent')}
 				</label>
-				<d2l-dropdown-button-subtle 
-					style="${this.htmlTemplatesHref ? '' : 'visibility:hidden;'}" 
+				<d2l-dropdown-button-subtle
+					style="${this.htmlTemplatesHref ? '' : 'visibility:hidden;'}"
```
coffeexcode, line 213: This will get cleaned up in another PR.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -84,6 +90,7 @@ class ContentEditorDetail extends MobxLitElement {
 				<d2l-activity-content-file-detail
 					.href="${contentActivityHref}"
 					.token="${this.token}"
+					?sortHTMLTemplatesByName="${this.sortHTMLTemplatesByName}"
 				>
```
prajapac, line 94: I'm unsure what/how the `?` here does, ternary stuff? 🤔

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -157,7 +158,12 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 	async _getHtmlTemplates() {
 		const htmlTemplatesResponse = await fetchEntity(this.htmlTemplatesHref, this.token);
 		const htmlTemplatesEntity = new ContentHtmlFileTemplatesEntity(htmlTemplatesResponse, this.token, { remove: () => { } });
-		const templates = htmlTemplatesEntity.getHtmlFileTemplates();
+		let templates = htmlTemplatesEntity.getHtmlFileTemplates();
+
+		if (this.sortHTMLTemplatesByName) {
+			templates = templates.sort((a, b) => a.properties.title.localeCompare(b.properties.title));
```
CodeBaboon, line 164: This is potentially at risk of a null object reference exception if `properties` does not exist on the returned item. Perhaps this isn't possible? I think the same could be said for `title` since we are attempting to directly call the `localeCompare()` function off it (which won't work if `title` is undefined).

If you are confident that we always know we have an object with these properties then this is fine, but I thought I'd call it out given the recent bug we fixed with a similar null object issue.

CodeBaboon (commented):
One comment around possible null object exceptions.

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
quinnbudan: No it's not ternary stuff, it means that the attribute is a boolean attribute, see [lit docs here](https://lit.dev/docs/templates/expressions/)

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
quinnbudan: Good catch! I'm inclined to believe `properties` will always be there (not necessarily `title` though), but I can' find where in the `siren-sdk` that should be defined, so we'll add null checks to both!

ianlaidlaw: nit: `Array.sort` sorts in place so there is no need to use `let` and reassign. you can do 
```
const templates = [...];
templates.sort(...); // templates is sorted now
```

quinnbudan: Thanks! I didn't know that

prajapac (dismissed):
LGTM

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -158,6 +159,11 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 		const htmlTemplatesResponse = await fetchEntity(this.htmlTemplatesHref, this.token);
 		const htmlTemplatesEntity = new ContentHtmlFileTemplatesEntity(htmlTemplatesResponse, this.token, { remove: () => { } });
 		const templates = htmlTemplatesEntity.getHtmlFileTemplates();
+
+		if (this.sortHTMLTemplatesByName) {
+			templates.sort((a, b) => a?.properties?.title?.localeCompare(b?.properties?.title));
```
ianlaidlaw, line 164: optional chaining is sweet. i noticed it gave a build issue and iirc @CodeBaboon had some issue using this in this repo for some reason? I may be misremembering. anyways, it's definitely way more verbose but you could get around this by writing your own sort function. something like this should work:
```
myArray.sort((x, y) => {
	const dataIsInvalid = !(x.properties && x.properties.title && y.properties && y.properties.title);

	if (dataIsInvalid) {
		return -1;
	}

	if (x.properties.title > y.properties.title) {
		return 1;
	}

	return 0;
});
```

ianlaidlaw (approved):
LGTM, also did some dev testing and it sorted

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
CodeBaboon: Iirc I suggested someone else use optional chaining and then they ran into the build issues and had to revert. Not sure why they don't work in this repo, isn't it 2021?

Anyways, Ian's solution looks legit.

ianlaidlaw: if it's an easy change and doesn't negatively affect other people using the repo then i'm all for it.

quinnbudan: Yeah I unfortunately think it's not an easy change, it has to do with polymer build using babel and polymer analyze not using babel. I like your sort function suggestion, thanks!

CodeBaboon (approved):
Dev tested, checks out. Respects config. When on, appears to sort goodly.

Off:
image

On:
image

quinnbudan:
> i've never used `localeCompare` for sorting, but my understanding is that it's a more powerful comparison that is configurable and it looks like we're not passing in any options. as such i recommend just doing a very dumb `>` check instead. this SO post may be relevant as to why we shouldn't bother with `localeCompare` if we don't need to use its features: https://stackoverflow.com/questions/14677060/400x-sorting-speedup-by-switching-a-localecompareb-to-ab-1ab10

That's a good point! We hadn't considered a performance impact. We can replace it for sure. Originally we had done it because that's how lessons does the sorting

quinnbudan:
> also, do we need to check for `b.title` too? personally i believe these checks don't serve much of a purpose anymore since you moved this to siren-sdk we can be reasonably certain the function exists based on the class we are importing and that it has a safe fallback, whether the api returned the right stuff or not. imo we can nix the defensive checks now that it's in siren-sdk land

I also agree it makes sense to get rid of the defensive checks now!

ianlaidlaw (approved):

ianlaidlaw (approved):

prajapac (approved):

PR #1862, opened 2021-06-30
Removed duplicate langterm

coffeexcode:
Removing a duplicate langterm causing OSLO to fail.


quinnbudan (approved):
Thanks for fixing this!

prajapac (approved):

PR #1867, opened 2021-07-07
Fix spacing on ip restrictions footer buttons

wongvincent:
https://trello.com/c/m9h6gil0/348-ip-restrictions-add-cancel-button-padding-is-inconsistent-from-other-face-quiz-dialogs


ronan-f (approved):
Inspirational

PR #1868, opened 2021-07-08
Use self hosted runners

mostlyfabulous:
No more github minutes


dustxd (approved):

svanherk:
So you can't use our D2L self-hosted runners on public repos, only private ones.  Public repos also don't count against our GitHub minutes.

PR #1869, opened 2021-07-08
fix: update score after linking to existing

mdulat:
This fixes the issue where: 
- select existing grade, click OK in dialog (GradeOutOf will be updated to that grad's maxPoints)
- change the GradeOutOf value
- Save
It goes back to existing grade's maxPoints. 

This fix only loads a new ActivityUsage entity after checkin if there is no `associateGradeHref` (i.e. if it is not an activity-usage type entity). Otherwise it just updates the `associateGradeHref` which is the only part of the activity-usage that is changed by working copy.

### DIFF
--- a/components/d2l-activity-editor/state/activity-usage.js
+++ b/components/d2l-activity-editor/state/activity-usage.js
@@ -222,5 +222,6 @@ decorate(ActivityUsage, {
 	setDates: action,
 	setAlignmentsHref: action,
 	setCanUpdateAlignments: action,
-	loadCompetencies: action
+	loadCompetencies: action,
+	setAssociateGradeHref: action
 });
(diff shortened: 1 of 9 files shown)


components/d2l-activity-editor/state/activity-usage.js
```diff
@@ -71,6 +71,9 @@ export class ActivityUsage extends WorkingCopy {
 	setAlignmentsHref(value) {
 		this.alignmentsHref = value;
 	}
+	setAssociateGradeHref(value) {
```
wongvincent, line 74: Also add it to MobX actions?

wongvincent (approved):

wongvincent (approved):

PR #1870, opened 2021-07-08
FACE HTML file - Support Template Selection - Selecting a template

prajapac:
## Relevant Rally Links

- US127822: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F602982229180

## Description

As part of our FACE initiative and revamp, we would like to support HTML template selection.

## Goal/Solution

This PR makes the necessary changes to fetch the HTML template's contents, once one has been selected from the dropdown, and replaces the editor's contents with it (New HTML Editor, Old HTML Editor and NoWYSIWYG Text Editor).

## Video/Screenshots

New HTML Editor Template File

Old HTML Editor Template File

Text HTML Editor Template File

## Related PRs

- https://github.com/Brightspace/lms/pull/12252

## Remaining Work

- [ ] Dev Testing: another developer will test this code on their machine (Note, test as a whole, you will need to keep the backend PR in mind). Lots to test: Test adding changes, then adding Template and make sure overwrites, then modify that, and then open the same template again, add changes, save, save and close etc. (I tested this stuff and it seems to work but testing by someone else would be great too!)

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
@@ -22,7 +22,7 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 
 // Index for the browse template button
-const browseTemplateKey = -1;
+const browseTemplateKey = '-1';
 const editorKeyInitial = 'content-page-content';
 
 class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingMixin(LocalizeActivityEditorMixin(EntityMixinLit(RtlMixin(ActivityEditorMixin(MobxLitElement))))))) {
@@ -178,31 +178,35 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 
 	async _handleClickTemplateMenutItem(e) {
 		const targetMenuItem = e.target.getAttribute('key');
-		const template = this.htmlFileTemplates[targetMenuItem];
-		const fileEntity = new FileEntity(template);
-		const contentUrl = fileEntity.getFileDataLocationHref();
-		const response = await window.d2lfetch.fetch(contentUrl);
 
-		if (response.ok) {
-			const content = await response.text();
-			this.pageContent = ''; // we need to reset this first, otherwise if same template is selected again, it won't re-render
-			this.pageContent = content;
-			this._savePageContent();
-			this.editorKey = `${editorKeyInitial  }-${Date.now().toString()}`; // key needs to be modified in order to re-render old HTML editor
+		if (targetMenuItem === browseTemplateKey) {
+			// TODO: Add Browse Template Button Support
+		}
+		else {
+			const template = this.htmlFileTemplates[targetMenuItem];
+			const fileEntity = new FileEntity(template);
+			const contentUrl = fileEntity.getFileDataLocationHref();
+			const response = await window.d2lfetch.fetch(contentUrl);
+
+			if (response.ok) {
+				const content = await response.text();
+				this._savePageContent(content);
+				this.editorKey = `${editorKeyInitial}-${Date.now().toString()}`; // key needs to be modified in order to re-render old HTML editor
+			}
 		}
 	}
 
 	_onPageContentChange(e) {
-		this.pageContent = e.detail.content;
-		this._savePageContent();
+		const htmlContent = e.detail.content;
+		this._savePageContent(htmlContent);
 	}
 
 	_onPageContentChangeDebounced(e) {
-		this.pageContent = e.detail.content;
+		const htmlContent = e.detail.content;
 		this._debounceJobs.description = Debouncer.debounce(
 			this._debounceJobs.description,
 			timeOut.after(ContentEditorConstants.DEBOUNCE_TIMEOUT),
-			() => this._savePageContent()
+			() => this._savePageContent(htmlContent)
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -169,21 +176,37 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
...
+		const fileEntity = new FileEntity(template);
+		const contentUrl = fileEntity.getFileDataLocationHref();
+		const response = await window.d2lfetch.fetch(contentUrl);
+
+		if (response.ok) {
+			const content = await response.text();
+			this.pageContent = ''; // we need to reset this first, otherwise if same template is selected again, it won't re-render
+			this.pageContent = content;
+			this._savePageContent();
+			this.editorKey = `${editorKeyInitial  }-${Date.now().toString()}`; // key needs to be modified in order to re-render old HTML editor
```
prajapac, line 191: This is needed for the old HTML editor, see: https://github.com/BrightspaceHypermediaComponents/activities/pull/1121

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -169,21 +176,37 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 		}
 	}
 
+	async _handleClickTemplateMenutItem(e) {
```
ChrisKraljevicD2L, line 179: extreme nit:

```suggestion
	async _handleClickTemplateMenuItem(e) {
```

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -13,18 +13,24 @@ import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
...
+import { FileEntity } from 'siren-sdk/src/files/FileEntity';
 import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { LocalizeActivityEditorMixin } from '../../mixins/d2l-activity-editor-lang-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 
+// Index for the browse template button
+const browseTemplateKey = '-1';
```
ChrisKraljevicD2L, line 25: Might want to leave a comment saying that this is a placeholder. I thought this was a tabindex at first.

Also, can we guarantee the index of the "browse templates" button is going to remain the same?

prajapac: Yes, it will always be guaranteed to be that, because we set it to it, and that's why we used the const so we have one spot where we set it. The templates themselves are indexed 0, 1, 2, ..., n-1. As far as the comment, the comment above the line indicates that it is the index, and the const shows it won't change.

coffeexcode (approved):
lgtm

ChrisKraljevicD2L (commented):
also LGTM, also will dev test

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -237,7 +262,7 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 			return html`<p class="d2l-menu-item-span d2l-body-small">${this.localize('content.noHtmlTemplates')}</p>`;
 		}
 
-		return this.htmlFileTemplates.map((template) => html`<d2l-menu-item text=${template.title()}></d2l-menu-item>`);
+		return this.htmlFileTemplates.map((template, index) => html`<d2l-menu-item text=${template.title()} key=${index}></d2l-menu-item>`);
```
ianlaidlaw, line 265: suggestion: so this is coming from React knowledge, but my instincts are telling me we shouldn't use `index` as they keys here, it should be something we can guarantee to be unique. This will also allow us to not have to use `-1` for the templates either. I'm wondering if we should just use the `path` to be the key instead? and then the templates can be whatever string we want. since -1 does seem a little random of a choice at first glance.

ianlaidlaw: I really can't find much information on Lit keys online though... so I dunno!

coffeexcode: Do we actually need the key here? Based on what's in [d2l-menu-item](https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item.js) and the [mixin](https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item-mixin.js) I don't see anything that actually uses/references the key.

ianlaidlaw: yeah so maybe index is fine just to keep it simple. i know in React using the index as the key is not recommended but i can't find anything for Lit, so maybe it's okay

ChrisKraljevicD2L (approved):
dev tested, looks great :+1:

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -280,12 +305,13 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 		this._debounceJobs[jobName] && this._debounceJobs[jobName].flush();
 	}
 
-	_savePageContent(pageContent) {
+	_savePageContent(htmlContent) {
```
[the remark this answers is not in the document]
prajapac: This was just to distinguish the variable passed in as a parameter to _savePageContent() from the global `this.pageContent`

coffeexcode (approved):

ChrisKraljevicD2L (approved):

coffeexcode (approved):

ChrisKraljevicD2L (approved):

ianlaidlaw (approved):

PR #1874, opened 2021-07-09
Content with invalid dates can still be saved if you collapse the date controls first

ianlaidlaw:
Brings the content dates editor inline with how assignments does it. That accordion component has an error property - so our content dates editor needed to add it's own error checking function and to use the MobxLitElement instead of LitElement so we can access the store.

invaliddates

Unfortunately there is a weird UI behavior where if the accordion is expanded with invalid dates and you attempt to save, it will close and then open. I think this is because the opened state is being controlled by that `hasError` value, which temporarily gets set to false while it is saving. This is in-line with how the Assignments Editor behaves as well so it might be best to leave it and ideally have the team that wrote the `d2l-activity-accordion-collapse` component to investigate.

invaliddates2

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
@@ -2,7 +2,6 @@ import '../d2l-activity-availability-dates-summary.js';
 import '../d2l-activity-availability-dates-editor.js';
 import '../d2l-activity-accordion-collapse.js';
 import { accordionStyles } from '../styles/accordion-styles';
-import { ActivityEditorMixin } from '../mixins/d2l-activity-editor-mixin';
 import { html } from 'lit-element/lit-element.js';
 import { LocalizeActivityEditorMixin } from '../mixins/d2l-activity-editor-lang-mixin.js';
 import { MobxLitElement } from '@adobe/lit-mobx';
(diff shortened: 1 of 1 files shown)


components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
```diff
@@ -20,22 +21,15 @@ class ContentAvailabilityEditor extends SkeletonMixin(LocalizeActivityEditorMixi
 	static get styles() {
 		return [
 			super.styles,
-			accordionStyles,
-			css`
```
ianlaidlaw, line 24: this is a refactor, these styles already exist in `accordionStyles`

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-availability-editor.js
```diff
@@ -47,8 +41,17 @@ class ContentAvailabilityEditor extends SkeletonMixin(LocalizeActivityEditorMixi
 		`;
 	}
 
-	_renderAvailabilityDatesEditor() {
+	_errorInAccordion() {
```
ianlaidlaw, line 44: more or less a copy + paste from `d2l-activity-assignment-availability-editor.js`

ChrisKraljevicD2L (approved):

brissons (approved):

PR #1876, opened 2021-07-12
Improve loading skeletons in Content editors

ianlaidlaw:
This pr sets out to improve the UX for the skeletons in the Content editors. 

It was noticed that the due date element was loading in way faster than everything else. This was because each component was handling it's own `skeleton` state by checking against their entity, and the due date was using the activity usage entity which is higher up the chain than everything else. 

As many of these components use the same entity, it made sense to let the `detail` part of the editor be in charge of the skeleton state and pass that down to each of the components. As `detail` is the one fetching the entity, the time for each sub-component refetching the same entity is negligible as it won't actually make another network request. The result of this is that each component will stop visual loading at the same time, with the exception of some other components like the LTI preview which was already happening anyways.

### BEFORE:

skeletonz1

### AFTER - MODULES:

skeletonzmod

### AFTER - WEBLINKS:

skeletonzlink

### AFTER - LTI LINKS:

skeletonzlti

### AFTER - FILES:

skeletonzfile

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
+++ b/components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
@@ -1,6 +1,5 @@
 import '../shared-components/d2l-activity-content-editor-title.js';
 import './d2l-activity-content-file-loading.js';
-import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { bodySmallStyles, labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { ContentFileEntity, FILE_TYPES } from 'siren-sdk/src/activities/content/ContentFileEntity.js';
 import { css, html } from 'lit-element/lit-element.js';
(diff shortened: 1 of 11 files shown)


ianlaidlaw:
The gifs in the description should have some editors have the due date be expanded by default - this was broken for those gifs and then fixed.

ChrisKraljevicD2L (dismissed):
Love me some synchronized skellingtons

doot_after

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -133,7 +134,7 @@ class ContentFileDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandlingM
 
 		this._saveOnChange('htmlContent');
 
-		const originalActivityUsageHref = contentFileActivity.activityUsageHref;
+		const originalActivityUsageHref = this.activityUsageHref;
```
ianlaidlaw, line 137: this is getting passed down now, so no need to try and get it off the `contentFileActivity`

components/d2l-activity-editor/d2l-activity-content-editor/d2l-activity-content-editor-detail.js
```diff
@@ -38,10 +38,6 @@ class ContentEditorDetail extends MobxLitElement {
 		];
 	}
 
-	constructor() {
```
ianlaidlaw, line 41: wasn't needed

ianlaidlaw:
I've found a weird problem with saving embeded LTI links... currently investigating

components/d2l-activity-editor/d2l-activity-content-editor/lti-link/d2l-activity-content-lti-link-detail.js
```diff
@@ -37,6 +42,7 @@ class ContentLTILinkDetail extends AsyncContainerMixin(SkeletonMixin(ErrorHandli
 		this._setEntityType(ContentLTILinkEntity);
 		this.skeleton = true;
 		this.saveOrder = 2000;
+		this.canEmbed = false;
```
brissons, line 45: This should also be set to 'true' since we are assuming that we can embed first.

ianlaidlaw: I think you're right. it doesn't seem to have the same problems as that other one we changed, but I can definitely default this to true

ChrisKraljevicD2L (commented):
LGTM, I'll dev test

ChrisKraljevicD2L (approved):
dev testing complete :+1:

brissons (approved):

coffeexcode (approved):
LGTM. Also dev tested because I didn't see that someone was already on it.

PR #1611, opened 2021-03-15
Update the activity and assignment editor to use the LocalizeDynamicMixin

kieranderson:
https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.md

PR #1891, opened 2021-07-16
Us129739/fix missing lang terms

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F603418168501

I don't think we've ever been getting lang terms for quizzes as we didn't have anything configured in our serge file. 

I also noticed the assignments localization mixin [had been updated here](https://github.com/BrightspaceHypermediaComponents/activities/pull/1611) and figured quizzing should likely use the same logic.

### DIFF
--- a/activities.serge.json
+++ b/activities.serge.json
@@ -219,5 +219,27 @@
       "sv-se sv",
       "tr-tr tr"
     ]
+  },{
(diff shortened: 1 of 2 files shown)


bearfriend (approved):
🥳

PR #1921, opened 2021-08-05
US128833 - Allowable filetype fixes

maddiegabriel:
Fixes related to [US128833 ](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F600937249791).

Now renders dropdown conditionally based on Submission Type.

Also fixing an error that was popping up when trying to save any submission types other than File submission.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
+++ b/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
@@ -168,7 +168,7 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 	}
 
 	_renderAllowableFileTypesDropdown(assignment) {
-		if (!assignment || !assignment.submissionAndCompletionProps || !assignment.submissionAndCompletionProps.showAllowableFileTypes) {
+		if (!assignment || !assignment.submissionAndCompletionProps || !assignment.submissionAndCompletionProps.showFilesSubmissionOptions) {
 			return html``;
 		}
 
@@ -247,7 +247,7 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 	_renderAssignmentFilesSubmissionLimit(assignment) {
 		if (!assignment ||
 			!assignment.submissionAndCompletionProps ||
-			!assignment.submissionAndCompletionProps.showFilesSubmissionLimit) {
+			!assignment.submissionAndCompletionProps.showFilesSubmissionOptions) {
 			return html``;
 		}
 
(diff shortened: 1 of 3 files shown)


components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-submission-and-completion.js
```diff
@@ -48,6 +48,10 @@ export class SubmissionAndCompletionProps {
 		this._setValidCompletionTypeForSubmissionType();
 	}
 
+	get showAllowableFileTypes() {
```
jeroach, line 51: Is this method not the exact same as `showFilesSubmissionLimit()` below?

I expected them to be different

maddiegabriel: You're right, they're identical because the Allowable Filetypes and Submission Limit options should both only appear when the File submission type is selected (which is what this function checks for).

I can combine into 1 new variable to avoid repeating the function 👍

jeroach: awesome, thanks :)

jeroach (approved):

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -168,12 +168,12 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
...
 
 	_renderAllowableFileTypesDropdown(assignment) {
-		if (!assignment || !assignment.submissionAndCompletionProps) {
+		if (!assignment || !assignment.submissionAndCompletionProps || !assignment.submissionAndCompletionProps.showFilesSubmissionOptions) {
 			return html``;
 		}
 
 		let allowableFileTypeContent = html``;
-		if (assignment.submissionAndCompletionProps.canEditSubmissionType) {
+		if (assignment.submissionAndCompletionProps.canEditAllowableFileType) {
```
emil-furniss, line 176: The conditions gating these two "permissions" in the LMS are not the same, meaning that this will potentially render the selectbox under different circumstances than before. Is that expected?

Conditions for including `canEditSubmissionType`: https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=f4e056be#1259

Conditions for including `canEditAllowableFileType`: https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=f4e056be#1280

maddiegabriel: Yes, the `canEditSubmissionType` from before was a mistake. The intended behavior here is to render the dropdown under the `canEditAllowableFileType` conditions.

emil-furniss (approved):

PR #1922, opened 2021-08-05
US130370: FACE > Genericize LTI preview component

prajapac:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F605438258067

## Description

LTI Face Pages have a Preview Component: https://search.d2l.dev/xref/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-content-editor/lti-link/d2l-activity-content-lti-link-preview.js?r=21973c0e that we aim to genericize to be reused for SCORM FACE pages.

## Goal/Solution

`d2l-activity-content-lti-link-preview.js` was renamed to `d2l-activity-content-link-preview.js` and other references within it to LTI have also genericized. The file has also been moved to the shared-components directory.

At the present moment the CSS in the component file is not being extracted. After discussing with @pbrown-d2l, we have established that at the present moment there is no need for such refactoring, and if such a requirement arises in the future, it will be okay to deal with then. Further, the CSS in the file doesn't seem to be LTI specific, and SCORM FACE pages will likely make use of the same.

## Remaining Work

- [ ] Dev Testing: another developer will test this code on their machine (Just make sure it still works as expected)


ianlaidlaw (approved):

coffeexcode (approved):
When I was stubbing the SCORM FACE page there was a couple things about the preview I noticed that I think could be nice to include here.
 
1. Pushing in the header label into the preview component. This way we can control the preview from the parent component rather than having a floating label above.

2. Is it at all worth genericizing the `d2l-activity-content-lti-link-external-activity.js` or is too far out of scope for this? (it's probably a good idea to do - for the same reason this was genericized)

Just thoughts, if not included that's good too.

ChrisKraljevicD2L (approved):

prajapac:
> When I was stubbing the SCORM FACE page there was a couple things about the preview I noticed that I think could be nice to include here.
> 
> 1. Pushing in the header label into the preview component. This way we can control the preview from the parent component rather than having a floating label above.
> 2. Is it at all worth genericizing the `d2l-activity-content-lti-link-external-activity.js` or is too far out of scope for this? (it's probably a good idea to do - for the same reason this was genericized)
> 
> Just thoughts, if not included that's good too.

Regarding both your points, based off the designs for the SCORM FACE page, while the portion above the Preview iFrame looks similar to LTI, it is still distinct enough. The display options are similar but the terminology is different, I imagine the changes to effect the functionality specific for SCORM would be different. For the Label itself, SCORM has the Package name and when it was last edited (something which LTI also doesn't have).

PR #1927, opened 2021-08-10
Mwang/us130450

cloudislife8899:
https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F605531810231


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -161,6 +191,22 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 			${assignment.submissionAndCompletionProps.submissionTypeOptions.map(option => html`<option value=${option.value} ?selected=${String(option.value) === assignment.submissionAndCompletionProps.submissionType}>${option.title}</option>`)}
 		`;
 	}
+	_isInvalidFileTypes(fileTypes) {
+		const fileTypesList = fileTypes.split(',').map(type => type.trim());
+		const minimumExtensionLength = 3;
```
JamieTran, line 196: I don't think this is true, it looks like there's the file extension `.RA` that is supported in assignments https://documentation.brightspace.com/EN/le/assignments/learner/assignments_intro_1.htm

cloudislife8899: Extension length includes the period. So (.RA).length is 3 which is the minimum length, unless there is a single character extension type lol.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -72,10 +79,12 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
 		super(store);
 		this.saveOrder = 2000;
 		this.allowableFileTypeCustomValue = '5'; // Custom allowable file type value
+		this.restrictedFileTypes = [];
 	}
 
 	render() {
 		const assignment = store.get(this.href);
+		const activity = activityStore.get(this.activityUsageHref);
```
JamieTran, line 87: i wonder if we should check if activity exists

JamieTran (dismissed):
we should probably get eyes from FACE on this PR

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
```diff
@@ -110,6 +119,27 @@ class ActivityAssignmentSubmissionAndCompletionEditor extends SkeletonMixin(Acti
...
 		await assignment.save();
 	}
+	/**
+	 * Follows a list of rels beginning at a specific entity.
+	 * @async
+	 * @param {String[]} relList List of rels to follow
+	 * @param {object} entity Beginning entity
+	 * @returns {object|null|undefined} The entity at the end of the rel path. {null|undefined} if an entity in the chain is missing or doesn't have the next rel.
+	 */
+	async _followRelPath(relList, entity) {
```
wongvincent, line 129: I think this can be done in the MobX activity object

wongvincent (commented):
If I understand it correctly, I think some of this code should be in the activity-usage MobX object. The `fetchEntity` call should be returning an entity that is an object type in `siren-sdk`. It can also be an object in the state.
It'll make it so that we won't have references like `activity._entity?._entity` or `restrictedFileTypesEntity.properties?.extensions` in the code.

Also, be careful to check for undefined cases. Just want to make sure none of these `split/some/includes/slice/store.get` calls throw an error.

wongvincent:
Please also double check that the "discard changes" dialog doesn't appear when the `RestrictFileExtensions` config is off. Right now, it is happening because of the Allowable File extensions stuff, but I've let Maddie know about it.

emil-furniss (approved):
Approving but as discussed, how the list of restricted file types is loaded should be refactored to use the mobx state and siren-sdk.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor-submission-and-completion.js
emil-furniss: Just as an aside, single char extensions definitely exist: `.c`, `.d`, and `.r` are used for C, D and R code files respectively, certain assembly code uses `.s`, and I think matlab has a `.m` extension? There are probably more...

PR #1945, opened 2021-08-19
DE43063 Disable "Reset to Ungraded" menu option if assignment has submissions

emil-furniss:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F517225304416


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -26,6 +26,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 	static get properties() {
 		return {
 			activityName: { type: String },
+			hasSubmissions: { type: Boolean },
```
mdulat, line 29: I wonder if it would be better to give this a more generic name, as the activity-score-editor can be used by any activity, and `hasSubmissions` is specific to assignments, as a reason for not allowing the resetting of a grade. Something like `enableResetGrade`?

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -164,6 +165,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 	constructor() {
 		super(store);
 		this.saveOrder = 500;
+		this.hasSubmissions = true;
```
mdulat, line 168: I think we want to initialize to false, so that other activity types that don't pass in the property won't have reset disabled.

components/d2l-activity-editor/d2l-activity-score-editor.js
emil-furniss: Good point, I missed that this is (trying to be) a generic component. Same goes for the initial value.

mdulat:
The other case I was wondering about: If the assignment has a grade and is inGrades, and has sumbissions. Then the user changes it to notInGradebook, then sets the grade to empty. Will this be caught?

emil-furniss:
> The other case I was wondering about: If the assignment has a grade and is inGrades, and has sumbissions. Then the user changes it to notInGradebook, then sets the grade to empty. Will this be caught?

Yes, that will show the "Grade Out Of is required" error in the grade field. There is no way to have an empty Grade Out Of field without getting that validation error.

mdulat (approved):

PR #1949, opened 2021-08-23
DE44810: Potential fix to empty module descriptions

ianlaidlaw:
tl;dr we are trying to use the full editor before it is ready, and sometimes module descriptions load very fast.

this pr will wait 2 seconds when fetching the module entity so that we can be reasonably confident that tinymce is ready.


coffeexcode (approved):

brissons:
I've sadly been able to recreate this bug locally. Interestingly, when I added a log to the `render()` function in the `htmleditor`, it doesn't look like it's getting anything... so I'm still inclined to believe that it has something to do with `live()` as you say.

Working (ignore console errors haha):
image
Not working:
image

ianlaidlaw:
> I've sadly been able to recreate this bug locally with the change. Interestingly, when I added a log to the `render()` function in the `htmleditor` and it fails, it doesn't look like the component is getting anything... so I'm still inclined to believe that it has something to do with `live()` as you say.
> 
> Working:
> image
> Not working:
> image

dang... well thanks for testing this at least! i honestly tried about 100 refreshes or edits and couldn't reproduce so I was hoping for the best.

ianlaidlaw:
I think Scott and I found the issue. going to add it to this pr, but as of now, what I have will not fix the issue

coffeexcode (approved):
LGTM

prajapac (approved):
I pretend I no see

coffeexcode:
I dev tested this and here are my results:

- 1000ms: Bad almost immediately
- 1250ms: Bad almost immediately
- 1500ms: Bad after ~10 tries
- 1750ms: Good, but failed when I loaded 8 module pages at the same time
- 2000ms: Never failed at <= 8 pages simultaneously. This seems like the winner to me

PR #1981, opened 2021-09-18
fix: US131442 wait for pending actions to complete before fetching wo…

mdulat:
…rking copy activity usage

fixes https://trello.com/c/IYlEGfc7/448-after-loading-an-ungraded-assignment-if-you-click-ungraded-and-enter-a-grade-and-save-quickly-before-all-the-cache-primers-have

The bug is because the `fetch` requests and siren `actions` are not coordinated, and we need to make sure that working-copy PATCH requests are completed before the `fetch` is done to see if the working-copy has a checkin action. The action is only present if there are changes, and changes are only made after the PATCH requests.

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -27,6 +27,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		return {
 			activityName: { type: String },
 			disableResetToUngraded: { type: Boolean },
+			pendingUpdates: { type: Array },
 			_focusUngraded: { type: Boolean },
 			_createSelectboxGradeItemEnabled: { type: Boolean },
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -166,6 +167,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		super(store);
 		this.saveOrder = 500;
 		this.disableResetToUngraded = false;
+		this.pendingUpdates = [];
```
wongvincent, line 170: Could the property could go in the `d2l-activity-editor-mixin.js` file

mdulat: yeah, that would make more sense, especially when other components start using working copy.

mostlyfabulous (approved):
LGTM

wongvincent (approved):

PR #2014, opened 2021-10-05
Rename submission-views-dialog to be submission-views-editor

wongvincent:
So that it matches timing, attempts, iprestrictions file naming format


martinwang-d2l (approved):
Looks good

PR #2019, opened 2021-10-05
US130775 Quizzes: Disable gradebook `button` while request is in-flight

mostlyfabulous:
[US130775](https://rally1.rallydev.com/#/?detail=/userstory/606249971567&fdp=true): [ui] add score-container to score-and-duedate-container

I'll try add a blur transition and filter

### DIFF
--- a/components/d2l-activity-editor/d2l-activity-score-editor.js
+++ b/components/d2l-activity-editor/d2l-activity-score-editor.js
@@ -199,6 +199,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		let scoreOutOfError;
 		let canEditScoreOutOf;
 		let canEditGradebookStatus;
+		let isGradebookStatusChanging;
 
 		if (this._createSelectboxGradeItemEnabled) {
 			const associateGradeEntity = associateGradeStore.get(this._associateGradeHref);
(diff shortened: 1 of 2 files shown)


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -161,6 +161,10 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 				color: inherit;
 				font-family: inherit;
 			}
+			.d2l-blur {
+				filter: opacity(60%);
+				transition: 1s -webkit-filter linear;
```
ronan-f, line 166: For transition do you need to specify a property? e.g. `transition: height 1s` or does leaving the property out implicitly set it? I'm not super familiar with transitions so just checking

ronan-f (commented):
Just one question about transitions but I think it looks good

components/d2l-activity-editor/d2l-activity-score-editor.js
mostlyfabulous: Great question, I'll find out

mdulat: I'm not familiar with it either, but property to transition would be `opacity` here I guess. Here's [an example](https://bavotasan.com/2011/a-simple-fade-with-css3/) I found, which also makes me think this may be something worth testing on other platforms (firefox, max safari?).

mostlyfabulous: If you don't specify, than all transition-able properties will transition. https://developer.mozilla.org/en-US/docs/Web/CSS/transition

mostlyfabulous: Transition is shorthand for four other transition attributes/styles

ronan-f: In that case should we specify opacity as the property to transition to avoid any unexpected transitions cropping up?

mostlyfabulous: Yes good shout, I'll give that a shot

mdulat (dismissed):

mdulat (approved):

PR #2078, opened 2021-10-22
DE45246: FACE > Can save erroneous dates after correcting erroneous dates

coffeexcode:
## Relevant Rally
- [DE45246](https://rally1.rallydev.com/#/?detail=/defect/609560685611&fdp=true): FACE > Can save erroneous dates after correcting erroneous dates

## Description
While the ticket saves only when fixing erroneous dates, the issue was that when using the 'save' button (save in place), when working on a topic that used working copy (HTML files or weblinks), a new href would be used for the page editor (with a brand new working copy id) that would basically unsynchronized the date picker from entity.

This means that when you went to save after changing the dates, the save would look successful because when it went to check the dates it was actually reading the dates of the ORIGINAL entity (which would be unchanged). This led the page to show a successful save, but the dates would never actually be different after the first save. This would happen anytime after the first use of the save button.

**It doesn't do this anymore.**

The fix:
1. Add some extra properties to the date editor to force a re-render when the underlying href changes
2. The save order for the 2 working copy editors was changed to 2000 to match the other editors
3. During the page save (screen whiteout), we silently fetch the updated entity to prevent the d2l-spinner from re-appearing

## Remaining work
- [x] Another developer will devtest this


components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -138,7 +145,21 @@ class ContentFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		});
 
 		await this.dispatchEvent(event);
+
+		// Tell the container it should fetch the new entity before finishing
+		// the saving animation
+		event = new CustomEvent('d2l-fetch-new-entity-after-save', {
```
ianlaidlaw, line 151: this is super nitpicky so no need to do it, but imo for readability sake i think it would be better to just create a new event variable rather than reuse the old one

Kiefer-Co-D2L (approved):
image

Tested both HTML file and web link:

When trying to save an invalid date, you get the whiteout, then a message saying the save was unsuccessful.

Upon fixing the date and clicking save once, you get the whiteout, then a message saying the save was successful.

This behaviour looks correct to me!

ianlaidlaw (approved):

brissons (approved):

PR #2108, opened 2021-11-01
DE45521 - Save custom allowable file types as lowercase

JamieTran:
[DE45521 ](https://rally1.rallydev.com/#/10153049633d/dashboard?detail=%2Fdefect%2F610904031713)

There is a defect where setting `allowableFyleType` to `.GIF` was not letting students submit ANY files to the assignment. 
Could not figure out why this only happens on `.GIF` during the investigation.

Saving the file ext. as lower case avoids this issue, this doesn't affect any other files as our file ext. comparisons are not case sensitive (eg. with `allowableFyleType` set to `.pdf`, students can submit both `.PDF` and `.pdf`)


maddiegabriel (approved):
LGTM!

Scy-D2L (approved):

PR #2111, opened 2021-11-02
Show captions from Manage Files on FACE media content page

ianlaidlaw:
Display the associated captions for a media file.

Note: adding captions via the dialog does not refresh the view, for now it will only show up on a page refresh. I have broken out a new task to handle this.

### Related prs:
lms: https://github.com/Brightspace/lms/pull/15522
siren-sdk: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/412

image

image

image


components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -80,11 +82,11 @@ export class ContentFile {
 		}
 
 		fileLocationHref = fileEntity.getFileLocationHref();
-		this.load(entity, fileContent, fileLocationHref);
+		await this.load(entity, fileContent, fileLocationHref);
 		return this;
 	}
 
-	load(contentFileEntity, fileContent, fileLocationHref) {
+	async load(contentFileEntity, fileContent, fileLocationHref) {
```
ianlaidlaw, line 89: I don't love this solution. I talked with @brissons and couldn't come up with a really great solution to this, right now `load` does some fetching _and_ some loading. I think this file needs a refactor.

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-media-file-detail.js
```diff
@@ -173,10 +169,20 @@ class ContentMediaFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeAc
 			return html``;
 		}
 
+		const captionsCSV = captions.reduce((acc, curr, index) => {
```
ianlaidlaw, line 172: There might be a better way to do this. I got rid of the `captions.join(', ')` as the values I'm actually interested in are nested inside objects

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -2,6 +2,7 @@ import { action, configure as configureMobx, decorate, observable } from 'mobx';
 import { ContentFileEntity, FILE_TYPES } from 'siren-sdk/src/activities/content/ContentFileEntity.js';
 import { ContentHtmlFileEntity } from 'siren-sdk/src/activities/content/ContentHtmlFileEntity.js';
 import { ContentMediaFileEntity } from 'siren-sdk/src/activities/content/ContentMediaFileEntity.js';
+import { ContentMediaFileCaptionsEntity } from 'siren-sdk/src/activities/content/ContentMediaFileCaptionsEntity.js';
```
ianlaidlaw, line 5: CI will fail until siren-sdk is merged and updated

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-media-file-detail.js
ianlaidlaw: I was mostly just using this for testing and it got left over. but open to easier solutions

coffeexcode: One liner could be something like
`captions.map( ({properties.langName]) => properties.langName).join(' ,  ');` where properties.langName would be the property you want to extract.

For example:
`
let objects = [{a: 5, b: 6}, {a:7, b:8}];
objects.map(({ a }) => a).join(', '); // returns '5, 7'
`

ianlaidlaw: i like that

brissons (dismissed):
LGTM! Dev. tested 👍

prajapac (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
prajapac: What exactly distinguishes a fetch from a load? A fetch makes me think of a network request/fetch, and perhaps a load to take the fetched response and populate different variables with those values?

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -105,6 +107,10 @@ export class ContentFile {
 			this.isMediaEmbedded = mediaFileEntity.embedMedia();
 			this.isAdvancedEditingEnabled = mediaFileEntity.isAdvancedEditingEnabled();
 			this.isContentServiceResource = mediaFileEntity.isContentServiceResource();
+			const mediaCaptionsHref = mediaFileEntity.getMediaFileCaptionsHref();
+			const sirenResponse = await fetchEntity(mediaCaptionsHref, this.token);
+			const mediaCaptionsEntity = new ContentMediaFileCaptionsEntity(sirenResponse, this.token, { remove: () => { } });
```
prajapac, line 112: What does the `{remove: () => { } });` do?

ianlaidlaw: it's a function that gets called when the entity is disposed - we don't want to do anything. but the code in `Entity.js` assumes it exists, so we pass in an empty function. At closer look I think we can get rid of this listener param and it will work, but we've just always included the empty function

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
ianlaidlaw: that's exactly right. what i change in this pr blurs that a bit, there's some fetching done in load now. it's not ideal but we decided this file needs a bit of a refactor, i think this approach is fine for now

brissons (approved):

PR #2116, opened 2021-11-03
TA158697 (Part 2): Add mediaFileName property

coffeexcode:
## Relevant Rally
- [TA158697](https://rally1.rallydev.com/#/?detail=/task/612430665105&fdp=true): Add properties onto HM response

## Description
This PR uses the new media file name property instead of parsing.


brissons (commented):
I think there is one more spot on line 134 (in `_renderAttachmentView`) that will need to change to `mediaFileEntity.mediaFileName`, but other than that lgtm!

ChrisKraljevicD2L (dismissed):

coffeexcode:
> I think there is one more spot on line 134 (in `_renderAttachmentView`) that will need to change to `mediaFileEntity.mediaFileName`, but other than that lgtm!

Nice catch, fixed!

brissons (approved):

PR #2123, opened 2021-11-04
DE45687 Work with `async` default scoring rubric option loading

mostlyfabulous:
[DE45687](https://rally1.rallydev.com/#/?detail=/defect/612303356763&fdp=true): [FACE] Removing indirectly associated rubrics may not store correct defaultScoringRubricId on an assignment

Previously the `defaultScoringRubric` dropdown would asynchronously add in the rubrics as they came in.

This change now causes the `defaultScoringRubric` dropdown to only re-render when all the rubrics that need to be fetched have returned.

May need to consider doing a disable and blur on the dropdown while that happens?


mdulat (approved):

mostlyfabulous:
There's a lot of cases to test on this but local is very slow for it.

PR #2126, opened 2021-11-04
Add lang files to eslint with their own rules and fix errors

bearfriend:
`lang/` files were being excluded from linting because they require a different set of rules (more like JSON), and maintaining overrides inside each individual file was infeasible.

I did a quick search and found that `.eslintrc.json` supports override rules with pattern matching, so added this and fixed any newly-presented errors.


jeroach (approved):
LGTM, thanks :)

PR #2158, opened 2021-11-21
Add messageformat-validator

bearfriend:
- Add `messageformat-validator`
- Sync `fr-on` with `fr`
- Sync all langs with `en`
- Fix remaining linting errors
- Add `lint:langs` to npm `lint` script


package.json
```diff
@@ -26,8 +29,8 @@
     "@babel/eslint-parser": "^7.12.17",
     "@babel/polyfill": "^7.0.0",
     "@babel/preset-env": "^7.8.3",
-    "@brightspace/content-components": "github:Brightspace/content-components#semver:^2",
     "@brightspace-ui/stylelint-config": "^0.0.3",
+    "@brightspace/content-components": "github:Brightspace/content-components#semver:^2",
```
KaiPrince, line 33: Not that it matters, cause it doesn't, but why was this line moved two lines down?

KaiPrince:
This is a soft-approve from me, who is unqualified to give a real approve.

package.json
bearfriend: Hah. No idea. I guess `npm i` decided to sort it differently when I installed `messageformat-validator`

mbardakovD2L: my best guess is that originally someone added it manually, but it got re-arranged with an `npm install` or something to be in alphanumeric order. ("-" comes slightly before "/" by ascii value)

mbardakovD2L:
Are the `split ___` warnings something to be concerned about? in `components/d2l-activity-editor/d2l-activity-quiz-editor/lang/en.js`. If not, I think this is good to go.

bearfriend:
> Are the `split ___` warnings something to be concerned about? in `components/d2l-activity-editor/d2l-activity-quiz-editor/lang/en.js`. If not, I think this is good to go.

They should absolutely be fixed long-term, but they do not break anything immediately.

mbardakovD2L (approved):

PR #2169, opened 2021-11-23
fix: DE46108 handle score when no grade permissions

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/615608209587&fdp=true

This fixes the handling of grade/ungraded states (i.e. scoreOutOf) when there is no grade (i.e. from grades tool) permissions.


mostlyfabulous (approved):

components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -207,10 +207,11 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		}
 
 		const gradeUnits = scoreOutOf === 1 ? this.localize('grades.gradeUnitsSingular') : this.localize('grades.gradeUnits');
-		const isUngraded = associateGradeEntity && associateGradeEntity.gradebookStatus === GradebookStatus.NotInGradebook && !scoreOutOf;
```
mdgbayly, line 210: Assuming this is working, but some of this conditionals feel more complicated that then need to be. Is it possible that `_focusUngraded` and `_focusInput` can both be `true` (or both be `false`)? But this is whole area makes my head spin :-).
It's not really clear when `scoreOutOf` would be just falsy vs `undefined` either and what the different implications are.

mdulat: `_focusUngraded` and `_focusInput` would never both be true but they could both be false. Their values are only checked when they change (in the `updated` function). `_focusInput` can change when we go from "in-gradebook" to "not-in-gradebook" and vice-versa, even though `_focusUngraded` does not change in these cases. We want to focus the input whenever the gradebook status changes to "in-gradebook" (i.e. inGrades), or with no grade permissions we want to focus it whenever we click the ungraded button. We want to focus the ungraded button whenever it initially appears (i.e. assignment has no grade, and with grade permissions gradebook status is not-in-gradebook).

`scoreOutOf` is `undefined` when you manually clear the "Grade Out Of" input. In this case we want to show the `Ungraded` button if gradebook status is also "not-in-gradebook" (or there is no permissions to see grades).

`scoreOutOf` is `null` if we click the `Ungraded` button before we enter an actual grade. In this case we don't want to go back to showing the `Ungraded` button before we have a chance to enter a grade.

PR #2178, opened 2021-11-26
fix: Disable advanced editing button for Content Service resources, when isAdvancedEditingEnabled is false

AdmiralCoco:
- The button will always remain visible for non Content Service resources
- isAdvancedEditingEnabled will be false if the user does not have permission to manage the Content Service resource
- Add help icon with tooltip explaining how to enable advanced editing (only visible if button disabled)
- Add langterm for tooltip

Corresponding LMS PR: https://github.com/Brightspace/lms/pull/16526


MichaelKovalchuk (dismissed):

MichaelKovalchuk (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-media-file-detail.js
```diff
@@ -89,6 +89,7 @@ class ContentMediaFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeAc
 					text=${this.localize('content.advancedEditing')}
 					class="d2l-skeletize d2l-action-button"
 					@click=${this._openDialog}
+					?hidden=${mediaFileEntity.isContentServiceResource && !mediaFileEntity.isAdvancedEditingEnabled}
```
coffeexcode, line 92: I would need to double check with our PM, but I think this might need to change. I think the requested fallback for when Producer is not enabled we should use the default LMS dialog for managing subtitles.

Also thinking about it a little bit more, that probably wouldn't be good either. If they have captions they uploaded via LMS dialog and then later enabled the advanced editing I guess they would lose all their saved captions when they switch. 🤔 

I'll bring this up at standup on Monday!

MichaelKovalchuk: LMS Captions dialog is not compatible with the new Content Service Videos. LMS Captions dialog is only good for captions uploaded to the LMS file system, and Content Service only knows of captions uploaded to Content Service. 
So, the logic here is next: if the video does not belong to Content Service - use legacy Captions dialog
If the video belongs to content service and you have permission to manage this video - Advanced Editing will show up

coffeexcode: Yeah I figured they wouldn't be compatible. We had a US at the beginning that said we should be using the LMS captions dialog as a fallback (without realizing that isn't going to work). I just wanted to bring it up to our PM to explain that there will be a change from the previously expected behavior.  Just wanted to bring it up to them incase they have anything to say about it.

MichaelKovalchuk: Sounds good. We will hold until confirmation from PM

AdmiralCoco:
Existing button:
image

---
New:
image

MichaelKovalchuk (dismissed):

AdmiralCoco:
Hey @coffeexcode - this is good to review again

Also, do you know what this error is about? (from CI)
`error missing  content.advancedEditingDisabledTooltip  String missing from locale file.`

This is a new langterm I added - am I missing an additional step?

coffeexcode (approved):
LGTM!

coffeexcode:
> This is a new langterm I added - am I missing an additional step?

Should just be adding it to the lang file. It looks like the same error when the branch was out of date, but this is even in the same branch. Not sure what really changed with CI recently causing all these headaches.

AdmiralCoco:
Hmm, no dice - my branch is up to date - I also tried opening another PR with a small whitespace change, but that failed as well 😕

coffeexcode:
Hmm. Maybe we can ask the creator of the lib to fill in some blanks.
@bearfriend Sorry to summon you, but do you have any insight on how we can get this lang-related error resolved?

AdmiralCoco:
[I added a blank placeholder](https://github.com/BrightspaceHypermediaComponents/activities/pull/2219/commits/9378f15c2864927f71cfad35463178f0fcdb57c8) for the other lang files in the other test PR, and it seems to have passed - not sure if this would impact translations though

AdmiralCoco:
I reached out to Jeff S. to see if adding the blank placeholders would affect Serge, this is what he said:
>AFAIK serge only looks at the source English file for diffs, so making changes to translation files should be ok. however serge will overwrite anything in the translation files that doesn’t match what it has in its DB (via PR though, which you could probably just ignore?)

In that case, I think it should be OK to add them in?

MichaelKovalchuk (approved):

coffeexcode (approved):

PR #2179, opened 2021-11-26
Fixed readonly categories name

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F521279527196

I did some permissions testing after yesterday's iteration review and noticed that we were always showing "No category" when a user lacked edit permissions whereas the expectation is that we show a readonly view of the selected category or "No category" if there is none selected. 

The issue was that we were deriving the selected category's name from the collection of categories that gets returned from the API however we don't return anything in the collection when a user is missing permissions and so the code was interpreting that as no category being selected. 

To fix this, instead of relying on the selected category entity being present, the readonly view now only[ relies on the `selectedCategory`](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/CategoriesEntity.js#L80) property which should be present regardless of permissions. 

image


dustxd (approved):

dustxd:
Good PR description 💯

PR #2186, opened 2021-11-30
Awaited completion of category reset

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F521276679604

David pointed out funky behaviour that we could only repro on non-chromium browsers in which sometimes categories weren't being reset correctly. This fix explicitly awaits for a category to be reset.


kikid2l (approved):
yes awaiting it is a good idea

PR #2188, opened 2021-11-30
DE46178: defaultScoringRubricId not being set as `-1` for 2 or more rubrics

mostlyfabulous:
[DE46178](https://rally1.rallydev.com/#/?detail=/defect/617169546655&fdp=true): [FACE] Adding two rubrics to assignment but not selecting a default gives defaultScoringId as null

Restores the functionality of [this](https://github.com/BrightspaceHypermediaComponents/activities/pull/2095/files#diff-be124ee4d5d66a162538f2e6628e8fc8e9f399f8ff74121d46fc5bbb212ffa2eR268-R270) from: #2095

Test notes from #2095 about the scenario being fixed:
> - When a new assignment is created and no rubrics are associated yet (or when less than 2 scoring rubrics are associated), the defaultScoringRubricId should be null. You can see this in the assignment entity's properties. You can update, save, reload this assignment, as long as you never associate 2 or more scoring rubrics, defaultScoringRubricId will be null.
> 
>- When you add 2 or more scoring rubrics and the default dropdown shows, when you save the assignment you should see a PATCH request on the folder with defaultScoringRubricId, value will be -1 if the dropdown option is set to "do default", or another value which is the rubricId of the selected default rubric. Now when you reload the assignment you will see defaultScoringRubricId = -1 (for no default) or the other value you selected. You can update the default to another rubric, back to "no default", etc. from now on defaultScoringRubricId will always have a non-null value. Even if at this point you delete all scoring rubrics, that value will simply become -1.


components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js
```diff
@@ -339,7 +339,7 @@ class ActivityRubricsListContainer extends ActivityEditorMixin(RtlMixin(Localize
 		}
 
 		const defaultScoringRubricOptions = this._dedupeDefaultScoringRubricOptions([...entity.defaultScoringRubricOptions, ...indirectAssociations.defaultScoringRubricOptions]);
-		if (assignment.defaultScoringRubricId === '-1' || assignment.defaultScoringRubricId === null) {
+		if (assignment.defaultScoringRubricId === '-1' || (assignment.defaultScoringRubricId === null && defaultScoringRubricOptions.length < 2)) {
```
mostlyfabulous, line 342: Adding this granularity to the conditional now let's [L352](https://github.com/BrightspaceHypermediaComponents/activities/blob/28475be16ee046f6f63ef20635f0e13be30fedcc/components/d2l-activity-editor/d2l-activity-rubrics/d2l-activity-rubrics-list-container.js#L352) be hit so that we PATCH `defaultScoringRubricId:  -1`  when `defaultScoringRubricId` is `null` and there's at least two rubric options available on save.

dustxd (approved):

dustxd:
Rally integration doesn't seem to work with defects...I tried changing the description to get it working, I'm sorry 😅

dustxd (approved):

PR #2192, opened 2021-11-30
De45353/fix face header hierarchy

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F609974365545

The goal with this change is to make the heading hierarchy accessible in FACE. The lang terms will potentially change but design has approved them for now. Refer to the screenshots below for the desired hierarchy. I'm not a big fan of adding the same lang terms in multiple files but due to the fact that different localize mixins are being used in the necessary files I'm not sure we have an easy way to share them. 

I tested this using NVDA and everything seemed to get announced correctly. 

The change for the immersive nav bar can be found here: https://github.com/Brightspace/lms/pull/16670

image

image


dlockhart (dismissed):
Looks good. If you did want to reduce the duplication of the lang terms and the markup (which I'm assuming will get worse as more and more activity types are converted), you could create a couple shared components -- e.g. `<d2l-activity-editor-core-properties-heading>` and `<d2l-activity-editor-advanced-properties-heading>`.

ronan-f:
> Looks good. If you did want to reduce the duplication of the lang terms and the markup (which I'm assuming will get worse as more and more activity types are converted), you could create a couple shared components -- e.g. `<d2l-activity-editor-core-properties-heading>` and `<d2l-activity-editor-advanced-properties-heading>`.

Lol that's a 🤦  moment. Making a component is much more straightforward than what I was thinking... Thanks Dave.

components/d2l-activity-editor/d2l-activity-quiz-editor/lang/en.js
```diff
@@ -243,5 +243,5 @@ export default {
 	"submissionViewsDialogSecondaryEditorGradeRangeTextAnd" : "and", // text in submission view secondary view edit view, grade bound text between lower and upper amounts
 	"submissionViewsDialogSecondaryEditorNumericalInputText" : "minute(s)", // text in submission view secondary view edit view, text after the time limit box.
 	"submissionViewsDialogSecondaryEditorTimeLimitText": "Time Limit", // text in submission view secondary view edit view, text above the time limit box.
-	"newQuizCategory": "New Quiz Category" // new quiz category dialog title
+	"newQuizCategory": "New Quiz Category", // new quiz category dialog title
```
dlockhart, line 246: Nit: I think this is still valid JS, but if you want to make the diff from this PR smaller you could remove it. There's a similar whitespace change in the other lang file as well.

dlockhart (dismissed):
Awesome!

dlockhart (approved):

PR #2193, opened 2021-12-01
US132833 Improvements to release date component

dustxd:
### Summary of Changes
* Updated lang term from `Release Date` to `Display Date` as per design
* Disabled the Release Date calendar input when permissions are missing
* Stopped sending PATCH requests when calendar input value hasn't changed
  * Context: to make a change in the calendar input, the `@change` event is triggered twice
    * Once on initial click
    * Once on the actual value change
 * Updated the `updateProperty()` function in quiz submission view mobx state to coordinate updates from the API and the optimistic UI changes - this fixes the flickering issues for the checkboxes in the dialog as well
   * Context: The flickering happens when the first PATCH request (for working copy) isn't done yet and the user quickly clicks again. The component will update to the newest value but flicker to the less new value as the first PATCH request returns and flicker back to the newest value once the second PATCH request returns => `this._entity = entity` instead of `this.load(entity)` fixes this issue.

### Rally Story
https://rally1.rallydev.com/#/?detail=/userstory/611715032115&fdp=true


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-submission-views-access-restrictions-editor.js
```diff
@@ -88,17 +88,20 @@ class ActivityQuizSubmissionViewsAccessRestrictionsEditor
 			canUpdateTimeLimitNumber,
 			timeLimitNumber,
 			gradeRestrictions,
-			releaseDate
+			releaseDate,
+			canUpdateReleaseDate
```
mostlyfabulous, line 92: I think we need to check `entity` is not undefined, the usual boilerplate checks we have 😄 in our code.

dustxd: Good point

mostlyfabulous (approved):

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-submission-view.js
```diff
@@ -213,7 +213,7 @@ export class QuizSubmissionView {
 
 	async updateProperty(updateFunc) {
 		this.saving = updateFunc();
```
stevenao, line 215: we usually have the `await` at the function call level. `this.saving = await updateFunc();`

dustxd: We're actually doing that on the next line 😄 `await this.saving;` is equivalent to `await updateFunc();`

stevenao: Ok. Just a nit. The saving var is holding a promise. I thought it was a value variable.

PR #2204, opened 2021-12-03
Change name of radio buttons

coffeexcode:
## Relevant Rally
- [DE46251](https://rally1.rallydev.com/#/?detail=/defect/618587325739&fdp=true): LTI > Non-embeddable link > Cant tab through to open in new tab

## Description
After a change to LTI links to allow FACE to change the embed status for links that can't be embedded (by disabled the embed option if it can't be embedded), you could no longer tab to select a link option. This is from a browser behavior that says 'you can't tab into a group of radio buttons if one is disabled'

This change separates that radio buttons into two groups. Since they were controlled components they don't actually need to be in the same group.


Kiefer-Co-D2L (approved):
> This is from a browser behavior that says 'you can't tab into a group of radio buttons if one is disabled'

Very interesting problem and solution.  LGTM!

brissons (commented):
I wonder, since they are now in different groups, does that mean that we would be able to select both at the same time?

coffeexcode:
> I wonder, since they are now in different groups, does that mean that we would be able to select both at the same time (if both are not disabled)?

We control the values of each button rather than rely on the browser controlling the group so we should be good. I haven't been able to select both.

brissons (approved):

PR #2213, opened 2021-12-07
DE46251: Fix LTI framing options for non-embeddable LTI links

coffeexcode:
## Relevant Rally
- [DE46251](https://rally1.rallydev.com/#/?detail=/defect/618587325739&fdp=true): LTI > Non-embeddable link > Cant tab through to open in new tab

## Description
I originally created PR https://github.com/BrightspaceHypermediaComponents/activities/pull/2204 that just updated the names of the radio buttons to make it tab-able. Even after @brissons asked if would let you select both buttons and I denied it, it turned out in-fact you could do it (I only tried changing it once, but if you changed it once, then again both would stay selected). This was because in Lit, the boolean `?checked` is only on page load, not the actually checked property. This PR only changes the name of the radio button groups when the LTI can't be embedded (when the first button is disabled) so you can no longer select both (and the expected up/down arrow behavior when both radio buttons are enabled)

TL:DR
This PR should actually fix the radio button tabbing issue without new bugs 🤞 

## Remaining Work
- [ ] Someone please give this a devtest


components/d2l-activity-editor/d2l-activity-content-editor/lti-link/d2l-activity-content-lti-link-options.js
```diff
@@ -20,8 +20,7 @@ class ContentEditorLtiLinkOptions extends SkeletonMixin(ErrorHandlingMixin(Local
 			skeleton: { type: Boolean },
 			onSave: { type: Function },
 			canEmbed: { type: Boolean },
-			_linkOptionsError: { type: String },
-			isExternalResource: { type: Boolean }
```
coffeexcode, line 24: Removing this as it's not being used at all

components/d2l-activity-editor/d2l-activity-content-editor/lti-link/d2l-activity-content-lti-link-options.js
```diff
@@ -66,14 +65,14 @@ class ContentEditorLtiLinkOptions extends SkeletonMixin(ErrorHandlingMixin(Local
 			: false;
 
 		return html`
-		<div id="content-link-options-container">
-			<div class="d2l-display-options-text">
-				<label class="d2l-label-text d2l-skeletize">
-					${this.localize('content.displayOptions')}
-				</label>
+			<div id="content-link-options-container">
```
coffeexcode, line 68: Whitespace only

brissons (approved):

Kiefer-Co-D2L (approved):

PR #2217, opened 2021-12-08
US132833 Add release date and restrict access options to create view

dustxd:
### Summary of Changes
* Refactored card header into its own class
* Updated card header styling and elements to align semantically better with the design
* Moved card header into the dialog card editor as it's used by both the create and edit workflow
  * The dialog card read-only summary view still uses card header

### Rally Story
https://rally1.rallydev.com/#/?detail=/userstory/611715032115&fdp=true


dustxd:
### Screenshots
Previously:
image

Create view now contains the additional options:
image

ronan-f (approved):
LGTM!

PR #2218, opened 2021-12-08
Pass filetype to media player for audio and video files

ianlaidlaw:
the media-player component was bumped to v2 in bsi, and it now requires you tell it whether it is an `audio` or `video` file or it will not render properly. thankfully we have access to this easily on the entity and can simply pass it in.

### BEFORE:

image

### AFTER:

image

image


brissons (approved):

coffeexcode (approved):
🥳

PR #2219, opened 2021-12-08
fix: Disable advanced editing button for Content Service resources, w…

AdmiralCoco:
…hen isAdvancedEditingEnabled is false

- The button will always remain visible for non Content Service resources

- isAdvancedEditingEnabled will be false if the user does not have
  permission to manage the Content Service resource
- Add help icon with tooltip explaining how to enable advanced editing (only visible if button disabled)
- Add langterm for tooltip


AdmiralCoco:
Was testing CI for this PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/2178

PR #2228, opened 2021-12-14
fix: DE46302 set GOO not-in-gradebook when missing ManagaGrade permis…

mdulat:
…sions

Second task from: https://rally1.rallydev.com/#/?detail=/defect/618761898521&fdp=true

When user is missing manage grade permissions, they should still be able to set a Grade Out Of for Assignments if gradebook status is not-in-gradebook. Previously they couldn't get it out of "Ungraded" state.


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -196,7 +196,7 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		let canEditScoreOutOf;
 
 		if (this.hasActivityScore) {
-			scoreOutOf = scoringEntity && scoringEntity.gradeMaxPoints;
+			scoreOutOf = scoringEntity && typeof scoringEntity.gradeMaxPoints !== 'undefined' ? scoringEntity.gradeMaxPoints : '';
```
ChrisKraljevicD2L, line 199: Is `typeof` required here? Is there a problem with `scoringEntity.gradeMaxPoints !== undefined`?

mdulat: Probably doesn't make a difference in this case. If the variable is undeclared then it makes a difference.

PR #2229, opened 2021-12-14
fix: US132834: fixed error tooltip for time limit input

martinwang-d2l:
Added check to ensure time limit stays within boundaries, and causes tooltip to appear when out of bounds. 

image

Fixed styling for label on input-percent components.


martinwang-d2l:
Checked with Patrick - reverted label to be left-aligned.

mostlyfabulous (approved):

PR #2235, opened 2021-12-14
upgrading visual-diff

dlockhart:
This upgrades activities to use the latest visual-diff/puppeteer/mocha in the visual-diff action (which is now on its `main` branch).

The upgrade required switching to ES modules. In other repos I've been converting them fully to ESM-by-default (using the `"type": "module"` in `package.json`), but the old Polymer stuff can't handle that. So unfortunately I had to switch the visual-diff tests to use the `.mjs` extension to tell Mocha that they're ESM.

I'm expecting some minor visual differences from the Puppeteer upgrade.


dbatiste (approved):

jeroach (approved):
lgtm

PR #2244, opened 2021-12-15
feat: switch various components to use `@brightspace-ui/core` equivilents

devpow112:
All of these are drop in replacements with a few exceptions/additional clean-ups.

* The `announce-text` is not an available property on `d2l-alert` and `d2l-alert-toast` anymore.
* Icon tier names don't need to start with `d2l-` anymore. In fact the code in `d2l-icon` strips it out.
  *  https://github.com/BrightspaceUI/core/blob/main/components/icons/icon.js#L63.
* The color `--d2l-color-white` is no longer defined and can just be replaced by `white`.

The `d2l-button`, `d2l-typography` and `d2l-offscreen` can't be removed as the polymer style shims and behaviours are used from them and are unavailable in `core`


dlockhart (approved):
LGTM!

PR #2248, opened 2021-12-16
DE46154 & DE46153 FACE: Respect `maxPoints` for Distributed Grade Categories

mostlyfabulous:
[DE46154](https://rally1.rallydev.com/#/?detail=/defect/616901920397&fdp=true): FACE Assignment > Updating Grade Out Of value on an assignment linked to a Grade item under grade category with distributed points, doesn't respect the distribution
[DE46153](https://rally1.rallydev.com/#/?detail=/defect/616899740963&fdp=true): FACE Assignment > When choose to link to existing grade item as part of grade category with distributed points, grade out of should not be editable

https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/438


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -191,13 +191,15 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 		const canSeeGrades = !!associateGradeEntity;
 		const canEditGradebookStatus = associateGradeEntity && associateGradeEntity.canCreateNewGrade;
 		const isGradebookStatusChanging = associateGradeEntity && associateGradeEntity.gradebookStatusChanging;
+		const isAutoPointsAndGraded = scoringEntity && scoringEntity.autoPoints && inGrades;
+		const isAutoPointsAndUngraded = scoringEntity && scoringEntity.autoPoints && !inGrades;
```
mdulat, line 195: It feels like this state should never happen. If there is no associated grade then autoPoints should never be true. I think instead of handling this case here we should ensure that is the case never happens (e.g. when we put it into not-in-gradebook state we should also set autoPoints to false at that point).

mostlyfabulous:
Tests will pass once BrightspaceHypermediaComponents/siren-sdk#438 is merged

mdulat (approved):

PR #2257, opened 2021-12-17
feat: missing manage grades can link to existing

mdulat:
Last task in https://rally1.rallydev.com/#/?detail=/defect/618761898521&fdp=true

Have to distinguish between permission to change gradebook status and permission to create new grade. Previously they were considered the same.


mostlyfabulous (approved):
Yes to splitting the permissions up!

PR #2260, opened 2021-12-20
fix: fixed issues with access restrictions, US132834.

martinwang-d2l:
Fixed several issues in https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F611715477831: (US132834 add access restrictions options). 

- PATCH request being sent with undefined payload if one of min/max grades are deleted through the UI. Resolved through fixing event handler. 
- Time limit value blanking upon first editing the input box. Was an issue with a mobX function.


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-submission-views-access-restrictions-editor.js
```diff
@@ -210,6 +210,11 @@ class ActivityQuizSubmissionViewsAccessRestrictionsEditor
 		`;
 	}
 
+	_isInputGradeValid(value) {
+		const targetValue = (value === undefined) ? null : String(value);
```
mostlyfabulous, line 214: value comes from `d2l-input-number` right?

martinwang-d2l: Yeah - this function is called with (e.target.value) which comes from d2l-input-number

components/d2l-activity-editor/d2l-activity-quiz-editor/state/quiz-submission-view.js
```diff
@@ -208,7 +208,7 @@ export class QuizSubmissionView {
 	}
 
 	setTimeLimitNumber(value) {
-		this.timeLimitNumber = value;
+		this.timeLimitNumber.value = value;
```
mostlyfabulous, line 211: Is `value` a property on the siren entity? I would have expected `timeLimitNumber` to be the just be a root level property only that has the value already without the need for more a `value`.

martinwang-d2l: The way we made timeLimitNumber was that it has timeLimitNumber.min, timeLimitNumber.max, and timeLimitNumber.value: the min and max being used in the d2l-input-number component to ensure the input doesn't exceed limits - in this case we originally forgot to specify the .value field

martinwang-d2l: ^ So that's for the mobx object: for Siren, yeah when PATCHing the entity only has a single value in timeLimitNumber

image

mostlyfabulous (approved):

PR #2266, opened 2021-12-21
LOR > Last modified time is inconsistent with the time listed in the Settings pop-up dialog

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F620789712357

## Description

The time shown under the LOR second heading banner on the FACE page does not match the time given in the settings pop-up dialog.  The settings dialog appears to be one hour ahead of UTC.

See initial screenshots from Gurpinder, who lives in Vaughn, Ontario (**UTC-5**):

image

The banner is **four** hours ahead of the pop-up dialog.

When repeated in Winnipeg (**UTC-6**), the banner is **five** hours ahead of the pop-up dialog:

Not Fixed LOR Time

## Goal/Solution

The LOR API shows that the time on the banner is the same as what comes from the back-end:

image

(See `DateCreated`, where 23:04 == 11:04 PM)

This implies the settings dialog pop-up is in UTC+1.  A function has been added to change the time to match the dialog's time zone.

The SCORM FACE page has also been modified to handle locales (not time zones) in a manner consistent with LOR's.

## Video/Screenshots

Fixed LOR Time

## Related PRs

- This PR will have a cert fix.

## Remaining Work


components/d2l-activity-editor/d2l-activity-content-editor/lor/d2l-activity-content-lor-activity-detail.js
```diff
@@ -144,6 +144,13 @@ class ContentLorActivityDetail extends SkeletonMixin(ErrorHandlingMixin(Localize
 		`;
 	}
 
+	convertToQuickLinkTime(date) {
+		// The time presented by the settings editing dialog pop-up is one hour ahead of UTC
+		const minutesPerHour = 60;
+		const msPerMinute = 60000;
+		return date - (new Date().getTimezoneOffset() - minutesPerHour) * msPerMinute;
```
prajapac, line 151: Would be better for readability to add brackets:
`return date - ((new Date().getTimezoneOffset() - minutesPerHour) * msPerMinute);`

prajapac (commented):
LGTM, though please verify if possible that the value returned from B/E will always be UTC+1.

components/d2l-activity-editor/d2l-activity-content-editor/lor/d2l-activity-content-lor-activity-detail.js
kristian-torres: we discussed what else could cause the mismatch of time here involving possibly backend date generation or Daylight saving maybe? is there a status on this?

Kiefer-Co-D2L: Still trying to figure out where it's coming from, I'll ping you if I find out!

kristian-torres: the changes look like they'll work as per our discussion though so I'm cool with it!

components/d2l-activity-editor/d2l-activity-content-editor/lor/d2l-activity-content-lor-activity-detail.js
```diff
@@ -144,6 +144,13 @@ class ContentLorActivityDetail extends SkeletonMixin(ErrorHandlingMixin(Localize
 		`;
 	}
 
+	convertToQuickLinkTime(date) {
+		// The time presented by the settings editing dialog pop-up is one hour ahead of UTC
```
brissons, line 148: I'm not sure about this part. User's can set their own timezone in `Account Settings`, and I think that the settings editing dialog reflects this. `getTimezoneOffset` however gets the bowser/desktop time. I imagine if the user's local is different than 1 hour ahead of UTC we will end up with the inconsistency again.

This almost feels like it should be a backend change. Are we localizing the time to the user preference in the response?

Kiefer-Co-D2L: I testing changing Account Settings timezone and here's what the pattern appears to be:

Time on LOR settings pop-up dialog = (date created in UTC) + (browser time zone - account settings time zone).

When the browser time zone is the same as the account settings time zone, the pop up shows exactly UTC+0.

brissons: Oh weird, I tried changing my browser timezone and it didn't look like it affected the LOR popup.

Kiefer-Co-D2L: Wait, I think what's going on.  The pop-up dialog is the correct time.  I thought the banner times were being localized wrong, but they're actually **not** being converted to local time zone, since

image

AND

https://desire2learn.atlassian.net/wiki/spaces/ALLOY/pages/1858437430/Dates+and+Time+Zones
(I'm assuming this applies to us too since it's database related?)

I need to find a way to get the user settings time zone.

Kiefer-Co-D2L:
Update: Phil and Scott and I jumped on a call and found a way to do it in the backend.  This PR will be closed.

PR #2277, opened 2021-12-24
fix: DE46314 allow link to existing grade when missing manage grades …

mdulat:
…permission

https://rally1.rallydev.com/#/?detail=/defect/618869797257&fdp=true

For quizzes


mostlyfabulous (approved):
It's growing bigger 🧟

PR #2285, opened 2022-01-05
DE46591 remove extra LOR scrollbar

ChrisKraljevicD2L:
[DE46591 rally link](https://rally1.rallydev.com/#/289692574792/detail/defect/620911538813)

I zoomed in, out, squashed and stretched my browser window as much as I could and I could not get the iframe to scroll. So I disabled scrolling on it entirely. I tried to be thorough in my dev testing, but I think this should get some extra scrutiny if we're applying this attribute on all LOR iframes.

TODO:
- [ ] Thorough dev testing


brissons (commented):
Interesting! It looks like the scrolling is being handled by the LOR renderer? LGTM. I can give it a dev test.

brissons (approved):
LGTM1 Tested with `lor-use-new-embed-view` enabled with a few different files and there was no extra scrollbar + scrolling was available.

PR #2295, opened 2022-01-10
Migrate to web-test-runner to remove open-wc/testing-karma dependency

annabelsegalld2l:


package.json
```diff
@@ -32,7 +33,8 @@
     "@brightspace-ui/stylelint-config": "^0.0.3",
     "@brightspace/content-components": "github:Brightspace/content-components#semver:^2",
     "@open-wc/testing": "^2.5.1",
-    "@open-wc/testing-karma": "^4.0.5",
+    "@web/test-runner": "^0.13",
+    "@web/test-runner-saucelabs": "^0.6",
```
dlockhart, line 37: I personally like installing this in the GitHub action [like this](https://github.com/BrightspaceUI/create/blob/main/src/generators/test-unit/templates/static/.github/workflows/ci.yml#L16). That way it's not taking up time installing it locally when you're not going to use it.

dlockhart (approved):

dlockhart:
Looks like there's a fair bit of CommonJS in this repo. Probably easier to use the `.mjs` extension for the config files and remove `"type": "module"` for now. You'll need to pass the `--config` flag to the test runner to tell it where the files are.

package.json
```diff
@@ -13,13 +13,13 @@
     "lint:wc": "polymer lint -i \"components/**/*.js\" \"demo/**/*.js\" \"test/**/*.js\"",
     "langs:rename": "mfv rename",
     "langs:sync": "mfv add-missing && mfv remove-extraneous",
-    "test:karma": "karma start",
-    "test:karma:watch": "karma start --auto-watch=true --single-run=false",
-    "test:polymer:local": "polymer test -p --skip-plugin sauce",
+    "test:headless": "web-test-runner --group default",
+    "test:headless:watch": "web-test-runner --group default --config web-test-runner.config.mjs",
+    "test:polymer:local": "polymer test -p --skip-plugin sauce  --config web-test-runner.config.mjs",
```
dlockhart, line 18: I think you want something like this:

```suggestion
    "test:headless": "web-test-runner --config web-test-runner.config.mjs",
    "test:headless:watch": "web-test-runner --config web-test-runner.config.mjs --watch",
    "test:polymer:local": "polymer test -p --skip-plugin sauce",
```

`polymer test` isn't going to understand the `@web/test-runner` config file -- it uses the older `wct.conf.json` and I think we just want to leave it alone unless we plan to rewrite all those test HTML files to the new format. Here's an [example of doing that](https://github.com/BrightspaceUI/navigation/pull/127) and it's a fair amount of work.

dlockhart (approved):
🎉

PR #2304, opened 2022-01-12
fix: US131498: call IQ clean up api when exiting IQ dialog. https://rally1…

alexli01:
….rallydev.com/#/?detail=/userstory/608682776473&fdp=true


mostlyfabulous (approved):
Any failure handling needed when `!response.ok`?

PR #2308, opened 2022-01-13
US134136: Show embed view for imported SCORM topics on FACE page

ianlaidlaw:
lms pr: https://github.com/Brightspace/lms/pull/17903

Adds the preview component for showing embedded imported SCORM packages. There is another view for showing the attachments view if it cannot be embedded, but as of right now the logic is hard coded to always show the embed view. ~I am still determining if a) that is something we can even determine, b) if it's even necessary or possible for it not to be embedded. With normal SCORM we always show the embed view and my current thinking is that this should be the same.~
We will be keeping the hardcoded `true` value in case these needs to change in the future, but for now we will be treating this the same as the normal SCORM page.

## SCREENSHOTS:

image

image

image

importedscorm


components/d2l-activity-editor/d2l-activity-content-editor/imported-scorm/d2l-activity-content-imported-scorm-activity-detail.js
```diff
@@ -89,6 +90,12 @@ class ContentImportedScormActivityDetail extends SkeletonMixin(ErrorHandlingMixi
 				?skeleton="${this.skeleton}"
 				?showLabel="${false}"
 			></d2l-activity-content-external-activity-container>
+
+			<d2l-activity-content-imported-scorm-preview
+				.entity="${importedScormActivityEntity}"
+				?skeleton=${this.skeleton}
+				?showActivityPreview=${true}
```
ianlaidlaw, line 97: hard coding this to true for now. as mentioned in the pr description I am still determining if this is something we can even do, or if it's even necessary to do.

components/d2l-activity-editor/d2l-activity-content-editor/imported-scorm/d2l-activity-content-imported-scorm-preview.js
```diff
@@ -0,0 +1,90 @@
...
+		`;
+	}
+
+	_renderAttachmentView(importedScormActivityEntity) {
+		if (!importedScormActivityEntity) {
+			return html``;
+		}
+
+		const attachment = {
+			name: this.localize('content.scormActivity'),
```
ianlaidlaw, line 67: confirming this with design

components/d2l-activity-editor/d2l-activity-content-editor/imported-scorm/d2l-activity-content-imported-scorm-preview.js
```diff
@@ -0,0 +1,90 @@
...
+			skeleton: { type: Boolean },
+			showActivityPreview: { type: Boolean }
+		};
+	}
+
+	static get styles() {
+		return  [
+			super.styles,
+			activityContentEditorStyles,
+			css`
```
coffeexcode, line 23: I think we're missing the border styles around the preview component.

coffeexcode (commented):
LGTM. Just a single thing

components/d2l-activity-editor/d2l-activity-content-editor/imported-scorm/d2l-activity-content-imported-scorm-preview.js
coffeexcode: I agree for LOR, but the regular SCORM one should have some spacing around it. Something looks weird on the page in your screenshot. 

The SCORM page should look like this (which has some padding)
image

ianlaidlaw: I believe the padding is only applied to the subtitle (last edited...) which looks to be optional?
```
				.d2l-external-activity-subtitle {
					display: inline-block;
					height: 1rem;
					margin: 12px 0;
				}

....

	_renderSubTitle() {
		return this.subTitle
			? html`
				<p class="d2l-body-small d2l-skeletize d2l-external-activity-subtitle">
					${this.subTitle}
				</p>`
			: html``;
	}
```

could that be why my screenshot looks weird, since I don't have a subtile?

ianlaidlaw: it's worth noting i just quickly tested with a broken quad SCORM topic, but the styling should be the same even if the topic is broken. unless subtitle isn't actually supposed to be optional?

coffeexcode: Oh yeah that would be it... I guess even better for the recent changes. It's not optional anymore so it should match up now.

ianlaidlaw:
pr description screenshots are slightly out of date, the preview now has a border:

# BEFORE:

image

# AFTER:

image

PR #2314, opened 2022-01-17
US135293 - remove old date-time picker - replace with d2l-input-datetime

annabelsegalld2l:
The old date-picker is deprecated, so we want to replace its last use with the new d2l-input-date-time. 

Made some slight UI changes to accommodate for the new date-picker: 

- Made the dialog fullscreen so we don't need to resize every time a radio button is selected, and the date-picker has more space to open.
- Removed the break between the body text and set of radio buttons to make sure the calendar footer is visible on most screens (otherwise, the users will need to scroll, which is still doable). Note on mobile the date-picker behaves differently, popping out from the bottom rather than dropping down from the input.

Before: 
old-calendar

After: 
new-calendar

Mobile mode: 
new-calendar-mobile


components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -52,9 +34,8 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
 		const groupID = this._genGroupID();
 
 		return html`
-			<d2l-dialog title-text="${this.localize('dismissActivity')}">
+			<d2l-dialog-fullscreen title-text="${this.localize('dismissActivity')}">
```
annabelsegalld2l, line 37: I think making this dialog a fullscreen rather than constantly resizing the dialog whenever we change radio options is a bit cleaner and prevents the dialog from jumping around the screen.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -52,9 +34,8 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
 				<div class="d2l-body-standard">${this.localize('dismissingAnActivityHides')}</div>
-				<br/>
```
annabelsegalld2l, line 57: Cleared this break element to make more space for the whole calendar to be seen on most screens. I can definitely put it back in but I thought it would be good to get the footer of the calendar visible without scrolling.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -113,41 +87,32 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
 	_getDialog() {
 		return this.shadowRoot.querySelector('d2l-dialog');
 	}
 
 	_updateProp(propName) {
 		this.selectedRadio = propName;
-		this._getDialog().resize();
-	}
-
-	_getIso8601Date(date) {
```
annabelsegalld2l, line 125: The new component takes care of all the localization for us 🎉 . Also deleted the tests for this function since they will no longer be needed.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -113,41 +87,32 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
-						placeholder="MM|DD|YYYY"
-						min="${minDate}"
-						hide-label
-						always-show-time
-					></d2l-datetime-picker>
+					<d2l-input-date-time
+						@change="${this._onDatetimePickerDatetimeChanged}"
+						min-value="${minDate}"
+						label-hidden
+						label="${this.localize('dismissUntil')}"
```
annabelsegalld2l, line 112: Labels are required on all the date pickers now, so I just used the "dismiss until..." string for now, but we can make a new langterm as well.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -1,33 +1,15 @@
...
-import '@brightspace-ui/core/components/dialog/dialog.js';
+import '@brightspace-ui/core/components/dialog/dialog-fullscreen.js';
 import '@brightspace-ui/core/components/button/button.js';
-import 'd2l-datetime-picker/d2l-datetime-picker.js';
+import '@brightspace-ui/core/components/inputs/input-date-time.js';
 import 'd2l-polymer-behaviors/d2l-id.js';
 import { DISMISS_TYPES } from './dismiss-types.js';
 
-// shim for vaadin calendar to overlay on top of the dialog
-(()=>{
```
annabelsegalld2l, line 14: We can delete this Vaadin shim styling code used for the old date-picker.

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
jeroach: Yeah this seems appropriate to me

For context:
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-quick-eval/lang/en.js#L20

jeroach (approved):
LGTM

Scy-D2L (dismissed):

components/d2l-quick-eval/dismiss/d2l-quick-eval-action-dismiss-dialog.js
```diff
@@ -113,41 +88,33 @@ class D2LQuickEvalActionDialog extends RtlMixin(LitQuickEvalLocalize(LitElement)
...
-			return date.format('YYYY-MM-DD');
-		}
 	}
 
 	renderDatePicker(selectedRadio) {
-		/* global moment:false */
-		const now = moment();
-		const minDate = this._getIso8601Date(now);
+		// minimum date-time allowed is now (to the minute)
+		const minDate = formatDateTimeInISO(convertLocalToUTCDateTime(getToday()));
```
annabelsegalld2l, line 105: So sorry for the re-review! Just made this quick fix here. It makes sure the minimum date required is now. So if it is 12:00PM, the date picker will only allow you to enter 12:01PM or later on today's date. This actually fixes a bug in the previous date-picker, where you could set the dismiss until to today's date but an earlier time, throwing an exception. (the code I had previously was using the local time instead of the UTC time, needed to add that conversion in the middle)

jeroach: No worries at all :)

jeroach (approved):

Scy-D2L (approved):

PR #2326, opened 2022-01-20
fix: DE46746 fix usage of ActivityEditorMixin

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/622230285877&fdp=true

I think the problem was that the `ActivityEditorMixin` wasn't getting the `store` in the constructor because it wasn't listed as the first mixin, so it wasn't being included in the list of components to wait for before going out of skeleton mode. 

I fixed this everywhere the mixin wasn't listed at the beginning, as well as some other small cleanups.

Will have to test this on quad because it's not reproducible locally, as it's a timing issue (i.e. order in which GETs are completed).


components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
```diff
@@ -14,15 +14,7 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 import { shared as store } from '../state/activity-store.js';
 
-class ActivityAssignmentAvailabilityEditor extends SkeletonMixin(LocalizeActivityAssignmentEditorMixin(ActivityEditorMixin(MobxLitElement))) {
-
-	static get properties() {
```
dustxd, line 19: would removing this affect calls like `this.href` and `this.token`?

components/d2l-activity-editor/d2l-activity-editor-footer.js
```diff
@@ -2,13 +2,12 @@ import './d2l-activity-editor-buttons.js';
 import './d2l-activity-visibility-editor.js';
 import { AsyncContainerMixin, asyncStates } from '@brightspace-ui/core/mixins/async-container/async-container-mixin.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
-import { ActivityEditorMixin } from './mixins/d2l-activity-editor-mixin.js';
 import { LocalizeActivityEditorMixin } from './mixins/d2l-activity-editor-lang-mixin';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SaveStatusMixin } from './save-status-mixin.js';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
-class ActivityEditorFooter extends SkeletonMixin(AsyncContainerMixin(SaveStatusMixin(ActivityEditorMixin(RtlMixin(LocalizeActivityEditorMixin(LitElement)))))) {
```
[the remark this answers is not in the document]
mdulat: Yeah, this component doesn't load anything directly, so it doesn't need this mixin.

components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-availability-editor.js
mdulat: Nope, those are defined within `ActivityEditorMixin` so don't need to be re-defined here.

dustxd: I see 😄

dustxd (approved):

PR #2347, opened 2022-02-02
Fix for Imported Scorm undefined grade name

ianlaidlaw:
The title passed to the grade editor was "undefined", and this is what gets put into working copy and goes all the way up the chain until the grade is created and associated. This change grabs the correct value, so we pass the activity title up to be the grade name.

## BEFORE:
image
image
image

## AFTER:

image
image
image


brissons (approved):

prajapac (approved):

Kiefer-Co-D2L (approved):

coffeexcode (approved):
t

PR #2362, opened 2022-02-08
US135527 Added reading FAD dates from the ActivityUsage siren entity

pureooze:
This PR allows FACE to load the FAD dates from the siren entity.

## Previous PR
1. https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/453
2. https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/454

TODO: Add tests in future PR
## Rally Story
https://rally1.rallydev.com/#/?detail=/userstory/623042532951&fdp=true


components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -121,16 +122,18 @@ class ActivityAvailabilityDatesEditor extends ActivityEditorMixin(LocalizeActivi
...
 			startDateType: null,
 			endDate: null,
 			endDateType: null,
 			startDateError: null,
 			endDateError: null,
-			defaultStartDateType: 0, //@todo: (accessRestricted) this default type should be fetched from the store once it is updated
-			defaultEndDateType: 0, //@todo: (accessRestricted) this default type should be fetched from the store once it is updated
-			displayInCalendar: false, //@todo: this value should be fetched from the store once it is updated
+			defaultStartDateType: 0,
+			defaultEndDateType: 0,
```
adkulas, line 134: Based on the way we decided to implement the activity usage, I don't think having default type here is necessary. 
I know in the previous PR we added the [`defaultStartType`](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/453/files#diff-b81c79c0e332ef29b7d3ef8ff523e2fe24ba38fa929791c2538a2a76c484a404R250) and `defaultEndType` methods to the class, but thinking about it more, the term "default" is misleading.

**More explanation:**
In dropbox when the usage is being created it will check is the date exists. if it does it will add for the date and type. However, if the date does not exist, it will make the date field null and the type field have the default value from the config settings. 
Might be easier to just look at the logic [here](https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox/Plugins/Activities/LMSAssignmentActivityUsageProvider.cs?r=0fba2854#236-237). 

The reason it's "confusing" IMO is because _sometimes_ a dateType is the actual type, and _other times_  it's the default from the config.

In terms of what the UI here actually needs, it just needs to display the type given by the usage object ([based on our team discussions - **warning it's long**](https://d2l.slack.com/archives/G50U54PGT/p1642445265275000)) and should be concerned whether the type comes from defaults or is legitimate etc. 

Definitely interested in everyone's thoughts though

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -140,8 +143,12 @@ class ActivityAvailabilityDatesEditor extends ActivityEditorMixin(LocalizeActivi
...
 		const dates = entity.dates;
 
-		if (!dates.canEditDates) {
-			datesEntity.canEditDates = false;
+		if (!dates.canEditStartDate) {
+			datesEntity.canEditStartDate = false;
+		}
+
+		if (!dates.canEditEndDate) {
+			datesEntity.canEditEndDate = false;
```
adkulas, line 151: Are we intending to use this for permissions checks?

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -263,7 +288,7 @@ class ActivityAvailabilityDatesEditor extends ActivityEditorMixin(LocalizeActivi
```
adkulas, line 288: Up above is where I'm suggesting that we can pass date type in blindly (sorry github won't let me select it)

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
pureooze: Hmm what happens if no date is set on load?

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
kieranderson: Or we move it to the sirenSDK and then we just references dates.endDateType

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -45,7 +45,8 @@ class ActivityAvailabilityDatesEditor extends ActivityEditorMixin(LocalizeActivi
 
 	render() {
 		const {
-			canEditDates,
+			canEditStartDate,
```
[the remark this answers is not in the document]
pureooze: Hmm well the actions are separate IIRC. Should I make a general availability dates property in the UI for now then? We can move it to siren next week (which is where it should be if we want it long term IMO).

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
```diff
@@ -55,9 +56,9 @@ class ActivityAvailabilityDatesEditor extends ActivityEditorMixin(LocalizeActivi
 			defaultStartDateType,
 			defaultEndDateType,
 			displayInCalendar,
+			hasActivityAvailabilityDates
 		} = this._getDateValues();
 
-		const renderDateTypeEditor = startDateType !== undefined && endDateType !== undefined;
 		const startType = (startDateType === undefined || startDateType === null)
```
adkulas, line 62: Something seems not right here...

components/d2l-activity-editor/d2l-activity-availability-dates-editor.js
adkulas: The action will contain the default type value for that date. I think we handled that logic in siren if I'm not mistaken

kieranderson (approved):

PR #2373, opened 2022-02-10
US136099 - Sub old d2l-offscreen with core version

annabelsegalld2l:
BrightspaceUI/offscreen is deprecated, so we are moving the last usages to the core equivalent. There should be no effect on the code.


components/d2l-activity-list-item/d2l-activity-list-item.js
```diff
@@ -218,7 +206,7 @@ class D2lActivityListItem extends mixinBehaviors([
 				<hr class="d2l-activity-list-item-top-line" />
 				<h2 class="d2l-activity-list-item-header-no-margin">
 					<a class="d2l-focusable" href$="[[_activityHomepage]]" aria-label$="[[_accessibilityDataToString(_accessibilityData)]]">
-						<span class="d2l-activity-list-item-link-text">[[_accessibilityDataToString(_accessibilityData)]]</span>
+						<d2l-offscreen class="d2l-activity-list-item-link-text">[[_accessibilityDataToString(_accessibilityData)]]</d2l-offscreen>
```
annabelsegalld2l, line 209: D2l-offscreen applies the offscreen styles automatically, so we can delete the extra CSS. The display is also not needed since the text is off the screen.

package.json
```diff
@@ -66,7 +66,7 @@
     "@adobe/lit-mobx": "^0.0.x",
     "@brightspace-hmc/foundation-components": "github:BrightspaceHypermediaComponents/foundation-components#semver:^0",
     "@brightspace-ui-labs/accordion": "^2",
-    "@brightspace-ui-labs/edit-in-place": "^1",
+    "@brightspace-ui-labs/edit-in-place": "^1.0.21",
```
annabelsegalld2l, line 69: Just making sure we always get the latest version of edit-in-place, as it was causing some CI issues earlier today.

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
```diff
@@ -5,7 +5,6 @@ import '@brightspace-ui/core/components/alert/alert.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import 'd2l-table/d2l-table-wrapper.js';
 import 'd2l-button/d2l-button.js';
-import 'd2l-offscreen/d2l-offscreen.js';
```
dlockhart, line 8: Weird, so they weren't using the old mixin but were using `<d2l-offscreen>` but not importing it eh.

dlockhart (approved):

components/d2l-quick-eval/d2l-quick-eval-submissions-table.js
dlockhart: Ah OK I see that makes more sense!

vix633 (approved):
Thanks!

PR #2382, opened 2022-02-14
Made Allocation options dropdown responsive

alexvlsem:
[DE46880](https://rally1.rallydev.com/#/?detail=/defect/622938061589&fdp=true): Allocation options dropdown is not responsive
Removed 1st column sticky behavior of the mapping table when the screen is narrower than 768px 
Added an entry to the activities serge file for the d2l-activity-evaluation-editor


activities.serge.json
```diff
@@ -274,5 +274,28 @@
...
       "tr-tr tr"
     ]
+  },
+  {
+    "name": "activityEvaluationEditor",
+    "parser_plugin": {
+      "plugin": "parse_js"
+    },
+    "source_match": "en\\.js$",
+    "source_dir": "components/d2l-activity-evaluation-editor/lang",
```
sarahboyd, line 284: Does this need to have the `d2l-activity-editor` as well?

```suggestion
    "source_dir": "components/d2l-activity-editor/d2l-activity-evaluation-editor/lang",
```

activities.serge.json
```diff
@@ -274,5 +274,28 @@
...
+    "output_file_path": "components/d2l-activity-evaluation-editor/lang/%LANG%.js",
```
sarahboyd, line 285: Same as above

```suggestion
    "output_file_path": "components/d2l-activity-editor/d2l-activity-evaluation-editor/lang/%LANG%.js",
```

activities.serge.json
alexvlsem: Good catch. Thank you!

activities.serge.json
alexvlsem: Updated

PR #2395, opened 2022-02-17
Fixed issue where IP restrictions appears incorrectly in submission views lang term. 

martinwang-d2l:
Fixed issue where IP restrictions appears incorrectly in submission views lang term.

Issue was with the "submissionViewReleaseDateSummary" langterm.


dustxd:
did you check if the string still appears when there are actually IP restrictions?

dustxd:
also do we need to update every lang file to remove this?

martinwang-d2l:
Yep - the permutations I removed were for the sentence incorrectly appearing when there are no IP restrictions - the sentence is still there for the correct ones

dustxd (approved):

PR #2428, opened 2022-03-02
remove no longer used code

mdulat:
Given that `d2l-activity-collection-editor` has been replaced by [d2l-activity-editor-collection-add](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/components/activity/editor/collection/d2l-activity-editor-collection-add.js), let's remove the no longer used code, as it does cause confusion when one's trying to figure what's what in the lms!


tenbradley (commented):
I'm happy to remove this code to prevent confusion. While we're removing this code, are you planning to clean up the lms code that is related to this?

mdulat:
> I'm happy to remove this code to prevent confusion. While we're removing this code, are you planning to clean up the lms code that is related to this?

Ideally yes, though this is more difficult as it is not obvious what is still being used. Most of the backend is being used by the new component in foundation-components. Though there is a duplicate serializer that we will be marking obsolete for now just to clarify which one should be expanded on going forward (this duplicate serializer in the lms was also a source of confusion).

mdgbayly (approved):

PR #2436, opened 2022-03-03
fix:  Line breaks in search results text

alexvlsem:
[DE47358](https://rally1.rallydev.com/#/?detail=/defect/627354229853&fdp=true): Evaluator Dialog > Line breaks in search results text not escaped


jmurray-d2l:
Looks okay to me, is this a common pattern?

jmurray-d2l (approved):

alexvlsem:
> 
I found the same approach at d2l-activity-quiz-editor-detail.js line #133

PR #2437, opened 2022-03-03
DE46888: FACE Media Skeletons

coffeexcode:
## Relevant Rally
- [DE46888](https://rally1.rallydev.com/#/?detail=/defect/622948038423&fdp=true): FACE Media > Loading skeleton sizes are misleading

## Description
We made the media FACE page branch of the content file page, which introduced the need for a generic skeleton page. However, we can do better than generic by getting a little extra context from Lessons. This PR gets a query param from lessons to check if the file being loaded is a media file, and renders a nice looking skeleton instead of a generic

## Screenshots
Before:
image

After:
skeletons


prajapac (approved):

connor-lowey (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-loading.js
```diff
@@ -39,11 +45,23 @@ class ContentFileLoading extends SkeletonMixin(MobxLitElement) {
 	}
 
 	render() {
+		const params = new Proxy(new URLSearchParams(window.location.search), {
```
brissons, line 48: It would be nice if we could just use the existing skeletons of each detail (since we would then have 1 source of skeleton truth), but this method works!

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-loading.js
```diff
@@ -17,13 +17,19 @@ class ContentFileLoading extends SkeletonMixin(MobxLitElement) {
...
 			}
-			.d2l-activity-label-container {
-				margin-bottom: 7px;
-			}
 			.d2l-file-content-load {
 				display: flex;
 				flex: 1;
 			}
+			.d2l-media-file-load {
+				max-height: 17rem;
```
brissons, line 25: I recall for the daylight design specs [everything is supposed to be multiples of 6px](https://daylight.d2l.dev/style-elements/layout-spacing/)? Are these "non-multiples-of-6" numbers needed since they aren't around components?

brissons: (woops I mean specifically for `d2l-media-file-banner-load`, not this one)

brissons (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-loading.js
coffeexcode: Yes... I looked at this size of the element the skeleton was for, and it had this as the min size so I copied it over for consistency. As for `.d2l-media-file-banner-load`, The LR margins are for the banner sizing to match what the actual banner is (copied over the css from the shared component) and the 28px bottom margin is some math to minimize layout shift, but I just realized I can't do math so I'll change it.

Does this max sense? Min margin = 18, Max margin = 56 (18 top + 18 bottom + 20 line height ) which means a margin of 37 (or we can even make it 36 for daylight) should minimize shift.

brissons (approved):

PR #2439, opened 2022-03-03
DE46889: Add alert role to save message alert toast

brissons:
## Relevant Rally Links

- [DE46889](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F622954793823&fdp=true): FACE > Screenreader does not automatically announce "Saved successfully" pop-up dialog

## Description

The screen reader does not announce that our FACE pages saved successfully.

The `d2l-alert-toast` actually [adds the alert role to the toast when it's opened](https://search.d2l.dev/xref/BrightspaceUI/core/components/alert/alert-toast.js?r=c77b1c36#251), which is strange why this wasn't working in the first place.

Turns out that `d2l-backdrop` was actually removing the alert role before the toast appeared on the page. When activated, this component [adds the presentation role](https://search.d2l.dev/xref/BrightspaceUI/core/components/backdrop/backdrop.js?r=f540cf3c#163) to all the child components, [then removes them ](https://search.d2l.dev/xref/BrightspaceUI/core/components/backdrop/backdrop.js?r=f540cf3c#208)(when `shown` is set back to `true`). It was this step that was removing the alert role, since between the `d2l-backdrop` activating and de-activating, it was unaware that the alert role was added to the toast. The removal of the role made it so the screen reader would not announce that the save was successful.

This PR directly adds the role to the component, since `d2l-backdrop` maintains existing roles. 

Tested with all combinations of FF, Chrome and JAWS, NVDA.

## Goal/Solution

Have the screenreader read the save successful message when the page is saved.


coffeexcode (approved):
This is pretty big, can you break it up into more than 1 PR next time?

ianlaidlaw (approved):

brissons:
> This is pretty big, can you break it up into more than 1 PR next time?

image

PR #2442, opened 2022-03-04
removing Serge config for activity-collection-editor

dlockhart:
#2428 removed the code associated with this config. This is currently breaking BSI builds:
https://github.com/Brightspace/brightspace-integration/runs/5425320498?check_suite_focus=true


mdulat (approved):
Thanks for taking care of this!

PR #2466, opened 2022-03-14
DE47655 Remove word "group" from start/end date widget in screen readers

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F628618321793

## Description

> Problem - Actual Behaviour:
>
> Navigating to the Due/Start/End date fields using a screenreader announces the label + selector group. Audio is "Group Start Date" or "Group Due Date" which is confusing as a group is an actual concept within Brightspace, but does not apply here. 
>
>
> Expected Behaviour:
>
> The component group should not be announced through the screenreader and only read out the actual label for Start Date and then the date selector.

### Problem Location

[The "group" keyword is associated with the fieldset tag in JAWS](https://freedomscientific.github.io/VFO-standards-support/html.html#toc-f).

The fieldset in question is found in [`d2l-input-fieldset`](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-fieldset.js), which is a core UI component within another core UI component, [`d2l-input-date-time`](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-date-time.js).

The problem occurs twice in [`d2l-activity-availability-dates-editor`](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js), once for the start date and end date each, and once in the [`d2l-activity-due-date-editor`](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-due-date-editor.js).

As the `d2l-input-date-time` component may be used elsewhere, the fix should be toggled off by default to maintain behaviour for other uses.

### Fix

The fix involves adding `role="presentation"` to the fieldset in `<d2l-input-fieldset>`, as well as making the "date/time" label in `<d2l-input-date>` and `<d2l-input-time>` more verbose.  The presentation role tells screen readers that the fieldset should not be read aloud, while allowing labels to be read for contained elements (unlike `aria-hidden="true"`).  These changes will only take place if the **suppress-group** property is true.

In JAWS, this removes the word "group".  In NVDA, this removes the word "grouping".
The text formerly read out by the fieldset's legend is now the label on the input date and time components themselves, preserving the information served while avoiding the term "group".

When **suppress-group** is not present, the components behave as they did without the changes.

#### JAWS

Before change/with fix toggled off:

> Edit-activity dash course dash google chrome, complimentary region, **start date group**, date edit has pop-up, m slash d slash yyyy, type and text, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

After change:

> Edit-activity dash course dash google chrome, complimentary region, **start date**, date edit has pop-up, m slash d slash yyyy, type and text, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

#### NVDA

Before change/with fix toggled off:

> Complimentary landmark, **start date grouping**, date edit submenu m slash d slash yyyy blank, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

After change:

> Complimentary landmark, **start date**, date edit submenu m slash d slash yyyy blank, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

## Video/Screenshots

Testing revealed the "group" keyword was associated with the fieldset tag, seen here within the `d2l-input-label-fieldset` class:

A11y Pre-Fix 3

With the changes described above:

A11y Fix 3

Note that when the optional fix is toggled, NVDA's element list becomes more navigable with the full labels.

Without `suppress-group`:

NVDA Pre-Fix

With `suppress-group`:

NVDA Fix 2

## Related PRs

- Brightspace-UI/core: https://github.com/BrightspaceUI/core/pull/2230

## Remaining Work


test/d2l-activity-editor/d2l-activity-availability-dates-editor.test.js
```diff
@@ -111,15 +111,15 @@ describe('d2l-activity-availability-dates-editor', function() {
 			dates.setErrorLangTerms('end-before-start-due-date-error');
 			await nextFrame();
 
-			expect(startDateInput.invalid).to.be.true;
-			expect(endDateInput.invalid).to.be.true;
+			await waitUntil(() => startDateInput.invalid, 'Start date did not become invalid', { timeout: 30000 });
```
ianlaidlaw, line 114: 30 seconds seems high. does this mean we are waiting 30 seconds until we check `startDateInput.invalid` ? this seems like it could add a ton of time to builds, tests, etc.

Kiefer-Co-D2L: Looks like that 30 seconds is just the timeout.  The default interval is polling every 50 ms until the input goes invalid.

`waitUntil` from the `@open-wc` package:

```js
/**
 * Waits until the given predicate returns a truthy value. Calls and awaits the predicate
 * function at the given interval time. Can be used to poll until a certain condition is true.
 *
 * @example
 * ```js
 * import { fixture, waitUntil } from '@open-wc/testing-helpers';
 *
 * const element = await fixture(html`<my-element></my-element>`);
 *
 * await waitUntil(() => element.someAsyncProperty, 'element should become ready');
 * ```
 *
 * @param {() => boolean | Promise<boolean>} predicate - predicate function which is called each poll interval.
 *   The predicate is awaited, so it can return a promise.
 * @param {string} [message] an optional message to display when the condition timed out
 * @param {{ interval?: number, timeout?: number }} [options] timeout and polling interval
 */
export function waitUntil(predicate, message, options = {}) {
  const { interval = 50, timeout = 1000 } = options;

  return new Promise((resolve, reject) => {
    let timeoutId;

    setTimeout(() => {
      clearTimeout(timeoutId);
      reject(new Error(message ? `Timeout: ${message}` : `waitUntil timed out after ${timeout}ms`));
    }, timeout);

    async function nextInterval() {
      try {
        if (await predicate()) {
          resolve();
        } else {
          timeoutId = setTimeout(() => {
            nextInterval();
          }, interval);
        }
      } catch (error) {
        reject(error);
      }
    }
    nextInterval();
  });
}
```

Kiefer-Co-D2L: Looks like that 30 seconds is just the timeout.  The default interval is polling every 50 ms until the input goes invalid.

`waitUntil` from the `@open-wc` package:

```js
/**
 * Waits until the given predicate returns a truthy value. Calls and awaits the predicate
 * function at the given interval time. Can be used to poll until a certain condition is true.
 *
 * @example
 * ```js
 * import { fixture, waitUntil } from '@open-wc/testing-helpers';
 *
 * const element = await fixture(html`<my-element></my-element>`);
 *
 * await waitUntil(() => element.someAsyncProperty, 'element should become ready');
 * ```
 *
 * @param {() => boolean | Promise<boolean>} predicate - predicate function which is called each poll interval.
 *   The predicate is awaited, so it can return a promise.
 * @param {string} [message] an optional message to display when the condition timed out
 * @param {{ interval?: number, timeout?: number }} [options] timeout and polling interval
 */
export function waitUntil(predicate, message, options = {}) {
  const { interval = 50, timeout = 1000 } = options;

  return new Promise((resolve, reject) => {
    let timeoutId;

    setTimeout(() => {
      clearTimeout(timeoutId);
      reject(new Error(message ? `Timeout: ${message}` : `waitUntil timed out after ${timeout}ms`));
    }, timeout);

    async function nextInterval() {
      try {
        if (await predicate()) {
          resolve();
        } else {
          timeoutId = setTimeout(() => {
            nextInterval();
          }, interval);
        }
      } catch (error) {
        reject(error);
      }
    }
    nextInterval();
  });
}
```

ianlaidlaw: ahh gotcha gotcha. so doing some reading it looks like it will keep checking for the condition, and if it succeeds it will stop. i initially read this as each test waiting 30s before checking the condition

brissons (commented):
Makes sense to me! Especially if it's specific for the availability dates side nav.

Just wondering though, is using the 'group' term bad for all of the date inputs? The due date field for our FACE pages (and all instances of this component) are still going to say 'group' instead of 'region'. Just wondering why specifically the start/end dates on the sidebar need to change and not the other ones.

Kiefer-Co-D2L:
Based on discussion in Slack:

> it looks like based on Carin's screenreader expertise we should be fine keeping the current "Start Date Group" and "Start Date Grouping" behaviour. So we are going to just end up closing this for now.

PR #2472, opened 2022-03-16
DE47748: Ensure google form links are treated as google drive links

coffeexcode:
## Relevant Rally
- [DE47748](https://rally1.rallydev.com/#/?detail=/defect/629540745405&fdp=true): Google Drive > Forms are not recognized as Google Drive topics, seen just as weblinks

## Description
The link structure for google forms is slightly different than for google drive. So we've added the regex necessary to detect google form links, and they are displayed as an iframe, rather than the google drive viewer component.

## Related PRs
1. LMS: https://github.com/Brightspace/lms/pull/20681
2. Lessons: https://github.com/Brightspace/smart-curriculum/pull/1521


brissons (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/shared-components/d2l-activity-content-link-preview.js
```diff
@@ -59,7 +61,7 @@ class ContentEditorLinkPreview extends SkeletonMixin(LocalizeActivityEditorMixin
 				<iframe
 					src=${link}
 					@load=${this._onLoad}
-					scrolling='no'
+					scrolling=${this.scrolling ? 'yes' : 'no'}
```
ianlaidlaw, line 64: am i missing something obvious or is this always false? do we need a ternary?

ianlaidlaw: oh, i totally missed it being passed in from the other component, whoops

brissons: do we also want it to be there all the time, or would 'auto' be more fitting than 'yes'?

ianlaidlaw: in that case would it just be easier to set it like so? no need to do it in the constructor. very minor nit.
```
scrolling: {
				type: Boolean,
				value: false
			},
```

ianlaidlaw: i think the fact that it was manually being set like that kind of threw me off a bit. when i see the default value that seems to point me more towards attributes that can come in as different values. i may be alone there though.

coffeexcode: 'auto' might make more sense. I will update that, and quickly check if I can set the default in the props

ianlaidlaw (approved):

brissons (approved):

PR #2475, opened 2022-03-17
DE47805: FACE HTML/Files > Clicking Back/Cancel on unmodified files causes "Discard changes?" warning to popup (after clicking in editor)

prajapac:
## Relevant Rally Links

DE47805: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F629716368649

## Description

When you are on the FACE edit page for an html file, and you click in the editor and then back/cancel, you get a prompt asking if you'd like to discard changes even though there are none. Further if you make changes, hit save and then hit back again, you still get the same prompt (perhaps a bigger issue in this scenario).

## Goal/Solution

It turns out that there are inconsistencies between what is persisted in file storage and the contents of the html editor, specifically it seems initially that the persistent storage has carriage returns for newlines: `/r/n` whereas the html editor source doesn't: `/n`. If you click save twice, they both end up with carriage returns and the back/cancel button works properly.

Note: This is a partial fix, there is an issue with the HTML editor where it changes the source for new files, so files need to be saved at least once before this fix will work (see attached .gif). 

## Video/Screenshots

Before:
Before

After:
After

Issue with the HTML editor:
HTML Editor Defect


ChrisKraljevicD2L (approved):

coffeexcode (approved):
LGTM

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -211,7 +211,11 @@ export class ContentFile {
 			return true;
 		}
 
-		return this.persistedFileContent === this.fileContent;
+		return this._getStringWithoutCarriageReturns(this.persistedFileContent) === this._getStringWithoutCarriageReturns(this.fileContent);
```
ianlaidlaw, line 214: hmm i think this will work in 99.9999999% of scenarios. i'm just imagining something weird where the user adds <cr> tags as their conent and this will think nothing has changed :P i think it's fine but i'm sure there's a more robust solution we can come back

brissons (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
ianlaidlaw: maybe i'm just being crazy though

Kiefer-Co-D2L (approved):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
prajapac: That should come up as a change, the tags should still be there, the only thing that this would miss is if the user pastes a carriage return as the only input, i.e. `/r`

ianlaidlaw: right, yeah. but should that be considered valid input? if im not mistaken this won't actually save anything in that case. but like i said, that is such an extreme edge case its really not worth worrying about.

ianlaidlaw: maybe this function is just for the discard dialog and not the dirty function. so really nbd at all

PR #2480, opened 2022-03-21
DE47531: Using Insert Stuff > Enter Embed Code results in blank page

coffeexcode:
## Relevant Rally
- [DE47531](https://rally1.rallydev.com/#/?detail=/defect/628157570873&fdp=true): Lessons> 20.22.2 > Using Insert Stuff > Enter Embed Code results in blank page


brissons (commented):
lgtm!

Just curious: does this happen in Lessons/Assignments/Modules etc. (anything else that uses the html editor?)? This is a good fix, but if it's a problem with the html editor itself we might want to fix it at that end/let the proper authorities know.

PR #2490, opened 2022-03-23
DE43778: [A11y] FACE invalid title does not read "Title is required" tooltip

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F600735049517

## Description

> **Expected behaviour**:
> "Title is required" tooltip should be read out by the screen reader so the user knows why the page is not saving
> 
> **Actual behaviour**:
> Title invalid entry just reads "invalid entry" but not the "Title is required" toolip.
> 
> **To reproduce**:
> 1. Go to a Course with Lessons FACE enabled
> 2. Using NVDA and keyboard navigation, create/edit a new module/html/weblink/lti in Lessons
> 3. Clear the title field
> 4. Attempt to Save the topic/module
> 5. Notice the error alert pops up and the screenreader will read out "invalid entry" instead of telling the user why it's invalid

## Goal/Solution

The following text from the tooltip should be added as an `aria-label` in addition to marking the field in question as `aria-invalid`:

> Your content item wasn't saved. Please correct the fields outlined in red.

As this overrides the existing label, the existing label is also appended to the error tooltip aria label.

The `aria-invalid` attribute is now correctly removed when the error is not present.

### NVDA

Previously, reproducing the above steps with NVDA read:

> main landmark, Name star edit invalid entry blank

Now, when reproducing the steps above, NVDA will read:

> main landmark, **Your content item wasn't saved. Please correct the fields outlined in red.** Name star edit invalid entry blank

### JAWS

Previously, reproducing the above steps with JAWS read:

> main region, Name star edit invalid entry blank, type and text

Now, when reproducing the steps above, JAWS will read:

> main region, **Your content item wasn't saved. Please correct the fields outlined in red.** Name star edit invalid entry blank, type and text

## Screenshots/Video

See ARIA attributes on the Name field:

image

## Remaining Work

- [x] Test with JAWS


kristian-torres (approved):

coffeexcode (approved):
The amount of detail you put in your PR is always crazy good!

PR #2493, opened 2022-03-24
DE47339: Add default gradeSchemeId handling

mostlyfabulous:
[DE47339](https://rally1.rallydev.com/#/?detail=/defect/627263379249&fdp=true): Grades > 20.22.1.34790 > Grade Item created in Assignments tool explicitly sets gradeScheme, rather than using the default

The `gradeSchemeId` should be set as `Null` by the `ActivityUsageGradeAssociationProvider` of the relevant tool to indicate that the org unit's default grade scheme is to be is used instead. This PR facilitates this by using `-1` to indicate this in the API calls.

- [x] Relies on: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/470

**Allows the option to choose between the current org unit default or the independent grade scheme**
image

Todo:

- [ ] Selectbox compatibility when the default scheme is Percentage type


mdulat (approved):

components/d2l-activity-editor/lang/en.js
```diff
@@ -123,7 +123,7 @@ export default {
 	"grades.selectboxDescription": "Grade users by selecting the grade scheme level that best matches their achievement.", // Description of selectbox grade type
 	"grades.selectboxDescriptionExample": "E.g. \"Very Good\" or \"B+\"", // Example of selectbox grade type
 	"grades.newGradeScheme": "Grade Scheme", // Label for the grade scheme
-	"grades.defaultGradeScheme": "--Default-- ({schemeName})", // name of default grade scheme
+	"grades.defaultGradeScheme": "-- Default Scheme -- ({schemeName})", // name of default grade scheme
```
mostlyfabulous, line 126: As per Joseph's agreement

PR #2501, opened 2022-03-29
Update CODEOWNERS

patel-kajal:
adding team usa as codeowners for date type dialog


CODEOWNERS
```diff
@@ -3,3 +3,4 @@ d2l-quick-eval-widget/ @BrightspaceHypermediaComponents/nimbus
 /components/d2l-work-to-do/ @BrightspaceHypermediaComponents/luna
 /demo/d2l-work-to-do/ @BrightspaceHypermediaComponents/luna
 d2l-activity-editor/d2l-activity-score-editor.js @mdulat @mostlyfabulous
+d2l-activity-date-type-dialog/lang/ @BrightspaceHypermediaComponents/team-usa-devs @BrightspaceHypermediaComponents/team-usa-senior-devs
```
adkulas, line 6: Hmm, we have a [TeamUsa team](https://github.com/orgs/BrightspaceHypermediaComponents/teams/team-usa), but it only has Uzair and Cressa.

adkulas (dismissed):

CODEOWNERS
adkulas: Ooooo it says "8 are members of child teams" I think we should just use the TeamUSA one in that case

adkulas: Yeah it actually was Jordan not Cressa 😂

But yeah teamUSA should include everyone
image

pureooze (approved):

CODEOWNERS
pureooze: The round robin doesnt work properly when you use the parent team

PR #2506, opened 2022-03-29
fix: DE47952 100 percent cut off

aid848:
[DE47952](https://rally1.rallydev.com/#/?detail=/defect/630634032315&fdp=true): [FACE Quiz] Completion tracking setting passing grade to 100% cuts off number with 1... %


components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-completion-tracking-editor.js
```diff
@@ -78,6 +79,7 @@ class ActivityQuizCompletionTrackingEditor extends ActivityEditorMixin(RtlMixin(
 							<d2l-input-percent
 								label="${this.localize('ariaPassingGradeMin')}"
 								label-hidden
+								input-width="6rem"
```
mostlyfabulous, line 82: Is this a css hack 😆 to fix the cut off? `input-width` is 6rem but `max-width` is 4.25rem?

aid848: At the min width (320px) it gets cut off without the larger input-width, but if both are set to 6rem the desktop textbox becomes too large

mostlyfabulous: There should be media queries you can use: https://www.w3schools.com/css/css_rwd_mediaqueries.asp
Not sure if that's the practice we use though!

mostlyfabulous (approved):

components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-completion-tracking-editor.js
mdulat: just wondering why didn't media queries work here?

mdulat: oh I guess it's not a css thing but a component property, never mind.

aid848: The d2l-input-percent already had an input-width property that was usable but the change to the styles was originally only done to accommodate very thin screens. It didn't seem necessary to keep the styling change and add media queries if the input-width could be adjusted.

aid848: Hmm the comments here seem delayed I didn't even see the second one until after I refreshed 😆

PR #2513, opened 2022-03-30
DE47850: FACE > Radio button label width expands to length of container

kristian-torres:
## Relevant Rally Links

[DE47850](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F628583327467&fdp=true) : FACE > Radio button label width expands to length of container

## Description

Problem - Current Behaviour

In FACE, input radio labels take up the width of its container, making it clickable from across the page. For example LTI links, in the picture below, the option is clickable where the red mouse cursor is placed (far right of radio input).

## Goal/Solution

Add a style to the d2l-input-radio-label class that will set the width to `fit-content`. This will prevent the label from setting it's width to it's parent container and fit it around it's content. This will fix the click area to the content only and not the width of the radio label.

## Video/Screenshots

image

image

image

## Remaining Work
- [x] Dev Testing: another developer will test this code on their machine


brissons (approved):
dev tested and lgtm!

coffeexcode (approved):

PR #2522, opened 2022-04-01
DE28017: FACE Document Template > Screenreaders do not read out Confirmation Dialog text

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fdefect%2F168190686908

## Description

> **Steps to reproduce**:
> Using a screenreader try to add a template to an HTML doc in Lessons
>
> **Actual Results**:
> When the confirmation dialog is shown, only the buttons are read out. 
>
> **Expected Results**:
> The text and buttons on the dialog should be read out by the screenreader

## Goal/Solution

An ARIA label has been added to the confirmation dialog pop-up.

### NVDA

Previously, reproducing the above steps with NVDA read:

> main landmark, **Your content item wasn't saved. Please correct the fields outlined in red.** Name star edit invalid entry blank

Now, when reproducing the steps above, NVDA will read:

> main landmark, **Name is required.** Name star edit invalid entry blank

### JAWS

Previously, reproducing the above steps with JAWS read:

> main region, **Your content item wasn't saved. Please correct the fields outlined in red.** Name star edit invalid entry blank, type and text

Now, when reproducing the steps above, JAWS will read:

> main region, **Name is required.** Name star edit invalid entry blank, type and text

## Screenshots/Video

See ARIA label attached:

image

## Remaining Work

- [ ] Test with JAWS
- [ ] Test with NVDA


prajapac (dismissed):

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-html-file-detail.js
```diff
@@ -252,6 +252,14 @@ class ContentHtmlFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeAct
 	_openReplaceHtmlTemplateDialog() {
 		const dialog = this.shadowRoot && this.shadowRoot.querySelector('d2l-dialog-confirm');
 		if (dialog) dialog.opened = true;
+
+		const innerDialog = dialog.shadowRoot && dialog.shadowRoot.querySelector('dialog');
+		if (innerDialog) {
+			const fullDialogText = `${this.localize('content.confirmDialogTitle')} ${this.localize('content.confirmDialogBody')}`;
```
prajapac, line 258: Do we want to merge the two texts, is there any reason why we shouldn't? Mostly just curious.

Kiefer-Co-D2L:
This one could really use a dev test!  This is the only way I've found to consistently read out the missing text, but it comes at the cost of making the buttons more cumbersome to read.  Suggestions very welcome.

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-html-file-detail.js
```diff
@@ -301,17 +307,22 @@ class ContentHtmlFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeAct
...
 			<d2l-dialog-confirm
-				title-text="${this.localize('content.confirmDialogTitle')}"
-				text="${this.localize('content.confirmDialogBody')}"
+				title-text="${titleText}"
+				text="${dialogText}"
 				@d2l-dialog-close=${this._handleReplaceHtmlTemplateDialogClose}
 			>
-				<d2l-button slot="footer" primary data-dialog-action="yes">
-					${this.localize('content.confirmDialogActionOption')}
+				<d2l-button slot="footer" primary data-dialog-action="yes" aria-label="${titleText} ${dialogText} ${actionText}">
```
brissons, line 321: It seems strange to read out all the texts when a user tabs between buttons.

I noticed a fairly recent PR to the dialog component which addresses a similar issue for Quzzing:
https://github.com/BrightspaceUI/core/pull/2030

I would check if this happens with the dialog in quizzing, then maybe try converting this dialog from `d2l-dialog-confirm` to `d2l-dialog`, then adding the` describe-content` property (and anything else that needs changing),

Kiefer-Co-D2L: I agree, felt janky but verbose seemed to be better than risking the text not being read.  That PR seems promising though, I'll give it a try!

Kiefer-Co-D2L: That `describe-content` property only works with NVDA, not JAWS.  Based on this and previous A11y defects where the problem reader was JAWS, I think maybe the UI core components are only tested for NVDA?

brissons: Oh shoot! This would mean that this defect would apply to the other confirmation windows (ex: the discard popups) using JAWS?

PR #2525, opened 2022-04-01
Mfurlong/us137648 manage special access basics

Mackenzie-Furlong:
- Created the first special access dialog skeleton.
- Dialog slightly differs from the design because we have decided to use the fullscreen dialog for sizing.
- The dialog features only blank state functionality and has all the necessary buttons but only the ones to close the dialog are funcitonal.
- The dialog also has a dynamic title that adjusts to the title of the activity

[Rally Story](https://rally1.rallydev.com/#/602620483013d/custom/602630203785?detail=%2Fuserstory%2F630828175943)


components/d2l-activity-editor/lang/en.js
```diff
@@ -86,6 +87,9 @@ export default {
 	"editor.assignment": "assignment", // Langterm for assignment or dropbox
 	"editor.hiddenHeaderCoreProperties": "Core properties", // Accessible text for hidden header describing core activity properties
 	"editor.hiddenHeaderAdvancedProperties": "Advanced properties", // Accessible text for hidden header describing advanced activity properties
+	"editor.btnDoneSpecialAccess": "Done", // Button text for being done with special access dialog
```
poojajoshi11, line 90: In the designs it says "Set" on the Special Access page. Should we follow that closely?

Mackenzie-Furlong: Nicole and I discussed this and she seems okay with the done button for now. It was some design liberty I took for now because it doesn't seem that we have finalized the language for this yet and the decision is still yet to be made as to when the changes are going to be saved. This may still yet turn into a "save" button in the future.

poojajoshi11 (approved):

PR #2534, opened 2022-04-05
fix: console error

mdulat:
Getting this console error when opening grades dialog:

image


mostlyfabulous (approved):
Thanks for catching that! Will this cause CI to complain with the optional chaining thing?

mdulat:
> Will this cause CI to complain with the optional chaining thing?

It shouldn't because we use it all over the place now.

PR #2538, opened 2022-04-06
fix: DE48085 save pasted images to course files

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/631259879407&fdp=true

Without the `ActivityEditorMixin` the component's `save()` function wasn't being called, so the files were never moved from the temp area into course files.


dustxd (approved):

mostlyfabulous (approved):
Wow, strange we never caught this

mdulat:
> Wow, strange we never caught this

It's something that I broke 3 months ago :( so it took that long for a customer to notice.

PR #2636, opened 2022-05-13
Switch to new Video/Audio renderer component

rcausey:


package.json
```diff
@@ -28,7 +28,6 @@
     "@babel/polyfill": "^7.0.0",
     "@babel/preset-env": "^7.8.3",
     "@brightspace-ui/stylelint-config": "^0.0.3",
-    "@brightspace/content-components": "github:Brightspace/content-components#semver:^3",
```
rcausey, line 31: This just seems like a bug to me... right? It probably only works because we use content-components in BSI dependencies

qwangd2l: Yeah, it probably works because of BSI. But I would keep this included in package.json

qwangd2l (approved):

PR #2640, opened 2022-05-17
Fix undefined org unit path for replace files dialog

ianlaidlaw:
When using the replace files dialog, the ou in the path was null, so instructors could not actually use it. This worked for d2lsupport, since that must ignore those restrictions.

I also added in a null check when attempting to fetch captions which was sometimes happening. I think this is not 100% needed since my fixes on the LMS side should fix it, but it is worth keeping.

# BEFORE:

image

# AFTER:

image


components/d2l-activity-editor/d2l-activity-content-editor/content-file/state/content-file.js
```diff
@@ -105,6 +105,11 @@ export class ContentFile {
 
 	async fetchCaptions() {
 		const sirenResponse = await fetchEntity(this.mediaCaptionsHref, this.token, true);
+
+		if (!sirenResponse) {
```
ianlaidlaw, line 109: Before I did my LMS changes: https://github.com/Brightspace/lms/pull/23274/files I was experimenting with fixing the captions issue here in activities. This turned out not to be a good-enough fix since you wouldnt be able to add new captions (it will still work off the old non-working copy) but I think this error could still happen. I think this is worth happening, if we do get an undefined sirenResponse we don't want to attempt to fetch things based off of that

components/d2l-activity-editor/d2l-activity-content-editor/content-file/d2l-activity-content-file-detail.js
```diff
@@ -59,7 +59,7 @@ class ContentFileDetail extends SkeletonMixin(ErrorHandlingMixin(LocalizeActivit
 		this.skeleton = true;
 
 		const context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
-		this.orgUnitId = context ? context.orgUnitId : '';
+		this._orgUnitId = context ? context.orgUnitId : '';
```
ianlaidlaw, line 62: 😳

prajapac (approved):

connor-lowey (approved):

PR #2641, opened 2022-05-17
Remove references to d2l-html-editor

grant-cleary:
We no longer support the new editor being disabled, so remove the connecting code from activities.


components/d2l-activity-editor/d2l-activity-text-editor.js
```diff
@@ -1,4 +1,5 @@
 import '@brightspace-ui/core/components/inputs/input-textarea.js';
+import './d2l-activity-html-new-editor';
```
dlockhart, line 2:
```suggestion
import './d2l-activity-html-new-editor.js';
```

dlockhart (approved):
Nice!

components/d2l-activity-editor/d2l-activity-text-editor.js
grant-cleary: Whoops, thanks for catching that!

dlockhart (approved):

PR #2644, opened 2022-05-18
DE48592: Don't attempt to delete modules on cancel if they are not new

ianlaidlaw:
There exists logic to cancel the creation of a module, since the module must be created first before it can be edited. Previously, if the user would cancel this creation (via the opt-out flyout) on a module that already existed, it would delete the module and the user would get stuck in a weird state. With these changes, this only happens on a "new" module.

## On an existing module:

Recording 2022-05-18 at 13 04 45

## On a new module:

Recording 2022-05-18 at 13 08 41


prajapac (commented):
How do the pieces connect, i.e. specifically where is the spot that determines whether or not the module `isNew`, the code changes pass it around but I don't see where it's value is determined?

Kiefer-Co-D2L (approved):
LGTM!  What a weird edge case.

prajapac (approved):

PR #2665, opened 2022-05-25
Clean up quick-eval-widget-localization

bearfriend:
Moving `localize-quick-eval-widget.js` out of the `lang` directory allows us to remove the hardcoded list of locales in `mfv.config.json` for the entire repo.

In addition:
- Move to `LocalizeDynamicMixin`
  - (Adds Hindi support)
- Add OSLO support
- Remove serge `output_lang_rewrite` config


dlockhart (approved):
So many awesome things all together!

PR #2676, opened 2022-05-31
DE48362 Hide update grade confirmation dialog for FACE Quiz

dustxd:
### Motivation/Context
This PR covers this part of the rally defect description
> grades.dialogConfirmUpdateGrade and grades.dialogConfirmTitle should be removed from the quizzing page as this dialog can no longer be reached through the quizzing workflow (DO NOT remove completely, just disassociate from quizzing as it is used in assignments)

### Rally Defect
[DE48362](https://rally1.rallydev.com/#/?detail=/defect/634194524971&fdp=true): [Face Quiz] > grades confirmation dialog lang terms not being picked up


components/d2l-activity-editor/d2l-activity-score-editor.js
```diff
@@ -329,10 +329,12 @@ class ActivityScoreEditor extends ActivityEditorMixin(SkeletonMixin(LocalizeActi
 					</div>
 				` : null}
 			</div>
-			<d2l-dialog-confirm title-text="${this.localize('grades.dialogConfirmTitle')}" text="${this.localize('grades.dialogConfirmUpdateGrade')}">
-				<d2l-button slot="footer" primary data-dialog-action="yes">${this.localize('editor.yesLabel')}</d2l-button>
-				<d2l-button slot="footer" data-dialog-action="no">${this.localize('editor.noLabel')}</d2l-button>
-			</d2l-dialog-confirm>
+			${canEditScoreOutOf ? html`
```
dustxd, line 332: I don't really like this because it modifies the file history but doesn't do much - this dialog is only opened during save if it's an assignment, has submissions, **and** the grade has changed in some sense (see [this block here](https://github.com/BrightspaceHypermediaComponents/activities/blob/45a749f12dcc3b3244114713b129c4086ba04481/components/d2l-activity-editor/d2l-activity-score-editor.js#L417-L426)). But I guess it's also safer to have a guard in the template?

mdulat (approved):

components/d2l-activity-editor/d2l-activity-score-editor.js
dustxd: Decided there's no harm adding a check, so we're proceeding with the PR 😄

PR #2679, opened 2022-06-02
DE48850 - Save and Copy - Change link type, add optional chaining

jguann:
[~~US139032~~](https://rally1.rallydev.com/#/431372673576d/search?detail=%2Fuserstory%2F636555822723)

Repurposing this PR to handle the change made for fixing [DE48850](https://rally1.rallydev.com/#/611579323773d/search?detail=%2Fdefect%2F639432268079)
- change to get link of rel type `https://api.brightspace.com/rels/valence-alternate` (backend change in https://github.com/Brightspace/lms/pull/23927)
- add optional chaining for safety


components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js
```diff
@@ -228,10 +228,10 @@ export const ActivityEditorContainerMixin = superclass => class extends Activity
 
 		const activityEntity = await activityStore.fetch(this.href);
 
-		const assignmentHref = activityEntity._entity._entity._linksByRel['https://api.brightspace.com/rels/assignment'][0].href;
+		const assignmentHref = activityEntity?._entity?._entity?._linksByRel['https://api.brightspace.com/rels/assignment'][0]?.href;
 		const assignmentEntity = await assignmentStore.fetch(assignmentHref);
 
-		const dropboxUrl = assignmentEntity._entity._entity._linksByRel.alternate.filter(x => x.type === 'application/json')[0].href;
+		const dropboxUrl = assignmentEntity?._entity?._entity?._linksByRel.alternate?.filter(x => x.type === 'application/json')[0]?.href;
```
jopnick, line 234: I like it! what do you do in the failure case, where one of these failed and you `dropboxUrl` is not set?

jguann: Hmm, in that case, it would only be able to Save since there's no dropboxId for the Copy portion.

@jorge-ramirez-arredondo Would it be acceptable behaviour to completely ignore the Copy if we can't fetch the dropboxId? Do we need to show something for that failure?

jorge-ramirez-arredondo: Hmmm, ideally, I think we would still want to redirect the user to the assignments view, and then show some error message, otherwise the user might be confused that they selected the save and copy, but the copy didn't do anything. Something like "Failed to open copy dialog".

A less ideal option would be to just send the user to the assignments view and not do anything else. Effectively just making the button work the same as the regular save button.

I would only choose the less ideal option if the effort to set that error message up is not worth it considering the likelihood of this scenario happening. Either way though, I think you might want to touch base with Sucheta/Chris/Josh about this.

jguann: For this PR, this case will be handled as if only Save were clicked (no Copy happens, and user is not informed). This will be properly handled in a separate ticket/PR

lweber-d2l (approved):

